PR #2, opened 2015-10-19
.gitignore file with values for initial setup

bryk:
Files from the following tools are ignored: npm, bower, gulp, InteliJ and Vi.


PeWu:
LGTM

PR #4, opened 2015-10-20
Define source code folder structure with sample app

bryk:
This is a sample Angular application written in ES6 with modules. It has
sample unit and integration tests. It was generated with a help of
Yeoman.

Don't mind what the application does right now. I'll use it for a next
few days to verify that build system is working. Then it will be
revamped.


bolu:
sweet

PR #8, opened 2015-10-21
Annotate JS source to make Closure Compiler work

bryk:
With this change advanced optimizations of the Closure Compiler work and
produce working output.


PeWu:
LGTM

PR #10, opened 2015-10-21
Protractor configuration file for the frontend

bryk:
This is a basic config that allows to run the tests. Subsequent pull
requests will use it in Gulp tasks to run protractor tests over
development and production code.


preillyme:
LGTM

PR #11, opened 2015-10-21
Define gulp test unit and integration test task

bryk:
Once all Gulp files are in the repository tests will be run with `# gulp
test` or `# gulp integration-test`.


bryk:
@mgr01 PTAL

PR #12, opened 2015-10-21
Gulp tasks for compiling JavaScript code

bryk:
There are two main tasks. One that creates dev bundle and the other that
compiles into prod version.


PeWu:
LGTM

PR #13, opened 2015-10-21
Create Gulp tasks that process stylesheets

bryk:
This will be used later by other Gulp tasks to create prod and dev
application versions.


preillyme:
LGTM

bryk:
Thank you Patrick for looking into this.

PR #14, opened 2015-10-21
Gulp tasks that wire deps to frontend index file

bryk:
This will be later used by other Gulp tasks that serve or consume
index.html file.


PeWu:
LGTM

PR #16, opened 2015-10-22
Create Go HTTP server that serves static files

bryk:
@gmarek Can you do the review? I'll do the merge if LGTMed. 

--- 
This is a simple implementaion of a HTTP server serving static files
from CWD. It'll be used for development for a next few weeks.

This can be tested with `# go build console.go` and then `# ./console
--port=8081`.

### DIFF
--- a/src/app/backend/console.go
+++ b/src/app/backend/console.go
@@ -1,0 +1,40 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+  "flag"
+  "fmt"
+  "log"
+  "net/http"
+)
+
+var (
+  argPort = flag.Int("port", 8080, "The port to listen to for incomming HTTP requests")
+)
+
+func main() {
+  flag.Parse()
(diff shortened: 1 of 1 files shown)


src/app/backend/console.go
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+  "flag"
```
gmarek: In k8s we're using spf13/pflag for flag handling. I'd be nice to be consistent , but it's not a hard requirement.

src/app/backend/console.go
```diff
@@ -0,0 +1,40 @@
...
+  "fmt"
+  "log"
```
gmarek: In the same spirit as above: we're using golang/glog for logging.

src/app/backend/console.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func main() {
+  flag.Parse()
+  log.Print("Starting HTTP server on port ", *argPort)
+
+  // Run a HTTP server that serves static files from current directory.
+  // TODO(bryk): Disable directory listing.
+  http.Handle("/", http.FileServer(http.Dir("./")))
+  err := http.ListenAndServe(fmt.Sprintf(":%d", *argPort), nil)
```
gmarek: We're using glog.Fatal(http.ListenAndServe(...))

gmarek:
Few nits, which you can change to be more consistent with 'main' K8s style (which I think is a good idea), or ignore if you don't think it'll give any benefit.

Other than that LGTM.

src/app/backend/console.go
bryk: Thanks for pointing this! Done.

src/app/backend/console.go
bryk: Done.

src/app/backend/console.go
bryk: Done.

PR #19, opened 2015-10-23
Wire Gulpfile which allows to run Gulp tasks

bryk:
This is the main file that wires up all dependencies for Gulp. When it
is in the repository along with the other 2 pull requests,
`# gulp test` or `# gulp build` will work.


PeWu:
LGTM

PR #27, opened 2015-10-27
Integrate build system tasks to make them work

bryk:
This is build integration commit after my recent set of separate changes with standalone tasks. Now, every task works and produces reliable output.

Here I, e.g., change paths so that files are served correctly and fix source maps in style tasks.


bryk:
Don't mind travis errors. It is not configured yet.

bryk:
Fixed travis errors. Now CI works.

bryk:
@PeWu Can you take a look?

PeWu:
LGTM

PR #28, opened 2015-10-28
Use standard Angular router instead of the new one

bryk:
This change modifies a lot of JS, but this is only to prove that the
code works. Later changes will cleanup the JS more.

PS. Ignore travis errors - it is not configured yet.


PeWu:
LGTM

PR #32, opened 2015-10-28
Project architecture design document

bryk:
This is architecture proposal for MVP version of the product. When MVP
is done it'll be revisited to verify whether it (still) works well.


bryk:
@bgrant0607 @PeWu @mgr01 @gmarek @jwforres @lavalamp @jimmidyson @preillyme @emacias

CCing folks that expressed interest in this project and its architecture.

bgrant0607:
LGTM

jimmidyson:
:+1:

preillyme:
LGTM

docs/design/architecture.md
```diff
@@ -0,0 +1,29 @@
...
+timeline”. Figure 1 outlines the architecture of the project.
+
+![Architecture Overview](architecture.png?raw=true "Architecture overview")
+
+*Figure 1: Project architecture overview*
+
+Rationale for having a backend that implements business logic:
+
+* Clear separation between the presentation layer (frontend) and business logic layer (backend).
+This is because every action goes through well defined API.
```
liggitt: is this a new API?

docs/design/architecture.md
```diff
@@ -0,0 +1,29 @@
...
+Kubernetes Console project consists of two main components. They are called here the
+frontend and the backend.
+
+The frontend is a single page web application that runs in a browser. It fetches all its
+business data from the backend using standard HTTP methods. It does not implement business logic;
+it only presents fetched data and sends requests to the backend for actions.
+
+The backend runs in a Kubernetes cluster as a kubernetes service. Alternatively, it may run anywhere
+outside of the cluster, given that it can connect to the master. The backend is a HTTP server that
+proxies data requests to appropriate remote backends (e.g., k8s apiserver or heapster) or implements
```
liggitt: adding another layer here will complicate authentication methods that are not transitive (like x509 certificate auth)

docs/design/architecture.md
```diff
@@ -0,0 +1,29 @@
...
+![Architecture Overview](architecture.png?raw=true "Architecture overview")
+
+*Figure 1: Project architecture overview*
+
+Rationale for having a backend that implements business logic:
+
+* Clear separation between the presentation layer (frontend) and business logic layer (backend).
+This is because every action goes through well defined API.
+* Transactional actions are easier to implement on the backend than on the frontend. Examples of
+such actions: "create a replication controller and a service for it" or "do a rolling update".
```
liggitt: I'm not convinced this is true if the backend is simply making API calls to additional distributed backends using APIs that would be accessible to the frontend client

docs/design/architecture.md
```diff
@@ -0,0 +1,29 @@
...
+* Possible code reuse from existing tools (e.g., kubectl) and upstream contributions to the tools.
+* Speed: getting composite data from backends is faster on the backend (if it runs close to the
+data sources). For example, getting a list of pods with their CPU utilization timeline
+requires at least two requests. Doing them on the backend shortens RTT.
```
liggitt: what about the speed cost of adding another hop, especially for things like logs and terminals in the browser? Those already have two hops (browser > master > node), this would add another to the stack

jwforres:
cc @smarterclayton

docs/design/architecture.md
```diff
@@ -0,0 +1,29 @@
...
+frontend and the backend.
+
+The frontend is a single page web application that runs in a browser. It fetches all its
+business data from the backend using standard HTTP methods. It does not implement business logic;
+it only presents fetched data and sends requests to the backend for actions.
+
+The backend runs in a Kubernetes cluster as a kubernetes service. Alternatively, it may run anywhere
+outside of the cluster, given that it can connect to the master. The backend is a HTTP server that
+proxies data requests to appropriate remote backends (e.g., k8s apiserver or heapster) or implements
+business logic. The backend implements business logic when remote backends APIs do not
```
jwforres: instead shouldn't we be pushing back on the api to support the use cases consoles want/need so that everyone can benefit from it instead of writing one-off versions just for kube dashboard?

docs/design/architecture.md
bgrant0607: Authorization is a problem for the dashboard, monitoring (heapster, kubedash), and logging (elasticsearch). It would be useful for someone to think about how it should work.

One idea a couple of us have discussed for the dashboard is to launch dashboards running on behalf of the user on demand, rather than running the dashboard as a shared service. @erictune

docs/design/architecture.md
bgrant0607: Yes, though that takes longer.

A Go server could also reuse our client-side caching library.

bgrant0607: But it's worth trying other caching solutions.

docs/design/architecture.md
bryk: @liggitt  On the other hand, when there is no backend for the UI, the browser has to have direct access to the kubernetes master (or, at least, apiserver) and store credentials on user's computer. This essentially means that master has to be world accessible, which is not true in all setups. 

Can you describe use case for non-transitive auths? Is this something hypothetical, or solution that people actually use?

docs/design/architecture.md
bryk: @liggitt Backend can make those actions transactional, doing such thing on frontend is very hard. Having transactions is important. Without them you'd have dangling resources when, e.g., your internet connection fails while doing rolling update.
Does this make sense?

docs/design/architecture.md
liggitt: > use case for non-transitive auths? Is this something hypothetical, or solution that people actually use?

x509 client certificates are a primary example, and yes, people use them :)

to support non-transitive methods like this, you have a few options:
* the intermediate backend handles auth/authz, then escalates to the API using its own credentials
* the intermediate backend handles auth, then makes calls to the API using its own credentials, on behalf of the user (requires delegation support in the API)

docs/design/architecture.md
liggitt: An intermediate backend is also subject to internet connection failure

docs/design/architecture.md
bryk: When the backend is running close to the k8s apiserver (e.g., in the same cluster), then the cost of another hop can be ignored. Right? 
Wen the backend runs far away from the k8s apiserver, but still in a data center, then it still speeds up things. Especially in cases where multiple requests are made to get the data or filters have to be applied. Of course, in cases where we simply get the data that apiserver returns, this is slower. But only a little.

In summary, cases where composite/processed data is fetched are sped up. This is what matters.

Does this make sense to you?

liggitt: > Especially in cases where multiple requests are made to get the data or filters have to be applied

That sounds like a batch API, and a filtering API. Again, I'm wary of stacking new APIs on top of the core one... I'd rather get batch, transactions, filtering, etc, in the core API

docs/design/architecture.md
bryk: @jwforres Yes, changes to API should be pushed whenever possible. The UI will, however, contain business logic, no matter what is in the API. We need to place the business logic somewhere and Go backend is a good place for it.

You might as, why business logic in the UI? Imagine a page that displays services, replication controllers and pods at the same time. The data has to be fetched from the apiserver with multiple requests, then joined and then perhaps processed. This is business logic the UI will have.

docs/design/architecture.md
bryk: Private API only for backend<->frontend communication. It has to be well defined, though. As every API.

bgrant0607: Do you expect the frontend and backend to be deployed together in the same pod? That would make upgrades simpler.

docs/design/architecture.md
bgrant0607: Note that we're going to do rolling update as a service in 1.2, so don't do that in the backend, even.

https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/deployment.md

docs/design/architecture.md
bgrant0607: Relevant:
https://github.com/kubernetes/kubernetes/issues/12143
https://github.com/kubernetes/kubernetes/issues/7311

So far, nobody has had the time to move these forward, unfortunately.

bgrant0607: Other than Deployment.

docs/design/architecture.md
gmarek: IIUC the next paragraph frontend will run in browser.

bryk: The plan is to have everything in a single package running in a single pod. In the pod, backend serves JavaScript frontend to the user and then the browser JavaScript communicates with the backend.

docs/design/architecture.md
bryk: Thank you @liggitt, this is helpful. I think the summary is that non-transitive auth methods are harder to handle with backend, but they aren't blockers. Right?

docs/design/architecture.md
bryk: @liggitt Yes, the backend can crash/loose connection as every distributed service. This does not prevent it from doing actions transactionally. Doing transactions purely on a browser frontend is much harder (if even possible).

bryk: @bgrant0607 Sure, we will not implement rolling update. This was just an example of a compound write action.

docs/design/architecture.md
bryk: @liggitt Re stacking APIs: If all this was done on the browser frontend, would then it be okay? So, I postulate that this is no different from having a backend that exposes a private API for the browser frontend. Nobody else, except from the browser frontend, would be able to use the private API. This is just a matter of where to put the business logic: frontend or backend.

Creating a web UI that directly talks with "base" APIs is a very tempting way to go. We've put a great deal of thoughts on whether to have backend or be backend-less and the decision was hard to make. The general conclusion is that if the UI is read-only and accesses single API, then making is backen-less is OKay. Otherwise, tons of problems arise. From aforementioned transactions, through mixing business logic with views, slow browsers to ops issues (e.g., "how can I know whether this application even works?"). I've seen large UIs that started with backend-less approach which now, when the UIs are out in the wild, causes so much troubles. And of course, the backend-less approach didn't live for a long time because of the troubles.

Does this make sense @liggitt? Do you know any non-trivial UIs that are happy with being backend-less?

bryk:
Merging this, as this is the architecture we ended with.

PR #33, opened 2015-10-28
Fix unit tests and prevent protractor from hanging

bryk:
Once unit and integration tests are hooked to travis, such errors should
not happen again.


bryk:
@bolu I've updated the branch. PTAL

PR #34, opened 2015-10-28
chore(build): cache node_modules in travis builds

SebastianM:
This sets the option in the Travis CI configuration to cache all the
downloaded Node.JS modules, which can senificantly speed up travis
builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@
 
 # Explicitly set the environment to be container-based. This makes the builds faster.
 sudo: false
+# Cache downloaded Node.JS modules in the node_modules directory for faster builds.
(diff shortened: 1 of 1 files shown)


bryk:
LGTM

Nice, thank you! I see than npm install takes 100+ seconds. Caching this is a good improvement.

bryk:
How about caching installation of bower, gulp and bower_dependencies? Will it work?

SebastianM:
Yeah, this will work :tada: . Will send a PR this evening (that there no global installed packages and all dependencies (node and bower) are cachable).

bryk:
Awesome, thank you!

PR #35, opened 2015-10-29
Rename Kubernetes Console to Kubernetes Dashboard

bryk:
This commit changes all occurrences of word "console" to "dashboard".

The rename was required due to collision with an other project.


bryk:
FYI @SebastianM @preillyme

PR #36, opened 2015-10-29
Add new rules to ESlint checker and update code

bryk:
This is to have unified codebase from the beginning of the project.


preillyme:
LGTM

SebastianM:
Would add `'eqeqeq': [2, "allow-null"]` (http://eslint.org/docs/rules/eqeqeq)
Everything else LGTM :+1:

bryk:
Using `'eqeqeq': [2, "allow-null"]` makes sense. Thank you @SebastianM

bryk:
PS. I hope that over time we add more and more lint rules. Our aim for this project is to be high quality and high velocity. So, the more automatic checks are done over the codebase the better. 
The same is with JavaScript compiler. We go with strict settings so that most of bugs are found during compilation time.

PR #38, opened 2015-10-29
Use webpack instead of jscompiler in dev builds

bryk:
Webpack is much faster and does not do extensive file processing. This
means that when there is a syntax error, webpack often works correctly.

One downside is that I had to disable ng-strict-di. Will work on this
later to reenable it.

### DIFF
--- a/build/script.js
+++ b/build/script.js
@@ -20,31 +20,10 @@ import gulpAngularTemplatecache from 'gulp-angular-templatecache';
 import gulpClosureCompiler from 'gulp-closure-compiler';
 import gulpEslint from 'gulp-eslint';
 import gulpMinifyHtml from 'gulp-minify-html';
-import lodash from 'lodash';
 import path from 'path';
+import webpackStream from 'webpack-stream';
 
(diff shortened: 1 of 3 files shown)


preillyme:
LGTM but I'd like to see ```ng-strict-di``` working as well.

SebastianM:
LGTM. `gulp-ng-annote` would be useful in this case: https://www.npmjs.com/package/gulp-ng-annotate/

Have you considered using TypeScript for easier migration to Angular2 in the future?

PeWu:
LGTM

bryk:
@SebastianM I did try to use ng-annotate, but as plugin to webpack [0]. It did not work with ES6 modules. I mean, it mostly worked, but there were corner cases where it did not recognize @ngInject. It was in the case:
`/** @ngInject */ export default function foo() {};` 

[0] https://github.com/huston007/ng-annotate-loader

bryk:
@preillyme Yes, turn it on either only on prod build or make ng-annotate work. Will do.

PR #43, opened 2015-10-30
Add service list and zero state views

bryk:
This change is mostly boilerplate for the two views and cleanup of test
code that has been laying around.

No actual business logic in this change.

### DIFF
--- a/src/app/frontend/chrome/chrome.controller.js
+++ b/src/app/frontend/chrome/chrome.controller.js
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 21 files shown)


src/app/frontend/service/service.module.js
```diff
@@ -0,0 +1,28 @@
...
+import stateConfig from './service.state';
+
+
+/**
+ * Angular module for the service view.
+ *
+ * The view shows services running in the cluster and allows to manage them.
+ */
+export default angular.module(
+    'kubernetesDashboard.service',
```
PeWu: Let's call this "serviceList"

bryk: Done.

bryk:
@PeWu Updated. PTAL

PeWu:
LGTM

PR #45, opened 2015-11-02
Use Godeps dependency management for the backend

bryk:
The Godep tool is de facto standard in the Go and Kubernetes
communities. So, the ealier this project switches to it, the better. The
setup uses recommended layout from the project docs:
https://github.com/tools/godep

You might wonder why to pull entire Godeps folder here. This is how
Godeps strongly advise to do.


bryk:
@mgr01 Can you take a look?

PR #50, opened 2015-11-05
Run test, build and lint checks on Travis CI

bryk:
This change makes it possible to run tests, integration tests and lint
on Travis CI. It also prepares a special Gulp task to perform the same
activities locally. This is handy when, e.g., sending a pull requrest.


PeWu:
LGTM

PR #55, opened 2015-11-06
Backend implementation for deploy app view

bryk:
The implementation is basically a test one. It'll to verify that
end-to-end setup of the application works.

Tested with: `curl -H "Content-Type: application/json" -X POST -d '{"appName": "foo", "containerImage": "bar"}' http://localhost:8080/api/deploy`

Result:
```
$ kubectl get pods
NAME                       READY     STATUS                                      RESTARTS   AGE
foo                        0/1       Error: image library/bar:latest not found   0          29m
foo-5963                   0/1       Error: image library/bar:latest not found   0          10s
foo-7378                   0/1       Error: image library/bar:latest not found   0          8s
```

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -15,26 +15,42 @@
 package main
 
 import (
-	"io"
 	"net/http"
 
 	restful "github.com/emicklei/go-restful"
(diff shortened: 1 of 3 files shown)


bryk:
@mgr01 Ping?

mgr01:
Thanks

PR #56, opened 2015-11-09
Update gulp-sass dependency so that it compiles on OSX

bryk:
There was an issue with the package that prevented it from compilation on OSX, see: https://github.com/sass/node-sass/issues/1086

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "gulp-replace": "~0.5.4",
     "gulp-rev": "~6.0.1",
     "gulp-rev-replace": "~0.4.2",
-    "gulp-sass": "~0.7.1",
+    "gulp-sass": "~2.1.0",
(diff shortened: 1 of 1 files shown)


markogresak:
AFAIK the linked issue was related to io.js v3.x only. It didn't occur in node v0.12 and was fixed for v4.0.0 release.

bryk:
@markogresak That's interesting. Previous version did not compile on osx yosemite with xcode 7, node 5  and npm 3. But anyway, this PR fixes the issue.

markogresak:
There is a chance that the previous gulp-sass version was using an older version of node-sass, I remember that there were some issues in the past. 
But it's nice to see that sometimes fix is as easy as updating the version, although that's very rare in my experience :smile:

PeWu:
LGTM

PR #67, opened 2015-11-13
Fix stdio:inherit option for building go code

bryk:
Previously it was in the wrong parameter.


bryk:
@taimir Please review

taimir:
Yes, thank you! Now it is correct.

PR #69, opened 2015-11-13
Gulp: add spawn-backend to serve task

cheld:
gulp spawn-backend had several issues:

- apiserver-host missing
- default port 8080 clashes with local cluster. I switched to 9090
- not called by serve task

Does this match the intention?

Open issue: 'serve:prod' does uses browsersync to serve files. backend in dist should serve the files.

### DIFF
--- a/src/app/frontend/zerostate/zerostate.controller.js
+++ b/src/app/frontend/zerostate/zerostate.controller.js
@@ -20,15 +20,15 @@
  */
 export default class ZeroStateController {
   /**
-   * @param {!angular.$timeout} $timeout
    * @ngInject
    */
-  constructor($timeout) {
-    /** @export {number} */
-    this.testValue = 7;
-
-    $timeout(() => {
-      this.testValue = 8;
-    }, 4000);
+  constructor() {
+    /** @export {!Array<{title:string, link:string}>} */
+    this.learnMoreLinks = [
+      {title: 'Dashboard Tour', link: "#"},
+      {title: 'Deploying your App', link: "#"},
+      {title: 'Monitoring your App', link: "#"},
+      {title: 'Troubleshooting', link: "#"},
+    ];
   }
 }
--- a/src/app/frontend/zerostate/zerostate.html
+++ b/src/app/frontend/zerostate/zerostate.html
@@ -14,9 +14,34 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div layout="vertical" layout-fill>
-  <header>
-    Zero state page. <a ui-sref="microservicelist">Go to microservices page</a> or
-    <a ui-sref="deploy">deploy an app</a>.
-  </header>
+<div layout="row" layout-align="center center" layout-padding class="kd-zerostate-content">
+  <md-content layout layout-align="center" class="kd-content">
+    <md-card flex="50" class="kd-zerostate-deploy-card">
+      <md-toolbar layout="row" class="md-primary kd-zerostate-card-header"
+                  layout-align="center center" flex>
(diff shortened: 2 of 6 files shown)


build/conf.js
```diff
@@ -47,7 +47,11 @@ export default {
     /**
      * Port number of the development server started by Gulp.
      */
-    devServerPort: 8080,
+    devServerPort: 9090,
```
bryk: You should change this default in the dashboard.go too. To have consistent default ports here and when you start application manually.

Other option is to pick here a random free port. This would allow us to run two `gulp serve` tasks at the same time. What do you think? Is there a nodejs api for picking random free port?

build/serve.js
```diff
@@ -120,7 +120,7 @@ gulp.task('serve:prod', ['build-frontend'], function () {
 gulp.task('spawn-backend', ['backend', 'kill-backend'], function () {
   runningBackendProcess = child.spawn(
       path.join(conf.paths.serve, conf.backend.binaryName),
-      [`--port=${conf.backend.devServerPort}`], {stdio: 'inherit'});
+      [`--apiserver-host=${conf.backend.apiServerHost}`,`--port=${conf.backend.devServerPort}`], {stdio: 'inherit'});
```
bryk: Style nitpicks: please limit width to 100 chars and have spaces after `,`. Also, rebase the commit because I've already modified this line.

bryk:
Looks nice. Can you change pull request title to something meaningful (70 chars or less...) and squash your commits into one.

build/serve.js
```diff
@@ -102,7 +102,7 @@ gulp.task('serve:watch', ['watch'], serveDevelopmentMode);
 /**
  * Serves the application in development mode.
  */
-gulp.task('serve', ['index'], serveDevelopmentMode);
+gulp.task('serve', ['spawn-backend','index'], serveDevelopmentMode);
```
bryk: This wasn't added here because backend would crash with no `apiserver-host` param. Thanks!

build/serve.js
```diff
@@ -102,7 +102,7 @@ gulp.task('serve:watch', ['watch'], serveDevelopmentMode);
 /**
```
bryk: How about adding 'spawn-backend' dependency to 'serve:watch' and 'serve:prod'?

cheld: spawn-backend uses the binary from .tmp. 

In case of 'serve:prod' the binary from dist should be used, I assume. 

In addition serve:prod starts browser-sync. I think, serve:prod should be 're-implemented' to use the dashboard binary from dist, only.

bryk: The reason that spawn-backend uses binary from .tmp is that the .tmp binary compiles much faster than dist one. Compilation options are different. What's more, the dist binary runs only on linux. This is because everything in dist is going to be packaged into a docker image. Do you have any idea on how to better arrange this?

bryk: Also, 'spawn-backend' hasnt been added before, because it does not work on Travis. I.e., integration-test:prod requires serve:prod, which in order to work on Travis needs Kubernetes apiserver host. We need development Kubernetes instance first to enable this.

Does this make sense?

cheld:
We have development-case and CI-case.

Development case: spawn-backend is nice

CI case: I assume the goal is to deploy the docker container to Kubernetes and run the integration tests like in production. As an intermediate step I would suggest run from dist

bryk:
Re CI case: Running on a real Kubernetes Cluster would be nice, but just running locally and connecting to a cluster via `--apiserver-host=` is good enough for next couple of weeks.

PR #83, opened 2015-11-19
Replica set view: added additional fields to cards

bryk:
The cards now display mostly fake data, but they look more complete. The
backend will be implemented soon.

replicasetlist-v2


mgr01:
LGTM

Issue #73, opened 2015-11-16
Implement namespace handling in deploy form


maciaszczykm:
This issue was created to implement namespace handling in deploy form:

![Design](https://raw.githubusercontent.com/kubernetes/dashboard/master/designs/11-11-2015/07_deploy%20form%20new%20namespace.png)

First step is to implement picking namespace from currently available ones, then new namespace creation etc.

I am working on this.

maciaszczykm:
Now I will implement adding new namespace from deploy form functionality:

![Design](https://raw.githubusercontent.com/kubernetes/dashboard/master/designs/11-11-2015/06_deploy%20form%20new%20namespace%20specified.png)

maciaszczykm:
Basic namespace functionality is implemented at the moment, however, still some improvements like validation process should be done. Closing issue for now.

PR #84, opened 2015-11-19
Implemented existing namespaces handling in deploy form

maciaszczykm:
Connected to #73 issue. Added:
- backend API to retrieve available namespaces
- frontend namespace field in deploy form
- backend handling of chosen namespace, it will be assigned to created services and replication controllers

New namespace creation from deploy form is still missing.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -46,6 +46,15 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(ReplicaSetList{}))
 	wsContainer.Add(replicaSetListWs)
 
+	namespaceListWs := new(restful.WebService)
+	namespaceListWs.Path("/api/namespace").
+		Produces(restful.MIME_JSON)
+	namespaceListWs.Route(
+		namespaceListWs.GET("").
+			To(apiHandler.handleGetNamespaceList).
+			Writes(NamespacesList{}))
+	wsContainer.Add(namespaceListWs)
+
 	return wsContainer
 }
 
@@ -73,6 +82,19 @@ func (apiHandler *ApiHandler) handleGetReplicaSetList(
 	request *restful.Request, response *restful.Response) {
 
 	result, err := GetReplicaSetList(apiHandler.client)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	response.WriteHeaderAndEntity(http.StatusCreated, result)
+}
+
+// Handles get namespace list API call.
+func (apiHandler *ApiHandler) handleGetNamespaceList(
+	request *restful.Request, response *restful.Response) {
+
+	result, err := GetNamespaceList(apiHandler.client)
 	if err != nil {
 		handleInternalError(response, err)
 		return
--- a/src/app/backend/namespacelist.go
+++ b/src/app/backend/namespacelist.go
@@ -1,0 +1,44 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+// List of Namespaces in the cluster.
+type NamespacesList struct {
+	// Unordered list of Namespaces.
+	Namespaces []string `json:"namespaces"`
+}
+
+// Returns a list of all namespaces in the cluster.
+func GetNamespaceList(client *client.Client) (*NamespacesList, error) {
+	list, err := client.Namespaces().List(labels.Everything(), fields.Everything())
+
+	if err != nil {
+		return nil, err
+	}
+
+	namespaceList := &NamespacesList{}
+
+	for _, element := range list.Items {
+		namespaceList.Namespaces = append(namespaceList.Namespaces, element.ObjectMeta.Name)
+	}
+
+	return namespaceList, nil
+}
--- a/src/app/externs/angularresource.js
+++ b/src/app/externs/angularresource.js
@@ -21,6 +21,14 @@
 
 
 /**
+ * @typedef {{
+ *   $promise: !angular.$q.Promise
+ * }}
+ */
+angular.ResourceClass;
+
+
+/**
  * @typedef {function(string):!angular.Resource}
  */
 angular.$resource;
@@ -38,6 +46,7 @@ angular.Resource = function() {};
  * @param {!T} data
  * @param {function(!T)=} opt_callback
  * @param {function(!angular.$http.Response)=} opt_errback
+ * @returns {!angular.ResourceClass}
  */
 angular.Resource.prototype.save = function(data, opt_callback, opt_errback) {};
 
@@ -45,5 +54,15 @@ angular.Resource.prototype.save = function(data, opt_callback, opt_errback) {};
 /**
  * @param {function(!T)=} opt_callback
  * @param {function(!angular.$http.Response)=} opt_errback
+ * @returns {!angular.ResourceClass}
  */
 angular.Resource.prototype.get = function(opt_callback, opt_errback) {};
+
+
+/**
+ * @param {function(!T)=} opt_callback
+ * @param {function(!angular.$http.Response)=} opt_errback
+ * @returns {!angular.ResourceClass}
+ */
+angular.Resource.prototype.query = function(opt_callback, opt_errback) {};
+
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -37,6 +37,9 @@ type AppDeployment struct {
 
 	// Whether the created service is external.
 	IsExternal bool `json:"isExternal"`
+
+	// Target namespace of the application.
+	Namespace string `json:"namespace"`
 }
 
 // Port mapping for an application deployment.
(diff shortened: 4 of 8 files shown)


maciaszczykm:
I signed it!

src/app/backend/namespacelist.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// List of namespaces in the cluster.
+type NamespaceList struct {
+	// Unordered list of namespaces.
+	Namespaces []Namespace `json:"namespaces"`
+}
+
+// Kubernetes namespace representation.
+type Namespace struct {
```
bryk: I think that a separate struct is not needed for namespaces. Even in kubernetes API is is just a string. So how about removing this struct and making `Namespaces` a slice of strings?

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -131,4 +143,24 @@ export default class DeployController {
...
+  initializeNamespaces_($log, $resource) {
+    /** @type {!angular.Resource<!backendApi.NamespaceList>} */
+    let resource = $resource('/api/namespace');
+
+    resource.get((namespaceList) => {
+      $log.info('Successfully fetched namespace list: ', namespaceList);
+      this.namespaces = namespaceList.namespaces;
+      this.namespace= this.namespaces[0];
+    }, (err) => {
+      $log.error('Error fetching namespace list: ', err);
```
floreks: It's better to use 'this.log_' here as it is already injected to constructor and assigned.

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -131,4 +143,24 @@ export default class DeployController {
...
+
+  /**
+   * Initializes namespace selector.
+   * @param {!angular.$log} $log
+   * @param {!angular.$resource} $resource
+   * @private
+   */
+  initializeNamespaces_($log, $resource) {
+    /** @type {!angular.Resource<!backendApi.NamespaceList>} */
+    let resource = $resource('/api/namespace');
```
bryk: In this file we have two patterns of using $resource now. The other is in line 61. Can you unify approaches so that both styles are the same? 

I mean. In the line 61 we create $resource('foo') object and then use it, and here we create $resource('foo') on demand. I like your solution more.

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -131,4 +143,24 @@ export default class DeployController {
...
     return !!portMapping.port && !!portMapping.targetPort;
   }
+
+  /**
+   * Initializes namespace selector.
+   * @param {!angular.$log} $log
+   * @param {!angular.$resource} $resource
+   * @private
+   */
+  initializeNamespaces_($log, $resource) {
```
bryk: Hmm... $resource and $log are already saved as private fields. I suggest using it when they are already available. With such approach we'll have consistent style across the file, which is very nice.

src/app/backend/namespacelist.go
maciaszczykm: I will do it.

src/app/frontend/deploy/deploy.controller.js
maciaszczykm: Already done.

src/app/frontend/deploy/deploy.controller.js
maciaszczykm: I will do it.

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -59,6 +68,8 @@ export default class DeployController {
 
     /** @private {boolean} */
     this.isDeployInProgress_ = false;
+
+    this.initializeNamespaces_($log, $resource);
```
bryk: I'd refrain from displaying entire form until the namespaces are loaded. What do you think? With that there'll be no flickering in the UI.

src/app/backend/apihandler.go
```diff
@@ -46,6 +46,15 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(ReplicaSetList{}))
 	wsContainer.Add(replicaSetListWs)
 
+	namespaceListWs := new(restful.WebService)
+	namespaceListWs.Path("/api/namespace").
+	Produces(restful.MIME_JSON)
```
bryk: Does this file needs gofmt? Please check

bryk:
Very nice piece of code. Just a few style comments.

src/app/frontend/deploy/deploy.controller.js
maciaszczykm: Already done.

bryk:
Btw, when this is ready for me to do a second round of review, add a comment "PTAL".

src/app/backend/apihandler.go
maciaszczykm: Applied gofmt.

src/app/frontend/deploy/deploy.controller.js
maciaszczykm: Already done.

maciaszczykm:
PTAL

src/app/frontend/deploy/deploy.state.js
```diff
@@ -26,6 +26,22 @@ export default function stateConfig($stateProvider) {
...
+
+  /**
+   * Resolves namespaces for the deploy view.
+   *
+   * @param {!angular.$resource} $resource
+   * @ngInject
+   */
+  function resolveNamespaces($resource) {
+    /** @type {!angular.Resource<!Array<!string>>} */
+    let resource = $resource('/api/namespace');
```
bryk: Style nit: primitives in JSDoc are always non-null, thus instead of `!string` write `string`.

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -131,4 +148,5 @@ export default class DeployController {
   isPortMappingEmpty_(portMapping) {
     return !!portMapping.port && !!portMapping.targetPort;
   }
+
```
bryk: Probably this NL is not needed.

src/app/externs/angularresource.js
```diff
@@ -38,6 +46,7 @@ angular.Resource = function() {};
  * @param {!T} data
  * @param {function(!T)=} opt_callback
  * @param {function(!angular.$http.Response)=} opt_errback
+ * @returns {angular.ResourceClass}
```
bryk: This always returns non null ResourceClass, i.e., prepend with `!`

src/app/backend/apihandler.go
```diff
@@ -46,6 +46,14 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(ReplicaSetList{}))
 	wsContainer.Add(replicaSetListWs)
 
+	namespaceListWs := new(restful.WebService)
+	namespaceListWs.Path("/api/namespace").
+		Produces(restful.MIME_JSON)
+	namespaceListWs.Route(
+		namespaceListWs.GET("").
+			To(apiHandler.handleGetNamespaceList))
```
bryk: Missing `Writes`?

src/app/backend/namespacelist.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Returns a list of all namespaces in the cluster.
+func GetNamespaceList(client *client.Client) ([]string, error) {
+	list, err := client.Namespaces().List(labels.Everything(), fields.Everything())
+
+	if err != nil {
+		return nil, err
+	}
+
+	namespaceList := make([]string, 0, 0)
```
bryk: If you're already creating the list manually, you can (should) preallocate its size to len(list.Items).

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -23,9 +23,10 @@ export default class DeployController {
    * @param {!angular.$resource} $resource
    * @param {!angular.$log} $log
    * @param {!ui.router.$state} $state
+   * @param {!Array<!string>} namespaces
```
bryk: Change `!string` to `string`

maciaszczykm:
PTAL

src/app/frontend/deploy/deploy.controller.js
```diff
@@ -45,11 +46,23 @@ export default class DeployController {
     /** @export {!Array<!backendApi.PortMapping>} */
     this.portMappings = [this.newEmptyPortMapping_(this.protocols[0])];
 
+    /**
+     * List of available namespaces.
+     * @export {!Array<!string>}
```
bryk: Remove `!` here too. With that you can also remove `|| []`, it is not needed.

src/app/externs/angularresource.js
```diff
@@ -21,6 +21,14 @@
 
 
 /**
+ * @typedef {{
+ *   $promise: Promise
```
bryk: The promise is technically `!angular.$q.Promise`

bryk:
@maciaszczykm I had one large comment which I didn't submit (accidentally). Wait a few minutes till I rewrite it.

src/app/backend/namespacelist.go
```diff
@@ -0,0 +1,38 @@
...
+package main
+
+import (
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+// Returns a list of all namespaces in the cluster.
+func GetNamespaceList(client *client.Client) ([]string, error) {
```
bryk: You return a list of strings here, instead of a struct instance that has a list inside. GetReplicaSetList api call does the opposite. It returns an object that has a list. This is an inconsistency that we should fix and agree on early in the project.

I am in favor of returning a struct instance here. This is because in client code it is easy to find out who produces this resource, you just need to grep for it. It is also possible to add further properties to the returned struct, which is not possible if you return an array. Think of properties such as "generation", "paginationStart", "timestamp", etc.

See also how Kubernetes code does this. They always return an object that has some auxiliary fields and the array inside. You can also do a research on the internet to check what is recommended. If nothing, then following Kubernetes core is good to me.

What do you think?

bryk:
Last large comment and we're ready to merge. PTAL

src/app/backend/namespacelist.go
maciaszczykm: I agree, I have simplified it too much. Previous version with one structure inside another also was not good. I will do it like you said, JSON object returned by API will contain list of strings (namespaces) and we will be able to add another properties like timestamp etc. in the future.

maciaszczykm:
Updated implementation. PTAL

bryk:
Awesome! Merging.

bryk:
@maciaszczykm One more thing. Replicaset list view does not use resolve mechanism as you do here. Can you migrate it, so that we have consistent code?

maciaszczykm:
@bryk Of course, I will do it.

PR #86, opened 2015-11-19
simplified gulp scripts

cheld:
I made three small changes. These are intended to make working with the scripts more convenient.

1. 'bower install' called implicitly from 'npm install'. (Copied from Angular-Seed project.)

2. added dependency from docker-image to build. This is needed to avoid empty docker image after 'gulp clean'

3. Renamed serve tasks: 'serve:watch' -> 'serve' and 'serve' -> 'serve:nowatch'. Reason: for development serving with watch is default choice. The name for this task should be short and obvious.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,5 +64,8 @@
   },
   "engines": {
     "node": ">=0.10.0"
+  },
+  "scripts": {
+    "postinstall": "./node_modules/.bin/bower install"
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -64,5 +64,8 @@
   },
   "engines": {
     "node": ">=0.10.0"
+  },
+  "scripts": {
+    "postinstall": "bower install"
```
bryk: `bower install` or `./node_modules/.bin/bower install`? I'd go with the latter and remove the requirement for developers to install bower globally. What do you think?

bryk:
Just a comment. PTAL

package.json
cheld: Much better

bryk:
LGTM

PR #87, opened 2015-11-19
Fix js_module_root compiler option

bryk:
Now it is correctly specified and allows for imports anchored at
frontend src, e.g., `import 'chrome/chrome';`.


bryk:
@cheld Can you review?

bryk:
Thanks! Merging

PR #88, opened 2015-11-20
Rename JS files from dot to underscore case

bryk:
This is a required change because with dot case we cannot reference
other modules inside JSDoc `@type` annotations. For example, if you want
to reference `@type {./deploy.controller.FooType}`, it does not work,
because the compiler does not know whether `deploy` is a module or
`deploy.controller` is. Changing dot case to underscore, camel or dash case
solves this problem.

I decided to go with underscore case because this is what golang uses
(e.g., `foo_test.js`) so we'll have higher similarity of backend and
frontend codebases.


bryk:
CC @zreigz @cheld @SebastianM @PeWu @mgr01 @floreks @maciaszczykm

PeWu:
LGTM

bryk:
Merging. If you folks have any outstanding changes, please rebase.

PR #90, opened 2015-11-23
Connect replica set list and detail views

bryk:
This change connects replica set list and detail views and applies
RESTful conventions to the API.

Now API and app URLs follow RESTful API conventions. This change is also
an example of how views hould be connected. See state params and state
name exports.


bryk:
@maciaszczykm Can you review?

maciaszczykm:
LGTM

PR #91, opened 2015-11-23
Bug fix. Prevent for double call isRunning() function

zreigz:


build/cluster.js
```diff
@@ -45,6 +45,13 @@ const kubernetesArchive = 'kubernetes.tar.gz',
  */
 let clusterProcess = null;
 let exec = childProcess.exec;
+/**
+ * A Number, representing the ID value of the timer that is set for function which periodically
+ * checks if cluster is running.
+ *
+ * @type {number}
```
bryk: The type should be `{?number}`, because number is a built in that is is always non null, but you assign null to it. Hence the `?`.

build/cluster.js
```diff
@@ -45,6 +45,13 @@ const kubernetesArchive = 'kubernetes.tar.gz',
  */
 let clusterProcess = null;
 let exec = childProcess.exec;
+/**
```
bryk: Two new lines required in a top level code block.

build/cluster.js
```diff
@@ -45,6 +45,13 @@ const kubernetesArchive = 'kubernetes.tar.gz',
...
+ * A Number, representing the ID value of the timer that is set for function which periodically
+ * checks if cluster is running.
```
bryk: Say that null means that no timer is running.

bryk:
Nice! A few style comments

bryk:
Is this ready for merging? Please remember to comment "PTAL" when you've addressed comments.

zreigz:
PTAL

bryk:
Thank you!

PR #93, opened 2015-11-23
Added labels handling in the deploy form. (View only)

floreks:
It's the frontend part of the labels handling in deploy form.

Current view:
screenshot_4

### DIFF
--- a/src/app/externs/backendapi.js
+++ b/src/app/externs/backendapi.js
@@ -106,3 +106,11 @@ backendApi.ReplicaSetPod;
  * }}
  */
 backendApi.NamespaceList;
+
+/**
+ * @typedef {{
+ *  key: string,
+ *  value: string
+ * }}
+ */
+backendApi.Label;
--- a/src/app/frontend/deploy/deploy_controller.js
+++ b/src/app/frontend/deploy/deploy_controller.js
@@ -16,6 +16,9 @@ import {stateName as zerostate} from 'zerostate/zerostate_state';
 import {stateName as replicasetliststate} from 'replicasetlist/replicasetlist_state';
 
 
+import DeployFormLabel, {APP_LABEL_KEY, VERSION_LABEL_KEY} from './deploy_form_label';
+import DeployLabelService from './deploy_label_service';
+
 /**
  * Controller for the deploy view.
  *
@@ -75,10 +78,20 @@ export default class DeployController {
 
     /** @private {boolean} */
     this.isDeployInProgress_ = false;
+
+    /** @private {!DeployLabelService} */
+    this.deployLabelService_ = new DeployLabelService();
+
+    /** @export {!Array<!DeployFormLabel>} */
+    this.labels = [
+      new DeployFormLabel(APP_LABEL_KEY, '', false),
+      new DeployFormLabel(VERSION_LABEL_KEY, '', false),
+      new DeployFormLabel(),
+    ];
   }
 
   /**
-   * Deploys the application based on the sate of the controller.
+   * Deploys the application based on the state of the controller.
    *
    * @export
    */
@@ -118,7 +131,7 @@ export default class DeployController {
    * @export
    */
   isDeployDisabled() {
-    return this.isDeployInProgress_;
+    return this.isDeployInProgress_ || this.deployLabelService_.hasDuplicationError(this.labels);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


src/app/frontend/deploy/deploy_controller.js
```diff
@@ -115,7 +128,7 @@ export default class DeployController {
    * @export
    */
   isDeployDisabled() {
-    return this.isDeployInProgress_;
+    return this.isDeployInProgress_ || this.deployLabelService_.hasDuplicationError(this.labels);
```
bryk: How about testing form.$valid bool here? I'd catch all future validation errors

src/app/externs/backendapi.js
```diff
@@ -78,3 +78,18 @@ backendApi.ReplicaSet;
  * }}
  */
 backendApi.NamespaceList;
+
+/**
+ * @typedef {{
+ *   labels: !Array<!backendApi.Label>
+ * }}
+ */
+backendApi.LabelsList;
```
bryk: I think you don't need this object; you can safely remove it from this change. Then, when implementing backend, you can add `labels: !Array<!backendApi.Label>` property to `AppDeployment`

bryk:
I'm playing with the UI and have a couple of notes:
1. When you hit enter on the label input, it deletes the row. I dont think this is OK. It should advance to next field or submit entire form. What do you think?
2. The trash icon appears only when I hover over the line. What about showing it when focused on one of the row inputs? This would allow to delete the row using keyboard 
3. There are some issues when I hit enter on an input: predefined labels get deleted.

bryk:
I did a quick look on the change. Will review fully tomorrow.

src/app/externs/backendapi.js
floreks: Good point.

src/app/frontend/deploy/deploy_controller.js
floreks: I can check if this will work. This depends if this custom validation based on label property changes form validation state.

floreks:
Re:

Good catch. :) I haven't actually thought of testing keyboard events as i've only used mouse events. I have to check why it executes delete method after hitting enter.

1. Submitting entire form might be not very intuitive. Advancing to the next field sounds good. I will see how hard it this to implement.
2. Actually that sounds good. With many labels it is easier to use keyboard to enter new one instead of mouse.
3. This has to be related to the first issue. I have to check keyboard events. I guess that it automatically executes button click event for the row and there is only 1 button with delete row action.

src/app/frontend/deploy/deploy_controller.js
bryk: You can hook custom validators to any form: https://docs.angularjs.org/guide/forms "Custom Validation
"

src/app/frontend/deploy/deploy_label_service.js
```diff
@@ -0,0 +1,160 @@
...
+    labelList.push(new DeployFormLabel());
+  }
+
+  /**
+   * Deletes row from labels list.
+   * @param {!Array<!DeployFormLabel>} labelList
+   * @param {!DeployFormLabel} label
+   */
+  deleteLabel(labelList, label) {
+    for (let i = 0; i < labelList.length; i++) {
```
bryk: Use Array.prototype.findIndex and then remove the element from the index. See http://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-array-prototype-object

Such code is less error prone, because you do not have iteration.

src/app/frontend/deploy/deploy_form_label.js
```diff
@@ -0,0 +1,58 @@
...
+    this.editable = editable;
+
+    /** @export {string} */
+    this.key = key;
+
+    /** @export {string} */
+    this.value = value;
+
+    /** @export {boolean} */
+    this.hasError = false;
```
bryk: Is this possible to use standard angular form validation here and remove this property?

src/app/frontend/deploy/deploy_label_service.js
```diff
@@ -0,0 +1,160 @@
...
+    return duplications > 1;
+  }
+
+  /**
+   * Returns true if there is at least 1 label with 'hasError' property set to true.
+   * @param {!Array<!DeployFormLabel>} labelList
+   * @returns {boolean}
+   */
+  hasDuplicationError(labelList) {
+    for (let i = 0; i < labelList.length; i++) {
```
bryk: Use Array.prototype.some or Array.protorype.every for such things. Such code is more idiomatic and easier to read.

bryk:
How about having a quick meeting to talk about this PR? I have a couple of questions about the patterns you've used.

floreks:
Sure. I'm available from 11.

floreks:
I have to apply some changes and this could take some time. I'll open again when it's ready.

bryk:
Thank you!

PR #94, opened 2015-11-23
Initial implementation for replica set detail view

bryk:
The implementation consists of backend code and sample frontend view
that dumps the data into an ugly looking table. Subsequent changes will
implement frontend so that things look pretty.

PR #95, opened 2015-11-24
Add backend support for logs

zreigz:
Retrieving logs for pod added.
For example you can test it like this:

curl -H "Content-Type: application/json" -X POST -d '{"podId":"redis-master-4h22s" "podNAmespace":"default"}' http://localhost:8082/api/logs

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -59,6 +59,15 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(NamespacesList{}))
 	wsContainer.Add(namespaceListWs)
 
+	logsWs := new(restful.WebService)
+	logsWs.Path("/api/logs").
+		Produces(restful.MIME_JSON)
+	logsWs.Route(
+		logsWs.GET("/{namespace}/{podId}").
+			To(apiHandler.handleLogs).
+			Writes(Logs{}))
+	wsContainer.Add(logsWs)
+
 	return wsContainer
 }
 
@@ -119,6 +128,19 @@ func (apiHandler *ApiHandler) handleGetNamespaceList(
 		return
 	}
 
+	response.WriteHeaderAndEntity(http.StatusCreated, result)
+}
+
+// Handles log API call.
+func (apiHandler *ApiHandler) handleLogs(request *restful.Request, response *restful.Response) {
+
+	namespace := request.PathParameter("namespace")
+	podId := request.PathParameter("podId")
+	result, err := GetPodLogs(apiHandler.client, namespace, podId)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
 	response.WriteHeaderAndEntity(http.StatusCreated, result)
 }
 
--- a/src/app/backend/logs.go
+++ b/src/app/backend/logs.go
@@ -1,0 +1,83 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"io/ioutil"
+	api "k8s.io/kubernetes/pkg/api"
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"strings"
+	"time"
+)
+
+// Log response structure
+type Logs struct {
+	// Pod id
+	PodId string `json:"podId"`
+	// Specific date (RFC3339) when logs started
+	SinceTime string `json:"sinceTime"`
+	// Logs string
+	Logs []string `json:"logs"`
+}
+
+// Return logs for particular pod or error when occurred.
+func GetPodLogs(client *client.Client, namespace string, podId string) (*Logs, error) {
+
+	pod, err := client.Pods(namespace).Get(podId)
+	if err != nil {
+		return nil, err
+	}
+
+	logOptions := &api.PodLogOptions{
+		Follow:     false,
+		Previous:   false,
+		Timestamps: true,
+	}
+
+	logString, err := getRawPodLogs(client, namespace, podId, logOptions)
+	if err != nil {
+		return nil, err
+	}
+	logs := &Logs{
+		PodId:     podId,
+		SinceTime: pod.CreationTimestamp.Format(time.RFC3339),
+		Logs:      strings.Split(logString, "\n"),
+	}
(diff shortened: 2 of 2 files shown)


src/app/backend/apihandler.go
```diff
@@ -55,6 +55,17 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(NamespacesList{}))
 	wsContainer.Add(namespaceListWs)
 
+	logsWs := new(restful.WebService)
+	logsWs.Path("/api/logs").
+		Consumes(restful.MIME_JSON).
+		Produces(restful.MIME_JSON)
+	logsWs.Route(
+		logsWs.POST("").
```
bryk: You're getting the list of logs using POST method. In RESTful APIs you get a list of objects using GET methods. There's a pull request that tackles the same problem and it follows RESTful specs. Take a look: https://github.com/kubernetes/dashboard/pull/94/files

So, if you change `logsWs.POST("")` to `logsWs.GET("/:namespace/:pod")` you'll have proper RESTful API. The other good thing about such approach is that this would be very similar to what Kubernetes core does. They have namespace and pod name in path param and all the rest in query params: http://kubernetes.io/third_party/swagger-ui/#!/api%2Fv1/readNamespacedPodLog

What do you think?

zreigz: Good idea. I will change it

src/app/backend/logs.go
```diff
@@ -0,0 +1,101 @@
...
+	// Pod id
+	PodId string `json:"podId"`
+	// Specific date (RFC3339) when logs started
+	SinceTime string `json:"sinceTime"`
+	// Logs string
+	Logs string `json:"logs"`
+}
+
+// Return logs for particular pod or error when occurred.
+func LogApp(logRequest *LogRequest, client *client.Client) (*Logs, error) {
```
bryk: "LogApp"? I don't know what this name means. How about something that says "get pod logs"?

src/app/backend/logs.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Log response structure
+type Logs struct {
+	// Pod id
+	PodId string `json:"podId"`
+	// Specific date (RFC3339) when logs started
+	SinceTime string `json:"sinceTime"`
+	// Logs string
+	Logs string `json:"logs"`
```
bryk: Maybe make this an array of lines? This would make it easier for format on the UI.

bryk:
PTAL

zreigz:
PTAL
Now result looks like this


logscreen

src/app/backend/logs.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		return logs, nil
+	}
+	return nil, err
+}
+
+// Handler method to retrieves a pod logs.
+func handleLog(client *client.Client, namespace string, podID string,
+	logOptions *api.PodLogOptions) (string, error) {
+	req := client.RESTClient.Get().
```
bryk: Why don't you use method `
func (c *pods) GetLogs(name string, opts *api.PodLogOptions) *Request {` from kubernetes/kubernetes/pkg/client/unversioned/pods.go ?

src/app/backend/logs.go
```diff
@@ -0,0 +1,88 @@
...
+			SinceTime: pod.CreationTimestamp.Format(time.RFC3339),
+			Logs:      strings.Split(logString, "\n"),
+		}
+		return logs, nil
+	}
+	return nil, err
+}
+
+// Handler method to retrieves a pod logs.
+func handleLog(client *client.Client, namespace string, podID string,
```
bryk: `handleLog` name does not say much about what this method does. It actually gets the log for the given pod and returs log string. Can you rename it so that the name says what it does?

src/app/backend/logs.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Return logs for particular pod or error when occurred.
+func GetPodLogs(client *client.Client, namespace string, podId string) (*Logs, error) {
+
+	pod, err := client.Pods(namespace).Get(podId)
+	if err != nil {
+		return nil, err
+	}
+
+	limitBytes := int64(1024 * 1024)
```
bryk: Do we need to limit bytes? Why this value? What does it mean?

src/app/backend/logs.go
```diff
@@ -0,0 +1,88 @@
...
+	logOptions := &api.PodLogOptions{
+		Follow:     false,
+		Previous:   false,
+		Timestamps: true,
+		LimitBytes: &limitBytes,
+	}
+
+	if logString, err := handleLog(client, namespace, podId, logOptions); err == nil {
```
bryk: Oh, this is tricky IF statement. What you normally do in Go is that inside this if you `return nil, err` and then outside this IF you have normal program flow. Please refactor, so that it looks like lines (39-42)

bryk:
Nice! PTAL

src/app/backend/logs.go
zreigz: It was added two weeks ego:
https://github.com/kubernetes/kubernetes/pull/10707/files
We don't have it in dashboard resources, but the approach is exactly the same what I did.

src/app/backend/logs.go
zreigz: For now I think we don't need limit. I will remove it

src/app/backend/logs.go
bryk: We can update our dependency to kubernetes code. This is done like with every other Godep dependency. I can do this if you want. What do you think?

And btw, They use `VersionedParams(opts, api.Scheme)` method in that PR. Can you use it too?

zreigz:
PTAL

src/app/backend/logs.go
zreigz: yes I can use VersionedParams, it works. In my opinion you can update dependency to kubernetes code. It always good to use the newest API. Maybe we find some useful new functionality for future work.
So I will wait until you add new files and then update my code.

src/app/backend/logs.go
```diff
@@ -0,0 +1,85 @@
...
+	logs := &Logs{
+		PodId:     podId,
+		SinceTime: pod.CreationTimestamp.Format(time.RFC3339),
+		Logs:      strings.Split(logString, "\n"),
+	}
+	return logs, nil
+}
+
+// Construct a request for getting the logs for a pod.
+// convertLogRequestToString(req)
```
bryk: Hmm... Is is a leftover from your tests? This line does not make sense. Remove?

src/app/backend/logs.go
```diff
@@ -0,0 +1,85 @@
...
+func requestForLogs(client *client.Client, namespace string, podID string,
```
bryk: This name still needs iteration. How about `getRawPodLogs`?

bryk:
PTAL

src/app/backend/logs.go
zreigz: Done

src/app/backend/logs.go
zreigz: Done

bryk:
I'm merging this. Including new version of Kubernetes golang dependency is broken at the moment.

zreigz:
OK. Thank you.

PR #96, opened 2015-11-24
Enforce clang-format formatting on all JS files

bryk:
This change reformats all files according to clang-format rules (the
same that Angular JS has) and forces the files to comply with the
formatting.


SebastianM:
Added one comment. Everything else LGTM.

bryk:
Thank you for the review. Merging, so that others can rebase.

Issue #66, opened 2015-11-12
Establish code styles and practices for the project 


bryk:
As the number of contributors to the project grows, we need to establish coding styles and practices so that the codebase stays unified and healthy for long time.

The areas that should be covered include (at least):
* how to create a view (what's the code layout, what to place where)
* what to do on the backend and what on the frontend
* how to communicate with backend
* what is the style of JS code (e.g., formatting, JSDoc annotations, patterns to use)
* what is the style of Go code (should be easy, if we follow gofmt)
* other trivia like: tabs/spaces, line width, comments 

Any other ideas?

cheld:
* Supported form factors
* Supported browsers
* Unit test coverage in backend and UI

cheld:
We could enhance eslint and/or check-in formating rules for WebStorm and Atom. What is your opinion?

bryk:
Check in the rules into the repository? Do you have idea whether this is anything generic or just for WebStrorm?

cheld:
I have never configured eslint myself, but some trivial things could be configured as warnings, e.g. line-width, camelcase, spacing,... These warnings can be shown in editors like webstorm or atom. So, this would be generic. However, my experience with eslint is limited. I cannot judge if it makes sense or not.

SebastianM:
My $0.02:

* How to create a view: I think we should create a "boilerplate" view (that is not visible of course) that can be used to create new views. It should be up to date all the time and should be well documented.
* how to communicate with backend: Yeah, this is really important (especially REST API design). We should document REST API patterns that get used in the project. I think it's also super helpful to document all REST related API with Swagger (like in Kubernetes but with Swagger v2)
* what is the style of JS code:
  * JSDoc
  * ESLint rules (with failing CI builds when rules are violated)
  * AngularJS patterns that make the migration to Angular 2 easy (beta is coming soon)
  * Clang formatted code (with Google style rules), it's also used by the [Angular team](https://github.com/angular/angular/blob/master/.clang-format). There's also a gulp plugin that can do format checking **and formatting** (ESLint plugins have very limited support to fix style errors)
* I think an [.editorconfig](https://github.com/angular/angular/blob/master/.editorconfig) is also a good idea.
* other trivia like: tabs/spaces, line width, comments: should be covered with ESLint and clang.
* Go code:
  * gofmt
  * [golint](https://github.com/golang/lint)
* [Greenkeeper](http://greenkeeper.io/) for NPM packages.
[...]

PR #98, opened 2015-11-24
Add more ESLint rules for JavaScript linting

bryk:
With this change I've gone through all ESLint rules at:
http://eslint.org/docs/rules/ and added all of them that I feel make
sense to the project.

This will allow us to focus on business logic of our code, rather than
on style discussions, which are now better guarded thanks to ESLint and
clang-format.

On the final note: it does not matter what concrete lint rules are
enforced (e.g., semicolons or nor). The important part is that they are
enforced by a tool and we do not spend time dicussing about them.

Addresses: #66

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,25 +17,59 @@
     "modules": true,
   },
   "rules": {
+    // Require parens in arrow function arguments.
+    "arrow-parens": [2, "always"],
     // Require dangling comma at the end of any  multiline list/object. This is to be consistent
     // with backend code.
     "comma-dangle": [2, "always-multiline"],
+    // Enforce newline consistency in member expressions. Dots are places on the same line as
+    // the property.
+    "dot-location": [2, "property"],
+    // Enforce newline at the end of file, with no multiple empty lines.
+    "eol-last": 2,
     // Force using === and !=== except when comparing to null. This is to prevent from bugs that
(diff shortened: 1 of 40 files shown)


bryk:
@SebastianM @cheld Can you take a look?

cheld:
LGTM. I suggest to merge in order to reduce conflicts.

@taimir 

You spent a bit of time collecting common review comments and converting to lint rules. Can you have a look?

BTW: I think the most common problem not covered is alphabetical ordering. However, no rule exists for that.

cheld:
As far as I remember, clean code principles recommend no empty lines inside methods and 2 empty lines between methods - code is easier to read.

bryk:
> As far as I remember, clean code principles recommend no empty lines inside methods and 2 empty lines between methods - code is easier to read.

Yes, but no ESLint nor clang-format can enforce such formatting. On the other hand, one newline can be enforced by ESLint, because people tend to separate code blocks that need to be separated by at least one newline.

I could include gjslint linter to enforce newlines between code blocks, but this would be 3rd JS style tool we have, which becomes dangerous. On the other hand, 1 newline between code blocks is what golang has, so frontend would be more aligned with the backend.

bryk:
> BTW: I think the most common problem not covered is alphabetical ordering. However, no rule exists for that.

Yes, I know this is an unsolved problem.

PR #100, opened 2015-11-26
Implemented create new namespace dialog

maciaszczykm:
Connected to #73 issue. Added:

- frontend dialog to create new namespaces
- backend API to create new namespaces
- basic validation to disable creation of duplicated namespaces

Few details may still be missing or need upgrade, please let me know if you see any area, that need improvement.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -33,8 +33,8 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 	deployWs.Route(
 		deployWs.POST("").
 			To(apiHandler.handleDeploy).
-			Reads(AppDeployment{}).
-			Writes(AppDeployment{}))
+			Reads(AppDeploymentSpec{}).
+			Writes(AppDeploymentSpec{}))
 	wsContainer.Add(deployWs)
 
 	replicaSetWs := new(restful.WebService)
@@ -50,14 +50,20 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(ReplicaSetDetail{}))
 	wsContainer.Add(replicaSetWs)
 
-	namespaceListWs := new(restful.WebService)
-	namespaceListWs.Path("/api/namespaces").
+	namespacesWs := new(restful.WebService)
+	namespacesWs.Path("/api/namespaces").
+		Consumes(restful.MIME_JSON).
 		Produces(restful.MIME_JSON)
-	namespaceListWs.Route(
-		namespaceListWs.GET("").
-			To(apiHandler.handleGetNamespaceList).
+	namespacesWs.Route(
+		namespacesWs.POST("").
+			To(apiHandler.handleCreateNamespace).
+			Reads(NamespaceSpec{}).
+			Writes(NamespaceSpec{}))
+	namespacesWs.Route(
+		namespacesWs.GET("").
+			To(apiHandler.handleGetNamespaces).
 			Writes(NamespacesList{}))
-	wsContainer.Add(namespaceListWs)
+	wsContainer.Add(namespacesWs)
 
 	logsWs := new(restful.WebService)
 	logsWs.Path("/api/logs").
@@ -77,17 +83,17 @@ type ApiHandler struct {
 
 // Handles deploy API call.
 func (apiHandler *ApiHandler) handleDeploy(request *restful.Request, response *restful.Response) {
-	cfg := new(AppDeployment)
-	if err := request.ReadEntity(cfg); err != nil {
+	appDeploymentSpec := new(AppDeploymentSpec)
+	if err := request.ReadEntity(appDeploymentSpec); err != nil {
 		handleInternalError(response, err)
 		return
 	}
-	if err := DeployApp(cfg, apiHandler.client); err != nil {
+	if err := DeployApp(appDeploymentSpec, apiHandler.client); err != nil {
 		handleInternalError(response, err)
 		return
 	}
 
-	response.WriteHeaderAndEntity(http.StatusCreated, cfg)
+	response.WriteHeaderAndEntity(http.StatusCreated, appDeploymentSpec)
 }
 
 // Handles get Replica Set list API call.
@@ -118,8 +124,24 @@ func (apiHandler *ApiHandler) handleGetReplicaSetDetail(
 	response.WriteHeaderAndEntity(http.StatusCreated, result)
 }
 
+// Handles namespace creation API call.
+func (apiHandler *ApiHandler) handleCreateNamespace(request *restful.Request,
+	response *restful.Response) {
+	namespaceSpec := new(NamespaceSpec)
+	if err := request.ReadEntity(namespaceSpec); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+	if err := CreateNamespace(namespaceSpec, apiHandler.client); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	response.WriteHeaderAndEntity(http.StatusCreated, namespaceSpec)
+}
+
 // Handles get namespace list API call.
-func (apiHandler *ApiHandler) handleGetNamespaceList(
+func (apiHandler *ApiHandler) handleGetNamespaces(
 	request *restful.Request, response *restful.Response) {
 
 	result, err := GetNamespaceList(apiHandler.client)
--- a/src/app/externs/backendapi.js
+++ b/src/app/externs/backendapi.js
@@ -43,7 +43,7 @@ backendApi.PortMapping;
  *   namespace: string
  * }}
  */
-backendApi.AppDeployment;
+backendApi.AppDeploymentSpec;
 
 /**
  * @typedef {{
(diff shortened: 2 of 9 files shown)


src/app/backend/apihandler.go
```diff
@@ -50,6 +50,17 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(ReplicaSetDetail{}))
 	wsContainer.Add(replicaSetWs)
 
+	namespaceCreationWs := new(restful.WebService)
+	namespaceCreationWs.Path("/api/namespacecreation").
+		Consumes(restful.MIME_JSON).
+		Produces(restful.MIME_JSON)
+	namespaceCreationWs.Route(
+		namespaceCreationWs.POST("").
```
bryk: To create a namespace you should use existing web service on path "/api/namespaces" and handle POST method there. This is how RESTful apis should work. Initial reading on this: https://en.wikipedia.org/wiki/Representational_state_transfer

src/app/backend/apihandler.go
```diff
@@ -90,6 +101,22 @@ func (apiHandler *ApiHandler) handleDeploy(request *restful.Request, response *r
 	response.WriteHeaderAndEntity(http.StatusCreated, cfg)
 }
 
+// Handles namespace creation API call.
+func (apiHandler *ApiHandler) handleNamespaceCreation(request *restful.Request,
+	response *restful.Response) {
+	cfg := new(NamespaceSpec)
```
bryk: This variable should not be named `cfg`, it does not describe what it really is. Maybe `namespaceSpec`? I see that line 91 has the same issue (probably a leftover). Can you fix both?

src/app/externs/backendapi.js
```diff
@@ -94,6 +94,13 @@ backendApi.ReplicaSetPod;
 
 /**
  * @typedef {{
+ *   name: string
+ * }}
+ */
+backendApi.NamespaceSpec;
```
bryk: You've defined a nice pattern here. The `Spec` suffix. I like it for POST REST handlers. Can you also rename `AppDeployment` to `AppDeploymentSpec`? This will make the code consistent.

src/app/frontend/deploy/namespace_dialog_controller.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {!angular.Scope} $scope
+ * @param {!md.$dialog} $mdDialog
+ * @param {!angular.$log} $log
+ * @param {!angular.$resource} $resource
+ * @param {!Array<string>} namespaces
+ * @constructor
+ */
+export function NamespaceDialogController($scope, $mdDialog, $log, $resource, namespaces) {
+  /** @export {string} */
+  $scope.namespace = '';
```
bryk: Do not use `$scope`. bindToController instead. See documentation for howto.

src/app/frontend/deploy/namespace_dialog_controller.js
```diff
@@ -0,0 +1,62 @@
...
+ * Namespace creation dialog controller.
+ *
+ * @param {!angular.Scope} $scope
+ * @param {!md.$dialog} $mdDialog
+ * @param {!angular.$log} $log
+ * @param {!angular.$resource} $resource
+ * @param {!Array<string>} namespaces
+ * @constructor
+ */
+export function NamespaceDialogController($scope, $mdDialog, $log, $resource, namespaces) {
```
bryk: Make this a ES6 class and export default: `export default class ...`

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -112,6 +117,34 @@ export default class DeployController {
   }
 
   /**
+   * Displays new namespace creation dialog.
+   *
+   * @param {!angular.Scope.Event} event
+   */
+  showNamespaceDialog(event) {
+    this.mdDialog_.show({
```
bryk: How about moving this object to `namespacfoo_dialog.js` and then `export default` it. This would follow the practice we have now with directives. That template, controller, and definition objects are in separate files.

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -112,6 +117,34 @@ export default class DeployController {
...
+                    controllerAs: 'ctrl',
+                    clickOutsideToClose: true,
+                    targetEvent: event,
+                    templateUrl: '/deploy/namespace_dialog.html',
+                    locals: {
+                      namespaces: this.namespaces,
+                    },
+                  })
+        .then(
+            function(answer) {
```
bryk: What is the type of `answer`? This is non trivial, can you annotate the function?

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -112,6 +117,34 @@ export default class DeployController {
...
+              if (answer) {
+                this.namespace = answer;
+                this.namespaces = this.namespaces.concat(answer);
+              } else {
+                this.namespace = this.namespaces[0];
+              }
+            }.bind(this),
```
bryk: Use fat arrow instead of bind(this). I.e., `(answer) => {}`.

src/app/frontend/deploy/namespace_dialog_controller.js
```diff
@@ -0,0 +1,62 @@
...
+// limitations under the License.
+
+/**
+ * Namespace creation dialog controller.
+ *
+ * @param {!angular.Scope} $scope
+ * @param {!md.$dialog} $mdDialog
+ * @param {!angular.$log} $log
+ * @param {!angular.$resource} $resource
+ * @param {!Array<string>} namespaces
```
bryk: Missing `@ngInject`

bryk:
PTAL. Very nice change. Did you test it on `:prod` environment?

src/app/backend/apihandler.go
maciaszczykm: Yes, totally forgot about second API. Updated implementation.

src/app/backend/apihandler.go
maciaszczykm: Sure, done.

src/app/externs/backendapi.js
maciaszczykm: Done.

src/app/frontend/deploy/namespace_dialog_controller.js
maciaszczykm: Added.

src/app/frontend/deploy/deploy_controller.js
maciaszczykm: Done.

src/app/frontend/deploy/deploy_controller.js
maciaszczykm: Added annotations.

src/app/frontend/deploy/namespace_dialog.html
```diff
@@ -0,0 +1,33 @@
+<!--
```
bryk: Re naming of the files. The general naming idea is to to have names with no separators and then suffix it with functional or angular suffixes. So we have `_test.js`, `_po.js` or `_controller.js`. How about naming this thing a "create namespace" and have files: `createnamespace.js` `createnamespace_controller.js` `createnamespace.html` and `createnamespace_dialog.js`. This is promoting `_dialog.js` to a set of suffixes we can use for angular constructs.

What do you think?

bryk:
One more comment. PTAL

src/app/frontend/deploy/namespace_dialog_controller.js
maciaszczykm: Done.

src/app/frontend/deploy/namespace_dialog.html
maciaszczykm: Completely agree, now it looks a look better than when it was `namespace_dialog`.

bryk:
Remember to comment "PTAL" when ready for rereview.

src/app/frontend/deploy/namespace_dialog_controller.js
maciaszczykm: `$scope` doesn't need to be used since controller is a class, at the moment I'm using `this` and there is no need to bind I think, please correct me if I'm wrong.

src/app/frontend/deploy/deploy_controller.js
maciaszczykm: Done.

maciaszczykm:
PTAL

src/app/frontend/deploy/createnamespace_dialog.js
```diff
@@ -0,0 +1,37 @@
...
+
+import NamespaceDialogController from 'deploy/createnamespace_controller';
+
+/**
+ * Displays new namespace creation dialog.
+ *
+ * @param {!md.$dialog} mdDialog
+ * @param {!angular.Scope.Event} event
+ * @param {!Array<string>} namespaces
+ *
```
bryk: No need for newline here :)

bryk:
One tiny-tiny comment and we're merging.

maciaszczykm:
Done :)

PR #101, opened 2015-11-26
Deploy using settings

batikanu:
This is the first draft of deploy using settings directive. In deploy.html there will be two directives like "deploy-using-settings" and "deploy-using-file". According to the selection of the radio button in deploy view the selected directive will be loaded in the page. In current state only deploy-using-settings directive is implemented and related other code blocks are refactored.

### DIFF
--- a/src/app/frontend/deploy/deploy.html
+++ b/src/app/frontend/deploy/deploy.html
@@ -17,8 +17,8 @@ limitations under the License.
 <div layout="column" layout-padding layout-align="center center">
   <md-whiteframe class="kd-deploy-whiteframe md-whiteframe-5dp" flex flex-gt-md>
     <h3 class="md-headline">Deploy a Containerized App</h3>
-    <form ng-submit="ctrl.deploy()">
-      <md-input-container>
+    <form ng-submit="ctrl.deployBySelection()">
+      <md-input-container class="md-block">
         <label>App name</label>
         <input ng-model="ctrl.name" required>
       </md-input-container>
@@ -30,50 +30,21 @@ limitations under the License.
           Upload a YAML or JSON file
         </md-radio-button>
       </md-radio-group>
-      <md-input-container>
-        <label>Container image</label>
-        <input ng-model="ctrl.containerImage" required>
-      </md-input-container>
-      <md-input-container>
-        <label>Number of pods</label>
-        <input ng-model="ctrl.replicas" type="number" required min="1">
-      </md-input-container>
-      <div layout="row" ng-repeat="portMapping in ctrl.portMappings">
-        <md-input-container flex>
-          <label>Port</label>
-          <input ng-model="portMapping.port" type="number" min="0">
-        </md-input-container>
-        <md-input-container flex>
-          <label>Target port</label>
-          <input ng-model="portMapping.targetPort" type="number" min="0">
-        </md-input-container>
-        <md-input-container flex="none">
-          <label>Protocol</label>
-          <md-select ng-model="portMapping.protocol" required>
-            <md-option ng-repeat="protocol in ctrl.protocols" ng-value="protocol">
-              {{protocol}}
-            </md-option>
-          </md-select>
-        </md-input-container>
-      </div>
-      <md-input-container>
-        <label>Namespace</label>
-        <md-select ng-model="ctrl.namespace" required>
-          <md-option ng-repeat="namespace in ctrl.namespaces" ng-value="namespace">
-            {{namespace}}
-          </md-option>
-          <md-option ng-click="ctrl.handleNamespaceDialog($event)">
-            Create a new namespace...
-          </md-option>
-        </md-select>
-      </md-input-container>
-      <md-switch ng-model="ctrl.isExternal" class="md-primary">
-        Expose service externally
-      </md-switch>
-      <md-input-container>
-        <label>Description (optional)</label>
-        <textarea ng-model="ctrl.description"></textarea>
-      </md-input-container>
+      <deploy-from-settings name="ctrl.name"
+                            namespace="ctrl.namespace"
+                            detail="ctrl.detail">
+      </deploy-from-settings>
+      <md-input-container class="md-block">
+      <label>Namespace</label>
+      <md-select ng-model="ctrl.namespace" required>
+        <md-option ng-repeat="namespace in ctrl.namespaces" ng-value="namespace">
+          {{namespace}}
+        </md-option>
+        <md-option ng-click="ctrl.handleNamespaceDialog($event)">
+          Create a new namespace...
+        </md-option>
+      </md-select>
+    </md-input-container>
       <md-button class="md-raised md-primary" type="submit" ng-disabled="ctrl.isDeployDisabled()">
         Deploy
       </md-button>
--- a/src/app/frontend/deploy/deploy_controller.js
+++ b/src/app/frontend/deploy/deploy_controller.js
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import {stateName as zerostate} from 'zerostate/zerostate_state';
-import {stateName as replicasetliststate} from 'replicasetlist/replicasetlist_state';
 import showNamespaceDialog from 'deploy/createnamespace_dialog';
 
 /**
@@ -34,25 +33,6 @@ export default class DeployController {
     /** @export {string} */
     this.name = '';
 
-    /** @export {string} */
-    this.containerImage = '';
-
-    /** @export {number} */
-    this.replicas = 1;
-
-    /** @export {string} */
-    this.description = '';
-
-    /**
-     * List of supported protocols.
-     * TODO(bryk): Do not hardcode the here, move to backend.
-     * @const @export {!Array<string>}
-     */
-    this.protocols = ['TCP', 'UDP'];
-
-    /** @export {!Array<!backendApi.PortMapping>} */
-    this.portMappings = [this.newEmptyPortMapping_(this.protocols[0])];
-
     /**
      * List of available namespaces.
      * @export {!Array<string>}
@@ -64,9 +44,6 @@ export default class DeployController {
      * @export {(string|undefined)}
      */
     this.namespace = this.namespaces.length > 0 ? this.namespaces[0] : undefined;
-
-    /** @export {boolean} */
-    this.isExternal = false;
 
     /** @private {!angular.$resource} */
     this.resource_ = $resource;
@@ -82,42 +59,23 @@ export default class DeployController {
 
     /** @private {boolean} */
     this.isDeployInProgress_ = false;
+
+    /**
+     * Contains the selected directive's controller which has its own deploy logic
+     *
+     * @export {{deploy:function()}}
+     */
+    this.detail = {deploy: function() {}};
   }
 
   /**
-   * Deploys the application based on the sate of the controller.
+   * Notifies the child scopes to call their deploy methods.
    *
    * @export
    */
-  deploy() {
-    // TODO(bryk): Validate input data before sending to the server.
-
-    /** @type {!backendApi.AppDeploymentSpec} */
-    let appDeploymentSpec = {
-      containerImage: this.containerImage,
-      isExternal: this.isExternal,
-      name: this.name,
-      description: this.description,
-      portMappings: this.portMappings.filter(this.isPortMappingEmpty_),
-      replicas: this.replicas,
-      namespace: this.namespace,
-    };
-
-    /** @type {!angular.Resource<!backendApi.AppDeploymentSpec>} */
-    let resource = this.resource_('/api/appdeployments');
-
+  deployBySelection() {
     this.isDeployInProgress_ = true;
-    resource.save(
-        appDeploymentSpec,
-        (savedConfig) => {
-          this.isDeployInProgress_ = false;
-          this.log_.info('Successfully deployed application: ', savedConfig);
-          this.state_.go(replicasetliststate);
-        },
-        (err) => {
-          this.isDeployInProgress_ = false;
-          this.log_.error('Error deploying application:', err);
-        });
+    this.detail.deploy().finally(() => { this.isDeployInProgress_ = false; });
   }
 
   /**
@@ -157,25 +115,4 @@ export default class DeployController {
    * @export
    */
   cancel() { this.state_.go(zerostate); }
-
-  /**
-   * @param {string} defaultProtocol
-   * @return {!backendApi.PortMapping}
-   * @private
-   */
-  newEmptyPortMapping_(defaultProtocol) {
-    return {
-      port: null,
-      targetPort: null,
-      protocol: defaultProtocol,
-    };
-  }
-
-  /**
-   * Returns true when the given port mapping hasn't been filled by the user, i.e., is empty.
-   * @param {!backendApi.PortMapping} portMapping
-   * @return {boolean}
-   * @private
-   */
-  isPortMappingEmpty_(portMapping) { return !!portMapping.port && !!portMapping.targetPort; }
 }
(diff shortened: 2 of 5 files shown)


src/app/frontend/deploy/deploy_settings_controller.js
```diff
@@ -0,0 +1,132 @@
...
+ */
+export default class DeploySettingsController {
+  /**
+   * @param {!angular.Scope} $scope
+   * @param {!angular.$log} $log
+   * @param {!ui.router.$state} $state
+   * @param {!angular.$resource} $resource
+   * @ngInject
+   */
+  constructor($scope, $log, $state, $resource) {
```
bryk: Quick comment: (almost) never use $scope, unless you're doing something that is application wide. All directives and views should have `bindToController=true` and `controllerAs:'ctrl'`.

Instead of communicating by events you can use plain old directive scope parameters. Communicating by events is hard to read and understand, so avoid it when you can and opt for explicit callbacks. E.g., make this directive look like this (deploy.html):
```
<deploy-from-settings deploy="ctrl.deployFn" />

Then the deploy_controller.js would have
/** @export {function():angular.$q.Promise} */
this.deployFn 

The deployFn would return a deploy promise. With the promise you can update this.isDeployInProgress_ easily. 
```

Does this make sense? What do you think?

batikanu: I tried using <deploy-from-settings deploy="ctrl.deployFn" /> but when deploy is with deploy button in parent scope triggered, this function can't access to the attributes of the DeploySettingsController class.

bryk: > In my case, I assigned the implementation of the deploy function in deploy-from-settings directive to the parent's deploy function.

This happened because you had name collision. You can easily fix this. Just use unique names for internal implementations and things that get passed by parameters.

batikanu: I have still the same problem. I think something is misunderstood. I created an example of the situation. http://plnkr.co/edit/GQtVQHAUMx4pUgLyBBSq?p=preview
In this example if you just assign the deploy function of parent scope in the deploy from settings directive, it can't access the containerImage attribute of the Ctrl class of deploy from settings  directive. 
I created 3 buttons for 3 cases. 
* deploySettings button: deploy function is called inside deploy from settings directive
* deploy button: deploy function of deploy from settings directive is called from parent scope
* deploy from instance button: deploy function of deploy from settings directive is called from parent scope by using the instance of child controller. 
I can't see here name collision.
What do you think?

bryk: I see. Execution scope of the function is $scope when you pass it by angular param. The solution with "deploy from instance button" sounds most reasonable. The controller of publish from settings publishes itself on the scope param and then the general deploy controller calls its method deploy.

batikanu: @bryk I made the changes after your feedback. Do you see any other improvements?

bryk:
Remember to comment 'PTAL' if ready for rereview.

batikanu:
PTAL

src/app/frontend/deploy/publishfromsettings_controller.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Deploys the application based on the state of the controller.
+   *
+   * @export
+   * @return {angular.$q.Promise}
+   */
+  deploy() {
+    let d = this.q_.defer();
```
bryk: Can you move this variable as far to the bottom of this method as possible? It is a good programming practice to have var initialization as close to its use as possible.

Also, re name: how about naming it deferred? We don't need to save a few keystrokes and `d` means little

src/app/frontend/deploy/publishfromsettings_controller.js
```diff
@@ -0,0 +1,119 @@
...
+export default class PublishFromSettingsController {
+  /**
+   * @param {!angular.$log} $log
+   * @param {!ui.router.$state} $state
+   * @param {!angular.$resource} $resource
+   * @param {!angular.$q} $q
+   * @ngInject
+   */
+  constructor($log, $state, $resource, $q) {
+    /** @export {Object} */
```
bryk: The type is `!PublishFromSettingsController`, not Object :)

bryk: Can you add a note that this initializes output scope param?

src/app/frontend/deploy/publishfromsettings_controller.js
```diff
@@ -0,0 +1,119 @@
...
+      replicas: this.replicas,
+      namespace: this.namespace,
+    };
+
+    /** @type {!angular.Resource<!backendApi.AppDeploymentSpec>} */
+    let resource = this.resource_('/api/appdeployments');
+    resource.save(
+        appDeploymentSpec,
+        (savedConfig) => {
+          d.resolve(false);  // Progress ends
```
bryk: Why not resolve with `savedConfig`? This is what you typically do with promises. Resolve with the object that has been saved.

src/app/frontend/deploy/publishfromsettings_controller.js
```diff
@@ -0,0 +1,119 @@
...
+          this.log_.info('Successfully deployed application: ', savedConfig);
+          this.state_.go(replicasetliststate);
+        },
+        (err) => {
+          d.resolve(false);  // Progress ends
```
bryk: Why not reject with `err`? With that the caller can do something with the error.

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -77,43 +57,22 @@ export default class DeployController {
...
-          this.state_.go(replicasetliststate);
-        },
-        (err) => {
-          this.isDeployInProgress_ = false;
-          this.log_.error('Error deploying application:', err);
-        });
+  deployBySelection() {
+    this.isDeployInProgress = true;
+    this.publisher.deploy().then(
+        ((deployCtrl) => ((status) => deployCtrl.isDeployInProgress = status))(this));
```
bryk: I seriously cannot read this line :) What is happening here? How about writing:

```
published.deploy()
    .then( (...) => { ..} )
    .finally( () => { deployCtrl.isDeployInProgress = false} ). 
```

src/app/frontend/deploy/publishfromsettings_directive.js
```diff
@@ -0,0 +1,33 @@
...
+import PublishFromSettingsController from './publishfromsettings_controller';
+
+/**
+ * Returns directive definition object for the publish from settings directive.
+ * @return {!angular.Directive}
+ */
+export default function publishFromSettingsDirective() {
+  return {
+    scope: {
+      namespace: '=deployNamespace',
```
bryk: Why not just `'='`? It is best to keep scope param names the same as controller's properties. Such code is easier to read.

And btw, you should move this to `bindToController: { ... }`. It is more idiomatic.

src/app/frontend/deploy/publishfromsettings_directive.js
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import PublishFromSettingsController from './publishfromsettings_controller';
+
+/**
+ * Returns directive definition object for the publish from settings directive.
+ * @return {!angular.Directive}
+ */
+export default function publishFromSettingsDirective() {
```
bryk: Re naming: why did you decide to go with `publish` instead of `deploy`?

bryk:
This looks nice! I've got a few style comments.

src/app/frontend/deploy/publishfromsettings_directive.js
batikanu: I actually changed it after your comment on 27th November "The controller of publish from settings publishes itself on the scope param and then the general deploy controller calls its method deploy." Should I change it back to deploy?

bryk: Oh, I'm sorry for my misleading comment. I was commenting about how to lay out directives and components, not about the name which was good :) Somehow I wrote 'publish from settings', perhaps because of the 'publishes itself' part.

src/app/frontend/deploy/deploy_controller.js
batikanu: This is equal to the code below.
```
this.publisher.deploy().then(
      (function(deployCtrl) {
        return function(status){
          deployCtrl.isDeployInProgress = status;
        };
      })(this));
```
It decorates the calllback function with the controller instance because it is modified in callback.
I may solve it in this way as well.
```
let deployCtrl = this;
this.publisher.deploy()
      .then(() => { deployCtrl.setDeployInProgress_(false); })
      .error(() => { deployCtrl.setDeployInProgress_(false); });
```

batikanu: or like that because in both cases deploy button has to be activated and in error case there is no required behavior defined yet. 
```
let deployCtrl = this;
this.publisher.deploy()
      .finally(() => { deployCtrl.setDeployInProgress_(false); })
```

bryk: Yeah, the `finally` approach looks more readable to me. This is what finally is designed to do anyway :)

src/app/frontend/deploy/publishfromsettings_controller.js
batikanu: done

src/app/frontend/deploy/publishfromsettings_controller.js
batikanu: done

src/app/frontend/deploy/publishfromsettings_controller.js
batikanu: done

src/app/frontend/deploy/publishfromsettings_controller.js
batikanu: done

src/app/frontend/deploy/deploy_controller.js
batikanu: done

src/app/frontend/deploy/publishfromsettings_directive.js
batikanu: done

batikanu:
PTAL

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -80,44 +57,22 @@ export default class DeployController {
     /** @private {!md.$dialog} */
     this.mdDialog_ = $mdDialog;
 
-    /** @private {boolean} */
-    this.isDeployInProgress_ = false;
+    /** @export {boolean}*/
+    this.isDeployInProgress = false;
+
+    /** @export {!Object}*/
```
bryk: The type of this should be "an object that has deploy() function". Right? This can be written as `{{deploy:function()}}`. Also, add a quick comment saying what this var is.

src/app/frontend/deploy/deploy.html
```diff
@@ -30,51 +30,22 @@ <h3 class="md-headline">Deploy a Containerized App</h3>
...
+      <md-select ng-model="ctrl.namespace" required>
+        <md-option ng-repeat="namespace in ctrl.namespaces" ng-value="namespace">
+          {{namespace}}
+        </md-option>
+        <md-option ng-click="ctrl.handleNamespaceDialog($event)">
+          Create a new namespace...
+        </md-option>
+      </md-select>
+    </md-input-container>
+      <md-button class="md-raised md-primary" type="submit" ng-disabled="ctrl.isDeployInProgress">
```
bryk: There used to be isDeployDisabled method. Can you still use it and make isDeployInProgress_ `@private`? The reason is that @floreks has a pull request that adds additional logic to this method (checks for form validity).

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -80,44 +57,22 @@ export default class DeployController {
...
-          this.state_.go(replicasetliststate);
-        },
-        (err) => {
-          this.isDeployInProgress_ = false;
-          this.log_.error('Error deploying application:', err);
-        });
+  deployBySelection() {
+    let deployCtrl = this;
+    deployCtrl.isDeployInProgress = true;
+    deployCtrl.detail.deploy().finally(() => { deployCtrl.isDeployInProgress = false; });
```
bryk: Use `this.isDeployInProgress` instead of `var foo = this`. Fat arrows have lexical scoping, so this just works. This projects generally uses fat arrows whenever possible and not `var self=this`. The reason is consistency across entire codebase for easier reading.

src/app/frontend/deploy/deployfromsettings_controller.js
```diff
@@ -0,0 +1,124 @@
...
+   * @param {!angular.$log} $log
+   * @param {!ui.router.$state} $state
+   * @param {!angular.$resource} $resource
+   * @param {!angular.$q} $q
+   * @ngInject
+   */
+  constructor($log, $state, $resource, $q) {
+    /**
+     * @export {!DeployFromSettingsController}
+     * It initializes the scope output parameter*/
```
bryk: Add new line. And move `@export` to the end of the comment. This is to have all comments look the same.

src/app/frontend/deploy/deployfromsettings_directive.js
```diff
@@ -0,0 +1,33 @@
...
+
+import DeployFromSettingsController from './deployfromsettings_controller';
+
+/**
+ * Returns directive definition object for the deploy from settings directive.
+ * @return {!angular.Directive}
+ */
+export default function deployFromSettingsDirective() {
+  return {
+    scope: {},
```
bryk: This is not needed.

bryk:
This looks nice now Batikan! I've got a few last style comments and we're merging.

src/app/frontend/deploy/deployfromsettings_directive.js
batikanu: Case1: If I don't use scope at all: It means `scope: false`
* Then I can't use `controllerAs: 'ctrl'`  in child directive. Because ctrl is already taken by parent controlller. If I use it anyway, then the ctrl from child directive overwrites the one from parent. The child ctrl should be %100 extending the parent ctrl to make it work even if it overwrites.  
* If I use different name like 'ctrlSettings' then directive-parameter binding doesn't work.

Case2: If I use `scope: true`

* Then still I can't use `controllerAs: 'ctrl'` because it creates a collision with parent ctrl and two ways directive-parameter binding doesn't work. I need to name it different again.

Case3: If I use `scope:{}` 

* Then I am free to use `controllerAs: 'ctrl'` and scope binding works fine. Nothing collides.

floreks: Maybe you could name it differently f.e. `controllerAs: 'deploySettingsCtrl'`. This way it won't override `ctrl`.

batikanu: In the previous implementation I named it differently but changed back to ctrl because it was requested to name it as ctrl only.

bryk: Oh, I was wrong. This is needed. I somehow assumed that when you have bindToController: {...} it is isolate scope by default. This is needed, sorry for confusion.

src/app/frontend/deploy/deploy_controller.js
batikanu: done

src/app/frontend/deploy/deploy.html
batikanu: done

src/app/frontend/deploy/deploy_controller.js
batikanu: done

src/app/frontend/deploy/deployfromsettings_controller.js
batikanu: done

src/app/frontend/deploy/deployfromsettings_directive.js
batikanu: No problem. I keep it as it is.

batikanu:
PTAL

PR #79, opened 2015-11-18
Add travis configuration to run local kubernetes cluster

zreigz:

PR #103, opened 2015-11-27
Resolved issues with the automatic cluster creation

taimir:
PTAL, this is a suggestion on how we could resove some of the problems that were discussed in:
https://github.com/kubernetes/dashboard/pull/79

@bryk I intentionally did not implement a "sudo" check when the local-up-cluster task is run. This could be done easily as I have already tried it out, but I am not sure if this is a hard requirement. What are your thoughts on this?

### DIFF
--- a/build/check.js
+++ b/build/check.js
@@ -23,12 +23,21 @@ import path from 'path';
 import conf from './conf';
 
 /**
- * Checks whether codebase is in a state that is ready for submission. This means that code
- * follows the style guide, it is buildable and all tests pass.
+ * Builds Dashboard and ensures that the following requirements are met:
+ *   * The code follows the style guidelines.
+ *   * Unit tests in frontend & backend are successful.
+ *   * Integration tests against Kubernetes are successful. The cluster is
+ *     expected to be up and running as a prerequisite.
  *
  * This task should be used prior to publishing a change.
  **/
 gulp.task('check', ['lint', 'build', 'test', 'integration-test:prod']);
+
+/**
+ * Checks the code quality of Dashboard. In addition a local kubernetes cluster is spawned.
+ * NOTE: This is meant as an entry point for CI jobs.
+ */
+gulp.task('check:create-cluster', ['local-up-cluster', 'check']);
 
 /**
  * Lints all projects code files.
--- a/build/cluster.js
+++ b/build/cluster.js
@@ -44,12 +44,30 @@ const upScript = `${conf.paths.kubernetes}/hack/local-up-cluster.sh`;
 let clusterProcess = null;
 
 /**
+ * @type {boolean} The variable is set when there is an error during cluster creation.
+ */
+let clusterSpawnFailure = null;
+
+/**
  * A Number, representing the ID value of the timer that is set for function which periodically
  * checks if cluster is running. The null means that no timer is running.
  *
  * @type {?number}
  */
 let isRunningSetIntervalHandler = null;
+
+/**
+ * Checks if there was a failure during cluster creation.
+ * Produces an error in case there was one.
+ * @param {function(?Error=)} doneFn - callback for the error
+ */
+function checkForClusterFailure(doneFn) {
+  if (clusterSpawnFailure) {
+    clearTimeout(isRunningSetIntervalHandler);
+    isRunningSetIntervalHandler = null;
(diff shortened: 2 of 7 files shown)


build/cluster.js
```diff
@@ -174,17 +192,32 @@ gulp.task(
       clusterProcess = childProcess.spawn(upScript, {stdio: 'inherit'});
 
       clusterProcess.on('exit', function() { clusterProcess = null; });
+
+      clusterProcess.on('close', function(code) {
```
bryk: Why not reuse 'exit'? I think this is what you want

build/cluster.js
```diff
@@ -52,6 +57,19 @@ let clusterProcess = null;
 let isRunningSetIntervalHandler = null;
 
 /**
+ * Checks if there was a failure during cluster creation.
+ * Produces an error in case there was one.
+ * @param {function(?Error=)} doneFn - callback for the eventual error
```
bryk: I guess this is not "eventual" error :) Safe to remove the word "eventual".

build/cluster.js
```diff
@@ -44,6 +44,11 @@ const upScript = `${conf.paths.kubernetes}/hack/local-up-cluster.sh`;
 let clusterProcess = null;
 
 /**
+ * The variable is set when there is an error during cluster creation.
+ */
```
bryk: Add `@type`?

build/check.js
```diff
@@ -23,14 +23,23 @@ import path from 'path';
 import conf from './conf';
 
 /**
- * Checks whether codebase is in a state that is ready for submission. This means that code
- * follows the style guide, it is buildable and all tests pass.
+ * Builds Dashboard and ensures that the following requirements are met:
+ *   * the code follows the style guidelines.
+ *   * unit tests in frontend & backend should be successful.
```
bryk: Can you improve formatting here? I.e., Make all bullets sentences (start with Capital letter and finish with dot) and use "is/are" instead of "should be". This will be easier to read :)

build/check.js
```diff
@@ -23,14 +23,23 @@ import path from 'path';
...
+ *     expected to be up and running as a prerequisite.
  *
  * This task should be used prior to publishing a change.
  **/
 gulp.task('check', ['lint', 'build', 'test', 'integration-test:prod']);
 
 /**
+ * Checks the code quality of Dashboard. In addition a local kubernetes cluster is spawned.
+ * NOTE: This is meant as an entry point for CI jobs.
+ **/
```
bryk: Just `*/`

bryk:
Nice, PTAL

taimir:
OK, I'm on it. BTW, right now there is no dependency between the availability of a cluster and the integration tests. The task 'integration-test' does not wait for the cluster to be there i.e. it has no (implicit) dependency on 'wait-for-cluster'. Is this intended?

taimir:
@bryk PTAL

bryk:
> OK, I'm on it. BTW, right now there is no dependency between the availability of a cluster and the integration tests. The task 'integration-test' does not wait for the cluster to be there i.e. it has no (implicit) dependency on 'wait-for-cluster'. Is this intended?
 This should be fixed.

bryk:
Can you add the fix to this PR?

taimir:
Sure, I'm on it ... BTW I think this might cause the build to no come to completion.

taimir:
PTAL

build/test.js
```diff
@@ -127,7 +127,7 @@ gulp.task('integration-test', ['serve:nowatch', 'webdriver-update'], runProtract
  * Runs application integration tests. Uses production version of the application.
  */
 gulp.task(
-    'integration-test:prod', ['local-up-cluster', 'serve:prod', 'webdriver-update'],
+    'integration-test:prod', ['serve:prod', 'webdriver-update', 'wait-for-cluster'],
```
bryk: How about 'integration-test' task?

taimir: Sorry, in my previous comment I meant 'integration-test:prod'. But I think you're right, the dev integration tests should also wait for the cluster to spawn. Should I add it?

taimir:
Added 'wait-for-cluster' to the 'integration-test' as well. PTAL.

bryk:
Thank you!

Issue #104, opened 2015-11-30
Implement events view


maciaszczykm:
This issue was created to implement events view:

![Design](https://github.com/kubernetes/dashboard/blob/master/docs/design/mockups/11-11-2015/19_events.png?raw=true)

First step is to implement backend, then frontend. I'm working on backend right now.

maciaszczykm:
@bryk To use `md-tab` in this view I think we need to switch from Material Design version `0.11.4` to at least `0.12.0-rc1` due to issue described at https://github.com/angular/material/issues/5240, it's the next version so impact shouldn't be too big. Otherwise, workaround will be required. Do you agree on upgrade to `0.12.0-rc1`?

bryk:
Yes, feel free to update ngMaterial. I hope that version 1.0 will be available before EOQ1 2016.

floreks:
I'm supporting marcin on this. I'll take left side of the view.

maciaszczykm:
@bryk Events view appear on replica set details page along with pods view. This page has two parameters at the moment - `namespace` and `replicaSet`. I wonder how we should filter events using `replicaSet` parameter, because `namespace` parameter is already considered when loading list of events from backend. 

I have just checked and Kubernetes API, available at http://127.0.0.1:8080/api/v1/events, returns JSON with structure like below:
```
{
  "kind": "EventList",
  "apiVersion": "v1",
  "metadata": {
    "selfLink": "/api/v1/events",
    "resourceVersion": "124873"
  },
  "items": [
    {
      "metadata": {
        "name": "asd-2skdj.141bbf69fadf45fd",
        "namespace": "asdasdsad",
        "selfLink": "/api/v1/namespaces/asdasdsad/events/asd-2skdj.141bbf69fadf45fd",
        "uid": "041815b5-980e-11e5-8194-901b0e5324e4",
        "resourceVersion": "124599",
        "creationTimestamp": "2015-12-01T09:29:27Z",
        "deletionTimestamp": "2015-12-01T14:25:49Z"
      },
      "involvedObject": {
        "kind": "Pod",
        "namespace": "asdasdsad",
        "name": "asd-2skdj",
        "uid": "f6586a19-980d-11e5-8194-901b0e5324e4",
        "apiVersion": "v1",
        "resourceVersion": "92530",
        "fieldPath": "spec.containers{asd}"
      },
      "reason": "Pulling",
      "message": "pulling image \"123\"",
      "source": {
        "component": "kubelet",
[...]

PR #105, opened 2015-11-30
Implemented backend events API

maciaszczykm:
Connected to #104 issue. Added backend API used to get event list from specific namespace.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -74,6 +74,15 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(Logs{}))
 	wsContainer.Add(logsWs)
 
+	eventsWs := new(restful.WebService)
+	eventsWs.Path("/api/events").
+		Produces(restful.MIME_JSON)
+	eventsWs.Route(
+		eventsWs.GET("/{namespace}").
+			To(apiHandler.handleEvents).
+			Writes(Events{}))
+	wsContainer.Add(eventsWs)
+
 	return wsContainer
 }
 
@@ -155,10 +164,20 @@ func (apiHandler *ApiHandler) handleGetNamespaces(
 
 // Handles log API call.
 func (apiHandler *ApiHandler) handleLogs(request *restful.Request, response *restful.Response) {
-
 	namespace := request.PathParameter("namespace")
 	podId := request.PathParameter("podId")
 	result, err := GetPodLogs(apiHandler.client, namespace, podId)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
+	response.WriteHeaderAndEntity(http.StatusCreated, result)
+}
+
+// Handles event API call.
+func (apiHandler *ApiHandler) handleEvents(request *restful.Request, response *restful.Response) {
+	namespace := request.PathParameter("namespace")
+	result, err := GetEvents(apiHandler.client, namespace)
 	if err != nil {
 		handleInternalError(response, err)
 		return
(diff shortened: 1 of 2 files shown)


src/app/backend/events.go
```diff
@@ -0,0 +1,82 @@
...
+	Source string `json:"source"`
+
+	// Sub-object.
+	SubObject string `json:"object"`
+
+	// Count.
+	Count int `json:"count"`
+
+	// First seen.
+	FirstSeen string `json:"firstSeen"`
```
bryk: Send timestamp object, not the string.

src/app/backend/events.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Last seen.
+	LastSeen string `json:"lastSeen"`
```
bryk: Send timestamp object, not the string.

src/app/backend/events.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	events := &Events{
+		Namespace: namespace,
+	}
+
+	for _, element := range list.Items {
+		events.Events = append(events.Events, Event{
+			Message:   element.Message,
+			Source:    "{" + element.Source.Component + " " + element.Source.Host + "}",
```
bryk: Why not both source component and source host. The client may then do whatever they want with this. Preformatting is almost always a bad thing. If you have raw data, send it.

src/app/backend/apihandler.go
```diff
@@ -74,6 +74,15 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			Writes(Logs{}))
 	wsContainer.Add(logsWs)
 
+	eventsWs := new(restful.WebService)
+	eventsWs.Path("/api/events").
+		Produces(restful.MIME_JSON)
+	eventsWs.Route(
+		eventsWs.GET("/{namespace}").
```
bryk: Why do we need only namespace? How about pod name? What does this WS actually return? Can you document it?

src/app/backend/events.go
```diff
@@ -0,0 +1,82 @@
...
+	// Message.
+	Message string `json:"message"`
+
+	// Source (following pattern {component host}).
+	Source string `json:"source"`
+
+	// Sub-object.
+	SubObject string `json:"object"`
+
+	// Count.
```
bryk: This documentation can (and should) be improved :) Say what count is, not that it is count. Try explaining this field to a five years old, and write your explanation here. 
Applies for all docs in this file.

bryk:
PTAL

src/app/backend/apihandler.go
maciaszczykm: I started implementation with only namespace as a parameter, but most probably this will have to be updated. I read the Kubernetes API documentation and to get events fot specific pod we will have to use `fieldSelectors` with specific parameters like `involvedObject.name=<pod>` or `involvedObject.uid=<pod_uid>`. I will continue to work on it, but if you want I will do it via only one pull request. More documentation will follow code updates.

maciaszczykm:
Updated documentation and timestamps, Now i will start updating UI and align backend to work with it. PTAL

maciaszczykm:
Updated documentation and timestamps, Now i will start updating UI and align backend to work with it. PTAL

src/app/backend/apihandler.go
bryk: I see. That's fair enough explanation. We can merge as it is and then continue work. I'll let you decide. Do what it better for you.

bryk: Ok, I'm going to merge this.

maciaszczykm: Thank you, I'm working on UI and backend update right now.

PR #106, opened 2015-11-30
Change healthz ping URL to make it work with proxy

bryk:
With this I can work with `kubectl proxy` clusters. Everything else
stays the same.

@taimir @zreigz Can you review?


cheld:
LGTM

zreigz:
LGTM

PR #108, opened 2015-12-01
Implement current logs form for replica set detail view

zreigz:


zreigz:
PTAL

bryk:
I'll do full review tomorrow. Quick notes:
1. Can you update commit and PR description to be more meaningful
2. This crashes in production mode. Can you ensure that it works with `gulp serve:prod`? (no errors on the browser console)

zreigz:
1 I've changed commit & PR description
2 @ngInject was missing for one of my controller

PTAL

bryk:
1. I'm thinking about the UI - the mocks show different behavior for "Logs" link. Take a look: https://raw.githubusercontent.com/kubernetes/dashboard/master/designs/11-11-2015/16_card%20variations.png

2. Your implementation does not work for pods that have multiple containers. Check it for, e.g., kubedns

3. I think that the route that identifies logs should be: /logs/:namespace/:replicaSetName/:podName/:containerName
ReplicaSet - to get list of possible pods
PodName - to identify what pod to query
ContainerName - to identify which container to query

What do you think?

src/app/frontend/chrome/chrome.html
```diff
@@ -23,10 +23,7 @@
       <h2>
         <span>kubernetes</span>
       </h2>
-      <span flex></span>
-      <md-button class="md-icon-button">
-        <md-icon md-font-library="material-icons">more_vert</md-icon>
-      </md-button>
+      <div flex ui-view="toolbar"></div>
```
bryk: I think it is best to create chrome_state.js file and `export const toolbarViewName = 'toolbar';` there. Then import it in places that use it. With that you have one magic string less and can do renaming and discovery quickly.

src/app/frontend/logs/logs_service.js
```diff
@@ -0,0 +1,57 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Service function for logs panel management.
+ * @return {{registerListener: Function, notifyPodChange: Function, notifyTextColorChange:
+ * Function}}
+ */
+export default function logsService() {
```
bryk: I think if you preserve all parameters required to display logs in the url (e.g., /logs/:namespace/:replicaSet/:pod/:container) you dont need this service at all. This would save a lot of code and complexity.

src/app/frontend/logs/logstoolbar/logstoolbar.scss
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+.md-logs-color-icon {
+  background-color: rgba(0, 0, 0, 1);
+  color: rgba(255, 255, 255, 1);
+}
+
+.md-logs-color-icon-invert {
+  background-color: rgba(255, 255, 255, 1);
+  color: rgba(0, 0, 0, 1) !important;
```
bryk: `!important`? Avoid it whenever possible. Why is it needed here? Cannot you rely on rule order or specificity?

src/app/frontend/replicasetlist/replicasetlist_controller.js
```diff
@@ -63,4 +64,13 @@ export default class ReplicaSetListController {
...
     return this.state_.href(stateName, new StateParams(replicaSet.namespace, replicaSet.name));
   }
+
+  /**
+   * @param {!backendApi.ReplicaSet} replicaSet
+   * @return {string}
+   * @export
+   */
+  getReplicaSetLogsHref(replicaSet) {
+    return this.state_.href(logsState, new StateParams(replicaSet.namespace, replicaSet.name));
```
bryk: Use StateParams for logs, not logs detail.

bryk:
This needs some more work. PTAL

Issue #80, opened 2015-11-18
Implement label handling in deploy form


floreks:
This issue is to implement label handling in deploy form for (RC, Pods, Services, etc.)

Based on this view:
https://raw.githubusercontent.com/kubernetes/dashboard/master/designs/11-11-2015/14_deploy%20more.png

I'm on it :)

floreks:
@bryk: I have some questions.

Should all defined labels be set for created RC, Service and Pods? Same labels for all?
What should we do when someone defines 2 labels with same key. Error, warning, block deploy button?

bryk:
Yes, apply the labels to everything. Take a look at: https://raw.githubusercontent.com/kubernetes/dashboard/master/designs/11-11-2015/14_deploy%20more.png

Re what to do in case of multiple labels with same key: This should be validation error. One of the inputs should be red and say what is wrong. The deploy button should be disabled too. Take a look on how validation is done in Angular and Angular Material: https://material.angularjs.org/latest/demo/select

CC @kenmoore

PR #110, opened 2015-12-01
Added labels handling in the deploy form.

floreks:
Connected to issue #80 

Labels handling with backend support. Labels will be added to created Pod/RC and Service if it is created.

Current view:
zrzut ekranu z 2015-12-01 12-36-38


bryk:
Fix travis?

floreks:
Done.

src/app/frontend/deploy/deploy.scss
```diff
@@ -18,3 +18,38 @@
...
+  }
+}
+
+.kd-deploy-labels-title {
+  padding-left: 5px;
+}
+
+.kd-deploy-labels-header {
+  font-size: 0.85em;
+  padding: 0 10px !important;
```
bryk: `!important`? It should be almost always avoided, unless you're writing framework-y stuff. Rely on rule specificity or order. Otherwise, we'll, sooner or later, run into problems caused by this. This always happens in web dev.

src/app/frontend/deploy/deploylabel_controller.js
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Validates label withing label form.
+   * Current checks:
+   *  - duplicated key
+   * @param {angular.FormController|undefined} labelForm
+   * @private
+   */
+  validateKey_(labelForm) {
+    if (angular.isDefined(labelForm)) {
+      /** @type {angular.NgModelController} */
```
bryk: Can the key be null? The type says yes, but you dont handle this case in the code

src/test/frontend/deploy/deploy_controller_test.js
```diff
@@ -0,0 +1,52 @@
...
+
+describe('Deploy controller', () => {
+  let vm;
+
+  beforeEach(() => {
+    angular.mock.module(deployModule.name);
+
+    angular.mock.inject(($resource, $log, $state, $mdDialog) => {
+      let namespacesMock = {namespaces: []};
+      vm = new DeployController($resource, $log, $state, $mdDialog, namespacesMock);
```
bryk: Use ctrl instead of vm. All other parts of the code use ctrl :) Consistency and code is what matters in long term projects. You'll see benefits of this soon, I hope :)

src/app/frontend/deploy/deploylabel_controller.js
floreks: I thought that it means that labelForm param is either defined or undefined. There has to be '!' in order to mark it non-null also?

src/test/frontend/.eslintrc
```diff
@@ -4,6 +4,7 @@
   },
   // Define missing global Angular testing environment variables.
   "globals": {
+    "angular": true,
     "inject": true,
     "module": true,
```
bryk: Remove module and inject from here? To prevent other from using them (they do not work)

src/app/frontend/deploy/deploylabel_directive.js
```diff
@@ -0,0 +1,33 @@
...
+ * Returns directive definition for deploy form label.
+ *
+ * @return {!angular.Directive}
+ */
+export default function labelDirective() {
+  return {
+    controller: DeployLabelController,
+    controllerAs: 'labelCtrl',
+    templateUrl: 'deploy/deploylabel.html',
+    restrict: 'E',
```
bryk: Do not restrict unless the restriction really matters, e.g., a directive needs to be an attribute. Otherwise, why bother about restricting.

src/app/frontend/deploy/deploy.scss
```diff
@@ -18,3 +18,38 @@
   // TODO(bryk): Find a better, application wide constant for the min width.
   min-width: 600px;
 }
+
+.kd-deploy-label-list {
```
bryk: I'm thinking about how to simplify this CSS and remove as much magic numbers as possible from here (e.g., 10px in a few places and this padding-bottom chain). I'll play with it tomorrow. But you may have your suggestions earlier.

The basic idea is to stick with stock Angular Material whenever possible.

bryk:
Nice piece of work! A few comments, PTAL

src/app/frontend/deploy/deploylabel_controller.js
bryk: ! - non null
Without ! - objects are nullable (e.g., ngmodelcontroller)
               - primitives are non nullable (string,number,boolean) 
To mark that this can be undefined, use {fooType|undefined}


(This is how it works, I did not invent it...)

src/app/frontend/deploy/deploy.scss
floreks: Styling using only stock Angular Material would be great. I'm still learning this so every workaround to get rid of this will be good for me.

floreks:
Resolved. PTAL

bryk:
Ooops, I tried testing this but got errors:
```
10angular.js:12477 TypeError: Cannot read property 'length' of undefined
    at DeployLabelController.isRemovable (http://localhost:9090/app-dev.js:1450:49)
    at Object.fn [as get] (eval at <anonymous> (http://localhost:9090/bower_components/angular/angular.js:13322:15), <anonymous>:4:259)
    at Scope.$digest (http://localhost:9090/bower_components/angular/angular.js:15818:40)
    at Scope.$apply (http://localhost:9090/bower_components/angular/angular.js:16097:24)
    at done (http://localhost:9090/bower_components/angular/angular.js:10546:47)
    at completeRequest (http://localhost:9090/bower_components/angular/angular.js:10744:7)
    at XMLHttpRequest.requestLoaded (http://localhost:9090/bower_components/angular/angular.js:10685:9)
```

Maybe you can create an integration test that navigates to deploy page and checks that nothing crashed (no errors on the console). It should be easy.

floreks:
Could you tell me what use case triggered this error? Labels array is never empty so there should be no way for it to be undefined. I've got no errors at all. I'll write a test for this.

bryk:
I just cloned your repo and opened deploy form on Linux on Chrome. Nothing more.

floreks:
So this is interesting. I'm running it both on serve:prod and watch tasks and there are no errors. This error could only happen if `labels` defined in DeployController are undefined, because they are passed further to DeployLabelController, but they are predefined in constructor.

bryk:
You need `scope: {},` on deploylabel_directive.js

bryk:
With this HTML and no CSS the input looks like on the presentation. Try using flex layout and as little css as possible. 
```
      <div>
        <div class="kd-deploy-labels-title">Labels (optional)</div>
        <div layout="column" class="kd-deploy-label-list">
          <div layout="row" class="kd-deploy-labels-header">
            <p flex>Key</p>
            <p flex>Value</p>
          </div>
          <div ng-repeat="label in ctrl.labels" class="kd-deploy-labels-row">
            <kd-label layout="row" flex label="label" labels="ctrl.labels"></kd-label>
          </div>
        </div>
      </div>
```

floreks:
PTAL

src/app/frontend/deploy/deploy.scss
```diff
@@ -18,3 +18,10 @@
   // TODO(bryk): Find a better, application wide constant for the min width.
   min-width: 600px;
 }
+
+.kd-deploy-labels-button {
```
bryk: Nice! Can you move this to deploylabel.scss? With this all deploylabel-related things are in deployalbel.* files. Then, e.g., when refactoring you know where to look for things.

bryk:
Last comment and we can merge :)

floreks:
Done and done :)

bryk:
Awesome, thank you!

PR #114, opened 2015-12-02
Bug fix. 'gulp check' hangs when tests fail

zreigz:


build/test.js
```diff
@@ -36,6 +36,10 @@ function runFrontendUnitTests(singleRun, doneFn) {
   };
 
   let server = new karma.Server(localConfig, function(failCount) {
+    if (failCount > 0) {
+      // Kill backend server and cluster, if running.
+      gulp.start('kill-all');
```
bryk: This does not fix the issue in all other cases. E.g., when lint or CSS styles rule fail, gulp will still hang up.

The only way to fix this is to remove 'wait-for-cluster' task. Otherwise this problem will come back every now and then.


How about creating new versions of integration-test tasks that also have local-up-cluster as a dependency. And make local-up-cluster finish only when cluster is healthly. I believe this would work.

@cheld What do you think?

cheld: Ok, lets make it save first and fancy later. We aim for the following synchronous sequence in CI:

local-up-cluster
check

To do so:
1. let's add wait-for-cluster dependency to local-up-cluster and add doneFN function to make the cluster creation synchronous. 

2. Currently, check:create-cluster has a dependency to local-up-cluster and check. Currently, I assume local-up-cluster and check would still run parallel. A dependency from check to local-up-cluster would solve that, however, I don't want this dependency. So, I guess we can't escape from some 'dirty hack'

e.g. pseudo code:

 check:create-cluster [local-up-cluster] {
   gulp.call(check)
}

or

 check:create-cluster {
     runInSequence(local-up-cluster,check)
}

Does it make sense?

zreigz: Looks good for me but one thing is missing. There is only one place where on error we kill backend and cluster. This check is only done in function runProtractorTests. So I will extend it for this code: 
```
import sequence from 'run-sequence';

gulp.task('check:create-cluster', function () {
  runSequence('local-up-cluster', ['check'],
    function () {
      gulp.start('kill-all');
  });
});

```
@bryk @cheld  what do you think  ?

zreigz:
I've moved wait-for-cluster to local-up-cluster
I've added sequence for local-up-cluster and check.
Additionally on any error backend server and cluster will be killed to finish gulp process. I've checked it with lint error, unit test error and it seems to work.

zreigz:
PTAL

cheld:
LGTM

build/check.js
```diff
@@ -34,10 +35,15 @@ import conf from './conf';
 gulp.task('check', ['lint', 'build', 'test', 'integration-test:prod']);
 
 /**
- * Checks the code quality of Dashboard. In addition a local kubernetes cluster is spawned.
- * NOTE: This is meant as an entry point for CI jobs.
+ * Waits until local kubernetes cluster is up and running and then build Dashboard.
```
bryk: Hmm... Previous documentation note was correct. I dont see a reason to change it to this one. Can you revert? And only change implementation.

bryk:
One comment. PTAL

bryk:
@zreigz @cheld 
I've took a stab on making local-up-cluster synchronous. Take a look: https://github.com/kubernetes/dashboard/commit/a2c2bf67ce6e747e4f0324f5de4f5a93bbd05f99

What do you think? Does it make sense to follow pattern in this PR or making local-up-cluster synchronous? 

In the linked commit I've also managed to simplity cluster.js a little (no global state, just the cluster process var)

zreigz:
I used your code and test it.
I like cluster.js code refactor. I've triggered some errors and it seems to doesn't work properly. Still error handling is missing for me. In this approach you need add error handling in all potential risk places.

 I've created two errors for
1 frontend test  - works fine gulp process exited with error code: https://travis-ci.org/zreigz/dashboard/builds/94579756
2 lint - gulp process stuck - https://travis-ci.org/zreigz/dashboard/builds/94580495

cheld:
@bryk I just looked at your code. I did not try it out. In general I like it. Clean and follows gulp concepts. 

I was a bit worried that an *optional* dependency from integration-test to local-up-cluster creates too much complexity in callers.

bryk:
You know what? Let's merge @zreigz proposal and I'll send a PR with just refactor in the cluster.js code. This is because I think that my proposal does not fix the problem entirely, but is a good step forward. Later, if this causes problems to us, we can continue the work. 
What do you think? If yes, can you @zreigz address the one comment I had in your PR?

zreigz:
Yes sure. I am doing it right now.

zreigz:
PTAL

PR #116, opened 2015-12-02
Tests for replica set list backend

bryk:
The tests cover all methods except the one that interacts with the
client. Test for it will be added later.

### DIFF
--- a/src/test/backend/replicasetlist_test.go
+++ b/src/test/backend/replicasetlist_test.go
@@ -15,10 +15,194 @@
 package main
 
 import (
+	"k8s.io/kubernetes/pkg/api"
+	"reflect"
 	"testing"
 )
 
-func TestFoo(t *testing.T) {
-	// TODO(bryk): Write tests here.
-	isServiceMatchingReplicaSet(nil, nil)
+func TestIsServiceMatchingReplicaSet(t *testing.T) {
+	cases := []struct {
+		serviceSelector, replicaSetSelector map[string]string
+		expected                            bool
+	}{
+		{nil, nil, false},
+		{nil, map[string]string{}, false},
+		{map[string]string{}, nil, false},
+		{map[string]string{}, map[string]string{}, false},
+		{map[string]string{"app": "my-name"}, map[string]string{}, false},
+		{map[string]string{"app": "my-name", "version": "2"},
+			map[string]string{"app": "my-name", "version": "1.1"}, false},
+		{map[string]string{"app": "my-name", "env": "prod"},
+			map[string]string{"app": "my-name", "version": "1.1"}, false},
+		{map[string]string{"app": "my-name"}, map[string]string{"app": "my-name"}, true},
+		{map[string]string{"app": "my-name", "version": "1.1"},
+			map[string]string{"app": "my-name", "version": "1.1"}, true},
+		{map[string]string{"app": "my-name"},
+			map[string]string{"app": "my-name", "version": "1.1"}, true},
+	}
+	for _, c := range cases {
+		actual := isServiceMatchingReplicaSet(c.serviceSelector, c.replicaSetSelector)
+		if actual != c.expected {
+			t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
+				c.serviceSelector, c.replicaSetSelector, actual, c.expected)
+		}
+	}
+}
+
+func TestGetServicePortsName(t *testing.T) {
+	cases := []struct {
+		ports    []api.ServicePort
+		expected string
+	}{
+		{nil, ""},
+		{[]api.ServicePort{}, ""},
+		{[]api.ServicePort{{Name: "foo", Port: 8080, Protocol: "TCP"}}, " 8080/TCP"},
+		{[]api.ServicePort{{Name: "foo", Port: 8080, Protocol: "TCP"},
+			{Name: "foo", Port: 9191, Protocol: "UDP"}}, " 8080/TCP,9191/UDP"},
+	}
+	for _, c := range cases {
+		actual := getServicePortsName(c.ports)
+		if actual != c.expected {
+			t.Errorf("getServicePortsName(%+v) == %+v, expected %+v", c.ports, actual, c.expected)
+		}
+	}
+}
+
+func TestGetExternalEndpoint(t *testing.T) {
+	cases := []struct {
+		serviceIp string
+		ports     []api.ServicePort
+		expected  string
+	}{
+		{"127.0.0.1", nil, "127.0.0.1"},
+		{"127.0.0.1", []api.ServicePort{{Name: "foo", Port: 8080, Protocol: "TCP"}},
+			"127.0.0.1 8080/TCP"},
+	}
+	for _, c := range cases {
+		actual := getExternalEndpoint(c.serviceIp, c.ports)
+		if actual != c.expected {
+			t.Errorf("getExternalEndpoint(%+v, %+v) == %+v, expected %+v",
+				c.serviceIp, c.ports, actual, c.expected)
+		}
+	}
+}
+
+func TestGetInternalEndpoint(t *testing.T) {
+	cases := []struct {
+		serviceName, namespace string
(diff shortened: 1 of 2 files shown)


bryk:

maciaszczykm:
After discussion with @floreks we think, that tests look fine, but maybe splitting them into smaller pieces would increase their readability. Also when some test fail it would be easier to find the issue, even by failed test name. Any future changes in application code would be then easier to cover in tests. The other way to enhance readability is to put some comments (regarding input data - given, operations tested - when, and expected results - then) inside. What do you think about it?

bryk:
I basically followed official examples of how to test in golang. See https://golang.org/doc/code.html paragraph testing. Also, this is the pattern that kubernetes core often uses, e.g., https://github.com/kubernetes/kubernetes/blob/master/pkg/admission/chain_test.go or https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/proxy_server_test.go

I'd say that such pattern is good for side-effect-less function that just have input and output parameters. I wouldn't use it for functions with side effects or ones that interact with mocked services.

> Also when some test fail it would be easier to find the issue, even by failed test name.

This is the error message I get when something fails. It does not use test method names, but gives an error message with input, output and expected. I think this is better than method names. And you get failure line numbers, of course.
```
--- FAIL: TestGetInternalEndpoint (0.00s)
	replicasetlist_test.go:109: getInternalEndpoint(my-service, my-namespace, [{Name:foo Protocol:TCP Port:8080 TargetPort:{Kind:0 IntVal:0 StrVal:} NodePort:0}]) == my-service.my-namespace 8080/TCP, expected my-service.my-namespace 8080/TCP 
FAIL
FAIL	github.com/kubernetes/dashboard	0.059s
```


> The other way to enhance readability is to put some comments (regarding input data - given, operations tested - when, and expected results - then) inside.

Sounds nice, can you give an example?

floreks:
I'm not sure how it would work in go but in java/js etc. we can write

```javascript
it('should return empty string when containerImage is undefined', () => {
    // Given
    ctrl.containerImage = undefined;

    // When
    let containerImageVersion = ctrl.getContainerImageVersion_();

    // Then
    expect(containerImageVersion).toEqual('');
  });
```

This way it is clear what is the input data, which functionality is beeing tested and what is the expected output based on the initial data. For now tests i've written for frontend are split into very small units and each test is for separate use case of given function. I see that Go doesn't really have this fluent api for tests so your way may be better here.

bryk:
I've conducted a small experiment. Converted current code to one with separate methods. Is this more readable? 

I've also done a survey around the office to ask what people think. Generally they say that current style is good when you test algorithm's input and output; you just specify a dozen of cases. "Given, when, then" approach with separate methods use for business logic or when you have complex interaction with an object.

Input:
```go
func TestIsServiceMatchingReplicaSet(t *testing.T) {
	cases := []struct {
		serviceSelector, replicaSetSelector map[string]string
		expected                            bool
	}{
		{nil, nil, false},
		{nil, map[string]string{}, false},
		{map[string]string{}, nil, false},
		{map[string]string{}, map[string]string{}, false},
		{map[string]string{"app": "my-name"}, map[string]string{}, false},
		{map[string]string{"app": "my-name", "version": "2"},
			map[string]string{"app": "my-name", "version": "1.1"}, false},
		{map[string]string{"app": "my-name", "env": "prod"},
			map[string]string{"app": "my-name", "version": "1.1"}, false},
		{map[string]string{"app": "my-name"}, map[string]string{"app": "my-name"}, true},
		{map[string]string{"app": "my-name", "version": "1.1"},
			map[string]string{"app": "my-name", "version": "1.1"}, true},
		{map[string]string{"app": "my-name"},
			map[string]string{"app": "my-name", "version": "1.1"}, true},
	}
	for _, c := range cases {
		actual := isServiceMatchingReplicaSet(c.serviceSelector, c.replicaSetSelector)
		if actual != c.expected {
			t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
				c.serviceSelector, c.replicaSetSelector, actual, c.expected)
		}
	}
}
```

Output:
```go
func TestIsServiceMatchingReplicaSet1(t *testing.T) {
	serviceSelector := nil
	replicaSetSelector := nil
	expected = false

	actual := isServiceMatchingReplicaSet(serviceSelector, replicaSetSelector)
	if actual != c.expected {
		t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
			serviceSelector, replicaSetSelector, actual, expected)
	}
}

func TestIsServiceMatchingReplicaSet1(t *testing.T) {
	serviceSelector := nil
	replicaSetSelector := map[string]string{}
	expected = false

	actual := isServiceMatchingReplicaSet(serviceSelector, replicaSetSelector)
	if actual != c.expected {
		t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
			serviceSelector, replicaSetSelector, actual, expected)
	}
}

func TestIsServiceMatchingReplicaSet1(t *testing.T) {
	serviceSelector := map[string]string{}
	replicaSetSelector := nil
	expected = false

	actual := isServiceMatchingReplicaSet(serviceSelector, replicaSetSelector)
	if actual != c.expected {
		t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
			serviceSelector, replicaSetSelector, actual, expected)
	}
}

func TestIsServiceMatchingReplicaSet1(t *testing.T) {
	serviceSelector := map[string]string{}
	replicaSetSelector := map[string]string{}
	expected = false

	actual := isServiceMatchingReplicaSet(serviceSelector, replicaSetSelector)
	if actual != c.expected {
		t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
			serviceSelector, replicaSetSelector, actual, expected)
	}
}

func TestIsServiceMatchingReplicaSet1(t *testing.T) {
	serviceSelector := map[string]string{"app": "my-name", "version": "2"}
	replicaSetSelector := map[string]string{"app": "my-name", "version": "1.1"}
	expected = false

	actual := isServiceMatchingReplicaSet(serviceSelector, replicaSetSelector)
	if actual != c.expected {
		t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
			serviceSelector, replicaSetSelector, actual, expected)
	}
}

func TestIsServiceMatchingReplicaSet1(t *testing.T) {
	serviceSelector := map[string]string{"app": "my-name", "env": "prod"}
	replicaSetSelector := map[string]string{"app": "my-name", "env": "prod"}
	expected = false

	actual := isServiceMatchingReplicaSet(serviceSelector, replicaSetSelector)
	if actual != c.expected {
		t.Errorf("isServiceMatchingReplicaSet(%+v, %+v) == %+v, expected %+v",
			serviceSelector, replicaSetSelector, actual, expected)
	}
}

<a few more methods>

```

floreks:
That's actually a valid point and we agree with that. Perhaps we can use both type of test templates. First can be used for simpler cases and second for tests where more complex object initialization is needed. 

Current tests are simple so they can be left as they are. We should be agile on this one :)

src/test/backend/replicasetlist_test.go
```diff
@@ -15,10 +15,194 @@
...
+				},
+				{
+					Spec: api.ServiceSpec{Selector: map[string]string{"app": "my-name-2", "ver": "2"}},
+					ObjectMeta: api.ObjectMeta{
+						Name:      "my-app-2",
+						Namespace: "namespace-2",
+					},
+				},
+			},
+			&ReplicaSetList{
```
floreks: Maybe we could delegate object creation to separate builder function? Service can be initialized by `createService("app", "my-name", "my-app", "namespace")` and then test will be easier to understand because this json-like format is sometimes hard to follow.

bryk:
Thanks, merging this then.

src/test/backend/replicasetlist_test.go
bryk: Ooops, I was too fast and already merged this... I really dont like GitHub code review system

PR #119, opened 2015-12-03
directive switch is created

batikanu:
Directive switch for deploy-from-file and deploy-from-settings directives is implemented.
File upload mechanism for deploy-from-file is added.

### DIFF
--- a/src/app/frontend/deploy/deploy_controller.js
+++ b/src/app/frontend/deploy/deploy_controller.js
@@ -69,6 +69,13 @@ export default class DeployController {
      * @export {{deploy:function()}}
      */
     this.detail = {deploy: function() {}};
+
+    /**
+     * Child directive selection model. The list of possible values is in template. This value
+     * represents the default selection.
+     * @export {string}
+     */
+    this.selection = "Settings";
   }
 
   /**
--- a/src/app/frontend/deploy/deployfromfile_controller.js
+++ b/src/app/frontend/deploy/deployfromfile_controller.js
@@ -1,0 +1,36 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Controller for the deploy from file directive.
+ *
+ * @final
+ */
+export default class DeployFromFileController {
+  /** @ngInject */
+  constructor() {
+    /**
+     * It initializes the scope output parameter
+     *
+     * @export {!DeployFromFileController}
+     */
+    this.detail = this;
+
+    /**
+     * Custom file model for the selected file
+     *
+     * @export {{name:string, content:string}} */
+    this.file = {name: '', content: ''};
+  }
+}
--- a/src/app/frontend/deploy/deploy.html
+++ b/src/app/frontend/deploy/deploy.html
@@ -22,18 +22,26 @@ limitations under the License.
         <label>App name</label>
         <input ng-model="ctrl.name" required>
       </md-input-container>
-      <md-radio-group ng-model="data.group1">
-        <md-radio-button class="md-primary">
+      <md-radio-group ng-model="ctrl.selection">
+        <md-radio-button value="Settings" class="md-primary">
           Specify app details below
         </md-radio-button>
-        <md-radio-button value="bar" class="md-primary">
+        <md-radio-button value="File" class="md-primary">
           Upload a YAML or JSON file
         </md-radio-button>
       </md-radio-group>
-      <deploy-from-settings name="ctrl.name"
-                            namespace="ctrl.namespace"
-                            detail="ctrl.detail">
-      </deploy-from-settings>
+      <div ng-switch="ctrl.selection">
+        <deploy-from-settings ng-switch-when="Settings"
+                              name="ctrl.name"
+                              namespace="ctrl.namespace"
+                              detail="ctrl.detail">
+        </deploy-from-settings>
+        <deploy-from-file ng-switch-when="File"
+                          name="ctrl.name"
+                          namespace="ctrl.namespace"
+                          detail="ctrl.detail">
+        </deploy-from-file>
+      </div>
       <md-input-container class="md-block">
       <label>Namespace</label>
       <md-select ng-model="ctrl.namespace" required>
--- a/src/app/frontend/deploy/deploy_module.js
+++ b/src/app/frontend/deploy/deploy_module.js
@@ -15,6 +15,9 @@
 import stateConfig from './deploy_state';
 import deployFromSettingsDirective from './deployfromsettings_directive';
 import deployLabelDirective from './deploylabel_directive';
+import deployFromFileDirective from './deployfromfile_directive';
+import fileReaderDirective from './filereader_directive';
+import uploadDirective from './upload_directive';
 
 /**
  * Angular module for the deploy view.
@@ -30,4 +33,7 @@ export default angular.module(
                           ])
     .config(stateConfig)
     .directive('deployFromSettings', deployFromSettingsDirective)
+    .directive('deployFromFile', deployFromFileDirective)
+    .directive('kdFileReader', fileReaderDirective)
+    .directive('kdUpload', uploadDirective)
     .directive('kdLabel', deployLabelDirective);
--- a/src/app/frontend/deploy/deployfromfile_directive.js
+++ b/src/app/frontend/deploy/deployfromfile_directive.js
@@ -1,0 +1,29 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import DeployFromFileController from './deployfromfile_controller';
+
+export default function deployFromFileDirective() {
+  return {
+    scope: {},
+    bindToController: {
+      name: '=',
+      namespace: '=',
+      detail: '=',
+    },
+    controller: DeployFromFileController,
+    controllerAs: 'ctrl',
+    template: '<kd-upload file="ctrl.file"></kd-upload>',
+  };
+}
--- a/src/app/frontend/deploy/filereader_directive.js
+++ b/src/app/frontend/deploy/filereader_directive.js
@@ -1,0 +1,50 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Provides ng-model binding for input elements with file type and updates the model
+ * on change with {name:string, content:string} custom object form of the selected file.
+ *
+ * @param {!angular.$log} $log
+ * @return {!angular.Directive}
+ * @ngInject
+ */
+export default function fileReaderDirective($log) {
+  return {
+    restrict: 'A',
+    require: ['^kdUpload', 'ngModel'],
+    link: function(scope, element, attrs, ctrls) {
+      // TODO: Annotate Type UploadController
+      let uploadController = ctrls[0];
+      uploadController.registerBrowseFileFunction(() => (element[0].click()));
+
+      element.bind('change', (changeEvent) => {
+        /** @type {File} */
+        let file = changeEvent.target.files[0];
+        if (!file) {
+          $log.error('Error invalid file: ', file);
+          return;
+        }
+        let reader = new FileReader();
+        reader.onload = () => {
+          /** @type {!angular.NgModelController} */
+          let ngModelController = ctrls[1];
+          ngModelController.$setViewValue({name: file.name, content: reader.result});
+        };
+        reader.onerror = (error) => $log.error('Error reading file:', error);
+        reader.readAsBinaryString(file);
+      });
+    },
+  };
+}
--- a/src/app/frontend/deploy/upload.html
+++ b/src/app/frontend/deploy/upload.html
@@ -1,0 +1,33 @@
+<!--
+Copyright 2015 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div layout="row" layout-align="space-between start">
+  <div flex>
+    <md-input-container class="md-block">
+      <label>YAML or JSON file</label>
+      <!--TODO: Browse file on focus doesn't work in Firefox. It is to be investigated.-->
+      <input ng-model="ctrl.file.name" ng-focus="ctrl.browseFile()" />
+      </md-input-container>
+  </div>
+    <md-button type="button" class="md-raised kd-upload-button" ng-click="ctrl.browseFile()">
+      <label class="kd-upload-label">
+        <md-icon>
+          <i class="material-icons">more_horiz</i>
+        </md-icon>
+        </label>
+    </md-button>
+    <input type="file" style="display:none" kd-file-reader ng-model="ctrl.file">
+</div>
--- a/src/app/frontend/deploy/upload.scss
+++ b/src/app/frontend/deploy/upload.scss
@@ -1,0 +1,24 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.kd-upload-button {
+  padding-left: 0;
+  padding-right: 0;
+}
+
+.kd-upload-label {
+  display: inline-block;
+  width: 100%;
+  cursor: inherit;
+}
--- a/src/app/frontend/deploy/upload_controller.js
+++ b/src/app/frontend/deploy/upload_controller.js
@@ -1,0 +1,51 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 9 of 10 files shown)


src/app/frontend/deploy/deploy_controller.js
```diff
@@ -69,6 +69,12 @@ export default class DeployController {
      * @export {{deploy:function()}}
      */
     this.detail = {deploy: function() {}};
+
+    /**
+     * child directive selection model
```
bryk: Add a note that the list of possible values is in the template. And this value represents default selection.

src/app/frontend/deploy/deployfromfile_controller.js
```diff
@@ -0,0 +1,58 @@
...
+   * @param {HTMLInputElement} element
+   * @export
+   */
+  updateFile(element) {
+    let file = element.files[0];
+    this.fileName = file.name;
+    let reader = new FileReader();
+    reader.onloadend = (event) => { this.file_ = event.target.result; };
+    reader.readAsBinaryString(file);
+    let defer = this.q_.defer();
```
bryk: Hmm... Why is deferred piece of code needed? You immediately resolve it.

src/app/frontend/deploy/deployfromfile.scss
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.kd-deployfromfile-button {
+  padding-left: 0;
+  padding-right: 0;
+}
+.kd-deployfromfile-label {
```
bryk: Please add newline here

src/app/frontend/deploy/deployfromfile.html
```diff
@@ -0,0 +1,35 @@
...
+  </div>
+  <div layout="column" flex="18">
+    <md-button type="button" class="md-raised kd-deployfromfile-button">
+      <label for="fileInput" class="kd-deployfromfile-label">
+        <md-icon>
+          <i class="material-icons">more_horiz</i>
+        </md-icon>
+      </label>
+    </md-button>
+    <input id="fileInput" type="file" style="display:none" onchange="angular.element(this).scope().ctrl.updateFile(this)">
```
bryk: 100 chars limit here?

bryk: Why not just ctrl.updateFile?

src/app/frontend/deploy/deployfromfile.html
```diff
@@ -0,0 +1,35 @@
...
+-->
+
+<div layout="row" layout-align="space-between start">
+  <div layout="column" flex="80">
+    <md-input-container class="md-block">
+      <label>YAML or JSON file</label>
+      <input ng-model="ctrl.fileName" ng-disabled="true"/>
+    </md-input-container>
+  </div>
+  <div layout="column" flex="18">
```
bryk: Flex = 18? Why not 20?

src/app/frontend/deploy/deployfromfile.html
```diff
@@ -0,0 +1,35 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div layout="row" layout-align="space-between start">
+  <div layout="column" flex="80">
```
bryk: Why do you need layout when there is only one thing inside?

src/app/frontend/deploy/deployfromfile_controller.js
batikanu: It is required to update view. There is no direct angular model binding. I don't want to use $apply() because we don't want to use $scope in controller.

src/app/frontend/deploy/deployfromfile.html
```diff
@@ -0,0 +1,35 @@
...
+    <md-input-container class="md-block">
+      <label>YAML or JSON file</label>
+      <input ng-model="ctrl.fileName" ng-disabled="true"/>
+    </md-input-container>
+  </div>
+  <div layout="column" flex="18">
+    <md-button type="button" class="md-raised kd-deployfromfile-button">
+      <label for="fileInput" class="kd-deployfromfile-label">
```
bryk: Hmm... I'm worried about uniqueness of this id. How about you generate it in the controller and then bind here? To generate use, e.g., let id = `fileInput-${new Date()}`;

bryk:
PTAL

src/app/frontend/deploy/deployfromfile_controller.js
bryk: Cannot you use ng-change?

src/app/frontend/deploy/deployfromfile.html
batikanu: Flex=20 is too much for the upload button. Then the upload button is has more width then the other buttons in the card like deploy and cancel. If I want to complete the total to 100 by using
Flex=82, Flex:18 then upload file input field is shrinked. It is kind of a fine adjustment. I don't want to assign an exact width to the button as well to make it compatible with the other buttons.

src/app/frontend/deploy/deployfromfile.html
batikanu: onchange is not angular js event handler and doesn't know about ctrl.updateFile. We need to find the ctrl object via element query

src/app/frontend/deploy/deployfromfile_controller.js
bryk: One additional note here is that you're accessing DOM inside a controller. This is a bad practice. How about implementing something like here (but simpler): http://stackoverflow.com/a/19121983 
This means: create a separate directive that manipulates DOM in link function.

batikanu: Yes actually it is bad practice but ng-change can't recognize input type=file change. I will check the link.

src/app/frontend/deploy/deployfromfile.html
maciaszczykm: @batikanu It seems, that `Currently, the flex attribute value is restricted to multiples of five and 33 or 66`. Look here: https://material.angularjs.org/latest/layout/grid.

batikanu: Yes that's right I will just remove Flex=18.

src/app/frontend/deploy/deployfromfile.html
batikanu: It is redundant. I think just a child div element with flex attribute is enough to use the grid. I will remove layout attribute.

bryk: Correct.

src/app/frontend/deploy/deployfromfile_controller.js
bryk: Thank you. If you use separate directive with link function that feeds ngModelController and ng-model on the file input this would be very clean and "the way it should be done in Angular" :)

batikanu: Actually, this directive can be reused by other views as well. Shall I locate it somewhere more common than deploy folder? If yes where is the right place for it?

bryk: Hmm... Until there is  second client for the directive, don't bother about where to place it. Place it here.

src/app/frontend/deploy/deploy_controller.js
batikanu: done

src/app/frontend/deploy/deployfromfile.scss
batikanu: done

src/app/frontend/deploy/deployfromfile.html
batikanu: it is removed with the implementation of file reader directive.

src/app/frontend/deploy/deployfromfile.html
batikanu: done

src/app/frontend/deploy/deployfromfile.html
batikanu: done

batikanu:
PTAL

bryk:
You're missing ngInject somewhere. I see errors in prod mode.

src/app/frontend/deploy/deployfromfile_controller.js
```diff
@@ -0,0 +1,42 @@
...
+     * Unique id for file input element
+     *
+     * @export {string} */
+    this.fileInputId = `fileInput-${new Date().getTime()}`;
+
+    /**
+     * Custom file model for the selected file
+     *
+     * @export {{name:string, content:string}} */
+    this.file = {name: '', content: ''};
```
bryk: Caveat: this will not work in serve:prod mode (please verify though). This is because this.file gets exported, but its props are renamed. Create two variables instead.

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -69,6 +69,13 @@ export default class DeployController {
      * @export {{deploy:function()}}
      */
     this.detail = {deploy: function() {}};
+
+    /**
+     * Child directive selection model. The list of possible values are in template. This value
```
bryk: "the list of (...) are" -> "the list of (...) is"

src/app/frontend/deploy/deployfromfile.html
```diff
@@ -0,0 +1,34 @@
+<!--Copyright 2015 Google Inc.All Rights Reserved.
+
+     Licensed under the Apache License,
```
bryk: Hmm... Problems with formatting here?

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {!angular.$log} $log
+ * @return {!angular.Directive}
+ */
+export default function kdFileReaderDirective($log) {
+  return {
+    restrict: 'A',
+    require: 'ngModel',
+    link: function(scope, element, attrs, ngModel) {
+      element.bind('change', (changeEvent) => {
+        let log = $log;
```
bryk: No need to alias $log

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+/**
+ * Provides ng-model binding for input elements with file type and updates the model
+ * on change with {name:string, content:string} custom object form of the selected file.
+ *
+ * @param {!angular.$log} $log
+ * @return {!angular.Directive}
+ */
+export default function kdFileReaderDirective($log) {
```
bryk: ngInject required here

batikanu:
Ok I check it

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    restrict: 'A',
+    require: 'ngModel',
+    link: function(scope, element, attrs, ngModel) {
```
bryk: `@type` ngModel that it is !angular.NgModelController

src/app/frontend/deploy/deployfromfile_controller.js
bryk: Hmm..., I see that you need this for ng-model. Let me think.

bryk: This works in prod mode. Nice! Sorry for confusion.

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,42 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Provides ng-model binding for input elements with file type and updates the model
+ * on change with {name:string, content:string} custom object form of the selected file.
```
bryk: Nice, that's good documentation.

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,42 @@
...
+      element.bind('change', (changeEvent) => {
+        let log = $log;
+        let file = changeEvent.target.files[0];
+        if (!file) {
+          log.error('Error invalid file: ', file);
+          return;
+        }
+        let reader = new FileReader();
+        reader.onload =
+            (event) => { ngModel.$setViewValue({name: file.name, content: event.target.result}); };
```
bryk: Use reader.result instead of event.target. This is easier for humans and compiler to comprehend.

src/app/frontend/deploy/deployfromfile.scss
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.kd-deployfromfile-button {
+  padding-left: 0;
+  padding-right: 0;
+}
+
+.kd-deployfromfile-label {
+  display: inline-block;
```
bryk: add cursor: inherit here. To make it look like button.

src/app/frontend/deploy/deployfromfile.html
```diff
@@ -0,0 +1,34 @@
...
+
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div layout="row" layout-align="space-between start">
+  <div flex="80">
+    <md-input-container class="md-block">
+      <label>YAML or JSON file</label>
+      <input ng-model="ctrl.file.name" ng-disabled="true"/>
```
bryk: Can you trigger the file input on focus on this field? Instead of disabling?

bryk:
Great progress. A couple of comments to make this even better. PTAL

src/app/frontend/deploy/kdfilereader_directive.js
batikanu: do you mean `@param ngModel {!angular.NgModelController}` which will be defined for link function?

bryk: Yes. You can do this outline or inline, see: https://github.com/google/closure-compiler/wiki/Using-JSDoc-Annotations-with-Closure-compiler

src/app/frontend/deploy/deployfromfile.html
batikanu: In this case two different elements will be interacting each other. Visible input field and hidden input field. There are 3 solutions. 
1- I can insert a custom attribute in the visible input field and query it from the kdFileReaderDirective's link function to register focus event. 
2- I can implement different directive for the visible input field which registers the focus event in its link function by using the id of the hidden input field. 
3- I can implement a new directive like kd-upload which contains all the components or elements (Input fields and button) required for the upload functionality. Those input fields interact each other via the scope of this directive. 
I think solution 3 is better. What do you think?

bryk: Yeah, solution 3 sounds best :) But perhaps of interacting via scope, interact via controller of the directive.

batikanu: I mean 2 input fields need to find each other anyway. Instead of writing a query, I may keep the reference of the other element in controller like scope.ctrl.kdFileReaderElement = element; in link function and the other element will use it to trigger scope.ctrl.kdFileReaderElement.click() in its own link function. This scope and ctrl will be under kd-upload directive so it won't be seen by the deploy from file directive

bryk: Yes, LGTM

bryk:
Btw, this needs rebase

src/app/frontend/deploy/kdupload_controller.js
```diff
@@ -0,0 +1,37 @@
...
+ * @final
+ */
+export default class KdUploadController {
+  /** @ngInject */
+  constructor() {
+    /**
+     * It will be set by kdFileReaderDirective to the instance of an input element
+     * which contains a click function
+     *
+     * @export {{click:Function}} */
```
batikanu: If there is a better suggestion for typing, I can change it.

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,47 @@
...
+ * @ngInject
+ */
+export default function kdFileReaderDirective($log) {
+  return {
+    restrict: 'A',
+    require: 'ngModel',
+    link: function(scope, element, attrs, /** @type {!angular.NgModelController} */ ngModel) {
+
+      element.bind('change', (changeEvent) => {
+        let file = changeEvent.target.files[0];
```
bryk: Hmm.... Add type information to this var if you can.

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,47 @@
...
+        let reader = new FileReader();
+        reader.onload = () => { ngModel.$setViewValue({name: file.name, content: reader.result}); };
+        reader.onerror = (error) => $log.error('Error reading file:', error);
+        reader.readAsBinaryString(file);
+      });
+
+      if (!scope.ctrl || !scope.ctrl.kdFileReaderElement) {
+        return;
+      }
+      scope.ctrl.kdFileReaderElement = element[0];
```
bryk: You know what? You can `require` KdUploadController here and register a callback in the controller. Such setup would be less fragile, than current one which bases on assumptions of the scope shape.

src/app/frontend/deploy/kdupload_controller.js
bryk: Hmm... Why an object containing a click function? How about making this just a function? And not even exported. It can be added via a method, e.g., `registerBrowseFileFunction`. The method can do also state checking to disallow multiple registrants.

src/app/frontend/deploy/kdupload_controller.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Controller for the kd upload directive.
+ *
+ * @final
+ */
+export default class KdUploadController {
+  /** @ngInject */
+  constructor() {
+    /**
+     * It will be set by kdFileReaderDirective to the instance of an input element
```
bryk: Use present tense in comments. Such comments are easier to read and always true. This is what this project uses and internet generally suggests (though, as always, there are disputes).

bryk: Say in the comment what the property does, not who and how uses it. I.e., here you say who sets it. I am sure that when somebody changes implementation of kdFileReaderDirective they would not update this comment. This is why you should generally avoid talking about clients of your API and what they do.

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {!angular.$log} $log
+ * @return {!angular.Directive}
+ * @ngInject
+ */
+export default function kdFileReaderDirective($log) {
+  return {
+    restrict: 'A',
+    require: ['^kdUpload', 'ngModel'],
+    link: function(scope, element, attrs, ctrls) {
+      let kdUploadController = ctrls[0];
```
batikanu: @type {KdUploadController} is unknown. If I import it then I need to use it otherwise I get another style error. Therefore I haven't set the type yet. Do you have any suggestion?

src/app/frontend/deploy/kdfilereader_directive.js
```diff
@@ -0,0 +1,49 @@
...
+        /** @type {File} */
+        let file = changeEvent.target.files[0];
+        if (!file) {
+          $log.error('Error invalid file: ', file);
+          return;
+        }
+        let reader = new FileReader();
+        reader.onload = () => {
+          /** @type {!angular.NgModelController} */
+          let ngModelController = ctrls[1];
```
batikanu: It is also possible to use ctrls[1].$setViewValue(...) but I think using extra variable makes it more understandable although it is used only once.

batikanu:
PTAL

batikanu:
One Note: Upload dialog won't work on focus in Firefox because it doesn't let click to an element programmatically. While both on focus and upload button work in chrome, only upload button is functional in Firefox.

src/app/frontend/deploy/kdfilereader_directive.js
bryk: Yes, I really like your approach. With that you, me and the compiler know the type of the variable :)

src/app/frontend/deploy/kdfilereader_directive.js
bryk: Try `@type {!./kdupload_controller}`. This gets default export of the file. `@type {!./kdupload_controller.Foo}` gets export named Foo from the file.

bryk:
Can you add a TODO to investigate the Firefox issue ?

bryk:
Nice. Add the type information and let's merge :)

src/app/frontend/deploy/kdfilereader_directive.js
batikanu:
```
src/app/frontend/deploy/kdfilereader_directive.js:28: ERROR - Bad type annotation. Unknown type module$deploy$kdupload_controller
      /** @type {!./kdupload_controller} */
```
Compiler doesn't like it.

bryk: Ok. I see. I did investigation. The problem is that you cannot reference types that are `default export`. Can you add a todo to the code and create an issue describing the problem? We'll merge without solving it.

bryk:
Last thing that applies to entire PR: you prefixed all names of your files with `kd`. Can you strip this prefix? No other file uses this pattern. You can leave the prefix in directive names. We'll talk about this issue next week in WAW

batikanu:
Ok I will omit this prefix everywhere but in directive name. Unfortunately, I won't be able to attend the meeting next week but I would like to hear about the outcomes of the discussions.

batikanu:
PTAL

src/app/frontend/deploy/kdfilereader_directive.js
batikanu: See issue #159.

batikanu: I will add code reference to the issue after merge

bryk:
Great work @batikanu :) I'm merging.

batikanu:
Thank you :)

PR #120, opened 2015-12-03
Added go environment prerequisite checks, godep is now installed via npm

taimir:
Added additional checks of the prerequisites to executing the go command in dashboard. This includes:

* check if go is on the PATH (if not, probably the user ran with sudo and didn't pass his path with PATH=$PATH)
* check if godep is on the PATH
* from now on, running npm install also installs godep as a local project dependency (under /tools/go)

@bryk PTAL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ bower_components/
 coverage/
 dist/
 kubernetes/
+.tools/
 
 ## IDE specific files
 # IntelliJ
--- a/build/gocommand.js
+++ b/build/gocommand.js
@@ -17,38 +17,109 @@
  */
 import child from 'child_process';
 import lodash from 'lodash';
+import q from 'q';
 
 import conf from './conf';
 
+// Add base directory to the gopath so that local imports work.
+const sourceGopath = `${conf.paths.backendTmp}`;
+// Add the project's required go tools to the PATH.
+const devPath = `${process.env.PATH}:${conf.paths.goTools}/bin`;
+
 /**
- * Spawns Go process wrapped with the Godep command. Backend source files must me packaged with
- * 'package-backend-source' task before running this command.
+ * The environment needed for the execution of any go command.
+ */
+const env = lodash.merge(process.env, {GOPATH: sourceGopath, PATH: devPath});
+
+/**
+ * Spawns a Go process wrapped with the Godep command after making sure all GO prerequisites are
+ * present. Backend source files must be packaged with 'package-backend-source' task before running
+ * this command.
  *
- * @param {!Array<string>} args
- * @param {function(?Error=)} doneFn
+ * @param {!Array<string>} args - Arguments of the go command.
+ * @param {function(?Error=)} doneFn - Callback.
  */
 export default function spawnGoProcess(args, doneFn) {
-  // Add base directory to the gopath so that local imports work.
-  if (!process.env.GOPATH) {
-    doneFn(new Error('GOPATH is not set. Go process will not be spawned.'));
-    return;
-  }
+  checkPrerequisites().then(() => spawnProcess(args)).then(doneFn).fail((error) => doneFn(error));
+}
 
-  let sourceGopath = `${process.env.GOPATH}:${conf.paths.backendTmp}`;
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@ bower_components/
 coverage/
 dist/
 kubernetes/
+tools/
```
bryk: The number of auxiliary directories in our project grows every week. I'm wondering whether we should merge them into one and clean them on "gulp clean". What do you think? 

Not that current solution is bad. I'm just looking for improvements.

build/gocommand.js
```diff
@@ -17,38 +17,120 @@
...
+/** Checks if the GOPATH variable has been passed.
+ * @return {Q.Promise} A promise object.
+ */
+function checkGopath() {
+  let deferred = Q.defer();
   if (!process.env.GOPATH) {
-    doneFn(new Error('GOPATH is not set. Go process will not be spawned.'));
-    return;
+    deferred.reject(
+        new Error('GOPATH environment variable is not set. Go process will not be spawned.'));
```
bryk: Do we still require GOPATH? I believe it is not needed anymore.

build/gocommand.js
```diff
@@ -17,38 +17,120 @@
...
+ * @return {Q.Promise} A promise object.
+ */
+function checkPrerequisites() {
+  return checkGopath().then(checkGo).then(checkGodep);
+}
+
+/** Checks if the GOPATH variable has been passed.
+ * @return {Q.Promise} A promise object.
+ */
+function checkGopath() {
```
bryk: This function does not need to be async.

build/gocommand.js
```diff
@@ -17,38 +17,120 @@
  */
 import child from 'child_process';
 import lodash from 'lodash';
+import Q from 'q';
```
bryk: Style note: let's name this `q` and rename all capital Q to q in this file. This is how I see it used most often and it aligns with what is on the frontend. Small issue though

build/gocommand.js
```diff
@@ -17,38 +17,120 @@
...
+ * @param {!Array<string>} args - Arguments of the go command.
+ * @param {function(?Error=)} doneFn - Callback.
  */
 export default function spawnGoProcess(args, doneFn) {
   // Add base directory to the gopath so that local imports work.
+  let sourceGopath = `${process.env.GOPATH}:${
+      conf.paths.backendTmp}`;
+  let devPath = `${process.env.PATH}:${
+      conf.paths.goTools}/bin`;
+  env = lodash.merge(process.env, {GOPATH: sourceGopath, PATH: devPath});
```
bryk: Hmm... I see something wrong. You're modifying global variable here. Why dont you do this on global scope and make the var constant ?

build/gocommand.js
```diff
@@ -17,38 +17,120 @@
  */
 import child from 'child_process';
 import lodash from 'lodash';
+import Q from 'q';
 
 import conf from './conf';
+let env = null;
```
bryk: This needs documentation.

.gitignore
taimir: Since the godep tool is installed via "npm install", I would think it should remain even after gulp clean is executed. Do you think it would help if we make the directory hidden (e.g. .tools)?

build/gocommand.js
taimir: True, I will correct this.

build/gocommand.js
taimir: Should have occured to me earlier :) will do

.gitignore
bryk: You're correct. gulp clean should not clean godep. The thing that struck me now is why we download kubernetes cluster in gulp and godep in npm. Should there be this difference? Maybe unify this? (possibly a work for future PR)

taimir: Good point, I will see what I can do about this.

taimir:
PTAL

bryk:
Nice! Merging.

bryk:
Can you update README.md docs and travis.yml not to set up GOPATH?

taimir:
Sure, I will do it later today (At the university right now, no access to a dev machine).

PR #124, opened 2015-12-04
Backend api for replica set pods (pod contains containers)

zreigz:
Backend API to retrieve pods with containers for replica set. Needed for this view:
logmenu

### DIFF
--- a/src/app/backend/replicasetpods.go
+++ b/src/app/backend/replicasetpods.go
@@ -1,0 +1,87 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	api "k8s.io/kubernetes/pkg/api"
+	types "k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+)
+
+// Information about a Container that belongs to a Pod.
+type PodContainer struct {
+	// Name of a Container.
+	Name string `json:"name"`
+
+	// Container state.
+	State types.ContainerState `json:"state,omitempty"`
+
+	// Number of restarts.
+	RestartCount int `json:"restartCount"`
+}
+
+// List of pods that belongs to a Replica Set.
+type ReplicaSetPods struct {
+	// List of pods that belongs to a Replica Set.
+	Pods []ReplicaSetPodWithContainers `json:"pods"`
+}
+
+// Detailed information about a Pod that belongs to a Replica Set.
+type ReplicaSetPodWithContainers struct {
+	// Name of the Pod.
+	Name string `json:"name"`
+
+	// Time the Pod has started. Empty if not started.
+	StartTime *unversioned.Time `json:"startTime"`
+
+	// List of Containers that belongs to particular Pod.
+	PodContainers []PodContainer `json:"podContainers"`
+}
+
+// Returns list of pods with containers for the given replica set in the given namespace.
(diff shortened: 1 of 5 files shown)


zreigz:
PTAL

src/app/backend/replicasetpods.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+// Information about a Container that belongs to a Pod.
+type PodContainer struct {
+	// Name of a Container.
+	Name string `json:"name"`
+
+	// Container state.
+	State string `json:"state"`
```
bryk: Is it a string? Maybe can you use type from kubernetes/pkg/api/types.go? What are valid values?

src/app/backend/replicasetpods.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// List of pods that belongs to a Replica Set.
+type ReplicaSetPods struct {
+	// List of pods that belongs to a Replica Set.
+	Pods []ReplicaSetPodWithContainers `json:"pods"`
+}
+
+// Detailed information about a Pod that belongs to a Replica Set.
+type ReplicaSetPodWithContainers struct {
```
bryk: Is this name to avoid name collision inside the package?

src/app/backend/replicasetpods.go
```diff
@@ -0,0 +1,93 @@
...
+	*ReplicaSetPods, error) {
+
+	replicaSetPods := &ReplicaSetPods{}
+
+	replicaSet, err := client.ReplicationControllers(namespace).Get(name)
+	if err != nil {
+		return nil, err
+	}
+
+	labelSelector := labels.SelectorFromSet(replicaSet.Spec.Selector)
```
bryk: Can you share the code that gets list of pods for a replica set with replicasetdetails.go? Maybe extract to a separate file or something

src/app/backend/replicasetpods.go
```diff
@@ -0,0 +1,93 @@
...
+			StartTime: pod.Status.StartTime,
+		}
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			podContainer := PodContainer{
+				Name:         containerStatus.Name,
+				RestartCount: containerStatus.RestartCount,
+				State:        "Waiting",
+			}
+			if containerStatus.State.Running != nil {
+				podContainer.State = "Running"
```
bryk: There are constants for this in kubernetes/pkg/api/types.go, please reuse them.

bryk: I think I'm wrong with this comment.

bryk: Maybe can you do state checking for all three cases here. This would improve readability.

if state.Running != nil {

} else if (state.Terminated != nil) {

} else if (state.Waiting != nil) {

} else {
 /// panic 
}

src/app/backend/replicasetpods.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
```
bryk: Can you write tests for this file? Remember, we've said to write tests from now on.

bryk:
Nice job, PTAL

src/app/backend/replicasetpods.go
zreigz: yes I took this name to avoid collision inside the package. Can it stay like this ?

src/app/backend/replicasetpods.go
zreigz: yes sure. Good point

src/app/backend/replicasetpods.go
zreigz: Sure I can use type from kubernetes/pkg/api/types.go? Then I will  avoid comment regarding assigning this value.

src/app/backend/replicasetpods.go
zreigz: If I use type from kubernetes/pkg/api/types.go it will be unnecessary.

bryk:
Should I rereview?

zreigz:
No yet. I have problem to execute unit tests for backend in SDK. Can not find tested functions.
test

zreigz:
Now I realized the gulp task is for it so it should help solve the problem. I will let you know.

zreigz:
PTAL

PR #126, opened 2015-12-04
Update angular material to v 1.0-rc6

bryk:
This also fixes a few breaking changes.

PR #127, opened 2015-12-04
Events view draft

maciaszczykm:
Connected to https://github.com/kubernetes/dashboard/issues/104 issue. Implemented initial version of events view with @floreks. Still a lot of mocks inside, but we will continue to work on that.

Backend loads all event for replica set and pods connected to it using labels. I'm not yet 100% sure it's good solution, but it seems to be working as expected.

@floreks implemented left side sidebar, I was working on events view and partially updated replica set detail view.

Moreover, from important things - updated Angular Material Design to 0.12.0-rc1 version.

### DIFF
--- a/src/app/assets/styles/_color_variables.scss
+++ b/src/app/assets/styles/_color_variables.scss
@@ -1,0 +1,18 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO(floreks): Find appropriate place for global color definitions.
+$primary: #326DE6;
+$line: #AAAAAA;
+$subline: #888888;
--- a/src/app/backend/events.go
+++ b/src/app/backend/events.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	api "k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/fields"
@@ -60,10 +61,65 @@ type Event struct {
 	Reason string `json:"reason"`
 }
 
-// Return events for particular namespace or error if occurred.
-func GetEvents(client *client.Client, namespace string) (*Events, error) {
+// Return events for particular namespace and replica set or error if occurred.
+func GetEvents(client *client.Client, namespace string, replicaSetName string) (*Events, error) {
+	events := &Events{
+		Namespace: namespace,
+	}
+
+	// Get events for replica set.
+	rsEvents, err := GetReplicaSetEvents(client, namespace, replicaSetName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	AppendEvents(rsEvents, events)
+
+	// Get events for pods in replica set.
+	podEvents, err := GetReplicaSetPodsEvents(client, namespace, replicaSetName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	AppendEvents(podEvents, events)
+
+	return events, nil
+}
+
+// Gets events associated to replica set.
+func GetReplicaSetEvents(client *client.Client, namespace, replicaSetName string) ([]api.Event,
+	error) {
+	fieldSelector, err := fields.ParseSelector("involvedObject.name=" + replicaSetName)
+
+	if err != nil {
+		return nil, err
+	}
+
 	list, err := client.Events(namespace).List(unversioned.ListOptions{
 		LabelSelector: unversioned.LabelSelector{labels.Everything()},
+		FieldSelector: unversioned.FieldSelector{fieldSelector},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return list.Items, nil
+}
+
+// Gets events associated to pods in replica set.
+func GetReplicaSetPodsEvents(client *client.Client, namespace, replicaSetName string) ([]api.Event,
+	error) {
+	replicaSet, err := client.ReplicationControllers(namespace).Get(replicaSetName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	pods, err := client.Pods(namespace).List(unversioned.ListOptions{
+		LabelSelector: unversioned.LabelSelector{labels.SelectorFromSet(replicaSet.Spec.Selector)},
 		FieldSelector: unversioned.FieldSelector{fields.Everything()},
 	})
 
@@ -71,12 +127,37 @@ func GetEvents(client *client.Client, namespace string) (*Events, error) {
 		return nil, err
 	}
 
-	events := &Events{
-		Namespace: namespace,
+	events := make([]api.Event, 0, 0)
+
+	for _, pod := range pods.Items {
+		fieldSelector, err := fields.ParseSelector("involvedObject.name=" + pod.Name)
+
+		if err != nil {
+			return nil, err
+		}
+
+		list, err := client.Events(namespace).List(unversioned.ListOptions{
+			LabelSelector: unversioned.LabelSelector{labels.Everything()},
+			FieldSelector: unversioned.FieldSelector{fieldSelector},
+		})
+
+		if err != nil {
+			return nil, err
+		}
+
+		for _, event := range list.Items {
+			events = append(events, event)
+		}
+
 	}
 
-	for _, element := range list.Items {
-		events.Events = append(events.Events, Event{
+	return events, nil
+}
+
+// Appends events from source slice to target events representation.
+func AppendEvents(source []api.Event, target *Events) {
+	for _, element := range source {
+		target.Events = append(target.Events, Event{
 			Message:         element.Message,
 			SourceComponent: element.Source.Component,
 			SourceHost:      element.Source.Host,
@@ -87,6 +168,4 @@ func GetEvents(client *client.Client, namespace string) (*Events, error) {
 			Reason:          element.Reason,
 		})
 	}
-
-	return events, nil
 }
--- a/src/app/backend/replicasetdetail.go
+++ b/src/app/backend/replicasetdetail.go
@@ -15,8 +15,11 @@
 package main
 
 import (
+	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/fields"
+	"k8s.io/kubernetes/pkg/labels"
 )
 
 // Detailed information about a Replica Set.
@@ -44,6 +47,9 @@ type ReplicaSetDetail struct {
 
 	// Detailed information about Pods belonging to this Replica Set.
 	Pods []ReplicaSetPod `json:"pods"`
+
+	// Detailed information about service related to Replica Set.
+	Services []ServiceDetail `json:"services"`
 }
 
 // Detailed information about a Pod that belongs to a Replica Set.
@@ -61,6 +67,22 @@ type ReplicaSetPod struct {
 	NodeName string `json:"nodeName"`
 }
 
+// Detailed information about a Service connected to Replica Set.
+type ServiceDetail struct {
+	// Internal endpoints of all Kubernetes services that have the same label selector as connected
+	// Replica Set.
+	// Endpoint is DNS name merged with ports.
+	InternalEndpoint string `json:"internalEndpoint"`
+
+	// External endpoints of all Kubernetes services that have the same label selector as connected
+	// Replica Set.
+	// Endpoint is external IP address name merged with ports.
+	ExternalEndpoints []string `json:"externalEndpoints"`
+
+	// Label selector of the service.
+	Selector map[string]string `json:"selector"`
+}
+
 // Returns detailed information about the given replica set in the given namespace.
 func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	*ReplicaSetDetail, error) {
@@ -72,6 +94,15 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	replicaSet := replicaSetWithPods.ReplicaSet
 	pods := replicaSetWithPods.Pods
 
+	services, err := client.Services(namespace).List(unversioned.ListOptions{
+		LabelSelector: unversioned.LabelSelector{labels.Everything()},
+		FieldSelector: unversioned.FieldSelector{fields.Everything()},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
 	replicaSetDetail := &ReplicaSetDetail{
 		Name:          replicaSet.Name,
 		Namespace:     replicaSet.Namespace,
@@ -79,6 +110,12 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 		LabelSelector: replicaSet.Spec.Selector,
 		PodsRunning:   replicaSet.Status.Replicas,
 		PodsDesired:   replicaSet.Spec.Replicas,
+	}
+
+	matchingServices := getMatchingServices(services.Items, replicaSet)
+
+	for _, service := range matchingServices {
+		replicaSetDetail.Services = append(replicaSetDetail.Services, getServiceDetail(service))
 	}
 
 	for _, container := range replicaSet.Spec.Template.Spec.Containers {
@@ -96,4 +133,21 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	}
 
 	return replicaSetDetail, nil
+}
+
+func getServiceDetail(service api.Service) ServiceDetail {
+	var externalEndpoints []string
+	for _, externalIp := range service.Status.LoadBalancer.Ingress {
+		externalEndpoints = append(externalEndpoints,
+			getExternalEndpoint(externalIp.Hostname, service.Spec.Ports))
+	}
+
+	serviceDetail := ServiceDetail{
+		InternalEndpoint: getInternalEndpoint(service.Name, service.Namespace,
+			service.Spec.Ports),
+		ExternalEndpoints: externalEndpoints,
+		Selector:          service.Spec.Selector,
+	}
+
+	return serviceDetail
 }
--- a/src/app/frontend/index_config.js
+++ b/src/app/frontend/index_config.js
@@ -19,11 +19,15 @@
 export default function config($mdThemingProvider) {
   // Create a color palette that uses Kubernetes colors.
   let kubernetesColorPaletteName = 'kubernetesColorPalette';
+  let kubernetesAccentPaletteName = 'kubernetesAccentPallete';
   let kubernetesColorPalette = $mdThemingProvider.extendPalette('blue', {
     '500': '326de6',
   });
 
   // Use the palette as default one.
   $mdThemingProvider.definePalette(kubernetesColorPaletteName, kubernetesColorPalette);
-  $mdThemingProvider.theme('default').primaryPalette(kubernetesColorPaletteName);
+  $mdThemingProvider.definePalette(kubernetesAccentPaletteName, kubernetesColorPalette);
+  $mdThemingProvider.theme('default')
+      .primaryPalette(kubernetesColorPaletteName)
+      .accentPalette(kubernetesAccentPaletteName);
 }
--- a/src/app/frontend/replicasetdetail/replicasetdetail.html
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.html
@@ -14,27 +14,177 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div layout="column" layout-margin layout-align="center center">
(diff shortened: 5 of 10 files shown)


maciaszczykm:
Regarding the Angular Material Design version update, I see, that there is already pull request https://github.com/kubernetes/dashboard/pull/126, which do the same, but to newer version. If it will be merged before this, we may need to rebase and fix some UI connected problems.

src/app/backend/events.go
```diff
@@ -72,8 +73,27 @@ func GetEvents(client *client.Client, namespace string) (*Events, error) {
 		Namespace: namespace,
 	}
 
-	for _, element := range list.Items {
-		events.Events = append(events.Events, Event{
+	// Get events for replica set.
+	selector, _ := fields.ParseSelector("involvedObject.name=" + replicaSet.Name)
```
bryk: You ignore error here. Just add `if err != nil {` boilerplate everywhere. It is easy. With that we'll not forget about this in the future.

src/app/frontend/index.scss
```diff
@@ -17,4 +17,4 @@
 @import url(//fonts.googleapis.com/css?family=Roboto:300,400,500,700,400italic);
 
 // Angular Material icon font. It allows to use the icons as fonts.
-@import url(//fonts.googleapis.com/icon?family=Material+Icons);
+@import url(//fonts.googleapis.com/icon?family=Material+Icons);
```
bryk: Revert?

src/app/frontend/index_config.js
```diff
@@ -17,13 +17,23 @@
  * @ngInject
  */
 export default function config($mdThemingProvider) {
+  let kubeBlue = '326de6';
```
bryk: I see that the palette configuration object is the same. How about extracting it here to a common variable, instead of just the kubeBlue color?

bryk:
Wow, this is a lot of work. You know what? Let's polish the and do not add more features to it until it is merged. Then, when it is merged, we can split the work and pixel-perfect it. 

Always try to send small pull requests and often. With that we can have high quality code reviews and parallel our work. 

<I'm reviewing>

src/app/backend/events.go
bryk: Try to remove "involvedObject.name" magic string. See how kubectl does this kubernetes/pkg/kubectl/describe.go

```go
func (d *PodDescriber) Describe(namespace, name string) (string, error) {
	rc := d.ReplicationControllers(namespace)
	pc := d.Pods(namespace)

	pod, err := pc.Get(name)
	if err != nil {
		eventsInterface := d.Events(namespace)
		selector := eventsInterface.GetFieldSelector(&name, &namespace, nil, nil)
		options := unversioned.ListOptions{FieldSelector: unversioned.FieldSelector{selector}}
		events, err2 := eventsInterface.List(options)
		if err2 == nil && len(events.Items) > 0 {
			return tabbedString(func(out io.Writer) error {
				fmt.Fprintf(out, "Pod '%v': error '%v', but found events.\n", name, err)
				DescribeEvents(events, out)
				return nil
			})
		}
		return "", err
	}
```

src/app/backend/events.go
```diff
@@ -72,8 +73,27 @@ func GetEvents(client *client.Client, namespace string) (*Events, error) {
...
+	list, _ := client.Events(namespace).List(labels.Everything(), selector)
+	appendEvents(list.Items, events)
+
+	// Get events for pods in replica set.
+	labelSelector := labels.SelectorFromSet(replicaSet.Spec.Selector)
+	pods, _ := client.Pods(namespace).List(labelSelector, fields.Everything())
+	for _, pod := range pods.Items {
+		selector, _ := fields.ParseSelector("involvedObject.name=" + pod.Name)
```
bryk: Ditto

src/app/backend/replicasetdetail.go
```diff
@@ -73,13 +81,33 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 		return nil, err
 	}
 
+	services, err := client.Services(namespace).List(labels.Everything(), fields.Everything())
+
+	if err != nil {
+		return nil, err
+	}
+
+	matchingServices := getMatchingServices(services.Items, replicaSet)
```
bryk: I see similar code block in replica set details. Reuse?

src/app/frontend/index_config.js
```diff
@@ -17,13 +17,23 @@
...
+    '500': kubeBlue,
   });
 
   // Use the palette as default one.
   $mdThemingProvider.definePalette(kubernetesColorPaletteName, kubernetesColorPalette);
-  $mdThemingProvider.theme('default').primaryPalette(kubernetesColorPaletteName);
+  $mdThemingProvider.definePalette(kubernetesAccentPaletteName, kubernetesAccentPalette);
+  $mdThemingProvider.theme('default')
+      .primaryPalette(kubernetesColorPaletteName)
+      .accentPalette(kubernetesAccentPaletteName);
```
bryk: Nice, thanks!

maciaszczykm:
Okay, next time we'll try to split work into smaller pieces. This time all of it was connected so it was harder task, so we went easier path. Moreover, I was working together with @floreks so we were able to do more working at the same time.

I've fixed the issues and rebased. PTAL

src/app/backend/replicasetdetail.go
floreks: This part is taken and adjusted from replica set list in order to add details about internal/external endpoints. Should it be changed?

src/app/backend/events.go
maciaszczykm: Already done.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+              <i class="material-icons">arrow_downward</i>
+              Last seen
+              <i class="material-icons">help</i>
+            </th>
+          </tr>
+          </thead>
+          <tbody>
+          <tr ng-repeat="event in ctrl.replicaSetEvents.events">
+            <td class="kd-replicasetdetail-table-dark">
+              {{ event.message | limitTo: 50 }}{{event.message.length > 50 ? '...' : ''}}
```
bryk: You can do this with pure CSS. See text overload or how replica set list implemented this.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+            </td>
+            <td>{<!>{{event.sourceComponent}} {{event.sourceHost}}<!>}</td>
```
bryk: What is this `<!>`

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+                {{source}}
+              </md-option>
+            </md-select>
+          </md-input-container>
+        </div>
+        <table flex class="kd-replicasetdetail-table" cellspacing="0" cellpadding="15">
+          <thead>
+          <tr>
+            <th>Message
+              <i class="material-icons">help</i></th>
```
bryk: This needs better formatting.

bower.json
```diff
@@ -8,7 +8,7 @@
     "angular": "~1.4.2",
     "angular-animate": "~1.4.2",
     "angular-aria": "~1.4.2",
-    "angular-material": "~0.11.4",
+    "angular-material": "~0.12.0-rc1",
```
bryk: There are errors on the console on the replica set list page. Remove flex attr from one of the buttons.

src/app/frontend/replicasetdetail/replicasetdetail.html
maciaszczykm: I've added it to escape `{` and `}`, which are displayed in view. When `{{{` or `}}}` are used Angular raises some kind of errors, so I have splitted them by `<!>` comments. Is there a better solution for it?

src/app/backend/replicasetdetail.go
bryk: I mean, you can create a function that returns list of endpoints for service list. And then use this method in two places.

floreks: I did not get that. I will refactor the code.

src/app/backend/replicasetdetail.go
```diff
@@ -46,6 +46,14 @@ type ReplicaSetDetail struct {
 
 	// Detailed information about Pods belonging to this Replica Set.
 	Pods []ReplicaSetPod `json:"pods"`
+
+	// Internal endpoints of all Kubernetes services have the same label selector as this Replica Set.
+	// Endpoint is DNS name merged with ports.
+	InternalEndpoints []string `json:"internalEndpoints"`
```
bryk: I think this model needs a submodel for the service. I.e.., replicaSetDetail contains service information. Replica list contain only endpoints, but that's all they're going to display. Here we have detailed information and label selector for the service. Later we'll add name and possibly other things. I think it is worth to create a nested or named struct here.

What do you think?

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+        <h1 class="md-body-2">Replica Sets</h1>
+        <h1 class="md-caption kd-font-1">Labels</h1>
+        <h1 class="md-caption kd-font-2"><b>{{ctrl.formatLabelString(ctrl.replicaSetDetail.labels)}}</b></h1>
+        <h1 class="md-caption kd-font-1">Images</h1>
+        <h1 class="md-caption kd-font-2" ng-repeat="image in ctrl.replicaSetDetail.containerImages">
+          <b>{{image}}</b>
+        </h1>
+        <h1 class="md-body-2">Service</h1>
+        <h1 class="md-caption kd-font-1">Label selector</h1>
+        <h1 class="md-caption kd-font-2"><b>{{ctrl.formatLabelString(ctrl.replicaSetDetail.labels)}}</b></h1>
```
bryk: This is not true. You should get and display the label selector of the service.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+      </a>
+      <span flex-offset="5" flex class="kd-align-mid md-title">
+        {{ctrl.replicaSetDetail.name}}
+      </span>
+    </div>
+    <div flex-offset="5" flex layout-align="start start" layout-padding layout="row">
+      <div flex layout="column">
+        <h3 layout="row" class="kd-primary md-body-2">
+          <div flex="55">
+            <i class="kd-align-mid material-icons">refresh</i>
```
bryk: I is for italic text, not info icon. Use span for inline elements and div for blocks, if there's nothing better.

bower.json
bryk: I was wrong. This was caused by my stale bower install. Ignore.

src/app/backend/replicasetdetail.go
floreks: With more information from service it is reasonable to create dedicated structure. We can create ServiceDetail and place all that in there.

src/app/frontend/replicasetdetail/replicasetdetail.html
maciaszczykm: Yes, I will refactor it. There is still a lot to do with table styling, most of all with cells size.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -0,0 +1,80 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.kd-replicasetdetail-sidebar {
+  background-color: #FAFAFA;
+  min-width: 310px;
+}
+
+.kd-replicasetdetail-table td,
```
bryk: Do not style th or td. Just add a class to specific elements. With that you'll be able to remove !important from the next rule.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -0,0 +1,80 @@
...
+.kd-replicasetdetail-table th {
+  text-align: left;
+  color: #777777;
+  font-size: 13px;
+  border-bottom: 1px solid #CCCCCC;
+}
+
+.kd-replicasetdetail-table-dark {
+  color: #000000 !important;
```
bryk: See my previous comment.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -0,0 +1,80 @@
...
+
+.kd-help-sm {
+  font-size: 1.2em;
+}
+
+.kd-open-in-new-sm {
+  font-size: 1.2em;
+}
+
+md-input-container {
```
bryk: Can you use class instead of element name? Input container should have an interesting class

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+          <i class="kd-editpen-md material-icons kd-primary kd-align-mid">
+            mode_edit
+          </i>
+        </h1>
+        <h1 class="md-caption kd-font-1">Label selector</h1>
+        <h1 class="md-caption kd-font-2">
+          <b>{{ctrl.formatLabelString(ctrl.replicaSetDetail.labelSelector)}}</b>
+        </h1>
+        <h1 class="md-body-2">Replica Sets</h1>
+        <h1 class="md-caption kd-font-1">Labels</h1>
```
bryk: I see that you're using h1-s and apply class md-caption to every row. Instead, it is better to place your visual elements into semantic boxes. For example:
```html
<div class="my-detail">
   <span class="my-detail-title">Label selector</span>
  <span class="my-detail-line">{{ctrl.formatLabelString(ctrl.replicaSetDetail.labelSelector)}}</span>
</div>
```
... and repeat this for all parts. 

With this you dont tie yourself to implementation details ("kd-font-1"). You just say what things are in real life and then style them in css.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+        <h1 class="md-caption kd-font-2"><b>{{ctrl.formatLabelString(ctrl.replicaSetDetail.labels)}}</b></h1>
+        <h1 class="md-caption kd-font-1">Images</h1>
+        <h1 class="md-caption kd-font-2" ng-repeat="image in ctrl.replicaSetDetail.containerImages">
+          <b>{{image}}</b>
```
bryk: Dont use `<b>`. Apply this in CSS styles.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -14,27 +14,145 @@
...
+      <a flex-offset="10" ui-sref="replicasets">
+        <i class="material-icons kd-align-mid kd-font-lg">keyboard_backspace</i>
+      </a>
+      <span flex-offset="5" flex class="kd-align-mid md-title">
+        {{ctrl.replicaSetDetail.name}}
+      </span>
+    </div>
+    <div flex-offset="5" flex layout-align="start start" layout-padding layout="row">
+      <div flex layout="column">
+        <h3 layout="row" class="kd-primary md-body-2">
```
bryk: Why did you use md-body-2? Can you remove it?

bryk: Hmm... h3 is for content heading. I dont think this is content heading. This is button bar.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -0,0 +1,80 @@
...
+  min-height: 48px;
+  max-height: 48px;
+}
+
+.kd-primary {
+  color: #326de6;
+}
+
+.kd-font-1 {
+  color: #AAAAAA;
```
bryk: I see that this CSS contains like 5 colors. Can you reduce their number? If not, extract to constants to give meaning?

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -0,0 +1,80 @@
...
+}
+
+.kd-font-2 {
+  color: #888888;
+}
+
+.kd-editpen-md {
```
bryk: `-md` and `-sm` suffixes are used by material to indicate rules that work on specified screen sizes. You should avoid using it unless you do the same.

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
```diff
@@ -20,12 +20,38 @@
...
+    this.eventSource = this.eventSourceFilter[0];
+  }
+
+  /**
+   * Formats labels object to readable string.
+   * @param {!Object} object
+   * @return {string}
+   * @export
+   */
+  formatLabelString(object) {
```
bryk: Can you reuse this in the details page?

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: The reason why i'm using headers(h1, h2, etc) is that they have padding(top, bottom) added by default to them and combined with `md-body` `md-caption` i don't have to write css to add vertical spacing myself.

Also I was thinking about font coloring because I've done investigation if this can be tied to mdThemingProvider and it seems that there is no option to do that now. There were few workarounds for that, but they involve some custom services. `kd-font-1` class is in actually in `my-detail` css file but to have consistent font colors across project we could extract that somewhere higher.

src/app/frontend/replicasetdetail/replicasetdetail_state.js
```diff
@@ -67,3 +68,16 @@ export function resolveReplicaSetDetails($stateParams, $resource) {
...
 }
+
+/**
+ * @param {!StateParams} $stateParams
+ * @param {!angular.$resource} $resource
+ * @return {!angular.$q.Promise}
+ * @ngInject
+ */
+export function resolveReplicaSetEvents($stateParams, $resource) {
+  /** @type {!angular.Resource<!backendApi.ReplicaSetDetail>} */
```
bryk: This has wrong type

bryk:
Nice work, PTAL.

src/app/frontend/replicasetdetail/replicasetdetail_state.js
maciaszczykm: Fixed.

src/app/frontend/replicasetdetail/replicasetdetail.scss
maciaszczykm: Fixed.

src/app/frontend/replicasetdetail/replicasetdetail.scss
maciaszczykm: Fixed.

src/app/frontend/replicasetdetail/replicasetdetail.scss
maciaszczykm: Fixed.

src/app/frontend/replicasetdetail/replicasetdetail.scss
maciaszczykm: Fixed.

src/app/frontend/replicasetdetail/replicasetdetail.html
maciaszczykm: Fixed.

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
floreks: I can extract this to export function then import in second controller and assign to fn variable. Is it ok or do you have something else in mid?

```javascript
import {formatLabelString} from 'replicasetdetail/replicasetdetail_controller';
.... in controller
this.formatLabelString = formatLabelString;
```

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: Changed

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: All the icon examples on material icons are within `<i>` tag. Is there a difference if we use span instead of `i` for icons?

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
bryk: Hmm... Maybe just add a TODO in both places to keep this PR short.

src/app/frontend/replicasetdetail/replicasetdetail.html
bryk: Hmm... If the official spec says so. Let's keep that in mind and add to a
style guide when we have one.

bryk: So: keep `<i>`

maciaszczykm:
PTAL

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: I've proposed solution for that.

maciaszczykm:
PTAL

src/app/assets/styles/_color_variables.scss
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+
+// Global colors
+$primary: #326DE6;
+$line: #AAAAAA;
+$subline: #888888;
+
+// Replica Set Details related colors
+$replicasetdetails-sidebar-bg: #FAFAFA;
```
bryk: Move all things that are specific only to replica set details page to its css. Follow the rule: "I want to be able to delete replica set detail feature folder to delete all things related to it." 

Also, I'm not sure whether location of this file is correct and its purpose. Please add a todo and create an issue to decide where to put global style variables and how to generally do global css changes.

bryk:
Last comment and let's merge this

maciaszczykm:
Done. PTAL

floreks:
Related theming issue #139

PR #129, opened 2015-12-04
Fix post-install npm hook

bryk:
Previously only bottom half was being executed. This was caused by
duplicate keys. I've moved postinstall to a script.


taimir:
Thank you for noticing this!

PR #132, opened 2015-12-04
Initial validators for create namespace dialog

cheld:
DO NOT MERGE. I created this pull request just for discussion.

I think so far, in dashboard there is almost no validation. In this dialog I tried to validate everything.

Please note:
- I removed html5 validations with 'novalidate'
- There is some flickering when the validation error is shown. Seems to be a broken transition.
- I created a uniqueness check as a reusable directive. We have another uniqueness check implemented differently. Not sure which version is better.
- Where should be store validation patterns. I don't have a clear opinion on this.
- We still need proper server error handling. e.g. a duplicate namespace could be created concurrently (not realistic, but QA will test it).

### DIFF
--- a/src/app/frontend/deploy/createnamespace.html
+++ b/src/app/frontend/deploy/createnamespace.html
@@ -19,10 +19,21 @@ limitations under the License.
     <h4 class="md-title" >Create a new namespace</h4>
     <p>The new namespace will be added to the cluster.</p>
     <!-- TODO(maciaszczykm): Missing detailed cluster info and learn more link. -->
-    <form id="namespaceForm" ng-submit="ctrl.createNamespace()">
+    <form name="namespaceForm" ng-submit="ctrl.createNamespace()" novalidate>
       <md-input-container class="md-block">
         <label>Namespace name</label>
-        <input ng-model="ctrl.namespace" required>
+        <input name="namespace" ng-model="ctrl.namespace"
+               md-maxlength="{{ctrl.namespaceMaxLength}}"
+               ng-pattern="ctrl.namespacePattern"
+               reject="ctrl.namespaces"
+               required>
+        <div ng-messages="namespaceForm.namespace.$error">
+          <!-- TODO(cheld): better error message -->
+          <div ng-message="pattern">Name must be alphanumeric</div>
+          <div ng-message="md-maxlength">Name is too long</div>
+          <div ng-message="reject">Namespace already exists</div>
+          <div ng-message="required" ng-if="namespaceForm.namespace.$touched">Name is required</div>
+        </div>
       </md-input-container>
       <div class="md-actions" layout="row">
         <md-button ng-disabled="ctrl.isDisabled()" class="md-primary" type="submit">OK</md-button>
--- a/src/app/frontend/deploy/createnamespace_controller.js
+++ b/src/app/frontend/deploy/createnamespace_controller.js
@@ -43,9 +43,16 @@ export default class NamespaceDialogController {
 
     /** @export {string} */
     this.namespace = '';
+
+    //TODO(cheld) move to some central place or html?
+    this.namespaceMaxLength = 63;
+
+    this.namespacePattern = "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$";
+
   }
 
   /**
+   * TODO(cheld) clean up
    * Returns true if new namespace name hasn't been filled by the user, i.e, is empty.
    * @return {boolean}
    */
--- a/src/app/frontend/deploy/deploy_module.js
+++ b/src/app/frontend/deploy/deploy_module.js
@@ -15,6 +15,7 @@
 import stateConfig from './deploy_state';
 import deployFromSettingsDirective from './deployfromsettings_directive';
 import deployLabelDirective from './deploylabel_directive';
+import rejectDirective from './reject_directive';
 
 /**
  * Angular module for the deploy view.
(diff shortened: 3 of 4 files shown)


cheld:
@maciaszczykm please have a look

maciaszczykm:
@cheld The idea looks very good. 

Ad. 3. Directive looks good. I would consider changing its name to `unique` instead of `reject`. In my opinion it would explain itself better.

Ad. 4. In my opinion we should store them locally (in the same file), but if there are any patterns, that will be applicable globally then maybe it really should be stored in one file. I would start with first solution.

Ad. 5. I think we shouldn't consider it at the moment, at least till all other functionality is working. Moreover, checking if namespace already exist just before creating will slow down creation process. Instead of this we could just handle server error and display message to the user (only if namespace was created concurrently, because we know earlier, that namespace already exists). Perhaps we should only handle server errors, without prevalidation?

Also like you said before `Cancel` button has to be fixed (there is missing `type="reset"` attribute on it and one more issue), but I believe it's not the place for that.

cheld:
@maciaszczykm

Ad. 3. I have to find a better name - agree.

Ad. 4. I could think of the following locations:

- html
- controller
- externs/backendapi 

Did you suggest html or controller?

Ad. 5: I wanted to express that proper server handling is the must have feature. We could drop validation and just show the errors from the server (however, not user friendly). We might want to do both, but I am a bit tempted to say, let's start with proper handling of server errors and then make it user friendly...

maciaszczykm:
@cheld 

Ad. 4. I would pick controller.

Ad. 5. I agree.

cheld:
I am mostly familiar with JSF. The JSF framework suggests a centralized way to handle errors. I searched for something similar in Angular and Material, but could not find anything.


I played with Gmail, Drive yesterday and tried to understand the error handling in these applications. I was surprised, because I could not find any server error handling. It seems to me these applications try to avoid generic errors at any cost. The applications do validation on field basis. In case of an error situation the applications swallow the exception and try to continue in some way. However, if one follows unrealistic use cases it is rather easy to find questionable or buggy behavior in gmail or drive, e.g. pressing ok buttons has no effect - no error shown, or changes are applied with success message, but disappear after browser refresh...


So let's apply the Google-way to the create namespace dialog:

Scenario 1:
- User enters namespace name that already exists
- On focus out, error message is displayed below field

Scenario 2:
- User enters namespace name that does not exist yet
- Focus out => validation ok
- User does not press ok, yet
- 2nd user creates namespace with the input of user1
- 1st user presses ok button (validation is not repeated)
- REST call fails, but error is swallowed by UI.
- Dashoard shows success message, and switches to namespace created by 2nd user
=> 1st user thinks he created the namespace, but actually it was created by 2nd user

(Of course, scenario2 is not realistic at all, but I am still worried a bit about QA)


Let's summarize:
- We follow Google style error handling
- We do not show server errors
- We only validate
- The backend will always try to continue and try to solve the situation somehow, even if the result might be questionable

@bryk do you agree?

bryk:
> Scenario 1:
This sounds OK.

> Scenario 2:
I'd actually display every error whenever it occurred. Let it be error in namespace creation or error while deploying the application. The reason Gmail and Drive avoid displaying errors at any cost is that they are consumer grade apps targeting newbie users. On the other hand, take a look at cloud console (console.developers.google.com). It is enterprise grade app and has very different behavior. It displays every operation you make and shows every error that occur (save for a few request retries to try to avoid errors).

So I'd say:
* always show server errors
* validate on the UI (to have good user experience) 
* validate on the backend (to prevent malicious client from breaking us)
* If request is questionable, reject it. 

What do you think?

bryk:
The other thing that cannot let me sleep at night is how should we do validation. Should it be implemented on the server side, client side or both?

I'd really love us to have validation on the server side only and call it from the client. This would make us secure and always aligned, because there is only one implementation on the backend. However, I fear that this is huge implementation overhead to do this.

Do you have any ideas how this might work?

cheld:
@bryk sounds good to me

I would like to split the problem in three parts (I think your concern was only about part 2):

1. User presses 'OK' and the form data is sent to server. Exception occurs and is sent back to client/browser. I think the error response must have some structure, in order to show it properly, e.g. was the error expected or unexpected (500), maybe we want some more detailed error code and have to think whether the message should be localized in client or server

2. Standard validation checks (required, maxlength, pattern, ...) must be consistent in UI and server. We could define a kind-of 'Meta-Model / DSL' (or call it simple property file) that defines the validation rules. The file is interpreted by server and a custom UI directive. (I have done this once in an application with extremely complex validation rules. It worked nicely.)

3. Asynchronous validation. e.g. Uniqueness checks should be asynchronously on every focus out to the backend. My uniqueness check is only offline - this is not so nice.

cheld:
I tried console.developers.google.com. I renamed a project after deleting it (I have no content and had a bit of a problem to create an error with a less obscure scenario). A pop-up is shown. It feels like an internal error (500). The message is a mixture of english and german. I assume that part of the message is from the server and part from the client. I assume Tracking-Nummer is an id to find the exception in the log.



Serverfehler

There's a problem at our end.

Bitte versuchen Sie es erneut. Falls das Problem weiterhin auftritt, informieren Sie uns über den unten stehenden Link "Feedback geben"

Tracking-Nummer: 1324798376947020947

Feedback geben

bryk:
Does this PR need further attention? We can have a discussion session on validation during our meeting.

cheld:
I close it

PR #134, opened 2015-12-07
Update docu to not require godep on the global path anymore

taimir:
Godep is no longer a global dependency of the project.


bryk:
Awesome! Our developer's guide is getting better and better.

PR #142, opened 2015-12-08
Added tests for DeployController and CreateNamespaceController

floreks:
More tests for deploy controllers.

### DIFF
--- a/src/test/frontend/deploy/deploy_controller_test.js
+++ b/src/test/frontend/deploy/deploy_controller_test.js
@@ -1,0 +1,70 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/test/frontend/deploy/deploy_controller_test.js
```diff
@@ -0,0 +1,65 @@
...
+    // when
+    let result = ctrl.isDeployDisabled();
+
+    // then
+    expect(result).toBeFalsy();
+  });
+
+  it('should change state to zerostate view on cancel', () => {
+    // given
+    spyOn(ctrl.state_, 'go');
```
bryk: This is private variable. You can save `$state` from beforeEach and spy on it.

src/test/frontend/deploy/deploy_controller_test.js
```diff
@@ -0,0 +1,65 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import DeployController from 'deploy/deploy_controller';
+import DeployFromSettingController from 'deploy/deployfromsettings_controller';
+import deployModule from 'deploy/deploy_module';
+import {stateName as zerostate} from 'zerostate/zerostate_state';
+
+describe('Deploy controller', () => {
+  let ctrl;
```
bryk: How about you add `@type` information here, like `@type {!DeployController}`. This has benefits, because the variable is initialized later and it is non trivial to know what's the type only by reading name.

bryk:
Very nice! Two small comments re style.

floreks:
Done. PTAL

PR #150, opened 2015-12-10
Fixed a typo that prevented the proper configuration of kubectl

taimir:
Minor fix. Now the kubectl utility should be configured properly.


bryk:
Ouch. Thanks for noticing.

PR #151, opened 2015-12-10
Added functionality to delete replica set from events page

floreks:
Related to issue #104.
- [x] Delete button on replica set detail page will open a dialog and ask user if he wants to delete current replica set.
- [x] Added check on replica set list page to prevent it from showing when there are not replica sets. 
- [x] Added check to hide `Service` block on events page sidebar when there are no service related to replica set.

zrzut ekranu z 2015-12-11 09-44-00

zrzut ekranu z 2015-12-11 09-44-10

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -49,6 +49,9 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			To(apiHandler.handleGetReplicaSetDetail).
 			Writes(ReplicaSetDetail{}))
 	replicaSetWs.Route(
+		replicaSetWs.DELETE("/{namespace}/{replicaSet}").
+			To(apiHandler.handleDeleteReplicaSet))
+	replicaSetWs.Route(
 		replicaSetWs.GET("/pods/{namespace}/{replicaSet}").
 			To(apiHandler.handleGetReplicaSetPods).
 			Writes(ReplicaSetPods{}))
@@ -135,6 +138,21 @@ func (apiHandler *ApiHandler) handleGetReplicaSetDetail(
 	}
 
 	response.WriteHeaderAndEntity(http.StatusCreated, result)
+}
+
+// Handles delete Replica Set API call.
+func (apiHandler *ApiHandler) handleDeleteReplicaSet(
+	request *restful.Request, response *restful.Response) {
+
+	namespace := request.PathParameter("namespace")
+	replicaSet := request.PathParameter("replicaSet")
+
+	if err := DeleteReplicaSetWithPods(apiHandler.client, namespace, replicaSet); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	response.WriteHeader(http.StatusOK)
 }
 
 // Handles get Replica Set Pods API call.
--- a/src/app/backend/replicasetdetail.go
+++ b/src/app/backend/replicasetdetail.go
@@ -135,6 +135,29 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	return replicaSetDetail, nil
 }
 
+// TODO(floreks): This should be transactional to make sure that RC will not be deleted without
+// TODO(floreks): Should related services be deleted also?
+// Deletes replica set with given name in given namespace and related pods
+func DeleteReplicaSetWithPods(client *client.Client, namespace string, name string) error {
+	pods, err := getRawReplicaSetPods(client, namespace, name)
+	if err != nil {
+		return err
+	}
+
+	if err := client.ReplicationControllers(namespace).Delete(name); err != nil {
+		return err
+	}
+
+	for _, pod := range pods.Items {
+		if err := client.Pods(namespace).Delete(pod.Name, &api.DeleteOptions{}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Returns detailed information about service from given service
 func getServiceDetail(service api.Service) ServiceDetail {
 	var externalEndpoints []string
 	for _, externalIp := range service.Status.LoadBalancer.Ingress {
--- a/src/app/backend/replicasetlist.go
+++ b/src/app/backend/replicasetlist.go
@@ -97,7 +97,7 @@ func GetReplicaSetList(client *client.Client) (*ReplicaSetList, error) {
 func getReplicaSetList(
 	replicaSets []api.ReplicationController, services []api.Service) *ReplicaSetList {
 
-	replicaSetList := &ReplicaSetList{}
+	replicaSetList := &ReplicaSetList{ReplicaSets: make([]ReplicaSet, 0)}
 
 	for _, replicaSet := range replicaSets {
 		var containerImages []string
--- a/src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
+++ b/src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
@@ -12,18 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import stateConfig from './replicasetdetail_state';
-
 /**
- * Angular module for the Replica Set details view.
- *
- * The view shows detailed view of a Replica Sets.
+ * Opens replica set delete dialog.
+ * @param {!md.$dialog} mdDialog
+ * @param {!function()} deleteReplicaSetFn - callback
  */
-export default angular.module(
-                          'kubernetesDashboard.replicaSetDetail',
-                          [
-                            'ngMaterial',
-                            'ngResource',
-                            'ui.router',
-                          ])
-    .config(stateConfig);
+export default function showDeleteReplicaSetDialog(mdDialog, deleteReplicaSetFn) {
+  mdDialog.show(
+              mdDialog.confirm()
+                  .title('Delete Replica Set')
+                  .textContent(
+                      'All related pods will be deleted. Are you sure you want to' +
+                      ' delete this replica set? ')
+                  .ok('Ok')
+                  .cancel('Cancel'))
+      .then(deleteReplicaSetFn);
+}
--- a/src/app/frontend/replicasetdetail/replicasetdetail.html
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.html
@@ -26,15 +26,16 @@ limitations under the License.
     </div>
     <div flex-offset="5" flex layout-align="start start" layout-padding layout="row">
       <div flex layout="column">
-        <div layout="row" class="kd-replicasetdetail-sidebar-buttons">
-          <span flex="55">
-            <i class="material-icons kd-replicasetdetail-sidebar-icon">refresh</i>
+        <div layout="row" flex>
+          <md-button class="md-primary kd-replicasetdetail-sidebar-button">
+            <md-icon class="material-icons">refresh</md-icon>
             ROLLING UPDATE
-          </span>
-          <span flex-offset="5" flex>
-            <i class="material-icons kd-replicasetdetail-sidebar-icon">delete</i>
+          </md-button>
+          <md-button class="md-primary kd-replicasetdetail-sidebar-button"
+                     ng-click="ctrl.handleDeleteReplicaSetDialog()">
+            <md-icon class="material-icons">delete</md-icon>
             DELETE
-          </span>
+          </md-button>
         </div>
         <div flex layout="column" class="kd-replicasetdetail-sidebar-info">
           <span class="kd-replicasetdetail-sidebar-title">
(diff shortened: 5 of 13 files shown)


src/app/backend/apihandler.go
```diff
@@ -52,6 +52,9 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 		replicaSetWs.GET("/pods/{namespace}/{replicaSet}").
 			To(apiHandler.handleGetReplicaSetPods).
 			Writes(ReplicaSetPods{}))
+	replicaSetWs.Route(
```
bryk: Move it 3 lines up near to the same route with GET?

src/app/backend/replicasetdetail.go
```diff
@@ -135,6 +135,18 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	return replicaSetDetail, nil
 }
 
+// Deletes replica set with given name in given namespace
+func DeleteReplicaSet(client *client.Client, namespace string, name string) error {
+	err := client.ReplicationControllers(namespace).Delete(name)
```
bryk: Just `return client.Repli....`?

src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Opens replica set delete dialog.
+ * @param {!md.$dialog} mdDialog
+ * @param {!function() : void} deleteReplicaSetFn - callback
```
bryk: No need for ` : void`

src/app/backend/apihandler.go
floreks: Sure

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
```diff
@@ -55,4 +73,47 @@ export default class ReplicaSetDetailController {
...
+   */
+
+  /**
+   * Deletes replica set based on current replica set namespace and name.
+   * @private
+   */
+  deleteReplicaSet_() {
+    /** @type {!angular.Resource<!backendApi.ReplicaSetDetail>} */
+    let resource = this.resource_(
+        `/api/replicasets/${this.replicaSetDetail.namespace}/${this.replicaSetDetail.name}`);
```
bryk: Hmm... We're creating this resource object in two places. Do you have any ideas on how to have it only in one place? Or the REST string at least.

src/app/frontend/replicasetlist/replicasetlist_controller.js
```diff
@@ -28,10 +29,24 @@ export default class ReplicaSetListController {
    */
   constructor($state, replicaSets) {
     /** @export {!Array<backendApi.ReplicaSet>} */
-    this.replicaSets = replicaSets.replicaSets;
+    this.replicaSets = replicaSets.replicaSets || [];
```
bryk: Hmm... Are you sure that when there are no replica sets the array is null? If yes, can you modify backend to always return a non null array instead of being defensive here?

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
floreks: huh... this could be hard. Maybe there is an option to pass object from resolve and bind it to controller somehow.

src/app/frontend/replicasetlist/replicasetlist_controller.js
```diff
@@ -28,10 +29,24 @@ export default class ReplicaSetListController {
...
 
     /** @private {!ui.router.$state} */
     this.state_ = $state;
+
+    this.initState_();
```
bryk: Hmm... I'm not sure whether this functionality belongs to the controller of replica set list. What do you think on checking this condition in onEnter function in the replicasetlist_state and transition to zerostate there?

src/app/frontend/replicasetlist/replicasetlist_controller.js
floreks: Ye, I've had to change this because of undefined error when I was checking length in `initState_()` method. I will try to change this in backend.

src/app/frontend/replicasetlist/replicasetlist_controller.js
floreks: Good idea. I will check that.

src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+export default function showDeleteReplicaSetDialog(mdDialog, deleteReplicaSetFn) {
+  mdDialog.show(
+              mdDialog.confirm()
+                  .title('Delete Replica Set')
+                  .textContent('Are you sure you want to delete this replica set?')
+                  .ok('Yes, I\'m sure!')
```
bryk: What are defaults for `Ok` and `Cancel`? Can we use them? How about using 'Ok' and 'Cancel'? I'm asking because `Yes, I'm sure!` does not look like for enterprise product ;)

bryk:
Very nice, PTAL

src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
floreks: Totally agree. :) I knew this will have to be changed.

src/app/frontend/replicasetlist/replicasetlist_controller.js
floreks: Next thing to remember is to initialize all arrays that belong to some structure. Problem here was that `ReplicaSetList` has `ReplicaSets []ReplicaSet` array and in fact when it is not initialized using `replicaSetList.ReplicaSets = make([]ReplicaSet, 0)` then this is just null pointer.

src/app/frontend/replicasetlist/replicasetlist_state.js
```diff
@@ -32,15 +33,32 @@ export default function stateConfig($stateProvider) {
...
+    $timeout(() => {$state.go(zerostate);});
+  }
+}
+
+/**
  * @param {!angular.$resource} $resource
  * @return {!angular.$q.Promise}
  * @ngInject
  */
-export function resolveReplicaSets($resource) {
```
floreks: I think we do not need to export these function as they should not be used outside this file.

src/app/frontend/replicasetdetail/replicasetdetail_state.js
```diff
@@ -57,16 +58,23 @@ export default function stateConfig($stateProvider) {
...
+}
 
-  return resource.get().$promise;
+/**
+ * @param {!angular.Resource<!backendApi.ReplicaSetDetail>} replicaSetDetailResource
+ * @return {!angular.$q.Promise}
+ * @ngInject
+ */
+export function resolveReplicaSetDetails(replicaSetDetailResource) {
+  return replicaSetDetailResource.get().$promise;
```
floreks: @bryk: What do you think about this solution to keep resource endpoint initialization in one place?

src/app/frontend/replicasetlist/replicasetlist_state.js
floreks: @bryk: What do you think?

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
```diff
@@ -55,4 +79,44 @@ export default class ReplicaSetDetailController {
...
+  }
+
+  /**
+   * Changes state back to replica set list after successful deletion of replica set.
+   * @private
+   */
+  onReplicaSetDeleteSuccess_() {
+    this.log_.info('Replica set successfully deleted.');
+    // State name can not be imported. Related issue: #153
+    this.state_.go('replicasets');
```
floreks: Related issue: #153

floreks:
PTAL

bryk:
@kenmoore What should we do when deleting a replica set? Should we also delete services associated with it?

src/app/backend/replicasetlist.go
```diff
@@ -98,6 +98,7 @@ func getReplicaSetList(
 	replicaSets []api.ReplicationController, services []api.Service) *ReplicaSetList {
 
 	replicaSetList := &ReplicaSetList{}
+	replicaSetList.ReplicaSets = make([]ReplicaSet, 0)
```
bryk: Move this statement to initialization phase, i.e., `&ReplicaSetList{ReplicaSets = make([]ReplicaSet, 0)}`. Try to initialize in the init object whenever possible.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -57,27 +58,28 @@
...
-            </i>
-            <span ng-hide="service.internalEndpoint">none</span>
-          </div>
-          <span class="kd-replicasetdetail-sidebar-line">External endpoint</span>
-          <div class="kd-replicasetdetail-sidebar-subline"
-               ng-repeat="service in ctrl.replicaSetDetail.services">
+            <span class="kd-replicasetdetail-sidebar-line">Internal endpoint</span>
+            <div class="kd-replicasetdetail-sidebar-subline"
+                 ng-repeat="service in ctrl.replicaSetDetail.services">
+              <span ng-show="service.internalEndpoint">
```
bryk: Hmm... You do two ng-shows on the same condition. How about wrapping it with div and having ng-show there?

src/app/backend/replicasetdetail.go
```diff
@@ -135,6 +135,12 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	return replicaSetDetail, nil
 }
 
+// Deletes replica set with given name in given namespace
+func DeleteReplicaSet(client *client.Client, namespace string, name string) error {
+	return client.ReplicationControllers(namespace).Delete(name)
```
floreks: It looks like this is not enough. It only deletes Replication Controller without deleting even Pods. I have to take a look how `kubectl delete rc...` works.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -57,27 +58,28 @@
...
+              <i ng-show="service.internalEndpoint"
+                 class="material-icons kd-replicasetdetail-sidebar-service-icon">
+                help
+              </i>
+              <span ng-hide="service.internalEndpoint">none</span>
+            </div>
+            <span class="kd-replicasetdetail-sidebar-line">External endpoint</span>
+            <div class="kd-replicasetdetail-sidebar-subline"
+                ng-repeat="service in ctrl.replicaSetDetail.services">
             <span ng-show="service.externalEndpoints">
```
bryk: Same comment applies here.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -31,7 +31,8 @@
             <i class="material-icons kd-replicasetdetail-sidebar-icon">refresh</i>
             ROLLING UPDATE
           </span>
-          <span flex-offset="5" flex>
+          <span flex-offset="5" flex class="kd-replicasetdetail-sidebar-button"
```
bryk: Can we make this md-button? See angular material spec.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -84,6 +84,11 @@ $replicasetdetails-table-cell: #777777;
   font-size: 0.85em;
 }
 
+.kd-replicasetdetail-sidebar-button {
```
bryk: If the button was md-button, this would not be needed. Try doing this.

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: I've tried that but in my opinion it did not look good.

src/app/frontend/replicasetdetail/replicasetdetail_state.js
bryk: This is nice :) Let's apply this pattern now. How about discuss this pattern with other folks during our meeting in Warsaw?

src/app/frontend/replicasetlist/replicasetlist_state.js
bryk: Definitely. Maybe a quick fix in other files? I see this pattern applied everywhere else.

src/app/frontend/replicasetlist/replicasetlist_state.js
```diff
@@ -32,15 +33,32 @@ export default function stateConfig($stateProvider) {
...
 /**
+ * Avoids entering replica set list page when there are no replica sets.
+ * Used f.e. when last replica set gets deleted.
+ * Transition to: zerostate
+ * @param {!ui.router.$state} $state
+ * @param {!angular.$timeout} $timeout
+ * @param {!backendApi.ReplicaSetList} replicaSets
+ * @ngInject
+ */
+function initState($state, $timeout, replicaSets) {
```
bryk: `initState`? This function does not do any initialization :) It redirects to zero state when there are no replica sets. Apply name accordingly

src/app/frontend/replicasetlist/replicasetlist_state.js
```diff
@@ -32,15 +33,32 @@ export default function stateConfig($stateProvider) {
...
+  if (replicaSets.replicaSets.length === 0) {
+    // allow original state change to finish before redirecting to new state to avoid error
+    $timeout(() => { $state.go(zerostate); });
```
bryk: Hmmmm.... $timeout... Can we avoid this? 

Check out this: https://github.com/angular-ui/ui-router/blob/d23e9fab3ee27b31f0e1adc5ec4578fc6f8ecbc3/test/stateSpec.js

src/app/frontend/replicasetdetail/replicasetdetail.html
bryk: You can make it look like it looks now and be md-button: https://material.angularjs.org/latest/CSS/button

See flat buttons.

bryk:
Nice piece of work. A few style comments and we can merge.

src/app/frontend/replicasetlist/replicasetlist_state.js
floreks: I've been looking for a better solution but haven't found any. Related issue:

https://github.com/angular-ui/ui-router/issues/1434
https://github.com/angular-ui/ui-router/issues/1234

Here i've found this solution:
http://stackoverflow.com/questions/28603620/cannot-read-property-of-null-when-using-state-in-onenter

Unless we move it to controller as it was, I've no other clean solution for this.

src/app/frontend/replicasetlist/replicasetlist_state.js
floreks: Renamed. This name is indeed bad.

src/app/frontend/replicasetdetail/replicasetdetail_state.js
floreks: Sure :)

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: Unfortunately it still requires removing some padding from button to make it look good, but  at least it looks like a button now.

src/app/frontend/replicasetlist/replicasetlist_state.js
floreks: Done

src/app/backend/replicasetdetail.go
```diff
@@ -135,6 +135,27 @@ func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	return replicaSetDetail, nil
 }
 
+// Deletes replica set with given name in given namespace and related pods
+func DeleteReplicaSetWithPods(client *client.Client, namespace string, name string) error {
```
floreks: This has changed to delete also related pods.

bryk: Add a todo that we need to do this in a transaction :) Otherwise, it may happen that we delete RC and not its pods.

src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Opens replica set delete dialog.
+ * @param {!md.$dialog} mdDialog
+ * @param {!function()} deleteReplicaSetFn - callback
+ */
+export default function showDeleteReplicaSetDialog(mdDialog, deleteReplicaSetFn) {
+  mdDialog.show(
+              mdDialog.confirm()
+                  .title('Delete Replica Set')
+                  .textContent('Are you sure you want to delete this replica set?')
```
bryk: Add a note that this will delete pods too.

bryk:
Quick PTAL.

src/app/backend/replicasetdetail.go
floreks: Done

bryk:
I've tried running it and got errors:
```
.
http://localhost:9090/api/replicasets/default/test-app Failed to load resource: the server responded with a status of 405 (Method Not Allowed)
angular.js:12520 
Object
config: Object
data: "405: Method Not Allowed"
headers: (name)
status: 405
statusText: "Method Not Allowed"
__proto__: Object

```

src/app/frontend/replicasetdetail/deletereplicaset_dialog.js
floreks: Done

floreks:
Hmm... what version of kubernetes are you running? I'm testing this now and it works.

bryk:
The error is while deleting the RC. The error happens on gulp serve. On the other hand, on gulp serve:prod the details page do not load at all.

floreks:
On my side gulp serve works fine, on the other hand on gulp serve:prod page does not want to load also, but there are no errors.

bryk:
Re gulp serve:prod:
This
```
resolve: {
      replicaSetDetailResource: getReplicaSetDetailsResource,
      replicaSetDetail: resolveReplicaSetDetails,
      replicaSetEvents: resolveReplicaSetEvents,
    },
```

Needs to be changed to:
```
resolve: {
      'replicaSetDetailResource': getReplicaSetDetailsResource,
      'replicaSetDetail': resolveReplicaSetDetails,
      'replicaSetEvents': resolveReplicaSetEvents,
    },
```

Renaming... Can you ensure that all other files have this pattern too?

floreks:
Sure I will change. This is very weird behavior.

floreks:
It's green. Should be ok now. PTAL

bryk:
Please rebase and we'll merge.

floreks:
Rebased

bryk:
Thank you for thins work. Looks really nice :)

PR #152, opened 2015-12-10
gulp check for go minimum version is added

batikanu:
gulp check fails if go version is older than `1.5.0`

### DIFF
--- a/build/gocommand.js
+++ b/build/gocommand.js
@@ -18,6 +18,7 @@
 import child from 'child_process';
 import lodash from 'lodash';
 import q from 'q';
+import semver from 'semver';
 
 import conf from './conf';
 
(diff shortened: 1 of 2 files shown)


build/gocommand.js
```diff
@@ -76,6 +81,37 @@ function checkGo() {
...
+ */
+function checkGoVersion() {
+  let deferred = q.defer();
+  child.exec(
+      'go version',
+      {
+        env: env,
+      },
+      function(error, stdout) {
+        let match = /[\d\.]+/.exec(stdout.toString());  // matches version number
```
bryk: Use https://github.com/npm/node-semver to compare versions :) It'll avoid a lot of troubles. I.e., I think current implementation does string comparison and is wrong when comparing e.g., `1.11.0` and `1.5.0`.

bryk:
Thank you. One comment to use different lib.

build/gocommand.js
batikanu: That's right :) I've fixed it.

batikanu:
PTAL

bryk:
Forgot to push most recent commit?

batikanu:
Sorry test version number is still there. I change it now

batikanu:
It should be latest now

bryk:
Thanks!

PR #163, opened 2015-12-11
Move namespace select to deploy from settings

bryk:
The reason for doing this is the fact that namespace selection does not
make sense in case of uploading YAML. This should be only in deploy from
settings.

### DIFF
--- a/src/app/frontend/deploy/deploy_controller.js
+++ b/src/app/frontend/deploy/deploy_controller.js
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import {stateName as zerostate} from 'zerostate/zerostate_state';
-import showNamespaceDialog from 'deploy/createnamespace_dialog';
 
 /**
  * Controller for the deploy view.
@@ -25,11 +24,10 @@ export default class DeployController {
    * @param {!angular.$resource} $resource
    * @param {!angular.$log} $log
    * @param {!ui.router.$state} $state
-   * @param {!md.$dialog} $mdDialog
    * @param {!backendApi.NamespaceList} namespaces
    * @ngInject
    */
-  constructor($resource, $log, $state, $mdDialog, namespaces) {
+  constructor($resource, $log, $state, namespaces) {
     /** @export {!angular.FormController} Initialized from the template */
     this.deployForm;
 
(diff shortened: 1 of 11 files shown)


bryk:
@batikanu Can you review?

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -43,10 +41,19 @@ export default class DeployController {
     this.namespaces = namespaces.namespaces;
```
batikanu: Can't you move it to deployfromsetttings_controller too ?

src/app/frontend/deploy/deployfromsettings_controller.js
```diff
@@ -12,12 +12,13 @@
...
 
+import showNamespaceDialog from './createnamespace_dialog';
 import DeployLabel from './deploylabel';
 import {stateName as replicasetliststate} from 'replicasetlist/replicasetlist_state';
 
 // Label keys for predefined labels
-export const APP_LABEL_KEY = 'app';
-export const VERSION_LABEL_KEY = 'version';
+const APP_LABEL_KEY = 'app';
+const VERSION_LABEL_KEY = 'version';
```
batikanu: why don't we use `/** @const {string} */` ?

src/app/frontend/deploy/deploy_controller.js
```diff
@@ -43,10 +41,19 @@ export default class DeployController {
...
 
     /**
-     * Currently chosen namespace.
-     * @export {(string|undefined)}
+     * Contains the selected directive's controller which has its own deploy logic
+     *
+     * Initialized from the template.
+     * @export {{deploy:function()}|undefined}
+     */
+    this.detail;
```
batikanu: This looks better.

batikanu:
Looks quite good. I have only a few questions.

src/app/frontend/deploy/deploy_controller.js
batikanu: I can modify [upload_controller](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/deploy/upload_controller.js#L30) in the same manner. :)

bryk: Ok, I've done this already :)

src/app/frontend/deploy/deploy_controller.js
bryk: I wanted to do so, but there is no resolve function for directives. So I left list of namespaces here. I do see, however, a room for improvement. For example, we could use Angular 1.5 activate method for directives. Until then, moving it to directive initialization phase would make things more complicated, I think

src/app/frontend/deploy/deployfromsettings_controller.js
bryk: Because these are ES6 constants, so compiler already knows that they are `@const` (this was needed for ES5). Also, the compiler knows that this is a string, because the const is initialized when declared. So compiler already knows `/** @const {string} */`.

On the other hand, if it was `let APP_LABEL_KEY` we would need type annotations to make sure the type always stays the same.

So this is, basically, a syntactic sugar that allows us to define constants easily.

bryk:
PTAL

src/app/frontend/deploy/deploy_controller.js
batikanu: Thank you :)

batikanu:
LGTM

batikanu:
Is there any problem with travis?

bryk:
Travis flaked. This is a bad sign. But this happened for the first time.

PR #166, opened 2015-12-11
Stringified bindToController and locals properties

floreks:
These properties needs to be stringified to avoid further errors.


bryk:
Thank you for fixing this :) Can you rebase?

floreks:
Rebased

PR #167, opened 2015-12-11
Create namespace dialog bugfix (production)

taimir:
###  This is not meant to be merged immediately.
There is currently a bug when running 'gulp serve:prod' - the namespace creation dialog in the deploy view does not respond (the dialog window does not open). Now looking at it, this was because of the following two issues:

* Some of the referenced functions were actually not @exported, and hence their names were minified => they couldn't be found after compilation in production. I've added the respective @export annotations.
* After fixing the first issue, another problem appeared: apparently the $mdDialog.show() method has problems resolving the template url '/deploy/createnamespace.html' when the app is run in production mode.

Pasting the template directly as shown here works, and this is a quick-fix workaround. Nevertheless, could anyone propose a better method of handling this?

### DIFF
--- a/src/app/frontend/deploy/createnamespace_controller.js
+++ b/src/app/frontend/deploy/createnamespace_controller.js
@@ -48,17 +48,22 @@ export default class NamespaceDialogController {
   /**
    * Returns true if new namespace name hasn't been filled by the user, i.e, is empty.
    * @return {boolean}
+   * @export
    */
   isDisabled() {
     return !this.namespace || /^\s*$/.test(!this.namespace) ||
            this.namespaces.indexOf(this.namespace) >= 0;
   }
 
-  /** Cancels the new namespace form. */
+  /**
+   * Cancels the new namespace form.
+   * @export
+   */
   cancel() { this.mdDialog_.cancel(); }
 
   /**
    * Creates new namespace based on the state of the controller.
+   * @export
    */
   createNamespace() {
     /** @type {!backendApi.NamespaceSpec} */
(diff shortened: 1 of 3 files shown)


taimir:
@bryk @zreigz @maciaszczykm @floreks

src/app/frontend/deploy/createnamespace_dialog.js
```diff
@@ -28,7 +28,23 @@ export default function showNamespaceDialog(mdDialog, event, namespaces) {
     controllerAs: 'ctrl',
     clickOutsideToClose: true,
     targetEvent: event,
-    templateUrl: '/deploy/createnamespace.html',
+    template: `<md-dialog aria-label="Create a new namespace" layout="column" layout-padding>
```
floreks: Leave the html template file and just change `/deploy/createnamespace.html` to `deploy/createnamespace.html`. This will work.

floreks:
Good catch. Just add `@export` where it's missing and change template path. Everything should work.

taimir:
Ok. thank you :+1:

taimir:
PTAL

src/app/frontend/deploy/createnamespace_controller.js
```diff
@@ -58,6 +58,7 @@ export default class NamespaceDialogController {
   cancel() { this.mdDialog_.cancel(); }
 
   /**
+   * @export
    * Creates new namespace based on the state of the controller.
    */
```
floreks: Everywhere else `@export` and `@private` is the last line of comment so maybe switch this to be consistent.

floreks:
Add `@export` also to `isDisabled` and `cancel` methods of `NamespaceDialogController` as they are used in template also.

taimir:
OK, just made it consistent. BTW, I notice the empty lines between the comment text and annotations are not really consistent... Could we somehow unify this (agree on some rule)?

floreks:
I agree, we should write this things down in some dev guide.

bryk:
Thank you for this fix! We need to come up with a clever way to verify this during QA. In other projects I've worked on we required to send a live production demo of a change. But this is hard in open source... Let's talk about this tomorrow.

PR #174, opened 2015-12-14
Added functionality to edit number of replicas in replica set

floreks:
Related issue: #104 
Next button on events page.

zrzut ekranu z 2015-12-14 11-31-55

Refreshing number of pods in the UI after update is not implemented. This has to be discussed as process of creating/deleting pods is not immediate.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -50,6 +50,10 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 			To(apiHandler.handleGetReplicaSetDetail).
 			Writes(ReplicaSetDetail{}))
 	replicaSetWs.Route(
+		replicaSetWs.POST("/{namespace}/{replicaSet}/update/pods").
+			To(apiHandler.handleUpdateReplicasCount).
+			Reads(ReplicaSetSpec{}))
+	replicaSetWs.Route(
 		replicaSetWs.DELETE("/{namespace}/{replicaSet}").
 			To(apiHandler.handleDeleteReplicaSet))
 	replicaSetWs.Route(
@@ -139,6 +143,28 @@ func (apiHandler *ApiHandler) handleGetReplicaSetDetail(
 	}
 
 	response.WriteHeaderAndEntity(http.StatusCreated, result)
+}
+
+// Handles update of Replica Set pods update API call.
+func (apiHandler *ApiHandler) handleUpdateReplicasCount(
+	request *restful.Request, response *restful.Response) {
+
+	namespace := request.PathParameter("namespace")
+	replicaSetName := request.PathParameter("replicaSet")
+	replicaSetSpec := new(ReplicaSetSpec)
+
+	if err := request.ReadEntity(replicaSetSpec); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	if err := UpdateReplicasCount(apiHandler.client, namespace, replicaSetName,
+		replicaSetSpec); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	response.WriteHeader(http.StatusAccepted)
 }
 
 // Handles delete Replica Set API call.
--- a/src/app/backend/replicasetdetail.go
+++ b/src/app/backend/replicasetdetail.go
@@ -83,6 +83,12 @@ type ServiceDetail struct {
 	Selector map[string]string `json:"selector"`
 }
 
+// Information needed to update replica set
+type ReplicaSetSpec struct {
+	// Replicas (pods) number in replicas set
+	Replicas int `json:"replicas"`
+}
+
 // Returns detailed information about the given replica set in the given namespace.
 func GetReplicaSetDetail(client *client.Client, namespace string, name string) (
 	*ReplicaSetDetail, error) {
@@ -152,6 +158,24 @@ func DeleteReplicaSetWithPods(client *client.Client, namespace string, name stri
 		if err := client.Pods(namespace).Delete(pod.Name, &api.DeleteOptions{}); err != nil {
 			return err
 		}
+	}
+
+	return nil
+}
+
+// Updates number of replicas in Replica Set based on Replica Set Spec
+func UpdateReplicasCount(client client.Interface, namespace string, name string,
+	replicaSetSpec *ReplicaSetSpec) error {
+	replicaSet, err := client.ReplicationControllers(namespace).Get(name)
+	if err != nil {
+		return err
+	}
+
+	replicaSet.Spec.Replicas = replicaSetSpec.Replicas
+
+	_, err = client.ReplicationControllers(namespace).Update(replicaSet)
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/src/app/frontend/replicasetdetail/replicasetdetail.scss
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.scss
@@ -93,6 +93,14 @@ $replicasetdetails-warning: orange;
   font-size: 0.85em;
 }
 
+.kd-replicasetdetail-sidebar-editpods-button {
+  margin: 0;
+  min-width: 25px;
+  > md-icon {
+    outline: none;
+  }
+}
+
 .kd-replicasetdetail-options {
   margin: 15px;
 }
--- a/src/app/frontend/replicasetdetail/replicasetdetail_controller.js
+++ b/src/app/frontend/replicasetdetail/replicasetdetail_controller.js
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import showDeleteReplicaSetDialog from 'replicasetdetail/deletereplicaset_dialog';
+import showUpdateReplicasDialog from 'replicasetdetail/updatereplicas_dialog';
 
 // Filter type and source values for events.
 const EVENT_ALL = 'All';
@@ -34,11 +35,12 @@ export default class ReplicaSetDetailController {
    * @param {!backendApi.ReplicaSetDetail} replicaSetDetail
    * @param {!backendApi.Events} replicaSetEvents
    * @param {!angular.Resource<!backendApi.ReplicaSetDetail>} replicaSetDetailResource
+   * @param {!angular.Resource<!backendApi.ReplicaSetSpec>} replicaSetSpecPodsResource
    * @ngInject
    */
   constructor(
       $mdDialog, $state, $resource, $log, replicaSetDetail, replicaSetEvents,
-      replicaSetDetailResource) {
+      replicaSetDetailResource, replicaSetSpecPodsResource) {
     /** @export {!backendApi.ReplicaSetDetail} */
     this.replicaSetDetail = replicaSetDetail;
 
@@ -47,6 +49,9 @@ export default class ReplicaSetDetailController {
 
     /** @private {!angular.Resource<!backendApi.ReplicaSetDetail>} */
     this.replicaSetDetailResource_ = replicaSetDetailResource;
+
+    /** @private {!angular.Resource<!backendApi.ReplicaSetSpec>} */
+    this.replicaSetSpecPodsResource_ = replicaSetSpecPodsResource;
 
     /** @export !Array<!backendApi.Event> */
     this.events = replicaSetEvents.events;
@@ -143,6 +148,15 @@ export default class ReplicaSetDetailController {
   }
 
   /**
+   * Handles update of replicas count in replica set dialog.
+   * @export
+   */
+  handleUpdateReplicasDialog() {
+    showUpdateReplicasDialog(
+        this.mdDialog_, this.replicaSetDetail, this.updateReplicas_.bind(this));
+  }
+
+  /**
    * Handles replica set delete dialog.
    * @export
    */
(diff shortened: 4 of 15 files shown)


src/test/integration/zerostate/zerostate_test.js
```diff
@@ -18,7 +18,7 @@ describe('Zero state view', () => {
   let page;
 
   beforeEach(() => {
-    browser.get('/index.html');
+    browser.get('#/zerostate');
```
floreks: Fix of test. It passed on travis in old form, but if there were any replica sets existing then it would redirect to `replicasetlist` page instead of `zerostate`.

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
```diff
@@ -156,6 +165,19 @@ export default class ReplicaSetDetailController {
...
 
   /**
+   * Updates replicas count in replica set
+   * @param {number} replicasCount
+   * @param {function(!backendApi.ReplicaSetDetail)=} opt_callback
+   * @param {function(!angular.$http.Response)=} opt_errback
+   * @private
+   */
+  updateReplicas_(replicasCount, opt_callback, opt_errback) {
+    this.replicaSetDetail.podsDesired = replicasCount;
```
floreks: I'm not sure about this. Should we use this structure or create new one?

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
```diff
@@ -156,6 +165,19 @@ export default class ReplicaSetDetailController {
...
+    this.replicaSetDetailResource_.save(this.replicaSetDetail, opt_callback, opt_errback);
+    // TODO(floreks): Think about refreshing data on this page after update.
```
floreks: Could we use three-way data binding here?

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
bryk: Only properties that have changed should be sent back to the server. There is many reasons to do so. First, if you send entire object you can have race conditions with other clients that already modify this object (e.g., imagine that in another tab you changed the name of the replica set). Other thing is that backend method needs to be generic enough to diff against all properties, which is harder to implement and read than to write specialized methods for concrete use cases.

bryk: The open question what REST endpoint should this have? I leave this to you for investigation. Think of `/repicasets/:namespace/:name/pods/update/` or `/repicasets/pods/:namespace/:name/update/`

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
bryk: We potentially could, but dont do this in this PR. Remember, the smaller PRs the better :) Open an issue about how to synchronize changes on modification.

For a nice start we can do resource refresh after a few seconds. This should work in most cases.

src/app/backend/apihandler.go
```diff
@@ -140,6 +145,24 @@ func (apiHandler *ApiHandler) handleGetReplicaSetDetail(
 	response.WriteHeaderAndEntity(http.StatusCreated, result)
 }
 
+// Handles update of Replica Set API call.
+func (apiHandler *ApiHandler) handleUpdateReplicaSet(
+	request *restful.Request, response *restful.Response) {
+
+	replicaSetDetail := new(ReplicaSetDetail)
```
bryk: Hmm... You've got name and namespace in the URL params? Which one should you use?

src/app/backend/replicasetdetail.go
```diff
@@ -157,6 +157,23 @@ func DeleteReplicaSetWithPods(client *client.Client, namespace string, name stri
 	return nil
 }
 
+// Updates Replica Set based on Replica Set specification
+func UpdateReplicaSet(client *client.Client, replicaSetDetail *ReplicaSetDetail) error {
```
bryk: Can you write a test for this? Verify that update has been called with everything the same except for pods desired.

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
floreks: I don't plan to add this on this PR. I was just investigating possible solutions. Refreshing after fixed time does not work very well when we are scaling down number of pods. Cleaning up and terminating pods takes sometimes over 10s even on my local cluster. Scaling up on the other hand is fast.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -93,6 +93,14 @@ $replicasetdetails-warning: orange;
   font-size: 0.85em;
 }
 
+.md-button.kd-replicasetdetail-sidebar-editpods-button {
```
bryk: Do not prefix with .md-button. Just use class that you've defined. This is a best practice to always just use classes with names that describe object's function. With that you're not coupled with HTML structure, nor with the implementation (md-button here).

src/app/frontend/replicasetdetail/updatereplicas_controller.js
```diff
@@ -0,0 +1,77 @@
...
+
+    /** @private {!angular.$log} */
+    this.log_ = $log;
+  }
+
+  /**
+   * Executes callback function to update replicas count in replica set.
+   * @export
+   */
+  updateReplicasCount() {
```
bryk: Can you test this? That error/info has been logged and dialog is hidden. Should be easy.

src/app/frontend/replicasetdetail/updatereplicas_controller.js
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * @param {!md.$dialog} $mdDialog
+   * @param {!angular.$log} $log
+   * @param {!function(number, function(!backendApi.ReplicaSetDetail),
+   * function(!angular.$http.Response)=)} updateReplicasFn
+   * @ngInject
+   */
+  constructor($mdDialog, $log, updateReplicasFn) {
+    /** @export {number} */
+    this.replicas = 1;
```
bryk: Hmm... Why 1? Why not actual number?

bryk:
PTAL

src/app/frontend/replicasetdetail/updatereplicas_controller.js
floreks: Should it be current number of replicas as starting point?

src/app/frontend/replicasetdetail/updatereplicas_controller.js
floreks: I will add test for backend and frontend when this PR is polished.

src/app/backend/apihandler.go
floreks: Yes they are in URL and I will use them when I change current solution to send only information about what should be updated.

src/app/frontend/replicasetdetail/updatereplicas_controller.js
bryk: I'd actually say that this should be current value. To see what you're changing from. Or, at least, have empty input here  and display current number of replicas in the description. 

The reason is I don't think that `1` is any better than `7` or `12312`.

src/app/frontend/replicasetdetail/replicasetdetail_controller.js
floreks: You think it is a good idea to actually send `backendApi.ReplicaSetDetail` but only with update property filled? Endpoint could be `/replicasets/:namespace/:name/update/pods`. Last part can be changed to what is actually updated, so `pods`, `name` etc.

bryk: I'd go with creating a new struct specialized only for this task. This is very cheap in our current architecture to create a new custom endpoint, so go for it. In return, you get clear, testable and maintainable API that can do only the thing it is designed to do.

src/app/frontend/replicasetdetail/updatereplicas_controller.js
floreks: Ok, I think that there should be some description to say that f.e. `5` that is displayed there, is actually current number of pods, because without any information, displaying some number (even current) doesn't mean anything.

src/app/frontend/replicasetdetail/updatereplicas.html
```diff
@@ -0,0 +1,33 @@
...
+-->
+
+<md-dialog aria-label="Create a new namespace" layout="column" layout-padding>
+  <md-content layout-padding>
+    <h4 class="md-title" >Set desired number of pods</h4>
+    <p>Replica set webapp will be updated to reflect the desired count.</p>
+    <form ng-submit="ctrl.updateReplicasCount()">
+      <md-input-container class="md-block">
+        <label>Number of pods</label>
+        <input type="number" min="1" ng-model="ctrl.replicas" placeholder="{{ctrl.curReplicasNr}}"
```
floreks: I've displayed it as a placeholder. I thought about couple of short messages to display near that like `Current: 5`, but none is good enough for me.

floreks:
PTAL

src/app/frontend/replicasetdetail/updatereplicas_controller.js
bryk: Ok, should we go with empty input and show current number in the description?

src/app/backend/replicasetdetail.go
```diff
@@ -157,6 +163,24 @@ func DeleteReplicaSetWithPods(client *client.Client, namespace string, name stri
 	return nil
 }
 
+// Updates number of replicas in Replica Set based on Replica Set Spec
+func UpdateReplicasCount(client *client.Client, namespace string, name string,
```
bryk: Use `client client.Interface` and then pkg/client/unversioned/testclient to be able to test this.

src/app/frontend/replicasetdetail/updatereplicas.html
bryk: I'd display this information as plain text in the window content and have this input start with no value.

bryk:
Let's wait for @maciaszczykm, but I think you can start writing tests.

src/app/backend/replicasetdetail.go
floreks: Ok, so i think other functions in this file should also use this interface to be testable. Update this in this PR?

floreks:
Tests added. PTAL

src/app/backend/replicasetdetail.go
```diff
@@ -83,6 +83,12 @@ type ServiceDetail struct {
 	Selector map[string]string `json:"selector"`
 }
 
+// Information needed to update replica set
+type ReplicaSetSpec struct {
+	// Replicas(pods) number in replicas set
```
maciaszczykm: Typo - missing space.

floreks: Done

maciaszczykm:
There is one thing, which may improve user experience. After edit dialog is displayed, there is placeholder with current number of pods, for example `12` and when user clicks arrow up then value is changed to `1` instead of `13` which may be more user-friendly. It's default Angular behaviour, but maybe we can change it.

src/app/frontend/replicasetdetail/updatereplicas.html
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<md-dialog aria-label="Create a new namespace" layout="column" layout-padding>
+  <md-content layout-padding>
+    <h4 class="md-title" >Set desired number of pods</h4>
+    <p>Replica set webapp will be updated to reflect the desired count.</p>
```
maciaszczykm: I think, that instead of `webapp` we should use real replica set name.

maciaszczykm:
One minor thing missing and one question regarding user experience, everything else looks good to me.

floreks:
I agree that information about current pods should be more clear. Any sugesstions are appreciated.
Should it look like `Current number of pods: x`? Where should we put it?

src/app/frontend/replicasetdetail/updatereplicas.html
floreks: Fixed

src/app/frontend/replicasetdetail/updatereplicas_controller.js
bryk: Any update on this? I still see the placeholder and no response :)

bryk:
> I agree that information about current pods should be more clear. Any sugesstions are appreciated.
Should it look like Current number of pods: x? Where should we put it?

How about you display info about current running and desired number of pods?

floreks:
I'm not convinced if this looks ok. What do you think?

zrzut ekranu z 2015-12-18 10-08-52

maciaszczykm:
@floreks For sure it's more user-friendlynow . What about @bryk's idea (to display desired number of pods)? It would give some kind of feedback about operation result to the user.

floreks:
I have 2 options that I think look ok and do not overwhelm rest of the dialog.

1.
zrzut ekranu z 2015-12-18 10-45-15

2. Works on mouse focus
zrzut ekranu z 2015-12-18 10-45-38

What do you think @maciaszczykm @bryk?

bryk:
How about adding "Current status: 5 running, 5 desired." under the normal text. The problem with hover is that this is not discoverable, i.e., how can you know that you'd get this info there? The problem with placeholder is that is is intended for example value, not documentation. And in material design, placeholders convert into labels once you click on them.

floreks:
How about now? @maciaszczykm @bryk 
zrzut ekranu z 2015-12-18 11-06-23

PS. Placeholder is converted into label only if there is no label defined before. This way it would disappear only when someone writes something in the input field, but I agree that there should be only example value shown there if it's used.

bryk:
LGTM

bryk:
@maciaszczykm Are you done with your review?

maciaszczykm:
@bryk Yes, LGTM.

bryk:
Merging. This is a nice piece of work.

PR #179, opened 2015-12-17
Validation design doc

cheld:
I will create three quick documents about the non-functional topics that have been discussed in the meeting. 

This one is about validation. I skipped the fancy topics, including async uniqueness check for the moment.

When we add the validation to the html a lot of hand-crafted ok-button enablement rules can be removed from the controllers. Also, most of the unit tests can be removed (or don't have to be written).

@floreks CC


cheld:
Sorry, wrong commit size

PR #180, opened 2015-12-17
Add command and command args to the deploy form

bryk:
The UI is as specified in the mocks, except for the names of the fields.
I decided to go with what is in the API.

Everything is extensively tested. I also added some tests for previously
untested cases.

### DIFF
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -32,6 +32,13 @@ type AppDeploymentSpec struct {
 	// Docker image path for the application.
 	ContainerImage string `json:"containerImage"`
 
+	// Command that is executed instead of container entrypoint, if specified.
+	ContainerCommand *string `json:"containerCommand"`
+
+	// Arguments for the specified container command or container entrypoint (if command is not
+	// specified here).
+	ContainerCommandArgs *string `json:"containerCommandArgs"`
+
 	// Number of replicas of the image to maintain.
 	Replicas int `json:"replicas"`
 
@@ -43,7 +50,7 @@ type AppDeploymentSpec struct {
 	IsExternal bool `json:"isExternal"`
 
 	// Description of the deployment.
-	Description string `json:"description"`
+	Description *string `json:"description"`
 
 	// Target namespace of the application.
 	Namespace string `json:"namespace"`
@@ -77,8 +84,11 @@ type Label struct {
 // App deployment consists of a replication controller and an optional service. Both of them share
 // common labels.
 // TODO(bryk): Write tests for this function.
-func DeployApp(spec *AppDeploymentSpec, client *client.Client) error {
-	annotations := map[string]string{DescriptionAnnotationKey: spec.Description}
+func DeployApp(spec *AppDeploymentSpec, client client.Interface) error {
+	annotations := map[string]string{}
+	if spec.Description != nil {
+		annotations[DescriptionAnnotationKey] = *spec.Description
+	}
 	labels := getLabelsMap(spec.Labels)
 	objectMeta := api.ObjectMeta{
 		Annotations: annotations,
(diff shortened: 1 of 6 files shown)


bryk:

src/app/backend/deploy.go
```diff
@@ -86,13 +96,23 @@ func DeployApp(spec *AppDeploymentSpec, client *client.Client) error {
...
+		Name:  spec.Name,
+		Image: spec.ContainerImage,
+	}
+
+	if spec.ContainerCommand != nil {
+		containerSpec.Command = []string{*spec.ContainerCommand}
+	}
+
+	if spec.ContainerCommandArgs != nil {
+		containerSpec.Args = []string{*spec.ContainerCommandArgs}
```
floreks: As backend actually has array of args and array of commands shouldn't we support it? Now there is array with 1 element only.

src/test/frontend/deploy/deployfromsettings_controller_test.js
```diff
@@ -104,4 +108,53 @@ describe('DeployFromSettings controller', () => {
...
+    ctrl.containerCommand = 'command';
+    ctrl.containerCommandArgs = 'commandArgs';
+    ctrl.description = 'desc';
+
+    // when
+    ctrl.deploy();
+
+    // then
+    expect(resourceObject.save).toHaveBeenCalled();
+  });
```
floreks: Nice tests! I've been looking for a good way to mock $resource.

floreks:
LGTM. Just one question.

src/test/frontend/deploy/deployfromsettings_controller_test.js
```diff
@@ -104,4 +108,53 @@ describe('DeployFromSettings controller', () => {
 
     expect(result).toEqual(expected);
   });
+
+  it('should not send empty description and container commands', () => {
```
cheld: From my understanding the description states: unmodified fields should not be sent as empty strings "", but stay as null value. Took me a while to understand the difference, but ok

cheld:
LGTM

src/app/backend/deploy.go
bryk: I know. I've been thinking about this. The reason I decided to go with just a string for our API is that this is presentation layer API, so it should be as close to the presentation layer as possible. It is implementation detail, that the actual backend uses array of commands. The thing that matters is that the UI has just input for one command.
When we change the UI to allow for multiple commands, then updating the presentation layer is OK.

bryk: Does this make sense?

src/test/frontend/deploy/deployfromsettings_controller_test.js
bryk: Thanks :)  Feel free to follow this pattern.

src/app/backend/deploy.go
floreks: For the command this makes sense but what about arguments?
When we provide multiple arguments will kubernetes manage it as only 1 element instead of multiple arguments split to multiple elements?

src/test/frontend/deploy/deployfromsettings_controller_test.js
bryk: This test is specifically about empty fields (or, call it, _falsey_). They can be modified, and then erased to empty values.

bryk: But rewrote the description a little

src/app/backend/deploy.go
bryk: The arguments are split by whitespaces: https://github.com/appc/docker2aci/pull/30

Does this solve the problem? I think so

floreks: Yep. It's ok.

PR #182, opened 2015-12-18
added simple validation rules to create namespace dialog

cheld:
I added the following validation rules:
- max length
- pattern
- required

I removed the following logic from the controller
- required check
- pattern check (btw: pattern did not match)

I removed one test because the logic was removed from the controller.

I intentionally skipped the uniqueness check to keep this PR smaller.

I also created a short design doc for validation. I will add in a separate PR.

@maciaszczykm CC


maciaszczykm:
Is there any way to display error message earlier? I mean, messages like `Name must be alphanumeric` are displayed when text field loses focus (at least in the first try before losing focus, in next ones it works fine).

Besides issue mentioned above LGTM.

cheld:
hmm, interesting. I think this is intentional. The question how would you feel on a larger dialog like deploy? I will check on some other sites like GMail

src/test/frontend/deploy/createnamespace_controller_test.js
```diff
@@ -26,14 +26,6 @@ describe('Deploy controller', () => {
     });
   });
 
-  it('should return true if namespace is empty', () => {
```
bryk: Hmm... Can you test namespacePattern regex? I know this is a constant, but with regexes I'm always worried that they are wrong. Just assert a few positive and negative cases.

src/app/frontend/deploy/createnamespace.html
```diff
@@ -19,10 +19,18 @@
     <h4 class="md-title" >Create a new namespace</h4>
     <p>The new namespace will be added to the cluster.</p>
     <!-- TODO(maciaszczykm): Missing detailed cluster info and learn more link. -->
-    <form id="namespaceForm" ng-submit="ctrl.createNamespace()">
+    <form name="namespaceForm" ng-submit="ctrl.createNamespace()" novalidate>
       <md-input-container class="md-block">
         <label>Namespace name</label>
-        <input ng-model="ctrl.namespace" required>
+        <input name="namespace" ng-model="ctrl.namespace"
+            md-maxlength="{{ctrl.namespaceMaxLength}}"
```
bryk: What's the difference between md-maxlength and ng-maxlength?

src/app/frontend/deploy/createnamespace_controller.js
```diff
@@ -43,6 +43,18 @@ export default class NamespaceDialogController {
...
+     * Max-length validation rule for namespace
+     * @export {number}
+     */
+    this.namespaceMaxLength = 63;
+
+    /**
+     * Pattern validation rule for namespace
+     * @export {string}
+     */
+    this.namespacePattern = "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$";
```
bryk: Can you have a RegEx instead of a string here? This would be faster and less error prone. Docs say that you can gave regex here: https://code.angularjs.org/1.4.7/docs/api/ng/directive/input

src/app/frontend/deploy/createnamespace_controller.js
```diff
@@ -50,10 +62,7 @@ export default class NamespaceDialogController {
    * @return {boolean}
    * @export
    */
-  isDisabled() {
-    return !this.namespace || /^\s*$/.test(!this.namespace) ||
-           this.namespaces.indexOf(this.namespace) >= 0;
-  }
+  isDisabled() { return this.namespaces.indexOf(this.namespace) >= 0; }
```
bryk: Can you test this method?

bryk:
I like this change a lot! A few minor style comments before we merge

src/test/frontend/deploy/createnamespace_controller_test.js
cheld: +1

good for documentation, too

src/app/frontend/deploy/createnamespace.html
cheld: ng-maxlength limits the input field. The user simply cannot type any longer after reaching the limit

md-maxlength shows a small hint below the input (e.g. 7/63) and a error in case the limit is exceeded

I think the md-maxlength is more user-friendly

src/app/frontend/deploy/createnamespace_controller.js
cheld: Done.

src/app/frontend/deploy/createnamespace_controller.js
cheld: Test case already exists. Anyway, I cleaned them up a bit.

cheld:
So, I compared the validation with Gmail. In gmail fields are only validated on focus out.

In our case (md) the fields are first validated on focus out and then on every input change. I think behavior is good.

Howerver, the behavior gets strange if multiple errors are involved in some way. Then, sometimes the error message is not shown. E.g. follow link below and type some non-sense into the postal code field

https://material.angularjs.org/latest/demo/input 

I seems to be a bit buggy. Nevertheless, I would suggest to move forward and investigate in this problem later

maciaszczykm:
@cheld I agree, let's work on it later. LGTM

cheld:
PTAL

src/app/frontend/deploy/createnamespace.html
bryk: Yes, that sounds better.

bryk:
@cheld You need to wrap `createNamespace` method with

```js
if (!this.isDisabled()) {
 ... <content> 
}
```

This is because you do validation on blur. The button may not yet be deactivated when the name is invalid. Please also test this :)

cheld:
@bryk You are right. It is a general question how to handle validation errors. 

1. Disable the submit button. 
2. Do nothing when pressing submit button. 

Gmail is following number 2 (https://accounts.google.com/SignUp). I think it is good usability pattern.

However, we end up with a mixed behavior until all input fields are refactored.

I tried three different implementations. They all work, but non I am fully satisfied. Versions:

* extend the html like this:
   `ng-submit="namespaceForm.namespace.$valid && ctrl.createNamespace()"`
* inject scope to the controller like this:
  `createNamespace() {
    if (!scope_.namespaceForm.namespace.$valid) return;`
* check if fields are undefined (because angular does not set value in case of error) like this:
 `createNamespace() {
    if (namespace===undefined) return;`

The last one might become difficult on larger forms to keep correct. Marcin suggested not to access scope from controller. Therefore, I chose version 1.

I started to draft some tests for createNamespace(), but I would like to keep them in a separate PR

cheld:
PTAL

bryk:
> inject scope to the controller like this: createNamespace() { if (!scope_.namespaceForm.namespace.$valid) return;

You can name the form "ctrl.namespaceForm" and then use it in controller as any other template initialized variable `/** @export {!angular.FormController} */ this.namespaceForm;` and then `if (this.namespaceForm.$valid) { ... } `. 

What do you think?

cheld:
Thanks for pushing me over the finishing line - maybe :)  I was not required to prefix the form name with "ctrl." 

PTAL

PR #187, opened 2015-12-18
Lint project SCSS stylesheet files

bryk:
This change adds linter for styles to the project and updates all styles
to conform with the default set of rules.

This should help us maintain a little bit more consistency in our
styles.


bryk:
@taimir @maciaszczykm Can you review?

taimir:
Running `gulp lint-styles` only produces a warning if e.g. the indentation of a rule is wrong, but the gulp task runs to completion. Is this intended?

taimir:
Looking at the default sass-lint rules on: https://github.com/sasstools/sass-lint/blob/master/lib/config/sass-lint.yml - it seems that all are set to warning (1) per default. If we want to force the build to break, they need to be redefined and set to 2. I do not know what the initial intention was, though :)

src/app/frontend/deploy/helpsection/userhelp.scss
```diff
@@ -11,22 +11,24 @@
...
-}
+  a {
+    color: $primary;
+    opacity: 1;
+    text-decoration: none;
 
-.kd-user-help b {
-  color: black;
-}
+    :hover {
```
taimir: This nested rule does not appear in the browser (Chrome / Firefox), neither in development, nor in production. Since the syntax is new to me, I cannot directly see the cause for the failure.

src/app/frontend/replicasetlist/replicasetlist.scss
```diff
@@ -11,28 +11,30 @@
...
+  min-height: inherit;
+  min-width: inherit;
+  padding: inherit;
+  text-decoration: none;
   text-transform: inherit;
-}
 
-.kd-replicaset-card-logs:after {
-  content: '\25BC';
+  :after {
```
taimir: Same issue as above, this is just nowhere to be seen in the browser. The nested rules which contain direct element selectors seem to function, but those with : do not.

taimir:
O.K., that's all I can see. The rest LGTM, AFAICT. Great style improvement :)

src/app/frontend/deploy/helpsection/userhelp.scss
```diff
@@ -11,22 +11,24 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-@import "../../../assets/styles/color_variables";
+@import '../../../assets/styles/color_variables';
```
maciaszczykm: Isn't a new line missing here?

src/app/frontend/replicasetlist/replicasetlist.scss
maciaszczykm: It works for me:

Logs arrow

Tested on Chrome with `gulp serve:prod`.

maciaszczykm:
@bryk Left two comments, rules for stylesheets will for sure improve code. LGTM

bryk:
> Running gulp lint-styles only produces a warning if e.g. the indentation of a CSS rule is wrong, but the gulp task runs to completion. Is this intended?

@taimir Yes. With this you can do development with ugly code, but you cannot merge it into master - travis will fail. In order to merge with master, you need to prettify the code. This is intended :)

src/app/frontend/deploy/helpsection/userhelp.scss
bryk: Thank you for pointing this. Indeed, I made mistake. The correct syntax is: `&:hover`. Fixed. Source: http://sass-lang.com/documentation/file.INDENTED_SYNTAX.html

src/app/frontend/replicasetlist/replicasetlist.scss
bryk: Thank you for pointing, this. Fixed. See my previous comment for explanation.

src/app/frontend/deploy/helpsection/userhelp.scss
bryk: Yes, fixed.

bryk:
> Looking at the default sass-lint rules on: https://github.com/sasstools/sass-lint/blob/master/lib/config/sass-lint.yml - it seems that all are set to warning (1) per default. If we want to force the build to break, they need to be redefined and set to 2. I do not know what the initial intention was, though :)

@taimir You're correct. This is only warnings. I think it is better to promote everything to errors, as you suggest. So I copied the default config file and changed all warnings to errors. There was no other way to promote warnings to errors.

bryk:
Thank you guys for the review. I've addressed all comments. Will merge soon.

PR #188, opened 2015-12-18
Changed default state to replica set list.

floreks:
Changed state transition when clicking k8s logo and on deploy form cancel to replica set list page.


bryk:
LGTM, merging

PR #190, opened 2015-12-21
Use compiler's externs for angular.$resource

bryk:
They just have been recently added. It allows us to delete some code.


bryk:

cheld:
Nice. LGTM

PR #195, opened 2015-12-22
Column layout for replica set list view

bryk:
This PR implements column layout for the replica set list view. The
layout is implemented through a directive that puts its transcluded
elements into a column layout.

There's some algorithmic code in this PR, so please comment if something
is not tested or easy to understand.

Please also verify on your machines that this works correctly.

### DIFF
--- a/src/app/frontend/replicasetlist/replicasetlistcontainer.js
+++ b/src/app/frontend/replicasetlist/replicasetlistcontainer.js
@@ -1,0 +1,132 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Computes optimal height of the given container that will fit all its child elements into N
+ * column layout. N is based on current media query from the $mdMedia service.
+ *
+ * @param {!Element} container Container element that has child elements to be arranged in columns.
+ *     It should be a flexbox container with columns wrapped.
+ * @param {function(string):boolean} mdMedia Angular Material $mdMedia service
+ * @return {number}
+ */
+export default function computeContainerHeight(container, mdMedia) {
+  /** @type {!Array<number>} */
+  let childHeights = Array.prototype.map.call(container.children, (child) => child.offsetHeight);
+
+  let columnCount = computeColumnCount(mdMedia);
+  return binarySearchOptimalHeight(childHeights, columnCount);
+}
+
+/**
+ * Returns optimal number of columns for current window size.
+ *
+ * @param {function(string):boolean} mdMedia Angular Material $mdMedia service
+ * @return {number}
+ */
+function computeColumnCount(mdMedia) {
+  if (mdMedia('gt-md')) {
+    return 3;
+  } else if (mdMedia('md')) {
+    return 2;
+  } else {
+    return 1;
+  }
+}
+
+/**
+ * Does binary search to find minimal integer I such that with height I elements fit into numColumns
+ * and with height I - 1 they do not.
+ *
+ * @param {!Array<number>} heights
+ * @param {number} numColumns
+ * @return {number}
+ */
+export function binarySearchOptimalHeight(heights, numColumns) {
+  let sum = Math.ceil(heights.reduce((a, b) => a + b, 0));
+  let height = 0;
+
+  let left = 0;
+  let right = sum;
+  for (;;) {
+    height = Math.floor((left + right) / 2);
+    let [leftChunks, rightChunks] = getActualColumnCount(heights, height - 1, height);
+
+    if ((leftChunks > numColumns && rightChunks <= numColumns) || (left === right)) {
+      break;
+    } else if (leftChunks > numColumns) {
+      left = height + 1;
+    } else {
+      right = height;
+    }
+  }
+
+  return height;
+}
+
+/**
+ * Returns actual column count for the given array of element heights and two actual heights.
+ * Two values are returned, first for leftHeight and the other for rightHeight. Infinity is
+ * returned if for any height (left of right) there is an item that does not fit it.
+ *
+ * @param {!Array<number>} heights
+ * @param {number} leftHeight
+ * @param {number} rightHeight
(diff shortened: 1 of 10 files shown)


bryk:

src/app/frontend/replicasetlist/replicasetlistcontainer.scss
```diff
@@ -0,0 +1,21 @@
...
+//   
+// Unless required by applicable law or agreed to in writing, software   
+// distributed under the License is distributed on an "AS IS" BASIS,   
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.    
+// See the License for the specific language governing permissions and   
+// limitations under the License.    
+
+.kd-replica-set-list-container {
+  align-content: center;
+  align-items: top;
```
floreks: I'm playing with UI now and when there is 1 column or 1 replica set only then it is aligned to the left side of the screen. With this changed to `center` it always tries to center cards and as height of card container is calculated to be as small as possible then only vertical center will be applied. What do you think?

cheld:
I just tried the code. I did not review it yet. I found a very, very small layout issue. 

Steps to reproduce:
1. create 5 pods
2. decrease window size until one column is shown
3. increase window size very slowly until layout is broken into two column. 

=> There is a space of 1-2 pixel where only the right column is shown, but not the left.

Remark: I could not reproduce this when switching from 2->3 columns


However, a browser resize/reload solves the issue. So, not really an issue that must be fixed.

broken-layout

BTW: I intuitively expected a row layout and that cards are sorted from left to right (in reading direction). Not up to down. But I assume this is intentional

src/app/frontend/replicasetlist/replicasetlistcontainer.scss
bryk: Agreed, thanks! I was looking for this solution :) 

Done.

bryk:
> I just tried the code. I did not review it yet. I found a very, very small layout issue.

I did my best to reproduce this, but couldnt find the issue. I applied one style change as @floreks suggested to the code. Can you rebase and check whether the issue is still there?

floreks:
Actualy this change to center elements fixes this issue :). I can't reproduce it either and before I was able to do it.

bryk:
> BTW: I intuitively expected a row layout and that cards are sorted from left to right (in reading direction). Not up to down. But I assume this is intentional

Cards can have different heights, hence we need column layout. Otherwise things look ugly. Take a look how pinterest, google plus or developers console do this. 

Re ordering things from the left to the right: this may be a next step of this layout. I wanted to have something useful with as little code as possible.

src/app/frontend/replicasetlist/replicasetlistcontainer.js
```diff
@@ -0,0 +1,132 @@
...
+ *     It should be a flexbox container with columns wrapped.
+ * @param {function(string):boolean} mdMedia Angular Material $mdMedia service
+ * @return {number}
+ */
+export default function computeContainerHeight(container, mdMedia) {
+  /** @type {!Array<number>} */
+  let childHeights = Array.prototype.map.call(container.children, (child) => child.offsetHeight);
+
+  let columnCount = computeColumnCount(mdMedia);
+  return binarySearchOptimalHeight(childHeights, columnCount);
```
floreks: This is a great non-trivial solution. I see that complexity here is `O(nlog(n))` so it's really good. I'm not sure how big sets of data will be used here, but for most cases this should be sufficient. 

I was thinking a bit about other possible solutions(I like algorithmic stuff :)). I think there are also solutions for this NP problem with complexity `O(n)`. As example, there could be used algorithm to split array of heights into P subarrays of balanced sum(`O(P*n)`). The subarray with highest partial sum would be our height.

This is very similar to bin packing problem.

bryk: :)

I bet there are better solutions for this problem, but this is pretty nice already. It is `O(n log sum_of_heights)` which is roughly `O(n log n)`. I did tests and on first layout it takes 10 ms to compute it (with 90% time spend on line 26 which is costly). On subsequent calls this function takes anything from 1 to 0.1 ms to compute. I had like 20 replica sets.

bryk: The takeaway here is that algorithm is very fast and `child.offsetHeight` call is the thing that may be a bottleneck.

cheld: BTW: yesterday resizing the browser felt very slow and sluggish. I have not investigated. I assume too many resize events. Maybe you consider a debounce function.

floreks: This algorithm is executed only on mdMedia change ( actual column number change ) and first when the view is called. Normally when resizing it is not called. The problem has to be somewhere else.

floreks: I'm not sure if `child.offsetHeight` is actually a bottleneck here. This is supposed to be calculated before and at the time this function is called be a read only value. Anyway I don't think that we are planning to show 1000+ replica sets :) Looks and works good to me.

bryk: > I'm not sure if child.offsetHeight is actually a bottleneck here.

There is a little bit of browser magic when you access this property. If you're in the middle of rendering, the browser has to paint the element to find out what's its height. This takes substantial amount of time. But then, when the element is rendered and hasn't changed, accessing this property is fast. You can see this in chrome debugging tools. Try it out, it is a nice exercise :)

bryk: >  Maybe you consider a debounce function.

Denouncing cannot be added in function that is called on $watch. What we can do is caching of old value. I expect cache hit ratio to be very high since most of the time width&height do not change. But let's leave it for future work. The layout does not feel slow at the moment. 
We can also call this function less often, e.g., only on resize, but this also may happen in the future.

cheld:
I have tested again, but this time on my notebook. My workstation is not operational, currently. Unfortunately, I cannot verify the problem under the exactly same situation. 
However, on my notebook:
* I did not notice any layout issue
* I did not notice any perfomance issue

src/app/frontend/replicasetlist/replicasetlistcontainer.js
```diff
@@ -0,0 +1,132 @@
...
+ * Does binary search to find minimal integer I such that with height I elements fit into numColumns
+ * and with height I - 1 they do not.
+ *
+ * @param {!Array<number>} heights
+ * @param {number} numColumns
+ * @return {number}
+ */
+export function binarySearchOptimalHeight(heights, numColumns) {
+  let sum = Math.ceil(heights.reduce((a, b) => a + b, 0));
+  let height = 0;
```
floreks: Intellij found that this doesn't have to be initialized as it's always initialized in the loop :)

floreks:
LGTM. It's tested and designed very well.

PR #197, opened 2015-12-23
Make serve:prod task use non-docker backend binary

bryk:
This speeds up serve:prod task a few times, yet has the same effect.
Backend:prod should only be used in dockerfiles.


bryk:
@cheld Can you take a look?

cheld:
I checked it, makes sense. 

LGTM

PR #199, opened 2015-12-23
Add run as privileged option to the deploy form

bryk:


bryk:
@zreigz Can you review?

zreigz:
Yes, I will do it Monday morning

zreigz:
When I enabled ```Run as privileged``` I have ```Internal Server Error```
```ReplicationController "p" is invalid: spec.template.spec.containers[0].securityContext.privileged: Forbidden: disallowed by policy```

Without this flag deploy works fine

bryk:
I see. Your cluster has settings to disable this option. See in `kubernetes/pkg/capabilities/capabilities.go` capabilities struct. 

What do you think about detecting this and hiding when not available?

zreigz:
I think it is good solution, otherwise user will be confused why can not deploy. 
Second solution is related to error handling to inform user that option is disabled in cluster settings. 
Because error handling is not implemented yet, so you proposition is the most reasonable here.

bryk:
@zreigz You know what? I think now is better to implement error handling than to work on conditional options in the deploy form. I want to merge this PR now and immediately start work on error handling. What do you think?

zreigz:
I agree

PR #203, opened 2015-12-28
Global styling framework: streamline variable handling

bryk:
Now variables are in variables.css in frontend directory. Previously they were in assets, which was wrong because it caused copy to production image.

### DIFF
--- a/src/app/frontend/_variables.scss
+++ b/src/app/frontend/_variables.scss
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO(floreks): Find appropriate place for global color definitions.
 $primary: #326de6;
 $delicate: #aaa;
 $muted: #888;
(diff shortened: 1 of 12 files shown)


bryk:
@zreigz Can you review?

zreigz:
yes

floreks:
Just had a quick look. One comment about variable file naming. I've intentionally used underscore there.

http://sass-lang.com/guide

`Partials` headline says:
`The underscore lets Sass know that the file is only a partial file and that it should not be generated into a CSS file. Sass partials are used with the @import directive.`

Rest I'll leave for @zreigz to review :)

src/app/frontend/variables.scss
```diff
@@ -20,6 +19,8 @@ $hover-primary: #1254df;
...
 $emphasis: #000;
 $content-background: #fff;
-// TODO(bryk): Reference angular_material/src/core/style/variables.scss instead of having
-// this variable here.
-$whiteframe-shadow-z1: 0 2px 5px 0 rgba(0, 0, 0, .26) !default;
+
+// TODO(bryk): Get those variables from Angular Material scss files.
+$layout-breakpoint-lg: 1280px;
+$whiteframe-shadow-1dp: 0 1px 3px 0 rgba(0, 0, 0, .2), 0 1px 1px 0 rgba(0, 0, 0, .14), 0 2px 1px -1px rgba(0, 0, 0, .12);
+$baseline-grid: 8px;
```
zreigz: For me ```$baseline-grid``` is misleading. I would suggest ```baseline-distance``` instead. What do you thing ?

zreigz:
Rest looks good for me. All views look like previously.

src/app/frontend/variables.scss
bryk: I'm also not a big fan of this name, but this is what Angular Material uses. Take a look at: https://github.com/angular/material/blob/HEAD/src/core/style/variables.scss

I hope that soon we'll be able to include angular material scss, so then this variable here will be deleted. And Material's one used. What do you think? Can I leave the name as it is?

bryk:
@floreks Thank you for the comment! I renamed the file

bryk:
PTAL

zreigz:
If it is common practice to use name ```$baseline-grid``` then LGTM

PR #204, opened 2015-12-29
Increase Karma no activiti timeout

bryk:
This is to prevent flakiness in our travis infrastructure.

Also, a few refactoring in the build pipeline re coverage.


bryk:
@zreigz Take a look

bryk:
I'm merging this because it blocks our entire Travis infra. I'll resolve any comments in a subsequent PR.

PR #205, opened 2015-12-29
Logs toolbar style adjustment to mock

zreigz:
toolbar

### DIFF
--- a/src/app/frontend/_variables.scss
+++ b/src/app/frontend/_variables.scss
@@ -12,6 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Function
+//-- Must be defined before variables
+@function rem($multiplier) {
+  $font-size: 10px;
+  @return $multiplier * $font-size;
+}
+
+$subhead-font-size-base:   rem(1.6) !default;
+
 $primary: #326de6;
 $delicate: #aaa;
 $muted: #888;
--- a/src/app/frontend/logs/logstoolbar/logstoolbar.scss
+++ b/src/app/frontend/logs/logstoolbar/logstoolbar.scss
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+@import '../../variables';
+
 $logs-toolbar-color-black: #000;
 $logs-toolbar-color-white: #fff;
 
@@ -32,4 +34,13 @@ md-toolbar {
       color: $logs-toolbar-color-black;
     }
   }
+}
+
+.kd-logs-toolbar-text {
+  font-size: $subhead-font-size-base;
+}
+
+.kd-logs-toolbar-select {
+  margin-bottom: 2 * $baseline-grid;
+  margin-left: auto;
 }
--- a/src/app/frontend/logs/logstoolbar/logstoolbar.html
+++ b/src/app/frontend/logs/logstoolbar/logstoolbar.html
@@ -18,25 +18,24 @@ limitations under the License.
   <span flex="10"></span>
 
   <div layout="row" layout-wrap layout-margin layout-align="center center">
-    <span>Logs from pod</span>
-    <span flex></span>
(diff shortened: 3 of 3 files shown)


src/app/frontend/logs/logstoolbar/logstoolbar.scss
```diff
@@ -14,6 +14,10 @@
 
 $logs-toolbar-color-black: #000;
 $logs-toolbar-color-white: #fff;
+$logs-toolbar-text-font-size: 17px;
+$logs-toolbar-title-font-weight: 500;
+$logs-toolbar-select-margin-bottom: 15px;
+$logs-toolbar-select-margin-left: auto;
```
bryk: This (and line 19) does not need to be in a variable. It is used only once and is pretty standard way of margining.

bryk: For 15px you could use 2*$baseline-grid. Will it look okay? The reason is that we should not use magic numbers unless really needed, and base everything on baseline grid.

src/app/frontend/logs/logstoolbar/logstoolbar.scss
```diff
@@ -14,6 +14,10 @@
 
 $logs-toolbar-color-black: #000;
 $logs-toolbar-color-white: #fff;
+$logs-toolbar-text-font-size: 17px;
```
bryk: Can you use any of font sizes from https://github.com/angular/material/blob/HEAD/src/core/style/variables.scss and define the font in src/app/frontend/_variables.scss ?

src/app/frontend/logs/logstoolbar/logstoolbar.scss
```diff
@@ -14,6 +14,10 @@
...
+$logs-toolbar-title-font-weight: 500;
```
bryk: I checked demo and without this 500 it looks better. How about removing this?

bryk:
Looks really nice. PTAL

src/app/frontend/logs/logstoolbar/logstoolbar.scss
zreigz: I removed this.

src/app/frontend/logs/logstoolbar/logstoolbar.scss
zreigz: done

src/app/frontend/logs/logstoolbar/logstoolbar.scss
zreigz: done

zreigz:
PTAL

src/app/frontend/_variables.scss
```diff
@@ -12,6 +12,20 @@
...
+// Function
+//-- Must be defined before variables
+@function rem($multiplier) {
+  $font-size: 10px;
+  @return $multiplier * $font-size;
+}
+
+// Typography
+// ------------------------------
+$font-family: Roboto, 'Helvetica Neue', sans-serif !default;
```
bryk: You dont need $font-family. All our texts are already with this font family.

src/app/frontend/_variables.scss
```diff
@@ -12,6 +12,20 @@
...
+$font-size:   10px;
```
bryk: You dont need to have it defined here. On line 18 is enough.

src/app/frontend/logs/logstoolbar/logstoolbar.scss
```diff
@@ -33,3 +35,13 @@ md-toolbar {
     }
   }
 }
+
+.kd-logs-toolbar-text {
+  font-family: $font-family;
```
bryk: Font family is already defined by angular material, so it is safe to remove it from here.

bryk:
Awesome. Last comments and let's merge. PTAL

zreigz:
PTAL

bryk:
LGTM

PR #207, opened 2015-12-29
Move name input to deploy from settings flow

bryk:
This is because name does not make sense in the YAML upload form.


bryk:
@zregiz can you review?

zreigz:
yes, I'am doing right now

zreigz:
Now deploy page looks much better. LGTM

PR #209, opened 2015-12-30
Validate app name in the deploy form

bryk:
This validates application name asynchronously on the server. Validation is synced with namespace select :)

### DIFF
--- a/src/app/backend/validateappname.go
+++ b/src/app/backend/validateappname.go
@@ -1,0 +1,70 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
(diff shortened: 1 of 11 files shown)


bryk:
@zreigz Can you review?

zreigz:
Yes. I am reviewing this

zreigz:
To make it complete you should also check if name contains illegal characters. For example when you enter name: ```some/text``` then you get 500  because ```/``` is forbidden. Also name containing ```%``` returns error. I think it should be done on client side before server side validation.

bryk:
Yes, definitely. But how about doing this in a next pull request? This one is already pretty large and I don't want to make it even larger.

src/app/backend/validateappname.go
```diff
@@ -0,0 +1,64 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	k8serrors "k8s.io/kubernetes/pkg/api/errors"
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+)
+
+type AppNameValiditySpec struct {
```
zreigz: Add small description

src/app/backend/validateappname.go
```diff
@@ -0,0 +1,64 @@
...
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+type AppNameValidity struct {
```
zreigz: Add small description

zreigz:
Yes I agree it can be done in other PR. Two small remarks and you can merge :)
PTAL

src/app/backend/validateappname.go
bryk: Done.

src/app/backend/validateappname.go
bryk: Done.

zreigz:
LGTM

PR #210, opened 2015-12-31
Unify service endpoint formatting through the project

bryk:
Now raw data is gathered on the backend and rendered on the frontend.

This PR looks large, but is is mostly moving things here and there.

### DIFF
--- a/src/app/backend/replicasetdetail.go
+++ b/src/app/backend/replicasetdetail.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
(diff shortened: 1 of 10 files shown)


bryk:
@zreigz Can you review?

zreigz:
Yes, last review in this year :)

src/app/backend/replicasetdetail.go
```diff
@@ -211,3 +229,41 @@ func getRestartCount(pod api.Pod) int {
...
 	return restartCount
 }
+
+// Returns internal endpoint name for the given service properties, e.g.,
+// "my-service.namespace 80/TCP" or "my-service 53/TCP,53/UDP".
+func getInternalEndpoint(serviceName string, namespace string, ports []api.ServicePort) Endpoint {
+
+	name := serviceName
+	if namespace != api.NamespaceDefault {
+		name = name + "." + namespace
```
zreigz: Very good practice is not using ```+``` for string concatenation.
Strings are immutable in Go - each time a string is assigned to a variable a new address is assigned in memory to represent the new value.  As the strings you are manipulating become large this process becomes increasingly slow. In fact, this method of naive appending is O(N2). (The same is in Java so StringBuffer is used instead ```+=```)
Here you should also use Buffer:
```
	bufferName := bytes.NewBufferString(name)
	bufferName.WriteString(".")
    bufferName.WriteString(namespace)
	name = bufferName.String()
```

zreigz:
Rest looks good :)

zreigz:
PTAL

src/app/backend/replicasetdetail.go
bryk: Oh, thanks for this comment. Makes sense. Fixed :)

bryk:
Thanks for the review :)

Issue #191, opened 2015-12-21
Relative time filter


maciaszczykm:
I will prepare `relative time` filter, which can be reused in replica set details table and other pages.

PR #215, opened 2016-01-04
Relative time filter

maciaszczykm:
Connected to issue #191.

I've implemented relative time filter which can be used in multiple views. At this point I've added it to replica set details and replica set list. Filter transforms date to relative date string, which can be seen on mocks. Added tests covering all important date units.

There is still one thing to do, but to keep this pull request small I think it will be better to move it to another pull request. The issue is, that at the moment filter uses client side current time and it's invalid in many cases. Date should be taken from server in all cases, that I know. **Please let me know if you have any suggestions how we should solve it**.

@floreks @bryk Can you review?

### DIFF
--- a/src/app/frontend/common/filters/relativetime_filter.js
+++ b/src/app/frontend/common/filters/relativetime_filter.js
@@ -1,0 +1,111 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Unit name constants (singular and plural form), that will be used by the filter.
+ * @enum {!Array<string>}
+ */
+const Units = {
+  SECOND: ['a second', 'seconds'],
+  MINUTE: ['a minute', 'minutes'],
+  HOUR: ['an hour', 'hours'],
+  DAY: ['a day', 'days'],
+  MONTH: ['a month', 'months'],
+  YEAR: ['a year', 'years'],
+};
+
+/**
+ * Unit conversion constants.
+ * @enum {number}
+ */
+const UnitConversions = {
+  MILLISECONDS_PER_SECOND: 1000,
+  SECONDS_PER_MINUTE: 60,
+  MINUTES_PER_HOUR: 60,
+  HOURS_PER_DAY: 24,
+  DAYS_PER_MONTH: 30,
+  DAYS_PER_YEAR: 365,
+  MONTHS_PER_YEAR: 12,
+};
+
+/**
+ * Time constants.
+ * @enum {string}
+ */
+const TimeConstants = {
+  NOT_YET: `didn't happen yet`,
+  NOW: `just now`,
+};
+
+/**
+ * Returns filter function to display relative time since given date.
+ * @return {Function}
+ */
+export default function relativeTimeFilter() {
+  /**
+   * Filter function to display relative time since given date.
+   * @param {string} value Filtered value.
+   * @return {string}
+   */
+  let filterFunction = function(value) {
+    // Current and given times in miliseconds.
+    let currentTime = (new Date()).getTime();  // TODO(maciaszczykm): Use server time.
+    let givenTime = (new Date(value)).getTime();
+
+    // Time differences between current time and given time in specific units.
+    let diffInMilliseconds = currentTime - givenTime;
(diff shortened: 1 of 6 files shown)


src/app/frontend/common/filters/relativetime_filter.js
```diff
@@ -0,0 +1,129 @@
...
+      return TIME_CONSTANTS.NOW;
+    } else if (diffInMinutes < 1) {
+      return formatOutputTimeString_(diffInSeconds, UNITS.SECOND);
+    } else if (diffInHours < 1) {
+      return formatOutputTimeString_(diffInMinutes, UNITS.MINUTE);
+    } else if (diffInDays < 1) {
+      return formatOutputTimeString_(diffInHours, UNITS.HOUR);
+    } else if (diffInMonths < 1) {
+      return formatOutputTimeString_(diffInDays, UNITS.DAY);
+    } else if (diffInYears < 1 && diffInMonths < UNIT_CONVERSIONS.MONTHS_PER_YEAR) {
```
maciaszczykm: One more thing to be fixed here.

src/app/frontend/common/filters/relativetime_filter.js
```diff
@@ -0,0 +1,129 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Unit name constants (singular and plural form), that will be used by the filter.
+ * @type {{
+ *   SECOND: !Array<string>,
```
bryk: Use `@enum{!Array<string>}`. See http://usejsdoc.org/tags-enum.html

src/app/frontend/common/filters/relativetime_filter.js
```diff
@@ -0,0 +1,129 @@
...
+ *   MINUTE: !Array<string>,
+ *   HOUR: !Array<string>,
+ *   DAY: !Array<string>,
+ *   MONTH: !Array<string>,
+ *   YEAR: !Array<string>
+ * }}
+ */
+const UNITS = {
```
bryk: Name this like a class, i.e., `Units`, when you convert it to `@enum`.

src/app/frontend/common/filters/relativetime_filter.js
```diff
@@ -0,0 +1,129 @@
...
+ *   SECONDS_PER_MINUTE: number,
+ *   MINUTES_PER_HOUR: number,
+ *   HOURS_PER_DAY: number,
+ *   DAYS_PER_MONTH: number,
+ *   DAYS_PER_YEAR: number,
+ *   MONTHS_PER_YEAR: number
+ * }}
+ */
+const UNIT_CONVERSIONS = {
+  MILISECONDS_PER_SECOND: 1000,
```
bryk: MI**LL**I...

src/test/frontend/common/filters/relativetime_filter_test.js
```diff
@@ -0,0 +1,213 @@
...
+       givenTime = givenTime.setMonth(givenTime.getMonth() - 1);
+
+       // when
+       let relativeTime = relativeTimeFilter(givenTime);
+
+       // then
+       expect(relativeTime).toEqual('a month');
+     }));
+
+  it("should return '11 months' string if given time is 11 months before current time",
```
bryk: Can you write a test for, e.g., 11 months 7 days 5 hours and 3 minutes?

src/app/frontend/replicasetlist/replicasetcard.html
```diff
@@ -62,8 +62,8 @@
         </div>
 
         <div flex="40" layout="column" class="kd-replicaset-card-section">
-          <span flex="initial">Creation time</span>
-          <span flex>{{::ctrl.replicaSet.creationTime}}</span>
+          <span flex="initial">Age</span>
+          <span flex>{{::ctrl.replicaSet.creationTime | relativeTime}}</span>
```
bryk: Can you provide original value on hover? Either as alt, title or md-tooltip.

bryk:
Looks really nice, a few comments. PTAL.

floreks:
There might be some problems with correct age calculation. I thought that we can check how age is retrieved with `kubectl` but there seems to be an issue with that also as described here:

https://github.com/kubernetes/kubernetes/issues/18183

Pod start time is taken directly from node as i see in `kubelet.go`. Nodes can be located on different machines in different time zones so this problem is becoming interesting. Syncing with master time then is also not correct. Currently I don't have any idea how to resolve this time diff problem.

maciaszczykm:
@bryk PTAL

@floreks Yes, it will be difficult to implement at this point. Therefore, I've left `TODO(maciaszczykm): Use server time` comment in the code to remember to update it later.

If you have any suggestions from where take server time and how to handle time zones then please share your ideas.

bryk:
LGTM. Can you rebase and apply @floreks comments (if any)

floreks:
LGTM also. This time issue is something we can deal with later.

maciaszczykm:
@bryk @floreks Done, thank you for reviews.

bryk:
There are conflicts. Can you rebase once more? Sorry for the mess...

maciaszczykm:
@bryk No problem, solved conflitcts and moved this filter to filters module :) PTAL

Issue #153, opened 2015-12-10
State name import - closure compiler dependency errors


floreks:
Currently there is a problem with importing `stateName`. I will give an example:

In `replicasetdetail_controller.js` we import `stateName` from `replicasetlist_state.js`.
In `replicasetlist_controller.js` we import `stateName` from `replicasetdetail_state.js`.

This results in frontend build error.

Unfortunately compiler does not provide good information about real cause of error. It's similar to this:
```
ERROR - required "module$replicasetlist$replicasetlist_state" namespace not provided yet import {stateName} from 'replicasetlist/replicasetlist_state';
```

Real cause is due to circual dependency between modules as I've described here: https://github.com/floreks/dashboard/commit/ad6c93803d09efd0fcc0020d931ee1cf507386e5#commitcomment-14899972

I've tested this in different case for `zerostate` and `replicasetlist` modules and the result is the same.

For now only solution I've in mind requires creation of new module or new file independent of current modules where we can keep state names. 

Any other ideas?

bryk:
Generally, the compiler is right. It does not allow for circular dependencies. However, the message is very misleading. Thanks @floreks for investigating. 

I guess that our design with having state name and logic in foo_state.js files is wrong and this causes circular deps. I think we should redesign this.

bryk:
@floreks Can you take this issue? As our code gets more and more complex, I see TODOs for this issues a lot. 

We basically need some kind of proposal of how to rearrange files so that we dont have circular deps (this is a benefit anywany).

At minimum, we need to place state names and state params in a file that does not import anything else (well, except for maybe some model classes).

bryk:
@cheld @floreks FYI

floreks:
I can take care of this but at the beginning of next week, so if anyone wants to take this sooner then be my guest ;)

floreks:
This is fixed. Closing.

PR #216, opened 2016-01-04
Resolved circural dependencies issue (#153)

floreks:
I've been thinking about some good way to resolve this issue (#153).

## Proposals
1. Introduce new module with module constant `states` that can be injected on demand.
2. Introduce new file with const object `states` not related to modules that can be imported on demand.

I've chosen second option as it seems to be cleaner and simpler. There is no need to define additional dependencies between modules and we don't have to inject constant to the function where we want it to be used.

## Architecture

I've externalized all state names to separate `states.js` file in order to get rid of circual dependencies between modules. Currently all modules can import on demand defined object and use its state properties.

## Usage

Usage is pretty straighforward. Just import `states` object and use predefined states f.e. `state.STATE_DEPLOY`.

## Things to improve

* Filename - As it is not related to any module I've chosen simple `states.js`. We can think about better name here
* Filepath - It's part of the global configuration, so I've decided to put it directly under `frontend` folder. Does it need new folder?

Questions? Comments? Any other ideas?

@bryk @zreigz @maciaszczykm @cheld

### DIFF
--- a/src/app/frontend/deploy/deploy_module.js
+++ b/src/app/frontend/deploy/deploy_module.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import stateConfig from './deploy_state';
+import stateConfig from './deploy_stateconfig';
 import deployFromSettingsDirective from './deployfromsettings_directive';
 import deployLabelDirective from './deploylabel_directive';
 import deployFromFileDirective from './deployfromfile_directive';
--- a/src/app/frontend/deploy/deploy_state.js
+++ b/src/app/frontend/deploy/deploy_state.js
@@ -12,37 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import DeployController from './deploy_controller';
-
 /** Name of the state. Can be used in, e.g., $state.go method. */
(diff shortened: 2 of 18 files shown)


bryk:
There is still a problem with StateParams classes. Am I correct? How can we fix this?

floreks:
I was not aware that we have a problem with that and this was actually not part of mentioned issue. With whole classes it may not be that easy.

bryk:
How about breaking state.js files into stateconfig.js and state.js? state.js would contain state name + params class and stateconfig.js would contain the config function.

floreks:
This may work. Let me check.

floreks:
It works pretty well. I thought that there will be still a problem with dependencies but splitting `state.js` file helped. PTAL

bryk:
Looks good to me. However, can you fix all places where issue #153 was referenced? E.g., in logstoolbar_controller.js where state name and state params class are needed.

floreks:
I thought I've changed it in all places. Thanks. I hope now it's everything. PTAL

bryk:
LGTM, thanks! I think we should merge this now to avoid collisions with others' work

floreks:
Sure. I've renamed this PR and changed commit message.

PR #217, opened 2016-01-04
Style replica set list page

bryk:


bryk:

src/app/frontend/replicasetlist/replicasetcard.scss
```diff
@@ -19,34 +19,54 @@
...
-  line-height: 0;
-  margin: inherit;
-  min-height: inherit;
-  min-width: inherit;
-  padding: inherit;
+  display: inline-block;
+  font-weight: $title-font-weight;
   text-decoration: none;
-  text-transform: inherit;
+  width: 100%;
```
floreks: I'm testing the UI and I've noticed that when name is short, clickable space to enter details page is very wide(it's always flexed). It doesn't feel very intuitive when I'm clicking on blank spot and I'm redirected to details page. Either we can make whole card area clickable or adjust card name width. It could be also more noticable that name can be clicked to perform action.

floreks:
Just one style comment. Rest LGTM

maciaszczykm:
It looks a lot better now! Three observations from my side, rest LGTM:

I would consider changing background or placement of `Logs` dropdown button, because now it looks like it's adjustment is not correct. 

It seems, that there is missing space between `Logs` and downwards arrow and I think it would look better with that space. 

The last thing I noticed is placement of right-top corner menu, it seems to be too far from right border, but it seems to be out of the box styling, so I guess it should stay like it's done at the moment.

src/app/frontend/replicasetlist/replicasetcard.scss
bryk: Makes sense. Done. I made only the title clickable.

floreks:
To resolve font size issues it would be good to define more variables. Now there is only 1 `subhead-font-size-base`. For normal text it would be good to have `font-size-base` set to `rem(1.2)` as default font size is 12px. Now fonts in logs menu are set to `.8em` and on replica set card to `12px`.

bryk:
> I would consider changing background or placement of Logs dropdown button, because now it looks like its adjustment is not correct.

Done.

> It seems, that there is missing space between Logs and downwards arrow and I think it would look better with that space.

Done.

> This view has a bit different styling than logs menu, which has bigger font and different colors. Should we use similar styling (with this one as a template - fonts, font colors etc.) for all other views and menus?

I've unified logs menu with this view. Basically, I'm taking what material design scss files have and using them in our code. This is what we should be doing.

bryk:
> To resolve font size issues it would be good to define more variables. Now there is only 1 subhead-font-size-base. For normal text it would be good to have font-size-base set to rem(1.2) as default font size is 12px. Now fonts in logs menu are set to .8em and on replica set card to 12px.

I added more variables.

bryk:
PTAL. I've also styled and unified logs menu

maciaszczykm:
Logs dropdown button looks better now. This time to test I've deployed app with version and version information in the UI is floating and it isn't aligned with anything.

floreks:
I think I've added todo to use common formatting for labels. I can prepare small PR and add filter for labels formatting. What do you think?

bryk:
> This time to test I've deployed app with version and version information in the UI is floating and it isn't aligned with anything. Maybe we can enhance it too?

Screenshot?

maciaszczykm:
@bryk Added screenshot.

bryk:
I see. Labels are not formatted pretty as of now. In next pull request I can format them in so called chips. Like here  . This requires additional work, however.

bryk:
> I think I've added todo to use common formatting for labels. I can prepare small PR and add filter for labels formatting. What do you think?

Yes, this should definitely be done :) But I'd prefer to have it in a separate PR. To make this smaller.

floreks:
Sure. I'll wait for this PR to be merged and I can prepare some filter for this. Chips are looking good so i may use that.

maciaszczykm:
The rest LGTM.

bryk:
@floreks Did you finish your review?

floreks:
Yes. LGTM now.

PR #220, opened 2016-01-05
Fix a bug where logs menu would crash when there were no statuses rep…

bryk:
…orted for containers


bryk:
@floreks Can you review?

floreks:
LGTM

PR #221, opened 2016-01-05
Minor styling updates of replica set details view

maciaszczykm:
Fixed some replica set view-related issues:

- Styled application name and arrow nearby, updated font weights.
- Unified dates (used`short` format).
- Fixed problems with missing stylesheet definitions (for example there was a problem with arrow near `Logs` text).
- Fixed issue with `{<!>` and `<!>}` displaying nearby source in events table - I think they are not neccessary anymore.

### DIFF
--- a/src/app/frontend/replicasetdetail/replicasetdetail.scss
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.scss
@@ -18,7 +18,11 @@ $replicasetdetails-sidebar-bg: #fafafa;
 $replicasetdetails-border: #ddd;
 $replicasetdetails-table-message: #000;
 $replicasetdetails-table-cell: #777;
-$replicasetdetails-warning: #ffa500;
+
+.kd-replicasetdetail-app-name {
+  color: $foreground-2;
+  font-weight: $regular-font-weight;
+}
 
 .kd-replicasetdetail-sidebar {
   background-color: $replicasetdetails-sidebar-bg;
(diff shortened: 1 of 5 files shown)


src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -19,6 +19,12 @@ $replicasetdetails-border: #ddd;
 $replicasetdetails-table-message: #000;
 $replicasetdetails-table-cell: #777;
 $replicasetdetails-warning: #ffa500;
+$replicasetdetails-message-font-weight: 400;
```
bryk: This is the same as $title-font-weight. Can you maybe unify this? Perhaps with changing the name of the variable to cover both cases.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -19,6 +19,12 @@ $replicasetdetails-border: #ddd;
...
+
+.kd-replicasetdetail-app-name {
+  color: $muted;
```
bryk: Please use `$foreground-2`. This is the same as muted. We need to unify those variables.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -37,13 +43,21 @@ $replicasetdetails-warning: #ffa500;
...
   font-size: 1.7em;
+  font-weight: $title-font-weight;
+  margin-right: 3px;
   vertical-align: middle;
 }
 
 .kd-replicasetdetail-table-message {
+  border-bottom: 1px solid $replicasetdetails-border;
   color: $replicasetdetails-table-message;
+  font-size: 13px;
```
bryk: Can you use $caption-font-size-base?

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -37,13 +43,21 @@ $replicasetdetails-warning: #ffa500;
 }
 
 .kd-replicasetdetail-sidebar-header-icon {
+  color: $muted;
   font-size: 1.7em;
```
bryk: Can you use $subhead-font-size-base? Or any other from https://github.com/angular/material/blob/8ef798f0fe242c555ef7c11fa0abe762fb218a02/src/core/style/variables.scss

bryk:
Looks really nice. A few comments on unifying styles even more.

bryk:
PTAL

maciaszczykm:
PTAL

bryk:
LGTM :)

PR #223, opened 2016-01-05
Added common label formatting (as chips) and unified across the project.

floreks:
This is how it looks:

zrzut ekranu z 2016-01-05 17-29-05

zrzut ekranu z 2016-01-05 17-29-11

I've changed default chips styling to fit application style. I think that similar look can be applied to deploy page labels if needed. Only 1 key/value input row would be left and all added labels would be displayed below as list of chips.

@bryk @maciaszczykm what do you think?

### DIFF
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -49,7 +49,7 @@ function getFileList() {
  */
 module.exports = function(config) {
   let configuration = {
-    basePath: conf.paths.base,
+    basePath: '.',
 
     files: getFileList(),
 
@@ -109,7 +109,8 @@ module.exports = function(config) {
     // karma-ng-html2js-preprocessor plugin config.
     ngHtml2JsPreprocessor: {
       stripPrefix: `${conf.paths.frontendSrc}/`,
-      moduleName: conf.frontend.moduleName,
+      // Load all template related stuff under ng module as it's loaded with every module.
+      moduleName: 'ng',
     },
   };
 
--- a/src/app/frontend/common/components/labels/labels.scss
+++ b/src/app/frontend/common/components/labels/labels.scss
@@ -12,16 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import middleEllipsisFilter from './middleellipsis_filter';
-import relativeTimeFilter from './relativetime_filter';
+@import '../../../variables';
 
-/**
- * Module containing common filters for the application.
- */
-export default angular.module(
-                          'kubernetesDashboard.common.filters',
-                          [
-                            'ngMaterial',
-                          ])
-    .filter('middleEllipsis', middleEllipsisFilter)
-    .filter('relativeTime', relativeTimeFilter);
+$label-height: $caption-font-size-base + 6;
+$label-margin: 0 $baseline-grid $baseline-grid 0;
+
+// Style taken(mostly) from angular material md-chip style.
+.kd-labels {
+  background-color: $body;
+  border-radius: $label-height / 2;
+  display: inline-block;
+  font-size: $caption-font-size-base;
+  line-height: $label-height;
+  margin: $label-margin;
+  padding: 0 $baseline-grid;
+  vertical-align: middle;
+}
--- a/src/app/frontend/replicasetdetail/replicasetdetail.scss
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.scss
@@ -18,6 +18,7 @@ $replicasetdetails-sidebar-bg: #fafafa;
 $replicasetdetails-border: #ddd;
 $replicasetdetails-table-message: #000;
 $replicasetdetails-table-cell: #777;
+$replicasetdetails-sidebar-width: 315px;
 
 .kd-replicasetdetail-app-name {
   color: $foreground-2;
@@ -26,7 +27,8 @@ $replicasetdetails-table-cell: #777;
 
 .kd-replicasetdetail-sidebar {
   background-color: $replicasetdetails-sidebar-bg;
-  min-width: 315px;
+  max-width: $replicasetdetails-sidebar-width;
+  min-width: $replicasetdetails-sidebar-width;
 }
 
 .kd-replicasetdetail-table {
@@ -76,7 +78,6 @@ $replicasetdetails-table-cell: #777;
 .kd-replicasetdetail-sidebar-subline {
   color: $muted;
   font-size: $caption-font-size-base;
-  padding-bottom: 8px;
 }
 
 .kd-replicasetdetail-sidebar-icon {
--- a/src/app/frontend/common/components/components_module.js
+++ b/src/app/frontend/common/components/components_module.js
@@ -12,16 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import middleEllipsisFilter from './middleellipsis_filter';
-import relativeTimeFilter from './relativetime_filter';
+import labelsDirective from './labels/labels_directive';
 
 /**
- * Module containing common filters for the application.
+ * Module containing common components for the application.
  */
 export default angular.module(
-                          'kubernetesDashboard.common.filters',
+                          'kubernetesDashboard.common.components',
                           [
                             'ngMaterial',
                           ])
-    .filter('middleEllipsis', middleEllipsisFilter)
-    .filter('relativeTime', relativeTimeFilter);
+    .directive('kdLabels', labelsDirective);
--- a/src/app/frontend/common/components/labels/labels.html
+++ b/src/app/frontend/common/components/labels/labels.html
@@ -1,0 +1,21 @@
+<!--
+Copyright 2015 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<md-content>
+  <div class="kd-labels" ng-repeat="(key, value) in ::labelsCtrl.labels">
+    {{key}}={{value}}
+  </div>
+</md-content>
--- a/src/app/frontend/common/components/labels/labels_controller.js
+++ b/src/app/frontend/common/components/labels/labels_controller.js
@@ -12,16 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import middleEllipsisFilter from './middleellipsis_filter';
-import relativeTimeFilter from './relativetime_filter';
-
 /**
- * Module containing common filters for the application.
+ * @final
  */
-export default angular.module(
-                          'kubernetesDashboard.common.filters',
-                          [
-                            'ngMaterial',
-                          ])
-    .filter('middleEllipsis', middleEllipsisFilter)
-    .filter('relativeTime', relativeTimeFilter);
+export default class LabelsController {
+  /**
+   * Constructs labels controller.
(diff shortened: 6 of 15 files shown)


src/app/frontend/common/components/label/label_controller.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @final
+ */
+export default class LabelController {
+  /**
+   * Constructs label controller.
+   * @ngInject
+   */
+  constructor() {
+    /** @export {!Array<Object.<string,string>>} Initialized from the scope and transformed to
```
bryk: No need for the dot: `Object.<`.

floreks: Done

src/app/frontend/common/components/label/label_controller.js
```diff
@@ -0,0 +1,44 @@
...
+    this.labels = this.transformLabels_();
+  }
+
+  /**
+   * Transforms labels object into array of simple key, value label objects. It is required
+   * in order to use md-chips for label display.
+   * @return {!Array<Object.<string,string>>}
+   * @private
+   */
+  transformLabels_() {
```
bryk: I believe you dont need the transformation. In md-chip-template you should have `$chip and $index` variables. Please verify.

bryk: See docs: https://material.angularjs.org/latest/api/directive/mdChips

src/app/frontend/common/components/label/label_directive.js
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+import LabelController from './label_controller';
+
+/**
+ * Returns directive definition for label.
+ *
+ * @return {!angular.Directive}
+ */
+export default function labelDirective() {
```
bryk: Is this **a** label or multiple labels? This should be reflected in the directive name.

src/test/frontend/common/components/label/label_controller_test.js
```diff
@@ -0,0 +1,58 @@
...
+    angular.mock.module(componentsModule.name);
+
+    angular.mock.inject(($controller, $rootScope) => {
+      // This is needed to mock variables initialized from scope with bindToController property.
+      let scope = $rootScope.$new();
+      let scopeData = {
+        labels: labelsMock,
+      };
+
+      ctrl = $controller(LabelController, {$scope: scope}, scopeData);
```
bryk: When you remove the logic from the controller, you can still have this test to verify that DOM tree has keys and values. There is at least one test in the codebase that does something similar. Look for `$compile`.

bryk:
Looks really nice. I've got a few comments on style. PTAL.

src/app/frontend/common/components/label/label_directive.js
floreks: Multiple labels. I'll change.

src/app/frontend/common/components/label/label.scss
```diff
@@ -0,0 +1,32 @@
...
+  > .md-chips {
+    box-shadow: none;
+    padding: inherit;
+  }
+
+  .md-chip {
+    border-radius: 0;
+    font-size: $caption-font-size-base;
+    height: $caption-font-size-base + 6;
+    line-height: $caption-font-size-base + 6;
```
maciaszczykm: This value (`+ 6`) repeats, can we export it to variable?

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -26,6 +26,7 @@ $replicasetdetails-table-cell: #777;
 
 .kd-replicasetdetail-sidebar {
   background-color: $replicasetdetails-sidebar-bg;
+  max-width: 315px;
```
maciaszczykm: Same as before, can we extract new variable (`315 px`)?

src/app/frontend/common/components/label/label_controller.js
floreks: The reason why I'm doing this is because md-chips needs array and what we get from backend is 1 object with lots of properties like `{app: 'name', version: 'v1'}`. I'm transforming object into array and md-chips iterates over this array. 

It's based on this demo:
https://material.angularjs.org/latest/demo/chips

maciaszczykm:
Got only two minor comments regarding stylesheets. LGTM

src/app/frontend/common/components/label/label.scss
floreks: How should it be called because I don't know. This is something like custom margin that we have in oder files without extracting to variable.

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: Sure

floreks:
PTAL.

Also what do you think about trying to apply similar look to labels in deploy page.

src/app/frontend/common/components/label/label_controller.js
bryk: How about using static chips, as at the bottom of the demo list: https://material.angularjs.org/latest/demo/chips ?

bryk: The reason I'm pushing this is the fact that this logic is not needed and non trivial. I.e., the controller is initialized with `labels` of type Foo which are then converted to type Bar. Meaning that `labels` are of two types, depending on when you look at them.

bryk:
PTAL

src/app/frontend/common/components/label/label.scss
maciaszczykm: I was thinking about local variable, which will show, that these values are somehow connected (equal). However, it's not mandatory.

maciaszczykm:
LGTM

src/app/frontend/common/components/label/label.scss
floreks: Sure. It can be refactored later when some appropriate name is proposed. Feel free to change it later in some PR. :)

src/app/frontend/common/components/label/label_controller.js
floreks: This was not as easy as I thought. I've had to fix karma `ng-html2js` configuration. It was not working. There were 2 issues:

1. `stripPrefix` path was wrong. It is merged with karma `basePath` and as result it was `../../src/app/frontend/`. I've changed `basePath` to point to current folder as all our `conf.paths` have `basePath` in them.
2. `moduleName` was pointing to non existing variable. It was undefined and not loaded. I've changed it to be loaded with `ng` module so it doesn't have to be mocked manually while writing tests.

Additionally in order to make `md-chips` work without additional label transformation I've had to use workaround. There are issues reported with making `ng-repeat` and `md-chips` work.

https://github.com/angular/material/issues/3482
https://github.com/angular/material/issues/2829

Last thing is that even after fixing karma and compiling directive it didn't show any `md-chip` in html code. It looks like another issue with `md-chips` directive. I've only checked if outer loop works and x `md-chips` tags are rendered.

floreks:
PTAL

bryk:
CC @kenmoore

bryk:
CC Ken, if you have any comments.

src/app/frontend/common/components/label/label_controller.js
floreks: I'm not sure if you have seen this comment because it's hidden. Can you confirm? @bryk

build/karma.conf.js
```diff
@@ -49,7 +49,7 @@ function getFileList() {
  */
 module.exports = function(config) {
   let configuration = {
-    basePath: conf.paths.base,
+    basePath: '.',
```
bryk: Thank you a lot for fixing this!

bryk:
Sorry for long review, but I'm experimenting with look and feel. I've a couple of questions/notes.

The problem I see is that the labels are too prominent and eye-catching. They are, kind of, secondary element of a card, yet they are bold and very visible. So I've attempted to make them less prominent. What I did is I reverted to rounded edges. And changed background to the same body has. Also, changed space between key and value to `=`.

Take a look and tell me what you think. 

selection_007

bryk:
@maciaszczykm @floreks See my previous comment.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -26,7 +27,8 @@ $replicasetdetails-table-cell: #777;
 
 .kd-replicasetdetail-sidebar {
   background-color: $replicasetdetails-sidebar-bg;
-  min-width: 315px;
+  max-width: $replicasetdetails-sidebar-width;
```
bryk: I believe this caused a bug:
selection_008

floreks: Ops actually I've set `$replicasetdetails-sidebar-width` variable to `315` instead of `315px`.

maciaszczykm:
@bryk I agree, that your proposal should be used. It's true, that bold label names were more visible than different (and more important) parts of view.

floreks:
Makes sense to me. They should not overwhelm user. I'll adjust.

bryk:
I did pre-merge verification of this and found troubling issues. Basically, md-chips is not designed for our use case. It is designed to work on input elements. Static chips work pretty bad, and even the docs say that this component is likely to change.

First thing I noticed is that every md-chips load asynchronously. See that in serve:prod the page does not load in one shot. Instead, the page loads and then chips load after a short while. Please test it out. I did an investigation in chrome and see what's happening:
selection_009
See more at: https://github.com/angular/material/blob/d1b7f79e92beced635190d23c307435aabcde9c5/src/components/chips/js/chipsDirective.js#L316


How about just styling our class to look like chips? In my tests it is actually less code and less CSS than current solution.

The template:
```html
<div ng-repeat="(key, value) in ::labelsCtrl.labels">
  <div class="kd-labels-chip">
    {{key}}={{value}}
  </div>
</div>
```
CSS copied and adjusted from chips. 

What do you think?

floreks:
Definitely. I didn't notice that. Anyway loading chips async does not fit our use case. I agree. Styling this to look as a chip should not be a problem. I'll change this. Thanks for taking a closer look at this.

floreks:
PTAL

src/app/frontend/common/components/labels/labels.scss
```diff
@@ -0,0 +1,31 @@
...
+@import '../../../variables';
+
+$label-height: $caption-font-size-base + 6;
+$label-margin: 0 $baseline-grid $baseline-grid 0;
+
+// Style taken(mostly) from angular material md-chip style.
+.kd-labels {
+  background-color: $body;
+  border-radius: $label-height / 2;
+  color: $muted;
```
bryk: You made the color $muted. How about removing color attribute here and just use default. $muted over $body is very hard to read. Normal over $body is easier to read.

bryk: What do you think?

bryk:
Looks great. I have last comment, I promise :)

src/app/frontend/common/components/labels/labels.scss
floreks: I was thinking about this. Pure black seemed a bit too distracting over the label background. Font color has to be set anyway because it will be overriden by other components if it's not set. Maybe something in the middle like `#444`?

bryk: If you omit color here it would be application default. Which is not full black, but dark grey. 

```css
md-content.md-default-theme, md-content {
    color: rgba(0,0,0,0.87);
}
```

floreks: Yes, but f.e. if I remove color at all from css then on replica set list page it will be black and on replica set details gray, because it inherits from more detailed classes like `kd-replicasetdetail-sidebar-subline`

bryk: I see. Can you take something darker from variables.scss? Or define there a variable for rgba(0,0,0,0.87). Or apply md-content class.

floreks:
I've surrounded it with `md-content` and it inherits color from it now. PTAL

bryk:
Awesome. Thank you. Will merge soon.

PR #227, opened 2016-01-07
Implement replica set card menu with details and delete links

bryk:


bryk:
@maciaszczykm Can you review?

maciaszczykm:
@bryk Sure, I'll do it tommorow morning.

src/app/frontend/replicasetdetail/deletereplicaset_service.js
```diff
@@ -0,0 +1,67 @@
...
+    let resource =
+        getReplicaSetDetailsResource(new StateParams(namespace, replicaSet), this.resource_);
+    let deferred = this.q_.defer();
+
+    this.mdDialog_
+        .show(
+            this.mdDialog_.confirm()
+                .title('Delete Replica Set')
+                .textContent(
+                    'All related pods will be deleted. Are you sure you want to delete this' +
```
maciaszczykm: I would consider adding replica set name to this message, just to confirm, which replica set will be deleted.

src/app/frontend/replicasetdetail/deletereplicaset_service.js
```diff
@@ -0,0 +1,67 @@
...
+
+import {StateParams} from './replicasetdetail_state';
+import {getReplicaSetDetailsResource} from './replicasetdetail_stateconfig';
+
+/**
+ * Opens replica set delete dialog.
+ *
+ * @final
+ */
+export class DeleteReplicaSetService {
```
maciaszczykm: Should it be so specific (service only for delete)? Maybe it could be named `ReplicaSetService` and contain other methods connected to operations on replica set.

maciaszczykm:
Looks and works good. I've left two comments.

src/app/frontend/replicasetdetail/deletereplicaset_service.js
bryk: I think that for now it can be for delete only. Later, when somebody wants to add new functionality to this service, the name can be changed. I think this is a better approach, because with it we do not design for the (unknown) future :) Only for what is now.

src/app/frontend/replicasetdetail/deletereplicaset_service.js
bryk: Makes sense. I've rephrased this sentence. PTAL

bryk:
PTAL

src/app/frontend/replicasetdetail/deletereplicaset_service.js
maciaszczykm: Makes sense to me.

maciaszczykm:
LGTM :)

bryk:
Thank you!

PR #229, opened 2016-01-08
Introduce Heapster to local Kubernetes cluster

zreigz:
This commit introduces Heapster to local Kubernetes cluster. Heapster listens on port 8082:
http://localhost:8082/api/v1/model/

To check if it is up and running:
http://localhost:8082/api/v1/model/stats
```
{
  "uptime": 99,
  "stats": {
   "cpu-limit": {
    "minute": {
     "average": 8000,
     "percentile": 8000,
     "max": 8000
    },
    "hour": {
   ...........


}
```

### DIFF
--- a/build/cluster.js
+++ b/build/cluster.js
@@ -27,6 +27,11 @@ import conf from './conf';
 const clusterHealthzUrl = `http://${conf.backend.apiServerHost}/healthz`;
 
 /**
+ * The validate URL of the heapster to check that it is running.
+ */
+const heapsterValidateUrl = `http://${conf.backend.heapsterServerHost}/validate`;
+
+/**
  * A Number, representing the ID value of the timer that is set for function which periodically
  * checks if cluster is running. The null means that no timer is running.
  *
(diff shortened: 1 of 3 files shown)


zreigz:
@cheld @bryk PTAL

cheld:
Nice, but I am not sure if wait-for-cluster is still valid. Is it required to wait for heapster to come up as well?

bryk:
I guess wait-for-cluster should also wait for heapster to go online.

build/hyperkube.sh
```diff
@@ -43,3 +43,7 @@ docker run \
 
 docker run -d --net=host --privileged gcr.io/google_containers/hyperkube:v${K8S_VERSION} \
     /hyperkube proxy --master=http://127.0.0.1:${PORT} --v=2
+
+# Runs Heapster in standalone mode which listen on port 8082
+docker run --net=host -d kubernetes/heapster \
```
bryk: Btw, can you specify the port here manually? This is not to rely on default port which may change

zreigz: Done

zreigz:
PTAL

build/cluster.js
```diff
@@ -76,9 +103,38 @@ gulp.task('spawn-cluster', function(doneFn) {
...
+      }
+    });
+  }
+});
+
+/**
  * Checks periodically if cluster is up and running.
  */
-gulp.task('wait-for-cluster', function(doneFn) {
+gulp.task('wait-for-cluster', ['wait-for-heapster'], function(doneFn) {
```
bryk: Should wait-for-cluster only start after wait-for-heapster is done? I dont think so. The tasks look like independent ones and can be separated. 

If you need one task, then you can create a task wait-for-FOO which depends on these two.

bryk:
Last comment. Sorry for lag in the review - the notification got buried somewhere.

bryk:
PTAL

build/cluster.js
zreigz: I will add wait-for-heapster to local-up-cluster to run those two tasks independently 
```gulp.task('local-up-cluster', ['spawn-cluster', 'wait-for-cluster', 'wait-for-heapster']);```
Is is OK ?

bryk: Looks good.

zreigz:
PTAL

bryk:
LGTM

Issue #228, opened 2016-01-07
Request logging on the backend


maciaszczykm:
I'll add logging for requests coming to the backend.

maciaszczykm:
@bryk @cheld @floreks @zreigz I am thinking about architecture of this implementation and it would be great if you would have any suggestions. Currently I can see two ways how it could be implemented:

- We're using [emicklei/go-restful](https://github.com/emicklei/go-restful) to provide restful webservices, and we can try to use logging provided by it. It needs to be configured in the way, that we want (still not sure about it, what do we need in logs?).
- We can log every request from `apihandler.go` file. It would be easier to configure and messages could be more specific, but it won't be as generic as first solution.

Do you have any tips? Maybe you can see another solution?

floreks:
I think it would be good to use what `go-restful` provides. We can define for every web service different log function or common for every ws. This example shows it pretty good.

https://github.com/emicklei/go-restful/blob/master/examples/restful-NCSA-logging.go

Only thing to establish is whether we need to log differently every WS or use commong log formatting for all.

bryk:
We should have two layers of logging. One generic, which basically logs all HTTP requests, without knowing what they do. This is actually very useful for debugging. It can show method, URI, cookies or headers. The other specific to api methods, which would help in debugging, like `log.info('creating replica set %#v', foo)`.

cheld:
I cannot help much as I have briefly looked in the backend code yet.

In Java I would recommend not to mix logs during concurrent rest calls and add some kind of session information to each log. So logs are clean and can easily be filtered for a specific session or specific action. 
But, in our case everything different so just ignore my comment :)

PR #230, opened 2016-01-08
Backend logging

maciaszczykm:
As it was discussed in issue #228 added logging in the backend. At the moment it's the part, that is specific to API methods. I will add generic part via another pull request. I've noticed also an issue with `glog` setup. There is following error before every log: `ERROR: logging before flag.Parse`. I will work on it in next pull request.

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -17,9 +17,9 @@ package main
 import (
 	"flag"
 	"fmt"
+	"log"
 	"net/http"
 
-	"github.com/golang/glog"
 	"github.com/spf13/pflag"
 )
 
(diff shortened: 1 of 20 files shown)


maciaszczykm:
@floreks @bryk Can you take a look? Should there be more log messages, do we need to use different log levels like trace (to log called method names for example) and so on?

src/app/backend/dashboard.go
```diff
@@ -32,6 +32,7 @@ var (
 )
 
 func main() {
+	//TODO(maciaszczykm): Resolve issue with logs (ERROR: logging before flag.Parse).
```
bryk: I think to solve this problem we can ditch glog package and start using just normal golang log package. What do you think?

bryk:
Just skimmed, looks really nice. Will do a full review on Monday.

bryk:
Thing that strikes me is whether on errors we should log and pass the error down. I think not. This is because with such patter we would end up with the same error logged in multiple places. This is the same as log-and-throw is considered antipattern, e.g., http://stackoverflow.com/questions/6639963/why-is-log-and-throw-condsidered-an-anti-pattern.

I'd suggest we do not log-and-throw, but instead log errors in error handler we have in the apihandler.

src/app/backend/dashboard.go
maciaszczykm: I agree. I've used `glog`, because it was already used in `dashboard.go` and there are logging levels available. Let's use golang log as you said.

maciaszczykm:
Used `goimports` on all Go files.

bryk:
LGTM :)

PR #239, opened 2016-01-11
Request and response logging in the backend

maciaszczykm:
Added request and response logging in the backend for all registered webservices. Related to #228 issue.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -15,13 +15,62 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"net/http"
 	"strconv"
+	"strings"
 
 	restful "github.com/emicklei/go-restful"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 )
+
+const (
+	Colon             = ":"
+	RequestLogString  = "Incoming %s %s %s request from %s:\n%v"
+	ResponseLogString = "Outcoming response to %s with %d status code"
+)
+
+// Web-service filter function used for request and response logging.
+func wsLogger(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
+	remoteAddr := GetRemoteAddr(req)
+	log.Printf(FormatRequestLog(req, remoteAddr))
+	chain.ProcessFilter(req, resp)
+	log.Printf(FormatResponseLog(resp, remoteAddr))
+}
+
+// Returns remote address of the request (without port number).
+func GetRemoteAddr(req *restful.Request) string {
+	if strings.Contains(req.Request.RemoteAddr, Colon) {
+		return strings.Split(req.Request.RemoteAddr, Colon)[0]
+	} else {
+		return req.Request.RemoteAddr
+	}
+}
+
+// Formats request log string.
+func FormatRequestLog(req *restful.Request, remoteAddr string) string {
+	reqBody := make([]byte, req.Request.ContentLength)
+
+
+	if req.Request.Body != nil {
+		req.Request.Body.Read(reqBody)
+	}
+
+	reqURI := ""
+	if req.Request.URL != nil {
+		reqURI = req.Request.URL.RequestURI()
+	}
+
+	return fmt.Sprintf(RequestLogString, req.Request.Proto, req.Request.Method,
+		reqURI, remoteAddr, reqBody)
+}
+
+// Formats response log string.
+// TODO(maciaszczykm): Display response content too.
+func FormatResponseLog(resp *restful.Response, remoteAddr string) string {
+	return fmt.Sprintf(ResponseLogString, remoteAddr, resp.StatusCode())
+}
 
 // Creates a new HTTP handler that handles all requests to the API of the backend.
 func CreateHttpApiHandler(client *client.Client) http.Handler {
@@ -29,6 +78,7 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 	wsContainer := restful.NewContainer()
 
 	deployWs := new(restful.WebService)
+	deployWs.Filter(wsLogger)
 	deployWs.Path("/api/appdeployments").
 		Consumes(restful.MIME_JSON).
 		Produces(restful.MIME_JSON)
@@ -45,6 +95,7 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 	wsContainer.Add(deployWs)
 
 	replicaSetWs := new(restful.WebService)
+	replicaSetWs.Filter(wsLogger)
 	replicaSetWs.Path("/api/replicasets").
 		Consumes(restful.MIME_JSON).
 		Produces(restful.MIME_JSON)
@@ -70,6 +121,7 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 	wsContainer.Add(replicaSetWs)
 
 	namespacesWs := new(restful.WebService)
+	namespacesWs.Filter(wsLogger)
 	namespacesWs.Path("/api/namespaces").
 		Consumes(restful.MIME_JSON).
 		Produces(restful.MIME_JSON)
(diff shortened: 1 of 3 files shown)


maciaszczykm:
@floreks @bryk Could you review?

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,35 @@ import (
 	"log"
 	"net/http"
 	"strconv"
+	"strings"
 
 	restful "github.com/emicklei/go-restful"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 )
 
+func WsLogger() restful.FilterFunction {
```
bryk: Can we also log JSON payload of the response and request?

bryk:
A question. PTAL

src/app/backend/apihandler.go
maciaszczykm: I've already added logging of JSON payload of the request, but for the response, which is more important, it's a little bit more complicated. It seems, that we don't have access to response content by `restful.Response` object and to change it, we would need to override/extend it somehow. Do you see any solution for this? Will it be acceptable to perform such modifications?

maciaszczykm:
Question from my side too. PTAL

src/app/backend/apihandler.go
bryk: If this is a lot of work, let's submit as it is. And later add response logging, when we find stronger need.

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,39 @@ import (
...
+		log.Printf("Incoming %s %s %s request from %s with %d content length:\n%v",
+			req.Request.Proto, req.Request.Method, req.Request.URL.RequestURI(), remoteAddr,
+			req.Request.ContentLength, reqBody,
+		)
+
+		chain.ProcessFilter(req, resp)
+		respBody := "[]" // TODO(maciaszczykm): Get response content.
+
+		log.Printf("Outcoming response to %s with %d status code and %d content length:\n%v",
+			remoteAddr, resp.StatusCode(), resp.ContentLength(), respBody,
```
bryk: ContentLength? I don't think this is useful. I'd remove.

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,39 @@ import (
...
+		req.Request.Body.Read(reqBody)
+		remoteAddr := strings.Split(req.Request.RemoteAddr, ":")[0]
+
+		log.Printf("Incoming %s %s %s request from %s with %d content length:\n%v",
+			req.Request.Proto, req.Request.Method, req.Request.URL.RequestURI(), remoteAddr,
+			req.Request.ContentLength, reqBody,
+		)
+
+		chain.ProcessFilter(req, resp)
+		respBody := "[]" // TODO(maciaszczykm): Get response content.
```
bryk: Let's revert this for now.

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,39 @@ import (
...
 
 	restful "github.com/emicklei/go-restful"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 )
 
+func WsLogger() restful.FilterFunction {
+	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
+		reqBody := make([]byte, req.Request.ContentLength)
+		req.Request.Body.Read(reqBody)
+		remoteAddr := strings.Split(req.Request.RemoteAddr, ":")[0]
```
bryk: Are you sure that remoteAddr always contain `:`? I'm not. You should check this and do nothing when there's no `:`.

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,39 @@ import (
...
 	"net/http"
 	"strconv"
+	"strings"
 
 	restful "github.com/emicklei/go-restful"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 )
 
+func WsLogger() restful.FilterFunction {
+	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
```
bryk: Can you test this?

bryk:
Looks very good. PTAL :)

src/app/backend/apihandler.go
maciaszczykm: Okay, I'll remove. I was using it before displaying content to check if there is any.

src/app/backend/apihandler.go
maciaszczykm: Okay, I'll leave just comment in the code.

src/app/backend/apihandler.go
maciaszczykm: You're right. It's fixed now.

src/app/backend/apihandler.go
maciaszczykm: I'm not sure how it should like. Do you have any suggestions?

maciaszczykm:
Adressed all comments, but not last one, where I'm not sure how to proceed. PTAL

bryk:
Which last one? The one about tests?

maciaszczykm:
Yes, how we should test logging function?

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,43 @@ import (
...
 
 	restful "github.com/emicklei/go-restful"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 )
 
+const (
+	Colon = ":"
+)
+
+func wsLogger() restful.FilterFunction {
```
bryk: Btw, why this is a function that returns a function? I'm pretty sure this can be just a function that matches the contract.

See the typedef: https://github.com/emicklei/go-restful/blob/master/filter.go

src/app/backend/apihandler.go
```diff
@@ -18,17 +18,43 @@ import (
...
+	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
```
bryk: Also, this needs a documentation note. What it does and why.

bryk:
You can test lines 32-42 to verify that correct log string is created. And then log it. Create a helper function that returns the string. Full coverage is not needed.

PTAL

cheld:
Hmm...shouldn't the logs be formatted in JSON? Otherwise - I assume - the logs will not parse properly in common logging solutions like loggly or the Kibana add-on of Kubernetes. Just full-text search will work. Anyway, it is ok not to care at the moment. We can leave it as an improvement for a future release.

src/test/backend/apihandler_test.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, c := range cases {
+		actual := GetRemoteAddr(c.request)
+		if !reflect.DeepEqual(actual, c.expected) {
+			t.Errorf("GetRemoteAddr(%#v) == %#v, expected %#v", c.request, actual, c.expected)
+		}
+	}
+}
+
+func TestFormatRequestLog(t *testing.T) {
+	// TODO(maciaszczykm): Implement test.
```
bryk: LGTM. Just implement the test :)

bryk:
Implement the tests and we're good to go :)

bryk:
> Hmm...shouldn't the logs be formatted in JSON?

@cheld I dont know. I've never seen other logs than plain text. AFAIK most logging tools is designed to parse output from "standard" loggers. But maybe I've been using different tooling.

cheld:
@bryk yes, traditionally, an admin would install the dashboard binary on a vm and install a log forwarder like logstash that will tokenize the logs and forward to the logging service. In Kubernetes case the logs are forwarded by fluentd without tokenizing. As a result, e.g. the timestamp in the log might be ignored and search result might not match expectation. 

Anyway, was just a comment. We should not spend time on investigation, currently.

src/test/backend/apihandler_test.go
maciaszczykm: Yes, I'll do it next week. I've left comments in code to remember about it :)

maciaszczykm:
@bryk Implemented tests. PTAL

bryk:
Awesome, looks really nice. Merging.

PR #241, opened 2016-01-12
Style and productionize replica set details page

bryk:
* make fonts larger (md-body instead of md-caption)
* use only color and sizes from variables.scss
* use less different colors
* simplify CSS rules

Please check out this change and compare to previous version.

### DIFF
--- a/src/app/backend/events.go
+++ b/src/app/backend/events.go
@@ -80,6 +80,7 @@ func GetEvents(client *client.Client, namespace, replicaSetName string) (*Events
 
 	events := AppendEvents(rsEvents, Events{
 		Namespace: namespace,
+		Events:    make([]Event, 0),
 	})
 
 	// Get events for pods in replica set.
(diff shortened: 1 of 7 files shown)


bryk:

floreks:
I've comparted with old version and this seems more readable. Part of css is also gone which is good. Looks good to me.

maciaszczykm:
I've checked it and I've got a few questions regarding the view. Take a look at following screenshot:

image

- red - I think it would look better if `margin-top` would be bigger for delete button.
- green - If size of picker is not fixed, then if we switch already chosen option, then the size changes. In my opinion we should avoid it.
- blue - Height of cells in events table is way bigger than in pods table. 

However, stylesheets look a lot better now :) PTAL

bryk:
* red - fixed - added padding of $baseline-grid
* green - fixed - set the width to 15 * $baseline-grid
* blue - fixed - now the heights are the same 
* outline of sort headers - fixed

bryk:
I plan to merge this now, since you're OOO. If you have any further comments, post them here and I'll address them.

maciaszczykm:
Sure, we'll notice if anything more needs further adjustments in the next week :) However, I think I didn't notice anymore issues in this view, which are connected to refactoring.

Issue #240, opened 2016-01-12
Deploy error when expose service set to true and protocol type is "UDP"


floreks:
I've found a bug as described in the title. According to documentation:
[Note that while NodePorts can be TCP or UDP, LoadBalancers only support TCP as of Kubernetes 1.0.](http://kubernetes.io/v1.1/docs/user-guide/services.html)

Currently I'm working on fixing this.

floreks:
It's done so I can close.

PR #242, opened 2016-01-12
Moved protocol type data to backend and added validation for chosen protocol type

floreks:
Fix for issue #240. I've moved protocols retrieval to backend also and it's synced with expose service switch.

### DIFF
--- a/src/app/externs/backendapi.js
+++ b/src/app/externs/backendapi.js
@@ -230,3 +230,25 @@ backendApi.AppNameValiditySpec;
  * }}
  */
 backendApi.AppNameValidity;
+
+/**
+ * @typedef {{
+ *    protocols: !Array<string>
+ * }}
+ */
+backendApi.Protocols;
+
+/**
+ * @typedef {{
+ *    valid: boolean
+ * }}
+ */
+backendApi.ProtocolValidity;
+
+/**
+ * @typedef {{
+ *    protocol: string,
+ *    isExternal: boolean
+ * }}
+ */
+backendApi.ProtocolValiditySpec;
--- a/src/app/frontend/deploy/deploy_stateconfig.js
+++ b/src/app/frontend/deploy/deploy_stateconfig.js
@@ -28,6 +28,8 @@ export default function stateConfig($stateProvider) {
     url: '/deploy',
     resolve: {
       'namespaces': resolveNamespaces,
+      'protocolsResource': getProtocolsResource,
+      'protocols': getDefaultProtocols,
     },
     templateUrl: 'deploy/deploy.html',
   });
@@ -43,4 +45,22 @@ function resolveNamespaces($resource) {
   let resource = $resource('api/namespaces');
 
   return resource.get().$promise;
+}
+
+/**
+ * @param {!angular.$resource} $resource
+ * @return {!angular.Resource<!backendApi.ReplicaSetSpec>}
+ * @ngInject
+ */
+function getProtocolsResource($resource) {
+  return $resource('api/appdeployments/protocols');
+}
+
+/**
+ * @param {!angular.Resource<!backendApi.Protocols>} protocolsResource
+ * @returns {!angular.$q.Promise}
+ * @ngInject
+ */
+function getDefaultProtocols(protocolsResource) {
+  return protocolsResource.get().$promise;
 }
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -68,11 +68,10 @@ export default class DeployFromSettingsController {
     this.description = '';
 
     /**
-     * List of supported protocols.
-     * TODO(bryk): Do not hardcode the here, move to backend.
-     * @const @export {!Array<string>}
+     * Initialized from the scope.
+     * @export {!Array<string>}
      */
-    this.protocols = ['TCP', 'UDP'];
+    this.protocols;
 
     /** @export {!Array<!backendApi.PortMapping>} */
     this.portMappings = [this.newEmptyPortMapping_(this.protocols[0])];
--- a/src/app/frontend/deploy/validprotocol_directive.js
+++ b/src/app/frontend/deploy/validprotocol_directive.js
@@ -1,0 +1,89 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/** The name of this directive. */
+export const validProtocolValidationKey = 'validProtocol';
+
+/**
+ * Validates that chosen protocol is valid for chosen service type.
+ *
+ * @param {!angular.$resource} $resource
+ * @param {!angular.$q} $q
+ * @return {!angular.Directive}
+ * @ngInject
+ */
+export default function validProtocolDirective($resource, $q) {
+  const isExternalParam = 'isExternal';
+
+  return {
+    restrict: 'A',
+    require: 'ngModel',
+    scope: {
+      [isExternalParam]: '=',
+    },
+    link: function(scope, element, attrs, ctrl) {
+      /** @type {!angular.NgModelController} */
+      let ngModelController = ctrl;
+      let externalChanged = false;
+
+      scope.$watch(isExternalParam, (newVal, oldVal) => {
+        externalChanged = newVal !== oldVal;
+        ctrl.$validate();
+      });
+
+      ngModelController.$asyncValidators[validProtocolValidationKey] = (protocol) => {
+        return validate(
+            ngModelController, externalChanged, protocol, scope[isExternalParam], $resource, $q);
+      };
+    },
+  };
+}
+
+/**
+ * @param {!angular.NgModelController} ngModelController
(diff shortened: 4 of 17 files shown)


src/app/externs/backendapi.js
```diff
@@ -218,3 +218,17 @@ backendApi.AppNameValiditySpec;
...
+ * }}
+ */
+backendApi.Protocols;
+
+/**
+ * @typedef {{
+ *    isExternal: boolean
+ * }}
+ */
+backendApi.ProtocolsSpec;
```
floreks: To meet half way I've added ProtocolsSpec to frontend so it can be used for documentation and in backend I've added comment explaining `isExternal`.

floreks:
@maciaszczykm @bryk review please :)

src/app/externs/backendapi.js
bryk: Nice:)

src/app/frontend/deploy/deploy_stateconfig.js
```diff
@@ -44,3 +46,23 @@ function resolveNamespaces($resource) {
...
+}
+
+/**
+ * @param {!angular.Resource<!backendApi.Protocols>} protocolsResource
+ * @returns {!angular.$q.Promise}
+ * @ngInject
+ */
+function getDefaultProtocols(protocolsResource) {
+  /** @type {!backendApi.ProtocolsSpec} */
+  let protocolsSpec = {isExternal: false};
```
bryk: Should default isExternal value be injected to the controller? I say yes. It should be easy to add. At least, create a constant for it.

src/app/frontend/deploy/deployfromsettings_controller.js
```diff
@@ -183,6 +188,30 @@ export default class DeployFromSettingsController {
...
+    /** @type {!backendApi.ProtocolsSpec} */
+    let protocolsQuery = {isExternal: this.isExternal};
+
+    this.protocolsResource.get(
+        protocolsQuery,
+        (protocols) => {
+          defer.resolve(protocols);
+          this.log_.info('Protocols successfuly retrieved: ', protocols.protocols);
+          this.protocols = protocols.protocols;
+          this.portMappings = [this.newEmptyPortMapping_(this.protocols[0])];
```
bryk: This line causes protocol to be cleared out when I have already something filled in. Can you fix this?

bryk:
PTAL

src/app/frontend/deploy/deploy_stateconfig.js
floreks: I see 2 solutions:

1. Add locals `isExternal` to state config, but then propagating this would require adding yet another binding to deploy from settings directive.
2. Create export const in deploy from settings controller then import and use in here.

I actually like first option better. What do you think?

src/app/frontend/deploy/deployfromsettings_controller.js
floreks: Sure. Didn't notice that.

src/app/frontend/deploy/deploy_stateconfig.js
floreks: Actually I don't see any locals that can be defined on state config. Must have mistaken it with something else. This leaves me with 2nd option.

floreks:
PTAL

src/app/frontend/deploy/deployfromsettings_controller.js
```diff
@@ -183,6 +190,33 @@ export default class DeployFromSettingsController {
   }
 
   /**
+   * @returns {!angular.$q.Promise}
+   * @export
+   */
+  updateAvailableProtocols() {
```
bryk: Hmmm... The more I think about this solution, the more problems I find. I.e., this code has race conditions when you change this.isExternal until this promise fulfills. Also, on changing this.isExternal, suddenly all your protocol selections get changed too. 

How about making this an async validator? Like the unique-name-validator? It can connect to angular form processing pipeline to take care of race conditions and would display message like: "this protocol is not supported for external services". The benefit of the error message is that user immediately sees what is wrong and what should/can be change. When we silently change protocols of all ports, this may be unnoticed.

bryk:
One major comment. PTAL

src/app/frontend/deploy/deployfromsettings_controller.js
floreks: Yes, I've even written in my first comment to you that I think that until data is retrieved from the server combobox should be disabled or marked with some progress bar. The switch also. Another solution is doing this async as you proposed.

I was considering validator at first, but this didn't seem like something that fits to be 'validator'

bryk: Yeah, I'm sorry I didn't make this comment at first. It only arose after I've been testing this on my machine. 

The validator solution seems better and better to me the more I think about it. Since this is like saying: "This protocol is invalid for external services", which looks good.

floreks: If you say it this way then sure, it looks like validator ;). I will change it to validator.. Unfortunately this has to wait till the next week.

bryk: No problems :) Take your time on the training

floreks:
@bryk Could you check new concept? Is this what you've had in mind?

bryk:
> I've got one problem. On page load protocol validation is fired. I'd like to avoid that but I can't find the solution to fix that. Could you take a look at that?

You can check whether the ng-model or even entire form is $touched or $pristline. And ignore validation when it is.

src/app/frontend/deploy/deploy_stateconfig.js
```diff
@@ -44,3 +46,21 @@ function resolveNamespaces($resource) {
...
   return resource.get().$promise;
 }
+
+/**
+ * @param {!angular.$resource} $resource
+ * @return {!angular.Resource<!backendApi.ReplicaSetSpec>}
+ * @ngInject
+ */
+function getProtocolsResource($resource) {
+  return $resource('/api/appdeployments/protocols');
```
bryk: Dont start api calls with slash. start with `('api/foo'`

bryk:
The concept looks really nice. Please go with final implementation. 

I've noticed bugs in it, though. Make sure to do monkey-like clicking and ensure that everything works as expected. If you run into any hard issues, let me know :)

floreks:
Thanks for pointing me to the right direction. I was looking for $touched. :) I've written tests and added some `debounce` to external service switch in order to avoid "visual bugs". PTAL

src/app/frontend/deploy/validprotocol_directive.js
```diff
@@ -0,0 +1,89 @@
...
+    scope: {
+      [isExternalParam]: '=',
+    },
+    link: function(scope, element, attrs, ctrl) {
+      /** @type {!angular.NgModelController} */
+      let ngModelController = ctrl;
+      let externalChanged = false;
+
+      scope.$watch(isExternalParam, (newVal, oldVal) => {
+        externalChanged = newVal !== oldVal;
```
bryk: This statement is always true.

Btw, why do you need it? Why is the variable `externalChanged` there? I think we don't need it. It is a small short-circut optimization that gives speedup in very few cases, but leaves business logic in the validator. Adding this code means that the validator assumes that unchanged isExternal is always valid. This is true assumption today, but may not be true in the future. And I bet that whoever changes this assumption, they will not look at this code.

bryk:
One minor comment, PTAL.

Also, can you cover entire src/app/frontend/deploy/validprotocol_directive.js file? There's one branch missing.

src/app/frontend/deploy/validprotocol_directive.js
floreks: Actually it is not always true. It is false on the page load. When te first time it's called.

screenshot from 2016-01-20 14-55-17

Without checks on page load validation is called 3 times. I wanted to avoid that. Now validation is called only when external service switch is clicked or protocol type is changed without additional validation when page is loaded.

bryk: Yeah, but on page load $touched is false, so the validation does not trigger, right?

floreks: Sure, but expose service switch does not affect ngModelController that is in protocols directive. Validation is not fired at all when switch is clicked until protocols form is touched. We can assume that our initial config is ok and leave it without this, but then if for some reason TCP and LoadBalancer service type will be also not supported then validation will not be fired. 

This is the only way to make switch and protocols form fire validation independent of each other.

floreks:
It should be covered now.

src/app/frontend/deploy/validprotocol_directive.js
bryk: I was talking only about removing `externalChanged`. Leaving $touched test here is still a problem, but I don't think we can overcome it at this moment.

floreks: Sure I can remove it. We have to remember however that service switch will not trigger validation until protocol is changed to UDP.

bryk: Hmm... If you have code like this, the validation will *always* trigger on isExternal change. And wait util you touch protocol.

```js
scope.$watch(isExternalParam, () => { ctrl.$validate() }
```
```js
 if (!ngModelController.$touched) {
   deferred.resolve();
   return deferred.promise;
 }
```

bryk: Am I correct?

floreks: It will be triggered on isExternal but there will be no call to the backend for validation. I will give an example:

Initially we have `TCP` and `false` set. That is ok we do not to validate that as we assume that initial config is correct. User then clicks on external service switch so we have `TCP` and `true`. If for some reason backend logic will change and `TCP` and `true` will no longer be a valid combination then validation will not be fired until user touches protocol combobox.

bryk: I see! Thanks for the persistence and explanation. Dont change this.

bryk:
LGTM. Will merge soon.

PR #246, opened 2016-01-13
Implement backend method to retrieve replica set pods CPU & memory usage

zreigz:


src/app/backend/apihandler.go
```diff
@@ -279,6 +284,20 @@ func (apiHandler *ApiHandler) handleEvents(request *restful.Request, response *r
 	response.WriteHeaderAndEntity(http.StatusCreated, result)
 }
 
+// Handles get Replica Set Pods metrics.
+func (apiHandler *ApiHandler) handleReplicaSetPodsMetrics(request *restful.Request,
```
bryk: Hmm... Why do we need a new api handler? Shouldn't the metrics be included in replicasetdetail response?

zreigz: OK I will move it.

build/serve.js
```diff
@@ -116,6 +116,7 @@ gulp.task('spawn-backend', ['backend', 'kill-backend'], function() {
       [
         `--apiserver-host=http://${conf.backend.apiServerHost}`,
         `--port=${conf.backend.devServerPort}`,
+        `--heapster-host=http://${conf.backend.heapsterServerHost}`,
```
bryk: I'm thinking about this a lot. The problem is that we want to talk with in-cluster heapster via service proxy. This is how clients would use. This is how my cluster works. This setup is only needed when you have our current dockerized cluster. 

Do you think it is possible to run heapster __inside__ the cluster? This would solve problems. 

If this is not possible, then we need to come up with a clever solution to make the Dashboard use in-cluster heapster   when available and external one when not available.

bryk:
One major comment. PTAL

build/serve.js
zreigz: 1 solution
It is possible run heapster inside cluster using kubectl command. Here you can find yaml files for it:
https://github.com/kubernetes/heapster/tree/master/deploy/kube-config/standalone
We need gulp task to get kubectl and second to create heapster.

2 solution 
Add heapster to hyperkube image and publish it as new huperkube-heapster image.

zreigz: PTAL

bryk:
Ping me when ready for rereview.

zreigz:
Ok I was struggling with service proxy but it was docker problem. When I test it I will ping you.

zreigz:
Please don't review. I still work on this

zreigz:
PTAL
One remark. Heapster does not give the containers' percentage of CPU usage:
https://github.com/kubernetes/heapster/issues/296

build/serve.js
```diff
@@ -136,6 +137,7 @@ gulp.task('spawn-backend:prod', ['build-frontend', 'backend', 'kill-backend'], f
       [
```
bryk: Can you move all three arguments of the binary to a common constant in the top of the file? The list of arguments is growing - let's avoid desyncing it.

src/app/backend/dashboard.go
```diff
@@ -42,9 +50,18 @@ func main() {
 		log.Print(err)
 	}
 
+	if *argHeapsterHost != "" {
```
bryk: Can you create heapsterclient.go file and do this logic there? The reason is that we really want to keep main function short and do not contain any logic.

src/app/backend/dashboard.go
```diff
@@ -42,9 +50,18 @@ func main() {
...
+		heapsterClient, err = CreateApiserverClient(*argHeapsterHost, new(ClientFactoryImpl))
+		if err != nil {
+			log.Print(err)
+		}
+	} else {
+		log.Print("Heapster not defined. Probably installed inside cluster")
```
bryk: Should we have a TODO here?

src/app/backend/dashboard.go
```diff
@@ -29,9 +30,16 @@ var (
...
 		"http://localhost:8080. If not specified, the assumption is that the binary runs inside a"+
 		"Kubernetes cluster and local discovery is attempted.")
+	argHeapsterHost = pflag.String("heapster-host", "", "The address of the Heapster Apiserver "+
+		"to connect to in the format of protocol://address:port, e.g., "+
+		"http://localhost:8082. If not specified, the assumption is that the binary runs inside a"+
+		"Kubernetes cluster and local discovery is attempted.")
 )
 
 func main() {
+        // When Heapster host not defined then heapster client will be null
```
bryk: It is only for the time being? Later there will be in cluster client? If yes, please add a todo.

src/app/backend/apihandler.go
```diff
@@ -167,7 +167,8 @@ func CreateHttpApiHandler(client *client.Client) http.Handler {
 }
 
 type ApiHandler struct {
-	client *client.Client
+	client         *client.Client
+	heapsterClient *client.Client
```
bryk: HeapsterClient should be RESTClient, not client.client. The latter is for services, pods, etc.

src/app/backend/replicasetpodsmetrics.go
```diff
@@ -0,0 +1,147 @@
...
+		metricName)
+}
+
+// Retrieves raw metrics from Heapster.
+func getRawMetrics(client *client.Client, heapsterClient *client.Client, metricPath string) ([]byte, error) {
+
+	var resultRaw []byte
+	var err error
+        // Service Proxy is used when heapster client is nil. Heapster probably installed inside cluster.
+	if heapsterClient == nil {
```
bryk: You should not be doing this logic here. Just create a RESTClient in dashboard.go (or heapsterclient.go) and use it here.

bryk:
Please make sure that the code works correctly when there is no heapster installed.

bryk:
PTAL

zreigz:
PTAL

src/app/backend/replicasetdetail.go
```diff
@@ -117,7 +121,7 @@ type ReplicaSetSpec struct {
 }
 
 // Returns detailed information about the given replica set in the given namespace.
-func GetReplicaSetDetail(client *client.Client, namespace, name string) (*ReplicaSetDetail, error) {
+func GetReplicaSetDetail(client *client.Client, heapsterClient *unversionedClient.RESTClient, namespace, name string) (*ReplicaSetDetail, error) {
```
bryk: Wrap this line?

src/app/backend/replicasetpodsmetrics.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+const (
+	cpuUsage                 = "cpu-usage"
+	memoryUsage              = "memory-usage"
+	megaByte                 = 1024 * 1024
+	defaultHeapsterNamespace = "kube-system"
+	defaultHeapsterScheme    = "http"
+	defaultHeapsterService   = "heapster"
+	defaultHeapsterPort      = "" // use the first exposed port on the service
```
bryk: Not needed anymore. As other fields here.

src/app/backend/replicasetpodsmetrics.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Return Pods metrics for Replica Set or error when occurred.
+func getReplicaSetPodsMetrics(podList *api.PodList,
+	heapsterClient *unversioned.RESTClient, namespace string, replicaSet string) (
+	*ReplicaSetMetricsByPod, error) {
+	log.Printf("Getting Pods metrics for Replica Set %s in %s namespace",
+		replicaSet,
+		namespace)
+	podNames := make([]string, 0)
+	emptyMetrics := &ReplicaSetMetricsByPod{
```
bryk: Should this be a constant?

src/app/backend/replicasetpodsmetrics.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Metrics map by pod name.
+type ReplicaSetMetricsByPod struct {
+	// Metrics map by pod name
+	MetricsMap map[string]PodMetrics `json:"metricsMap"`
+}
+
+// Pod metrics structure.
+type PodMetrics struct {
+	// Cumulative CPU usage on all cores in nanoseconds.
```
bryk: Hmm... The return value should be something directly display-able in the UI. For example, milicores.

src/app/backend/replicasetpodsmetrics.go
```diff
@@ -0,0 +1,151 @@
...
+	CpuUsage uint64 `json:"cpuUsage"`
+	// Pod memory usage in MB.
```
bryk: How about send bytes? Then the UI can show its fraction, when the usage is small.

src/app/backend/replicasetpodsmetrics.go
zreigz: I totally agree with you. But as I mentioned  there is still open issue kubernetes/heapster#296
We have to figure out how to recalculate it. I can create also issue for Dashboard project regarding this CPU usage. Unfortunately I have to switch to another project next week so someone else can handle this.

src/app/backend/replicasetpodsmetrics.go
zreigz: Const can only be numbers, characters (runes), strings or booleans.

src/app/backend/replicasetpodsmetrics.go
zreigz: done

src/app/backend/replicasetpodsmetrics.go
zreigz: done

src/app/backend/replicasetdetail.go
zreigz: done

zreigz:
PTAL

bryk:
Ok, let's merge this. The code does not quite work for in-cluster heapster, but I'll debug it and send a PR.

PR #251, opened 2016-01-14
Considerations for i18n in dashboard - Overview & angular-translate

taimir:
### DO NOT MERGE - ONLY MEANT FOR DISCUSSION

With this pull request I want to start the discussion for the i18n & l10n of the Dashboard project, using 2 concrete tools as a "proof of concept" of sorts:

* angular-translate
* Google Closure Templates for translation

I've summarized my knowledge on those two frameworks in the following design document:
https://github.com/taimir/dashboard/blob/i18n-angular-translate/docs/design/i18n.md

They are also discussed in this AngularJS design document, dealing with i18n:
https://docs.google.com/document/d/1mwyOFsAD-bPoXTk3Hthq0CAcGXCUw-BtTJMR4nGTY-0

This pull request contains a simple integration of angular-translate, please check the changes in the code to get an idea of the basic usage. For a Closure Templates example, please refer to this pull request: https://github.com/kubernetes/dashboard/pull/252

From my perspective, angular-translate is easier to integrate in our project, but lacks message extraction (into XLIFF files, good for human translators). Closure Templates has this, but does not fit our project that nicely. I really want to hear other opinions on this topic and am open for other tools I could try out.

@bryk @cheld @floreks @maciaszczykm @SebastianM

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,7 +12,8 @@
     "angular-messages": "~1.4.8",
     "angular-ui-router": "~0.2.15",
     "angular-resource": "~1.4.8",
-    "angular-sanitize": "~1.4.8"
+    "angular-sanitize": "~1.4.8",
+    "angular-translate": "~2.8.1"
   },
   "devDependencies": {
     "angular-mocks": "~1.4.8",
--- a/docs/design/i18n-angular-translate.md
+++ b/docs/design/i18n-angular-translate.md
@@ -1,0 +1,19 @@
+# Using angular-translate
+Angular JS is currently an external project supporting AngularJS 1.X and is entirely based on the Angular functionality. It relies heavily on the data-bindings and no build step is required (in contrast to the Closure Templates). A fully dynamic framework (all translations are available at runtime without reload), which also makes it a bit more demanding performance-wise.
+
+To use it:
+* in some structure (e.g. a JSON file) define a key-value object, containing all the translated messages.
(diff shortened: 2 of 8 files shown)


bryk:
This is awesome work @taimir! Thank you for this. Give me and others a time to digest this and think through.

bryk:
1. Do you know any way to extract messages for translation using angular-translate solution? 

2. Do you have any numbers on performance impact? E.g., can you assess how many new bindings does the solution add?

3. The problem of deciding what language to use is still unsolved as in the closure solution, right? Still it has to be server to decide what language to use.

bryk:
@cheld @floreks @maciaszczykm @zreigz What's your look on this?

floreks:
I actually like solution with `goog.getMsg` better(#252). It allows us to keep translations close to controllers. It doesn't involve additional code dependencies. Generating different translated versions of js is also good, because it doesn't have to be done on client side. Server can just serve different js on demand.

cheld:
About automatic text extraction: 

Not sure if such a functionality is needed. We never used such tools because context is lost when extracting text. Typically translations are stored as key -value pairs. The key is not just an ID. Most projects add a prefix and a suffix to keys, like deployPage.xxxxx.xxxx.title. The prefix is a hint for the place and the suffix is a hint for the sentence style (e.g. gerund)

Atanas noted that the closure message extractor will pick up annotations and add it as meta data to the the xliff output. So closure is potentially better than the tools I saw before.

PR #252, opened 2016-01-14
Considerations for i18n in dashboard - Closure Templates

taimir:
### DO NOT MERGE - ONLY MEANT FOR DISCUSSION

This would be the counterpart of angular-translate for i18n, as a follow up to https://github.com/kubernetes/dashboard/pull/251. The pull request contains a mock integration of the Closure Templates into our project. However it has many issues, I still couldn't get the compiled templates to pass through the Closure Compiler for example. 

Please refer to https://github.com/taimir/dashboard/blob/i18n-angular-translate/docs/design/i18n.md for the overview of my thoughts on both tools and check my notes in the first pull request https://github.com/kubernetes/dashboard/pull/251.

Notice that i18n is not complete here, as I need a way for the server to determine which .js file (for which locale) should be served.

As stated previously, we could go for any of those two options or I could even try some new one if anyone has an idea.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/252)


bryk:
Did you think about using `goog.getMsg` (which is a compiler primitive) instead of Soy templates? I agree that Soy templates are cumbersome and hard to use for our case, because we use Angular and Angular templates. However, `goog.getMsg` does not look that bad (assuming that we solve other problems of the compile-time translation). The benefit is that the messages are close to the place they are used (i.e., in the controller of the template).

Such solution means putting all messages in template controllers, like:
```js
class DeployController {
  constructor() {
    /**
     * @desc Explains the purpose of the app name field in the deploy form
     * @export
     */
    this.MSG_APP_HELP = goog.getMsg("An 'app' label with this value will be added to the Replica Set and Service that get deployed.");
  }
}
```

And then use it in HTML template:
```html
<kd-user-help>
  {{ctrl.MSG_APP_HELP}}
</kd-user-help>
```


The messages are replaced during compilation. XTB translation files can be generated easily: https://github.com/kuzmisin/xtbgenerator

taimir:
Interesting, I wasn't aware of the XTB Generator. Let me look a bit closer into this.

taimir:
O.K., correct me if I'm wrong: from what I've gathered as information, using `goog.getMsg()` would have the following effects:
* the soyutils.js and Soy in general is not needed any more
* the instances where goog.getMsg() is called will be parsed by the Closure Compiler at compile time. Given a translation bundle file (passed via the flag ` --translations_file`), the compiler is going to again generate multiple copies of the code, replacing the messages with the translations for each locale from the bundle file. Only XTB bundles are supported.
* we can auto-extract all messages into a XTB bundle via https://github.com/kuzmisin/xtbgenerator. Further there's a chance we could convert to XLIFF from there (back and forth) if XLIFF is required by any chance (I read somewhere that XTB is not a standard format, that's why I'm mentioning this).
* the problem how the different JS versions are served **remains** the same and will have to be solved somehow
* Where goog.getMsg() is called, one can add @descr annotations to provide context to the translators. If the XTB generator can parse those, this is actually good. Otherwise it would be a bit of a setback.

AFAICS, if the above is all correct, particularly the last point and the convertion XLIFF - XTB - XLIFF, there are no major differences with respect to i18n with Soy and we could use the goog.getMsg()? I need to test proof all of this.

Thoughts? Remarks?

bryk:
No additional thoughts from me.

taimir:
I checked the XtbGenerator yesterday - there is a small problem. It currently has no option to specify the input JS language version (in our case ES6) and therefore does not extract the messages. Since it is a wrapper around Google's `com.google.javascript.jscomp.JsMessage`, it is possible to specify the input language. I will have to change the Java code of the XtbGenerator for this. @bryk are you interested in me trying this out?

bryk:
For tests you can transpile the code to ES3 with `gulp scripts` and then try to apply XtbGenerator.

bryk:
Or traspile with Closure Compiler and optimization_options: WHITESPACE_ONLY

taimir:
True, didn't think of that. Will try it out, thanks.

floreks:
@taimir do we need this and #251 open? I think #528 is our solution for i18n.

bryk:
Yeah, close?

taimir:
Yes, you can close the other two pull requests.

PR #253, opened 2016-01-14
Make CPU and memory requirement inputs span across enite row.

bryk:
Previously they were very small.


bryk:
@batikanu Can you review?

batikanu:
LGTM

PR #263, opened 2016-01-18
Add more details and redesign replica set details view

bryk:
* Add more details: name & namespace, service name
* Refresh the view on replicas change
* Overall design refresh of the left menu

selection_012

### DIFF
--- a/src/app/backend/replicasetcommon.go
+++ b/src/app/backend/replicasetcommon.go
@@ -27,6 +27,24 @@ type ReplicaSetWithPods struct {
 	Pods       *api.PodList
 }
 
+// ReplicaSetPodInfo represents aggregate information about replica set pods.
+type ReplicaSetPodInfo struct {
+	// Number of pods that are created.
+	Current int `json:"current"`
+
+	// Number of pods that are desired in this Replica Set.
+	Desired int `json:"desired"`
+
+	// Number of pods that are currently running.
+	Running int `json:"running"`
+
+	// Number of pods that are currently waiting.
+	Pending int `json:"pending"`
+
+	// Number of pods that are failed.
+	Failed int `json:"failed"`
+}
+
 // Returns structure containing ReplicaSet and Pods for the given replica set.
 func getRawReplicaSetWithPods(client *client.Client, namespace, name string) (
 	*ReplicaSetWithPods, error) {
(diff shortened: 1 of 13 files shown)


bryk:
@maciaszczykm Can you review?

maciaszczykm:
@bryk Of course, sorry for delay.

maciaszczykm:
As you can see on image below, there is an issue with styling of message cells in events table. Perhaps you could update it in this pull request (it's probably not connected to it, but it would be nice thing to do)? Otherwise, I can create pull request for it.

image

`Replica set` and `Services` labels should have bigger top and bottom margins in my opinion. 

There is an information about desired number of pods, but there is no information about their current number. This information is only present in edit pods number dialog.

I didn't notice any issues in the code. PTAL

bryk:
> Replica set and Services labels should have bigger top and bottom margins in my opinion.

Yeah, sounds good. How about adding $baseline-grid (8xp) to the top?

bryk:
> There is an information about desired number of pods, but there is no information about their current number. This information is only present in edit pods number dialog.

How about adding a new section called, like, `Pod statuses` which will have info `3 running, 4 waiting`.

bryk:
> As you can see on image below, there is an issue with styling of message cells in events table. Perhaps you could update it in this pull request (it's probably not connected to it, but it would be nice thing to do)? Otherwise, I can create pull request for it.

Can you create a separate PR with this? Sounds like a small task.

maciaszczykm:
> Can you create a separate PR with this? Sounds like a small task.

Yes, of course.

maciaszczykm:
> Yeah, sounds good. How about adding $baseline-grid (8xp) to the top?

Sounds okay, should be fine.

maciaszczykm:
> How about adding a new section called, like, Pod statuses which will have info 3 running, 4 waiting?

Yeah, something like pods status (not sure which name would be better) would be more intuitive I think.

bryk:
New version. PTAL. 

selection_013

maciaszczykm:
LGTM :)

floreks:
New version look really nice. Pod statuses is really useful. It's good that you've added reload on dialog close. I've found 2 small style issues.

screenshot from 2016-01-19 13-38-48

The pencil should be a bit lower. Centered with the text. Now it's higher over the bottom of the text.

When status of pods is waiting only then there is just comma hanging on the end.

bryk:
PTAL. I've fixed the comma issue and aligned the pencil with bottom text.

bryk:
Also, renamed `waiting` to `pending` to use official API wording.

floreks:
Perfect. LGTM :)

bryk:
Thanks for the review!

PR #266, opened 2016-01-19
Fix go build process to actaully use customizes env variables

bryk:


bryk:
@luxas Can you look at this?

luxas:
Could you remove this also? 
```
// Scratch docker image is linux.
GOOS: 'linux',
```

It isn't required as `go` automatically will use `GOHOSTOS` and `GOHOSTARCH` (official builds will be on `linux/amd64`)
I will push some cross-compilation code soon

bryk:
Makes sense, done. PTAL

build/backend.js
```diff
@@ -66,8 +66,6 @@ gulp.task('backend:prod', ['package-backend-source'], function(doneFn) {
                 doneFn, {
                   // Disable cgo package. Required to run on scratch docker image.
                   CGO_ENABLED: '0',
```
luxas: I hope it don't crashes with this extra `,` Does `js` allow that?

bryk: It it okay. JS and golang allow (and like) trailing commas.

luxas:
LGTM. The Travis build also passed

PR #267, opened 2016-01-19
Remove leading slash from all XHR calls

bryk:
This is to be able to run when the application is rooted not at the
domain name. E.g.,
http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/kubernetes-dashboard/#/replicasets


bryk:
@zreigz Can you review?

zreigz:
LGTM

PR #268, opened 2016-01-19
Word-break for events and pods tables and logs view

maciaszczykm:
Added small fix for issue mentioned in pull request #263. Now message cells size shouldn't cause anymore problems. However, column widths are still an issue. The only solution which comes to my mind at the moment is making their values fixed (in percents). Do you have any suggestions?


bryk:
Check lint errors

bryk:
PTAL

maciaszczykm:
Fixed lint errors. PTAL

bryk:
LGTM

PR #270, opened 2016-01-20
Disabled validation fix

maciaszczykm:
It fixes issue caused by code from pull request #239. There was some strange behavior during logging of request body for validation. It blocked validation on all pages, so it's critical fix. At the moment I've removed logging of request body as it was done with response content. When I'll have good solution for it I'll create pull request. Sorry, for not testing it before merge of #239, but didn't expected such problems in this area.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	Colon             = ":"
-	RequestLogString  = "Incoming %s %s %s request from %s:\n%v"
+	RequestLogString  = "Incoming %s %s %s request from %s"
 	ResponseLogString = "Outcoming response to %s with %d status code"
 )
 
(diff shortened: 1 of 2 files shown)


maciaszczykm:
@bryk PTAL, as it's very important.

bryk:
LGTM

bryk:
Do you know why is this happening? Can you write a short explanation?

maciaszczykm:
> Do you know why is this happening? Can you write a short explanation?

It seems, that there is some problem with request body content for validation, but unfortunately I'm not sure what is wrong. 

```
reqBody := make([]byte, req.Request.ContentLength)
if req.Request.Body != nil {
	req.Request.Body.Read(reqBody)
}
```

This code causes the issue, exacly it's `req.Request.Body.Read(reqBody)` method call. I guess, there is a problem with length of request body or with unexpected/missing EOF. Is there any preparation needed in Go to perform I/O operations?

bryk:
I don't have any top of my head answer for your question.

PR #273, opened 2016-01-20
Deployment configuration build rules

bryk:
* now it is possible to build and push images to gcr.io with gulp
* there's yaml file to run the dashboard in your cluster with kubectl create -f
* the image is already pushed to gcr, so you can test it out


bryk:
@floreks @cheld Can you take a look?

floreks:
I've played a bit with that. Deployed dashboard using dashboard. Everything works. This is really nice feature. LGTM.

cheld:
Very nice. One comment:
I noticed that the dashboard docker image is not working in the hypercube, because the volume with certificates is not mounted (I think it is called service account functionality). So, basically it is not working for the same reason as heapster is not working in hypercube. However, I assume for the moment we do not plan to run integration tests in this setup.

BTW: It does work if the shell script is used to create a local cluster (hack/local-up-cluster.sh)

bryk:
Thanks, that's good to know. We may revert to previous cluster setup if hyperkube keeps causing problems.

PR #277, opened 2016-01-21
Allow for specifying multiple port mappings

bryk:
This also does:
* unify styles of port mappings and label inputs
* make label input not auto-delete rows that are empty - this is because
  I find such behavoir too aggresive (e.g., when you erase the text by
  hand you go to previous page when the row is being deleted)


bryk:
@floreks Can you look at this?

floreks:
I've looked at it shortly and tried testing. There is a problem now with deploying application when multiple port mappings are provided.

Server responses with:
`Service "kube-dashboard2" is invalid: [spec.ports[0].name: Required value, spec.ports[1].name: Required value]`

I've taken a look at that and this should explain it:
```Go
type ServicePort struct {
	// Optional if only one ServicePort is defined on this service: The
	// name of this port within the service.  This must be a DNS_LABEL.
	// All ports within a ServiceSpec must have unique names.  This maps to
	// the 'Name' field in EndpointPort objects.
	Name string `json:"name"`
```
This is not filled in our backend and with multiple ports this has to be unique. I'm not sure whether we should generate that or add option for use to name this mapping. We could also create string from information about port mapping and protocol.

bryk:
Thanks for catching this. How about autogenerating names for ports? In the format, e.g, port + "-" targetPort + "-" + protocol.

floreks:
That may be a good idea, but I'm not sure how kubernetes reacts when we provide multiple same mapping but with different names. This could work if there will be some validation on frontend like on labels to forbid creation of same mapping.

bryk:
I'll test this and if apiserver accepts identical port mappings, but with different autogenerated names I'll not validate this in the UI.

bryk:
Tested. Name generation works. Multiple identical ports are handled OKay by the master.

PTAL

src/app/backend/deploy.go
```diff
@@ -203,6 +207,13 @@ func GetAvailableProtocols() *Protocols {
 	return &Protocols{Protocols: []api.Protocol{api.ProtocolTCP, api.ProtocolUDP}}
 }
 
+func geteratePortMappingName(portMapping PortMapping) string {
```
floreks: Typo here :)

bryk: Ooops, fixed.

floreks:
Rest LGTM. Tested and works like a charm.

bryk:
Awesome, thank you. Will merge soon.

Issue #274, opened 2016-01-21
Edit pod count on replica set card menu


floreks:
![screenshot from 2016-01-21 10-19-38](https://cloud.githubusercontent.com/assets/2285385/12476378/34bc2690-c029-11e5-829e-342f1cbc85f0.png)

I'm working on this.

floreks:
Merged. Closing.

PR #280, opened 2016-01-22
Added edit replica set option to replica set card menu

floreks:
Related to #274.

I have also refactored replica set service and extended its responsibility. 

screenshot from 2016-01-22 09-52-06

@zreigz @bryk could you review?

### DIFF
--- a/src/app/frontend/replicasetdetail/replicaset_service.js
+++ b/src/app/frontend/replicasetdetail/replicaset_service.js
@@ -14,13 +14,14 @@
 
 import {StateParams} from './replicasetdetail_state';
 import {getReplicaSetDetailsResource} from './replicasetdetail_stateconfig';
+import showUpdateReplicasDialog from './updatereplicas_dialog';
 
 /**
  * Opens replica set delete dialog.
  *
  * @final
  */
-export class DeleteReplicaSetService {
+export class ReplicaSetService {
   /**
    * @param {!md.$dialog} $mdDialog
    * @param {!angular.$resource} $resource
@@ -39,7 +40,7 @@ export class DeleteReplicaSetService {
   }
 
   /**
-   * Opens a replica set delete dialog. Returns a promise that is resolved/rejeced when user wants
+   * Opens a replica set delete dialog. Returns a promise that is resolved/rejected when user wants
    * to delete the replica. Nothing happens when user clicks cancel on the dialog.
    *
    * @param {string} namespace
@@ -63,5 +64,20 @@ export class DeleteReplicaSetService {
         .then(() => { resource.remove(deferred.resolve, deferred.reject); });
 
     return deferred.promise;
+  }
+
+  /**
+   * Opens an update replica set dialog. Returns a promise that is resolved/rejected when user wants
+   * to update the replicas. Nothing happens when user clicks cancel on the dialog.
+   *
+   * @param {string} namespace
+   * @param {string} replicaSet
+   * @param {number} currentPods
+   * @param {number} desiredPods
+   * @returns {!angular.$q.Promise}
+   */
+  showUpdateReplicasDialog(namespace, replicaSet, currentPods, desiredPods) {
+    return showUpdateReplicasDialog(
+        this.mdDialog_, namespace, replicaSet, currentPods, desiredPods);
   }
 }
--- a/src/app/frontend/replicasetdetail/updatereplicas_controller.js
+++ b/src/app/frontend/replicasetdetail/updatereplicas_controller.js
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {StateParams} from './replicasetdetail_state';
+import {getReplicaSetSpecPodsResource} from './replicasetdetail_stateconfig';
+
 /**
  * Controller for the update replica set dialog.
  *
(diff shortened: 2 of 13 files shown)


src/app/frontend/replicasetdetail/replicaset_service.js
```diff
@@ -64,4 +66,41 @@ export class DeleteReplicaSetService {
...
   }
+
+  /**
+   * Opens an update replica set dialog. Returns a promise that is resolved/rejected when user wants
+   * to update the replicas. Nothing happens when user clicks cancel on the dialog.
+   *
+   * @param {!backendApi.ReplicaSetDetail} replicaSetDetail
+   * @returns {!angular.$q.Promise}
+   */
+  showUpdateReplicasDialog(replicaSetDetail) {
```
bryk: These two methods (on this service and on the dialog) take too broad arguments. They take entire replica set detail, whereas they only need name, current, and desired numbers of pods. This volatiles law of demeter that says that you should only have direct dependencies that you use in params list. Having broader arguments makes the API harder to use. 
If you change this to have only directly used parameters, the usage in src/app/frontend/replicasetlist/replicasetcardmenu_controller.js will be easy. The test will be cleaner as well. 

I think this wasn't noticed before, because this was called from details page which had details object and everything looked good.

src/app/frontend/replicasetdetail/replicaset_service.js
```diff
@@ -64,4 +66,41 @@ export class DeleteReplicaSetService {
...
+        new StateParams(replicaSetDetail.namespace, replicaSetDetail.name), this.resource_);
+    let deferred = this.q_.defer();
+
+    /** @type {!backendApi.ReplicaSetSpec} */
+    let replicaSetSpec = {
+      replicas: replicasCount,
+    };
+
+    resource.save(replicaSetSpec, opt_callback, opt_errback);
+    // TODO(floreks): Think about refreshing data after update.
```
bryk: Hmm... Should this todo be still here? 

Btw, you can call this.state_.refresh() on the details page to reflect changes on success.

src/app/frontend/replicasetdetail/replicaset_service.js
```diff
@@ -64,4 +66,41 @@ export class DeleteReplicaSetService {
...
+  /**
+   * Updates number of replicas in replica set.
+   *
+   * @param {number} replicasCount
+   * @param {!backendApi.ReplicaSetDetail} replicaSetDetail
+   * @param {!function(!backendApi.ReplicaSetSpec)=} opt_callback
+   * @param {!function(!angular.$http.Response)=} opt_errback
+   * @return {angular.$q.Promise}
+   */
+  updateReplicas(replicasCount, replicaSetDetail, opt_callback, opt_errback) {
```
bryk: Hmm... I think that you can move this method to UpdateReplicasDialog. With this we would get rid of the lengthy callback. You can just make showUpdateReplicasDialog return the promise of this function. The same as delete dialog. 
What do you think?

src/app/frontend/replicasetlist/replicasetcardmenu_controller.js
```diff
@@ -63,8 +62,26 @@ export default class ReplicaSetCardMenuController {
...
+    this.kdReplicaSetService_.showDeleteDialog(this.replicaSet.namespace, this.replicaSet.name)
         .then(() => this.state_.reload(), () => this.log_.error('Error deleting replica set'));
   }
+
+  /**
+   * @export
+   */
+  showUpdateReplicasDialog() {
+    /** @type {!backendApi.ReplicaSetDetail} - Prepare replica set detail object */
+    let replicaSetDetail = {
```
bryk: See my other comment on how to improve this call.

bryk:
Looks really nice. A few comments to make the code better.

bryk:
PTAL

src/app/frontend/replicasetdetail/replicaset_service.js
floreks: Maybe not. I've left this because I still thought that on some point we will make 3(4)-way data binding in order to reflect changes in backend. After decreasing number of pods page is refreshed and we can see that desired number of pods is 1, but still in logs there are more pods and current number of them is higher.

src/app/frontend/replicasetdetail/replicaset_service.js
floreks: Totally agree. I will make it simpler. On details page it had more sense to leave it like that.

src/app/frontend/replicasetdetail/replicaset_service.js
bryk: All right, sounds good :) State.reload() is a cheap method to show updated info until we have proper way of doing this.

src/app/frontend/replicasetdetail/updatereplicas_controller.js
```diff
@@ -22,40 +25,67 @@ export default class UpdateReplicasDialogController {
...
+    let deferred = this.q_.defer();
+
+    /** @type {!backendApi.ReplicaSetSpec} */
+    let replicaSetSpec = {
+      replicas: this.replicas,
+    };
+
+    resource.save(
+        replicaSetSpec, this.onUpdateReplicasSuccess_.bind(this),
         this.onUpdateReplicasError_.bind(this));
```
floreks: @bryk I'd really want to test callback functions but I can't find a way to reach them through resource. Do you know a way to do that?

floreks:
PTAL

src/app/frontend/replicasetdetail/updatereplicas_controller.js
```diff
@@ -22,40 +25,67 @@ export default class UpdateReplicasDialogController {
...
+    return deferred.promise;
```
bryk: You know that this promise never resolves, right?

src/app/frontend/replicasetdetail/updatereplicas_controller.js
bryk: Look for angular documentation. AFAIK it is `resource.save(...).$promise`.

bryk: Or better, see how showDeleteDialog does this. I.e., on successful change the promise is resolved (this would happen here in this.onUpdateReplicasSuccess_)

bryk:
Last thing about the promise and lets merge.

src/app/frontend/replicasetdetail/updatereplicas_controller.js
floreks: Ooops. It's not called by callback function. Yeah. Should I change that and use defer or just remove it?

bryk: If there are no clients of the promise, then sure, remove it. Somebody will add when needed.

If there are clients, then resolve the promise somewhere ;)

src/app/frontend/replicasetdetail/updatereplicas_controller.js
floreks: Both helped. Thanks!

src/app/frontend/replicasetdetail/updatereplicas_controller.js
floreks: Good to know. I will remember that :)

bryk:
Gulp check fails. Please fix.

bryk:
LGTM. Fix compilation errors and let's merge.

floreks:
It's fixed.

src/test/frontend/replicasetdetail/replicasetdetail_controller_test.js
```diff
@@ -89,15 +91,37 @@ describe('Replica Set Detail controller', () => {
...
 
     // when
     ctrl.handleUpdateReplicasDialog();
 
     // then
-    expect(mdDialog.show).toHaveBeenCalled();
+    expect(kdReplicaSetService.showUpdateReplicasDialog).toHaveBeenCalled();
+  });
+
+  it('should create logs href', () => {
```
bryk: Hey, this test is copied from line 127 :) Please revert.

bryk:
Found one more thing :) Please fix.

src/test/frontend/replicasetdetail/replicasetdetail_controller_test.js
floreks: I have looked at this code like 20 times and didn't catch that. huh... good eye

floreks:
I hope that's everything :)

bryk:
Done! :)

PR #281, opened 2016-01-22
Godeps for kubectl

maciaszczykm:
Added Go dependencies (and entries in Godeps.json) to enable usage of `kubectl create -f <path-to-yaml-file>` command equivalent in the backend. I'm still not sure about how Godeps should be managed, so if you have any suggestions please let me know.

I'll be used to deploy from files:

```go
create "k8s.io/kubernetes/pkg/kubectl/cmd"
cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"

...

cmd := create.NewCmdCreate(cmdutil.NewFactory(nil), os.Stdout)
cmd.Flags().Set("filename", "<path-to-yaml-file>")
cmd.Flags().Set("output", "name")
cmd.Run(cmd, []string{})
```

I've already checked it, and above code works if there is no issue with Godeps. We only need to handle file upload and then use it.

@batikanu @bryk PTAL and let me know what you think about it, because it's pretty complicated change. I'm still not sure if all dependencies are correct, because of their number.

### DIFF
--- a/Godeps/_workspace/src/k8s.io/heapster/api/v1/types/model_types.go
+++ b/Godeps/_workspace/src/k8s.io/heapster/api/v1/types/model_types.go
@@ -1,0 +1,59 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


maciaszczykm:
Please don't review it, because few files have been ignored during commit.

bryk:
Ok, let me know when to review. I'll do this ASAP to avoid merge conflicts, as I know Godeps are tricky. Thanks!

PR #285, opened 2016-01-22
Deploy file

batikanu:
First working draft of Upload file backend. Uploaded file is deployed by using kubectl code in backend.

### DIFF
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -16,16 +16,14 @@ package main
 
 import (
 	"fmt"
-	ioutil "io/ioutil"
 	"log"
-	"os"
 	"strings"
 
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/resource"
-	create "k8s.io/kubernetes/pkg/kubectl/cmd"
-	cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
+	cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
+	kubectlResource "k8s.io/kubernetes/pkg/kubectl/resource"
 	"k8s.io/kubernetes/pkg/util/intstr"
 )
 
@@ -246,26 +244,36 @@ func getLabelsMap(labels []Label) map[string]string {
 }
 
 // Deploys an app based on the given yaml or json file.
-func DeployAppFromFile(spec *AppDeploymentFromFileSpec, client *client.Client) error {
-	// Creates temp file to use it in kubectl command
-	file, err := ioutil.TempFile("", spec.Name)
-	if err != nil {
-		return err
-	}
-	_,err = file.WriteString(spec.Content)
-	if err != nil {
-		return err
-	}
-	// Runs kubectl create command with the temp file
-	options := &create.CreateOptions{
-		Filenames: []string {file.Name()},
-	}
-	cmd := create.NewCmdCreate(cmdutil.NewFactory(nil), os.Stdout)
-	cmd.Flags().Set("filename", file.Name())
-	cmd.Flags().Set("output", "name")
-	err = create.RunCreate(cmdutil.NewFactory(nil), cmd, os.Stdout, options)
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec) error {
+	const (
+		validate      = true
+		emptyCacheDir = ""
+	)
 
-	// Removes the created temp file after using it
-	os.Remove(file.Name())
-	return err
+	factory := cmdutil.NewFactory(nil)
+	schema, err := factory.Validator(validate, emptyCacheDir)
+	if err != nil {
+		return err
+	}
+
+	mapper, typer := factory.Object()
+	reader := strings.NewReader(spec.Content)
+
+	r := kubectlResource.NewBuilder(mapper, typer, factory.ClientMapperForCommand()).
+		Schema(schema).
+		NamespaceParam(api.NamespaceDefault).DefaultNamespace().
+		Stream(reader, spec.Name).
+		Flatten().
+		Do()
+
+	return r.Visit(func(info *kubectlResource.Info, err error) error {
+		// creates an object from input info
+		_, err = kubectlResource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
+		if err != nil {
+			return err
+		}
+		log.Printf("%s is deployed", info.Name)
+		return nil
+	})
 }
+
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -16,11 +16,15 @@ package main
 
 import (
 	"fmt"
+	ioutil "io/ioutil"
 	"log"
+	"os"
 	"strings"
 
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/resource"
+	create "k8s.io/kubernetes/pkg/kubectl/cmd"
+	cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/util/intstr"
 )
@@ -74,6 +78,15 @@ type AppDeploymentSpec struct {
 
 	// Whether to run the container as privileged user (essentially equivalent to root on the host).
 	RunAsPrivileged bool `json:"runAsPrivileged"`
+}
+
+// Specification for deployment from file
+type AppDeploymentFromFileSpec struct {
+	// Name of the file
+	Name string `json:"name"`
+
+	// File content
+	Content string `json:"content"`
 }
 
 // Port mapping for an application deployment.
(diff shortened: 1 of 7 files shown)


src/app/backend/deploy.go
```diff
@@ -224,3 +237,28 @@ func getLabelsMap(labels []Label) map[string]string {
...
 }
+
+// Deploys an app based on the given yaml or json file.
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec, client *client.Client) error {
+	// Creates temp file to use it in kubectl command
+	file, err := ioutil.TempFile("", spec.Name)
+	if err != nil {
+		return err
+	}
+	_,err = file.WriteString(spec.Content)
```
batikanu: spec.content can be byte[] instead of string. In case it is needed, frontend should be modified as well.

src/app/backend/deploy.go
```diff
@@ -224,3 +237,28 @@ func getLabelsMap(labels []Label) map[string]string {
 
 	return result
 }
+
+// Deploys an app based on the given yaml or json file.
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec, client *client.Client) error {
+	// Creates temp file to use it in kubectl command
+	file, err := ioutil.TempFile("", spec.Name)
+	if err != nil {
+		return err
```
batikanu: Right now all those error messages are sent to the user with error popup. We might improve it by hiding some error messages coming form server.

src/app/backend/deploy.go
```diff
@@ -224,3 +237,28 @@ func getLabelsMap(labels []Label) map[string]string {
 
 	return result
 }
+
+// Deploys an app based on the given yaml or json file.
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec, client *client.Client) error {
+	// Creates temp file to use it in kubectl command
+	file, err := ioutil.TempFile("", spec.Name)
```
bryk: I'm very worried about creating the temporary file. I dont even know if it is supported in our docker scratch image. 

How about we do copy->paste->remove_unneeded_stuff from the RunCreate method? https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/create.go#L90

I think that we can have a RunCreate method that reads from a stream in a dozen of lines, because most of the stuff there is not needed for us. 

The data would be read from a stream: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/resource/builder.go#L150
Instead from a file. Basically the part of RunCreate needs to change:
```go
	r := resource.NewBuilder(mapper, typer, f.ClientMapperForCommand()).
		Schema(schema).
		ContinueOnError().
		NamespaceParam(cmdNamespace).DefaultNamespace().
		FilenameParam(enforceNamespace, options.Filenames...).
		Flatten().
		Do()
```

What do you think?

batikanu:
PTAL

src/app/frontend/deploy/deployfromfile_controller.js
```diff
@@ -33,5 +40,55 @@ export default class DeployFromFileController {
...
+    /** @type {!backendApi.AppDeploymentFromFileSpec} */
+    let deploymentSpec = {
+      name: this.file.name,
+      content: this.file.content,
+    };
+
+    let defer = this.q_.defer();
+
+    /** @type {!angular.Resource<!backendApi.AppDeploymentFromFileSpec>} */
+    let resource = this.resource_('/api/appdeploymentfromfile');
```
bryk: Dont start api calls with leading `/`. With this the app would not work when non rooted at the domain.

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+
+// Deploys an app based on the given yaml or json file.
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec, client *client.Client) error {
+	cmd := create.NewCmdCreate(cmdutil.NewFactory(nil), os.Stdout)
+	return RunCreate(cmdutil.NewFactory(nil), cmd, os.Stdout, spec)
+}
+
+// Validates and runs the given command
+func RunCreate(f *cmdutil.Factory, cmd *cobra.Command, out io.Writer, spec *AppDeploymentFromFileSpec) error {
+	schema, err := f.Validator(cmdutil.GetFlagBool(cmd, "validate"), cmdutil.GetFlagString(cmd, "schema-cache-dir"))
```
bryk: Dont use `cmdutil.GetFlagBool(cmd, "validate"), cmdutil.GetFlagString(cmd, "schema-cache-dir")`. Create constants here in the code.

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+	if err != nil {
+		return err
+	}
+	cmdNamespace, _, err := f.DefaultNamespace()
```
bryk: Do you need this?

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+	if err != nil {
+		return err
+	}
+
+	mapper, typer := f.Object()
+	reader := strings.NewReader(spec.Content)
+
+	r := kubectlResource.NewBuilder(mapper, typer, f.ClientMapperForCommand()).
+		Schema(schema).
+		ContinueOnError().
```
bryk: Should we continue on error? I think not

batikanu: It means we will stop creating every other objects if any of them fails in the deployment. So all or nothing is the preferred behavior?

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+		if err != nil {
+			return err
+		}
+
+		if err := createAndRefresh(info); err != nil {
+			return cmdutil.AddSourceToErr("creating", info.Source, err)
+		}
+
+		count++
+		cmdutil.PrintSuccess(mapper, true, out, info.Mapping.Resource, info.Name, "created")
```
bryk: Use logging not cmdutil here. Basically, use as little from kubectl as possible :)

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	if count == 0 {
+		return fmt.Errorf("no objects passed to create")
```
bryk: Use logging, not fmt.

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+	return nil
+}
+
+// createAndRefresh creates an object from input info and refreshes info with that object
+func createAndRefresh(info *kubectlResource.Info) error {
+	obj, err := kubectlResource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
+	if err != nil {
+		return err
+	}
+	info.Refresh(obj, true)
```
bryk: Do we need refreshing? What is this?

src/app/backend/deploy.go
```diff
@@ -231,3 +246,65 @@ func getLabelsMap(labels []Label) map[string]string {
...
+		Do()
+
+	count := 0
+	err = r.Visit(func(info *kubectlResource.Info, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if err := createAndRefresh(info); err != nil {
+			return cmdutil.AddSourceToErr("creating", info.Source, err)
```
bryk: Do we need this?

bryk:
Nice! Can you write tests for remaining code? The error dialog and sending deploy spec to the backend.

bryk:
Can you rename this pull request? It is not only backend anymore :)

bryk:
PTAL

src/app/backend/deploy.go
bryk: Yes :)

src/app/backend/deploy.go
batikanu: schema-cache-dir is created in NewCmdCreate method by the constants defined in loader.go. If I create copy of these constants we can't get any update of these constants in case somebody changes them. For validate I can use a constant which is set to true since it is just a boolean value.

src/app/backend/deploy.go
batikanu: Namespace has to be set in the command. Is there any better way to retrieve the DefaultNamespace?

src/app/backend/deploy.go
bryk: Why do we need schema-cache-dir here? The docs say that we can leave it empty. If with this our code still works, then leave it empty :)

src/app/backend/deploy.go
bryk: `api.NamespaceDefault`

See other files.

bryk:
Replied to comments, PTAL

src/app/frontend/deploy/deployfromfile_controller.js
batikanu: Done

src/app/backend/deploy.go
batikanu: Done

src/app/backend/deploy.go
batikanu: Thank you

src/app/backend/deploy.go
batikanu: Done

src/app/backend/deploy.go
batikanu: Done

src/app/backend/deploy.go
batikanu: Done

src/app/backend/deploy.go
batikanu: It is not needed. I've removed it.

src/app/backend/deploy.go
```diff
@@ -231,3 +242,38 @@ func getLabelsMap(labels []Label) map[string]string {
 
 	return result
 }
+
+// Deploys an app based on the given yaml or json file.
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec) error {
```
bryk: This looks really nice now!

bryk:
Is this ready for rereview?

bryk:
Btw, update PR's description. It is invalid at the moment

bryk:
Btw, tested on gulp serve:prod. It works! This is awesome!

batikanu:
A test is missing for sending deploy spec to the backend. Deploy method is actually tested. Do you want a test which verifies whether the backend api is called after deploy method is called?

batikanu:
The rest can be rereviewed already.

bryk:
LGTM, merging :)

PR #286, opened 2016-01-22
Remove menu from pods table

maciaszczykm:
Removed menu from pods table, as there were no menu items inside it. It was supposed to be used to access logs of previous instances (for example after pod restart), but it's not possible at the moment I think. Left comment in the code to take care of it later.

Fixed issue with not closed `<span>` element and modified logs tooltip, because it was sometimes outside page and horizontal scroll-bar was shown.


bryk:
LGTM :)

Issue #288, opened 2016-01-25
Event filtering by source


maciaszczykm:
On replica set detail view mock (image below) there is an event source filter, however, in Kubernetes API there is no more information about event than its source component and host. Filtering is based on 3 options: all, system and user. How we can tell if event is coming from user or system?

![Mock](https://raw.githubusercontent.com/kubernetes/dashboard/master/docs/design/mockups/11-11-2015/19_events.png)

I think at the moment we should hide this filter and add it again when it will be clear how filtering should work.

bryk:
Yeah, LGTM. The idea was that we cleverly distinguish system and user events, but this hasn't been formalized. I'd go for removing this for now and doing other, well-defined, work.

PR #289, opened 2016-01-25
Hide event source filter

maciaszczykm:
Connected to #288. Hidden event source filter, because backend logic is not implemented yet.


bryk:
LGTM

PR #295, opened 2016-01-26
UX changes and styling of replica set detail page

floreks:
Style update:

* Changed sub-header height to 66px (same as toolbar)
* Moved delete button up and changed text to `Delete Replica Set`
* Adjusted vertical spacing between sidebar panel elements
* Reverted accent pallete to blue (button color on replica set list page will be adjusted by @maciaszczykm)
* Removed animation when switching tabs
* Added `No events found` information when there are no events. (Please check if it looks ok)
* Removed some tooltips messages from events table that are not very meaningful.

2 things still to do, but in other PRs:
* Deleting service when deleting replica set. (Needs discussion)
* Combining information about pod statuses.

### DIFF
--- a/src/app/frontend/replicasetdetail/replicasetdetail.html
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.html
@@ -34,9 +34,12 @@ limitations under the License.
       <div layout="row" flex="nogrow">
         <md-button class="md-primary" ng-click="ctrl.handleDeleteReplicaSetDialog()">
           <md-icon class="material-icons">delete</md-icon>
-          DELETE
+          DELETE REPLICA SET
         </md-button>
       </div>
+      <span class="kd-replicasetdetail-sidebar-title kd-replicasetdetail-sidebar-info">
+        Replica set
+      </span>
       <div flex layout="column" class="kd-replicasetdetail-sidebar-info">
         <span class="kd-replicasetdetail-sidebar-line">Name</span>
         <span class="kd-replicasetdetail-sidebar-subline">
@@ -251,7 +254,7 @@ limitations under the License.
       </md-content>
     </md-tab>
     <md-tab label="Events">
-      <md-content flex>
+      <md-content flex ng-if="ctrl.hasEvents()">
         <div class="kd-replicasetdetail-options" layout="row">
           <md-input-container class="kd-replicasetdetail-option-picker">
             <label>Type</label>
@@ -270,7 +273,7 @@ limitations under the License.
             <th class="kd-replicasetdetail-table-header">
               <kd-sorted-header currently-selected-column="ctrl.sortEventsBy"
                                 currently-selected-order="ctrl.eventsOrder"
-                                column-name="message" tooltip="Event message">
+                                column-name="message">
                 Message
               </kd-sorted-header>
             </th>
@@ -285,7 +288,7 @@ limitations under the License.
             <th class="kd-replicasetdetail-table-header">
               <kd-sorted-header currently-selected-column="ctrl.sortEventsBy"
                                 currently-selected-order="ctrl.eventsOrder"
-                                column-name="object" tooltip="Sub-object of the event">
+                                column-name="object">
                 Sub-object
               </kd-sorted-header>
             </th>
@@ -336,6 +339,15 @@ limitations under the License.
           </tr>
           </tbody>
         </table>
+      </md-content>
+      <md-content class="kd-replicasetdetail-no-events" ng-if="!ctrl.hasEvents()">
+        <md-icon class="material-icons kd-replicasetdetail-no-events-icon">info_outline</md-icon>
+        <div class="kd-replicasetdetail-no-events-text">No events were found</div>
+        <span class="kd-replicasetdetail-no-events-info">
+          There are no events to display. It's possible that all of them have expired.
+          <a href="">Learn more</a>
+        </span>
+      </md-content>
     </md-tab>
   </md-tabs>
 </md-content>
--- a/src/app/frontend/replicasetdetail/replicasetdetail.scss
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.scss
@@ -53,18 +53,13 @@ $table-cell-height-half: $table-cell-height / 2;
   font-weight: $bold-font-weight;
   padding-bottom: $baseline-grid;
   padding-top: $baseline-grid;
-
-  &.kd-replicasetdetail-sidebar-info {
-    padding-bottom: 0;
-    padding-top: 2 * $baseline-grid;
-  }
 }
 
 .kd-replicasetdetail-sidebar-line {
   color: $foreground-2;
   display: block;
   font-size: $body-font-size-base;
-  padding-bottom: $baseline-grid;
+  padding-bottom: 0;
 }
 
 .kd-replicasetdetail-sidebar-subline {
@@ -133,4 +128,28 @@ $table-cell-height-half: $table-cell-height / 2;
 .kd-replicasetdetail-help-icon {
   color: $foreground-2;
   cursor: default;
+}
+
+.kd-replicasetdetail-no-events {
+  margin-top: 4 * $baseline-grid;
+  text-align: center;
+}
+
+md-icon {
+  &.kd-replicasetdetail-no-events-icon {
+    color: $muted;
+    font-size: $display-1-font-size-base;
+    margin: 0 (2 * $baseline-grid) (2 * $baseline-grid) 0;
+  }
+}
+
+.kd-replicasetdetail-no-events-text {
+  color: $muted;
+  font-size: $headline-font-size-base;
+  margin-bottom: $baseline-grid;
+}
+
+.kd-replicasetdetail-no-events-info {
+  color: $delicate;
+  font-size: $subhead-font-size-base;
 }
--- a/src/app/frontend/_variables.scss
+++ b/src/app/frontend/_variables.scss
@@ -20,6 +20,7 @@
 }
 
 $display-4-font-size-base: rem(11.2) !default;
+$display-1-font-size-base: rem(3.4) !default;
 $headline-font-size-base:  rem(2.4) !default;
 $subhead-font-size-base:   rem(1.6) !default;
 $body-font-size-base:      rem(1.4) !default;
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -81,3 +81,8 @@ a {
     }
   }
 }
+
+// Disables animation on md-tab switch. Avoids some flickering and is simpler.
+[role='tabpanel'] {
+  transition: none;
+}
--- a/src/app/frontend/index_config.js
+++ b/src/app/frontend/index_config.js
@@ -19,11 +19,15 @@
 export default function config($mdThemingProvider) {
   // Create a color palette that uses Kubernetes colors.
   let kubernetesColorPaletteName = 'kubernetesColorPalette';
+  let kubernetesAccentPaletteName = 'kubernetesAccentPallete';
   let kubernetesColorPalette = $mdThemingProvider.extendPalette('blue', {
     '500': '326de6',
   });
 
   // Use the palette as default one.
   $mdThemingProvider.definePalette(kubernetesColorPaletteName, kubernetesColorPalette);
-  $mdThemingProvider.theme('default').primaryPalette(kubernetesColorPaletteName);
+  $mdThemingProvider.definePalette(kubernetesAccentPaletteName, kubernetesColorPalette);
+  $mdThemingProvider.theme('default')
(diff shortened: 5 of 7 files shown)


bryk:
@maciaszczykm Can you review?

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -63,7 +63,7 @@ <h1 flex class="md-title kd-replicasetdetail-app-name">
             </md-button>
           </div>
         </div>
-        <span class="kd-replicasetdetail-sidebar-line">Pods status</span>
+        <span class="kd-replicasetdetail-sidebar-line">Pods Status</span>
```
bryk: Hey, I think this is the opposite. The previous version is correct. I think Ken wanted to change labels on the cards list, which I already done. Ref: https://msdn.microsoft.com/en-us/library/windows/desktop/bb246428(v=vs.85).aspx

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -336,6 +336,11 @@ <h1 flex class="md-title kd-replicasetdetail-app-name">
           </tr>
           </tbody>
         </table>
+      </md-content>
+      <md-content class="kd-replicasetdetail-no-events" ng-if="!ctrl.events">
+        <md-icon class="material-icons kd-replicasetdetail-no-events-icon">info_outline</md-icon>
+        <div class="kd-replicasetdetail-no-events-text">No events were found</div>
```
bryk: Can you add explanation here that events expire after some time. And maybe a learn more link?

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: Nice! There is even a documentation for that.  :)

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: I can't find any information in documentation about event expiration. Could you provide me some link?

About message. Should it be some tooltip on icon or just plain text? Something similar to help section on deploy page?

maciaszczykm:
@bryk Yes, I'll check it.

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -251,7 +251,7 @@ <h1 flex class="md-title kd-replicasetdetail-app-name">
       </md-content>
     </md-tab>
     <md-tab label="Events">
-      <md-content flex>
+      <md-content flex ng-if="ctrl.events">
```
maciaszczykm: This `if` statement doesn't cover empty events list case.

maciaszczykm:
One issue with `ng-if="ctrl.events"``, everything else LGTM.

src/app/frontend/replicasetdetail/replicasetdetail.html
floreks: Done

floreks:
PTAL

maciaszczykm:
LGTM

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -102,6 +97,26 @@ $table-cell-height-half: $table-cell-height / 2;
...
 
+.kd-replicasetdetail-tab-header {
+  md-tabs-canvas,
+  md-pagination-wrapper {
+    height: $toolbar-height-size-base;
+  }
+
+  md-tab-item {
+    padding-top: $toolbar-height-size-base / 2;
+    text-transform: none;
```
bryk: Why this?

floreks: I was just going with what the mock shows. It's not capitalized there.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -102,6 +97,26 @@ $table-cell-height-half: $table-cell-height / 2;
...
+  }
+
+  md-tab-content {
+    &.md-left,
+    &.md-right {
+      animation: none;
```
bryk: Can you maybe disable the animations with `$animate.enabled(...)`? If this does not work, please comment why we have this here.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -134,3 +149,21 @@ $table-cell-height-half: $table-cell-height / 2;
...
+
+.kd-replicasetdetail-no-events {
+  margin-top: 4 * $baseline-grid;
+  text-align: center;
+}
+
+md-icon {
+  &.kd-replicasetdetail-no-events-icon {
+    color: $muted;
+    font-size: 1.5 * $headline-font-size-base;
```
bryk: Hmm... Why this? With this line removed it still looks okay.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -53,18 +53,13 @@ $table-cell-height-half: $table-cell-height / 2;
...
-    padding-top: 2 * $baseline-grid;
-  }
 }
 
 .kd-replicasetdetail-sidebar-line {
   color: $foreground-2;
   display: block;
   font-size: $body-font-size-base;
-  padding-bottom: $baseline-grid;
+  padding-bottom: $baseline-grid / 4;
```
bryk: You know me... I'm opposed to everything that is not a natural multiplier of $baseline-grid :) How about making this 0 and line 70 $baseline grid. This would be pretty much the same as on replica set cards.

bryk: And continuation: How about making gutter between elements in cards to $baseline grid (now it is 2 * $baseline-grid)? CC @maciaszczykm 
This would make all gutters the same.

src/app/frontend/replicasetdetail/replicasetdetail.scss
```diff
@@ -102,6 +97,26 @@ $table-cell-height-half: $table-cell-height / 2;
   width: 15 * $baseline-grid;
 }
 
+.kd-replicasetdetail-tab-header {
+  md-tabs-canvas,
+  md-pagination-wrapper {
+    height: $toolbar-height-size-base;
```
bryk: Hmm... Can you revert this 66px change from this PR. This does not look good to me, because the tabs are misaligned now.

Basically, I'm opposed to everything that is not standard Angular Material, unless it brings real benefits. The reason is that we have limited engineering resources and we cannot be implementing (and maintaining) our own UI components. 

What do you think?

@maciaszczykm

src/app/frontend/replicasetdetail/replicasetdetail.scss
bryk: I see. I think we should be going with standard angular material components, unless it gives benefits for users to change them. With this we'll save engineering resources to implement features, rather than play with frameworks. Does this make sense?

bryk:
A few discussion comments. PTAL.

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: I've changed it and it looks good, but I'm not that picky when it comes to UI :)

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: I also do think that it looked better the old way. If it's ok to everyone then I'll revert.

src/app/frontend/replicasetdetail/replicasetdetail.scss
maciaszczykm: I agree and I'll unify it in my pull request.

maciaszczykm: Done!

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: Sure, I'm thinking the same. Looking at Ken's feedback felt like he wants view to be closer to mocks. That's why I've changed it.

src/app/frontend/replicasetdetail/replicasetdetail.scss
bryk: Btw, more reading on baseline grid: https://www.google.com/design/spec/layout/metrics-keylines.html

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: I'm not sure. Without this icon seems really small compared to the text.

screenshot from 2016-01-26 15-03-14

vs

screenshot from 2016-01-26 15-03-25

What do you think?

src/app/frontend/replicasetdetail/replicasetdetail.scss
maciaszczykm: In my opinion it looks better when we don't change height here (use standard Angular Material), but color should definitely be changed here (is it pink?) :)

src/app/frontend/replicasetdetail/replicasetdetail.scss
floreks: I've looked a bit more and found a better solution. `$animate.enabled()` doesn't seem to be working.

https://github.com/angular/material/issues/811#issuecomment-65164436

This looks to be a good way to do it. Do you think we can disable it globally for tabs?

src/app/frontend/replicasetdetail/replicasetdetail.scss
bryk: Then maybe use $headline-font-size-base for the size?

floreks: Without specyfing font-size it will be the same as $headline-font-size-base (first picture).

On the first picture both text and icon are 24px.
On second icon is 36px and text 24px.

src/app/frontend/index.scss
```diff
@@ -18,3 +18,7 @@
 
 // Angular Material icon font. It allows to use the icons as fonts.
 @import url('//fonts.googleapis.com/icon?family=Material+Icons');
+
+[role="tabpanel"] {
+  transition: none;
+}
```
floreks: Is it ok to disable transition for tabs globally?

bryk: Yes, sounds good.

bryk: Add a comment, of course :)

src/app/frontend/index.scss
```diff
@@ -18,3 +18,7 @@
 
 // Angular Material icon font. It allows to use the icons as fonts.
 @import url('//fonts.googleapis.com/icon?family=Material+Icons');
+
+[role='tabpanel'] {
+  transition: none;
+}
```
floreks: Is it ok if we disable tab switch animation globally?

src/app/frontend/replicasetdetail/replicasetdetail.html
```diff
@@ -336,6 +336,15 @@ <h1 flex class="md-title kd-replicasetdetail-app-name">
           </tr>
           </tbody>
         </table>
+      </md-content>
+      <md-content class="kd-replicasetdetail-no-events" ng-if="!ctrl.hasEvents()">
+        <md-icon class="material-icons kd-replicasetdetail-no-events-icon">info_outline</md-icon>
+        <div class="kd-replicasetdetail-no-events-text">No events were found</div>
+        <span class="kd-replicasetdetail-no-events-info">
+          There are no events to display. It's possible that all of them have expired.
+          <a href="">Learn more</a>
```
floreks: Is this message ok?

floreks:
@bryk @maciaszczykm PTAL

src/app/frontend/index.scss
bryk: Yes, but add a comment explaining what and why you're doing here.

src/app/frontend/replicasetdetail/replicasetdetail.html
bryk: SG

src/app/frontend/replicasetdetail/replicasetdetail.scss
bryk: Take font size from _variables.scss or add a new one copied from https://github.com/angular/material/blob/HEAD/src/core/style/variables.scss

$display-1-font-size-base or $display-2-font-size-base should work

bryk:
Last style comments and let's merge.

src/app/frontend/index.scss
floreks: Of course. Just wanted to make sure that it is ok.

floreks:
Done. PTAL.

bryk:
Nice work, thanks! :)

PR #296, opened 2016-01-26
Validation and layout on deploy page

digitalfishpond:
Validation:
1) validation - message not showing when 2 different errors caused consecutively - fixed
2) validation - regex check to ensure lowercase and hyphens only
3) validation - increased responsiveness by replacing ng-model-options with ng-if="<model>.$touched"
4) validation - removed md-progress linear which was causing page to jitter (showing progress bar - creating a new line in the form - for fractions of a second and then hiding again). Validation for unique-name still works.
5) validation - for 'number of pods', added validation message for 'required' as it is possible to remove default value.

Layout:
6) layout - set default font-size to 14px
7) layout - adjusted alignment between components within a single row by removing .kd-deploy-input-row formatting in scss and adding md-input-container.
8) layout - renamed 'more options / less options' to 'show advanced options / hide advanced options' to avoid grammatical error and clumsiness of 'more options / fewer options'
9) moved 'show advanced options / hide advanced options' toggle to top of advanced options section to avoid having to scroll down to hide, but also added a 'hide advanced options' button at bottom of advanced options section
10) replaced switches for checkboxes contained in 'div' tags to preserve alignment between checkbox and label.

@bryk, please forgive my inevitable first-timer mistakes, and let me know what I need to change. Many thanks!

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings.html
+++ b/src/app/frontend/deploy/deployfromsettings.html
@@ -17,18 +17,22 @@ limitations under the License.
 <kd-help-section>
   <md-input-container class="md-block" md-is-error="ctrl.isNameError()">
     <label>App name</label>
-    <input ng-model="ctrl.name" name="name" namespace="ctrl.namespace" required kd-unique-name
-        ng-model-options="{ updateOn: 'default blur', debounce: { 'default': 500, 'blur': 0 } }">
-    <md-progress-linear class="kd-deploy-form-progress" md-mode="indeterminate"
-        ng-show="ctrl.form.name.$pending">
+    <input ng-model="ctrl.name" name="name" namespace="ctrl.namespace" required ng-pattern="ctrl.namePattern"
+           ng-model-options="{ updateOn: 'default blur', debounce: { 'default': 500, 'blur': 0 } }" kd-unique-name
+           ng-maxlength="ctrl.nameMaxLength">
+    <md-progress-linear class="kd-deploy-form-progress" md-mode="indeterminate" ng-show="ctrl.form.name.$pending">
     </md-progress-linear>
     <ng-messages for="ctrl.form.name.$error" role="alert" multiple>
       <ng-message when="required">Application name is required.</ng-message>
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
+      <ng-message when="uniqueName">Application with this name
+        already exists within namespace <i>{{ctrl.namespace}}</i>.</ng-message>
+      <ng-message when="pattern">Application name should contain only lowercase letters, numbers, and '-' between words
       </ng-message>
+      <ng-message when="maxlength">Application name should have no more than 63 characters</ng-message>
     </ng-messages>
+
   </md-input-container>
+
   <kd-user-help>
     An 'app' label with this value will be added to the Replication Controller and Service that get deployed.
     <a href="http://kubernetes.io/v1.1/docs/user-guide/labels.html" target="_blank">Learn more</a>
@@ -55,8 +59,9 @@ limitations under the License.
     <label>Number of pods</label>
     <input ng-model="ctrl.replicas" type="number" required min="1" name="replicas">
     <ng-messages for="ctrl.form.replicas.$error" role="alert" multiple>
+      <ng-message when="required">Number of pods is required.</ng-message>
       <ng-message when="number">Number of pods must be a positive integer.</ng-message>
-      <ng-message when="min">Number of pods must be positive.</ng-message>
+      <ng-message when="min">Number of pods must be at least 1.</ng-message>
     </ng-messages>
   </md-input-container>
   <kd-user-help>
@@ -81,11 +86,12 @@ limitations under the License.
 
 <kd-help-section>
   <md-checkbox ng-model="ctrl.isExternal" class="md-primary"
-             ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }">
+               ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }">
     Expose service externally
   </md-checkbox>
 </kd-help-section>
 
+<!-- advanced options -->
 <div ng-show="ctrl.isMoreOptionsEnabled()">
   <kd-help-section>
     <md-input-container>
@@ -99,18 +105,18 @@ limitations under the License.
   </kd-help-section>
 
   <kd-help-section>
-    <div>
-      <div>Labels</div>
-      <div layout="column">
-        <div layout="row">
-          <p flex>Key</p>
-          <p flex>Value</p>
-        </div>
-        <div ng-repeat="label in ctrl.labels">
-          <kd-deploy-label layout="row" flex label="label" labels="ctrl.labels"></kd-deploy-label>
-        </div>
+    <div layout="column">
+      <div class="kd-label-title md-body-2">Labels</div>
+      <div layout="row" class="kd-label-header-row">
+        <div flex="45">Key</div>
+        <div flex="5"></div>
+        <div flex="40">Value</div>
+      </div>
+      <div ng-repeat="label in ctrl.labels">
+         <kd-deploy-label layout="row" flex label="label" labels="ctrl.labels"></kd-deploy-label>
       </div>
     </div>
+
     <kd-user-help>
       The specified labels will be applied to the created Replication Controller, Service (if any) and Pods.
       Common labels include release, environment, tier, partition and track.
@@ -165,6 +171,7 @@ limitations under the License.
           <ng-message when="min">CPU requirement must be given as a positive number.</ng-message>
         </ng-messages>
       </md-input-container>
+      <div flex="5"></div>
       <md-input-container flex>
         <label>Memory requirement (MiB)</label>
         <input ng-model="ctrl.memoryRequirement" type="number" name="memoryRequirement" min="0">
@@ -200,9 +207,11 @@ limitations under the License.
   </kd-help-section>
 
   <kd-help-section>
-    <md-switch ng-model="ctrl.runAsPrivileged" class="md-primary">
-      Run as privileged
-    </md-switch>
+    <div class="md-block">
+      <md-checkbox ng-model="ctrl.runAsPrivileged" class="md-primary">
+        Run as privileged
+      </md-checkbox>
+    </div>
     <kd-user-help>
       Processes in privileged containers are equivalent to processes running as root on the host.
     </kd-user-help>
@@ -219,14 +228,14 @@ limitations under the License.
   </kd-help-section>
 </div>
 
+<!-- show/hide advanced options toggle -->
 <md-button class="md-primary kd-deploy-moreoptions-button" type="button"
            ng-click="ctrl.switchMoreOptions()"
            ng-hide="ctrl.isMoreOptionsEnabled()">
-  <md-icon>arrow_drop_down</md-icon>More options
+  <md-icon>arrow_drop_down</md-icon>show advanced options
 </md-button>
-
 <md-button class="md-primary kd-deploy-moreoptions-button" type="button"
            ng-click="ctrl.switchMoreOptions()"
            ng-show="ctrl.isMoreOptionsEnabled()">
-  <md-icon>arrow_drop_up</md-icon>Less options
+  <md-icon>arrow_drop_up</md-icon>hide advanced options
 </md-button>
--- a/src/app/frontend/deploy/deployfromsettings.scss
+++ b/src/app/frontend/deploy/deployfromsettings.scss
@@ -13,11 +13,10 @@
 // limitations under the License.
 
 @import '../variables';
+$bar-height: 2px;
 
 md-progress-linear {
   &.kd-deploy-form-progress {
-    $bar-height: 2px;
-
     clear: left;
     height: $bar-height;
     margin-bottom: -$bar-height;
--- a/src/app/frontend/deploy/environmentvariables.html
+++ b/src/app/frontend/deploy/environmentvariables.html
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div class="kd-environment-variables-title">Environment variables</div>
+<div class="kd-environment-variables-title md-body-2">Environment variables</div>
 <div ng-repeat="variable in ctrl.variables">
   <ng-form name="variablesForm" layout="row">
     <md-input-container flex class="kd-deploy-input-row">
@@ -25,6 +25,7 @@ limitations under the License.
         <ng-message when="pattern">Variable name must be a valid C identifier.</ng-message>
       </ng-messages>
     </md-input-container>
+    <div flex="5"></div>
     <md-input-container flex class="kd-deploy-input-row">
       <label>Value</label>
       <input ng-model="variable.value" name="value">
--- a/src/test/frontend/deploy/deployfromsettings_controller_test.js
+++ b/src/test/frontend/deploy/deployfromsettings_controller_test.js
@@ -359,4 +359,83 @@ describe('DeployFromSettings controller', () => {
       expect(ctrl.imagePullSecret).toEqual('');
     });
   });
+
+  /**
+   * The value entered for ‘App Name” is used implicitly as the name for several resources (pod, rc,
+   * svc, label). Therefore, the app-name validation pattern is based on the RC-pattern, but must
+   * conform with all validation patterns of all the created resources.
+   * Currently, the ui pattern that conforms with all patterns is alpha-numeric with dashes between.
+   */
+  it('should allow strings that conform to the patterns of all created resources', () => {
+    // given the pattern used by the App Name field in the UI
+    let namePattern = ctrl.namePattern;
+    // given the patterns of all the names that are implicitly created
+    let allPatterns = {
+      servicePattern: '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*',
+      labelPattern: '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?',
+      rcPattern: '[a-z0-9]([-a-z0-9]*[a-z0-9])?',
+      appNamePattern: namePattern,
+    };
+
+    // then
+    for (let pattern in allPatterns) {
+      expect('mylowercasename'.match(allPatterns[pattern])).toBeDefined();
+      expect('my-name-with-dashes-between'.match(allPatterns[pattern])).toBeDefined();
+      expect('my-n4m3-with-numb3r5'.match(allPatterns[pattern])).toBeDefined();
+    }
+  });
+
+  /**
+   * The value entered for 'App Name' is used implicitly as the name for several resources (pod, rc,
+   * svc, label).
+   * Remark: The app-name pattern excludes some service names and label values, which could be
+   * created manually. This is a restriction of the current design.
+   */
+  it('should reject names that fail to conform to appNamePattern', () => {
+
+    // given
+    let appNamePattern = ctrl.namePattern;
+
+    // then the following valid service names will be rejected
+    expect('validservice.com'.match(appNamePattern)).toBeNull();
+    expect('validservice/path'.match(appNamePattern)).toBeNull();
+
+    // then the following valid label names will be rejected
+    expect('valid_label'.match(appNamePattern)).toBeNull();
+    expect('ValidLabel'.match(appNamePattern)).toBeNull();
+    expect('validLabel'.match(appNamePattern)).toBeNull();
+
+    // then these input values will be rejected
+    expect('@myname-with-illegal-char-prefix'.match(appNamePattern)).toBeNull();
+    expect('myname-with-illegal-char-suffix#'.match(appNamePattern)).toBeNull();
+    expect('myname-with-illegal_char-in-middle'.match(appNamePattern)).toBeNull();
+    expect('-myname-with-hyphen-prefix'.match(appNamePattern)).toBeNull();
+    expect('myname-with-hyphen-suffix-'.match(appNamePattern)).toBeNull();
+    expect('Myname-With-Capital-Letters'.match(appNamePattern)).toBeNull();
+    expect('myname-with-german-umlaut-äö'.match(appNamePattern)).toBeNull();
+    expect('my name with spaces'.match(appNamePattern)).toBeNull();
+    expect('  '.match(appNamePattern)).toBeNull();
+  });
+
+  /**
+   * The data from the App Name field of the deploy form is used implicitly for the creation of
+   * Service Name, Label Name and RC name. Pod names are truncated by the api server and therefore
+   * ignored.
+   * Remark: The maximum characters number should match all three, thereby excluding
+   * service names of more than 63 chars via this form, while it is possible to create service names
+   * of 253 chars manually. This is a restriction of the current design.
+   *
+   * ctrl.maxNameLength = 63
+   */
+  it('should limit input that conforms to all created resources', () => {
+
+    // service names are max. 253 chars
+    expect(ctrl.maxNameLength <= 253);
+
+    //  label are max. 63 chars. the 256 prefix cannot be entered
+    expect(ctrl.maxNameLength <= 63);
+
+    // RC name are max. 63 chars.
+    expect(ctrl.maxNameLength <= 63);
+  });
 });
--- a/src/app/frontend/deploy/deploy.html
+++ b/src/app/frontend/deploy/deploy.html
@@ -31,10 +31,11 @@ limitations under the License.
       </kd-help-section>
 
       <div ng-switch="ctrl.selection">
-        <deploy-from-settings ng-switch-when="Settings" name="ctrl.name"
+        <deploy-from-settings ng-switch-when="Settings" class="md-body-1" name="ctrl.name"
             namespaces="ctrl.namespaces" detail="ctrl.detail" form="ctrl.deployForm" protocols="ctrl.protocols">
         </deploy-from-settings>
-        <deploy-from-file ng-switch-when="File" name="ctrl.name" detail="ctrl.detail" form="ctrl.deployForm">
+        <deploy-from-file ng-switch-when="File" class="md-body-1" name="ctrl.name" detail="ctrl.detail"
+            form="ctrl.deployForm">
         </deploy-from-file>
       </div>
 
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -122,6 +122,20 @@ export default class DeployFromSettingsController {
     this.name = '';
 
     /**
+     * Checks that a name begins and ends with a lowercase letter
+     * and contains nothing but lowercase letters and hyphens ("-")
+     * (leading and trailing spaces are ignored by default)
+     * @export {RegExp}
+     */
+    this.namePattern = new RegExp('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$');
+
+    /**
+     * Maximum length for Application name
+     * @export {string}
+     */
+    this.nameMaxLength = '63';
+
+    /**
      * Whether to run the container as privileged user.
      * @export {boolean}
      */
--- a/src/app/frontend/deploy/environmentvariables.scss
+++ b/src/app/frontend/deploy/environmentvariables.scss
@@ -15,5 +15,5 @@
 @import '../variables';
 
 .kd-environment-variables-title {
-  margin: $baseline-grid 0;
+  margin: $baseline-grid * 2 0;
 }
--- a/src/app/frontend/deploy/portmappings.html
+++ b/src/app/frontend/deploy/portmappings.html
@@ -26,6 +26,7 @@ limitations under the License.
         <ng-message when="max">Port must less than 65536.</ng-message>
       </ng-messages>
     </md-input-container>
+    <div flex="5"></div>
     <md-input-container flex class="kd-deploy-input-row">
       <label>Target port</label>
       <input ng-model="portMapping.targetPort" ng-change="ctrl.addProtocolIfNeeed()"
@@ -36,6 +37,7 @@ limitations under the License.
         <ng-message when="max">Target port must less than 65536.</ng-message>
       </ng-messages>
     </md-input-container>
+    <div flex="5"></div>
     <md-input-container flex="none" class="kd-deploy-input-row">
       <label>Protocol</label>
       <md-select ng-model="portMapping.protocol" name="protocol" is-external="ctrl.isExternal"


src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,17 +17,19 @@
...
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
-    </ng-messages>
+    <input ng-model="ctrl.name"
+           name="name"
+           namespace="ctrl.namespace"
+           required
+           kd-unique-name
+           ng-pattern="/^[a-z]+[a-z\-]*[a-z]*$/">
```
bryk: We tend to have regexes in controllers. This is because we can compile them and test. Can you move it to the controller and write tests?

bryk:
@batikanu @maciaszczykm Can you also review?

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,17 +17,19 @@
...
-        ng-show="ctrl.form.name.$pending">
-    </md-progress-linear>
-    <ng-messages for="ctrl.form.name.$error" role="alert" multiple>
-      <ng-message when="required">Application name is required.</ng-message>
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
-    </ng-messages>
+    <input ng-model="ctrl.name"
+           name="name"
```
bryk: ng-model-options were here to reflect name change not on every keystroke. When you removed this, the help text  "
Ports are optional. If specified, a Service will be created mapping the Port (incoming) to a target Port seen by the container. The internal DNS name for this Service will be" updates on every keystroke. This is distracting. With model options this updates on blur or after you stop typing, which is better. Revert? What do you think? @maciaszczykm

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -79,12 +82,25 @@
 </kd-help-section>
 
 <kd-help-section>
-  <md-switch ng-model="ctrl.isExternal" class="md-primary"
+  <md-checkbox ng-model="ctrl.isExternal" class="md-primary no-ink"
```
bryk: Why no-ink? :)

bryk: Default to default angular material spec unless there's a good reason not to.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -140,18 +156,18 @@
       <md-input-container flex>
         <label>CPU requirement (cores)</label>
         <input ng-model="ctrl.cpuRequirement" type="number" name="cpuRequirement" min="0">
-        <ng-messages for="ctrl.form.cpuRequirement.$error" role="alert" multiple>
-          <ng-message when="number">CPU requirement must be a valid number.</ng-message>
-          <ng-message when="min">CPU requirement must be positive.</ng-message>
-        </ng-messages>
+        <div ng-messages="ctrl.form.cpuRequirement.$error" ng-if="ctrl.form.cpuRequirement.$touched">
```
bryk: Why change from `<ng-messages` to `<div ng-messages`? I'm curious.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,17 +17,19 @@
...
-      </ng-message>
-    </ng-messages>
+    <input ng-model="ctrl.name"
+           name="name"
+           namespace="ctrl.namespace"
+           required
+           kd-unique-name
+           ng-pattern="/^[a-z]+[a-z\-]*[a-z]*$/">
+
+    <div ng-messages="ctrl.form.name.$error" ng-if="ctrl.form.name.$touched">
```
bryk: This `ng-if="ctrl.form.name.$touched"` changes the behavior that is default for angular material. Normally, when you type the value in the input for the first time, you don't get errors until you're done with typing your value. This makes sense, because you don't want to get message like "this is invalid email" before you've even typed `@`. What do you think about reverting this? And staying with material design defaults. 

Note that on line 18 there is `md-is-error="ctrl.isNameError()"` that shows unique-name errors nevertheless. This is because unique-name errors are async.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,17 +17,19 @@
...
-      <ng-message when="required">Application name is required.</ng-message>
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
-    </ng-messages>
+    <input ng-model="ctrl.name"
+           name="name"
+           namespace="ctrl.namespace"
+           required
+           kd-unique-name
```
bryk: Can you bring back the md-progress-linear and make it not jitter? This would be better :)

src/app/frontend/deploy/deployfromsettings.html
bryk: @maciaszczykm

bryk:
PTAL

maciaszczykm:
image

Font in the toolbar seems to be too small right now.

src/app/frontend/deploy/deployfromsettings.html
maciaszczykm: I agree, now when long name is being typed there is a noticeable flickering of mentioned message. I think previous version was fine in that matter.

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -14,6 +14,15 @@
 
 @import '../variables';
 
+* {
+  font-size: 14px;
```
maciaszczykm: It changes font size not only here, but on other pages too. It has to be changed to specific class.

maciaszczykm:
> 9) moved 'show advanced options / hide advanced options' toggle to top of advanced options section to avoid having to scroll down to hide, but also added a 'hide advanced options' button at bottom of advanced options section

Still not sure if we should have same button twice, it looks strange, but of course can be useful.

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: I added no-ink in order to get the blue color (when checked) consistent with the rest of the page. Angular's default checkbox is pink: https://material.angularjs.org/latest/demo/checkbox

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -118,7 +134,7 @@
   </kd-help-section>
 
   <kd-help-section>
-    <md-input-container class="md-block">
+    <md-input-container class="md-block"  style="margin-bottom: 20px;">
```
cheld: we should not have styles inline, I think. Any reason?

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: I have just checked - the bug I thought it was fixing was actually fixed by ng-if="ctrl.form.name.$touched" (see upcoming answer relevant to this point). This change actually has no effect. I will revert.

src/app/frontend/deploy/deployfromsettings.html
maciaszczykm: If defaults work (I think they do) then I would stick with them.

maciaszczykm:
image

Could you make space between option picker and name field bigger?

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: This addition still does not cause validation until the first blur. The user only starts seeing any validation after the first time he/she clicks away. After this however is where there is a difference: without this, if there are two errors in a row, the error mesage disappears. The input area remains red, but with no help text. For example, I type " %34_invalidName ", the pattern error shows. I back-space to remove all the characters one by one. The last character that I remove was "%" - (still an error character, so I haven't trggered a valid state yet). When I delete this last invalid character, the input component goes straight from one error state to the next ('required' is now not met). However the error message disappears altogether. 
this `ng-if="ctrl.form.name.$touched"` fixes that.

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: Yes, of course :)

maciaszczykm:
Added a few comments. PTAL

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: I will do that.

src/app/frontend/deploy/deployfromsettings.scss
bryk: Correct. This changes font size of entire application :) The solution for this may be to wrap deploy form with a class and apply font size there. Or do this in the chrome/chrome.html{scss}

src/app/frontend/deploy/deployfromsettings.html
bryk: md-primary is for blue color. no-ink is for disabling the fancy droplet-like animation on click.

src/app/frontend/deploy/deployfromsettings.html
batikanu: Does it work when you submit the form without touching form.name?

maciaszczykm: So it seems, that Angular defaults are not 100% working as expected. There is some issue with error messages and `ng-if="ctrl.form.name.$touched"` fixes it. I'm not sure if it's the only one solution.

bryk:
@digitalfishpond Btw, when you're ready for us to rereview, comment PTAL :)

src/app/frontend/deploy/deployfromsettings.html
bryk: Hmm.... I see this problem only for name input. This may be caused by ` md-is-error="ctrl.isNameError()"`. Please check this out.

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: There are some side-effects to having `ng-model-options`, and side effects to not having it. I'll enumerate the negative impact with and without here for clarity: 
WITH: the progress bar only occasionally shows for a fraction of a second, and most of the time it doesn't show at all. This is because it only starts showing from the moment the 'name' is permitted to update, removing a minimum of 500 ms from its total showing time which is most often less than that.
WITH: validation appears laggy to the user who can sometimes get a 'app must have a name' error message still showing for 1/2 second after he/she has been imputing valid characters, or a 'wrong characters' message for 1/2 second after his/her input is valid etc. 1/2 is a long time in terms of UX.
WITHOUT: calls are made to the server with every keystroke to check for unique name.
without: the aforementioned can cause validation (for uniqueName) to take longer while the calls to the server are made, but the progress bar shows during the entire time of that process.
WITHOUT: the name in the help section you mentioned doesn't flicker, but does update to reflect the user's input as soon as each validation has occured which can be on every keystroke if they are slow enough.
I have tried implementing a solution which shows the name in the help section only when the 'name' input does not have focus, but when the user changes focus the name has not always finished validating, causing the old name to show until validation (slow in this case) has finished, at which point it updates. 
@bryk @maciaszczykm @cheld, which solution should I implement?

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: @bryk, no, unfortunately removing `md-is-error="ctrl.isNameError()"` altogether changes nothing to this behaviour. This behaviour is not visible on the 'Container Image' field as there are currently not enough error cases to be able to cause two consecutive errors, but on the 'Number of Pods' field, if you remove `ng-if="ctrl.form.replicas.$touched"`, and then enter a negative value, then click out to make the error message show, and then backspace to empty, you can see the same behaviour - the 'required' message does not appear. This 'Number of Pods' field has neither `...isNameError()` nor any asynchronous validation, so for the moment I am still inclined to think it is an AngularJS thing.
@batikanu , you are right, when you submit the form without touching the form.name, the line turns red, but no message appears. I can fix this though, if we are going to keep the `$touched` solution.

src/app/frontend/deploy/deployfromsettings.html
cheld: For the unique name check the debounce is required. I assume this delay affects all validations on the field. 

The check on focus out with debounce is fine and done everywhere like this. However, the continues check after the first failure is new. I have not seen this on any other web page. The check as you type together with debounce might be a bit strange. 

Anyway, we should stick to the default behavior (debounce) of the framework  and leaf it for the UX expert to think about improvements

src/app/frontend/deploy/deployfromsettings.html
cheld: The described behavior can be found on any input field, even the material design demos. I consider it as a bug in md. We should not fix it on our side. Someone with good JS knowledge might want to investigate and create an issue. But not so urgent, I think

cheld: @bryk 
BTW: I don't understand the purpose of md-is-error="ctrl.isNameError()". I checked documentation and the test cases, but could not make any clue out of it. For me it seems buggy and superfluous, but probably I just lack understanding...

bryk: @digitalfishpond Thanks for through investigation. I agree with @cheld that this is Angular Material bug. It shouldn't behave like this. Having said that, I don't think we should fix this. To me, the best way is just to wait until Angular Material fixes it. 
What do you think?

bryk: > BTW: I don't understand the purpose of md-is-error="ctrl.isNameError()". I checked documentation and the test cases, but could not make any clue out of it. For me it seems buggy and superfluous, but probably I just lack understanding...

Normally errors are shown only when focusing-out. For names we show errors at the time of typing. This is because name validation is async. Does this make sense?

digitalfishpond:
I signed it!

digitalfishpond:
PTAL

src/app/frontend/deploy/deployfromsettings.scss
bryk: Still not fixed. 

Just apply .md-body-1 class to the deploy form. See more at: https://material.angularjs.org/latest/CSS/typography

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -14,6 +14,19 @@
...
+}
+
+md-input-container {
+  margin-bottom: 5px;
```
bryk: Base all margins on $baseline-grid. Make them natural multipliers of this value. Learn more at: https://www.google.com/design/spec/layout/responsive-ui.html#responsive-ui-grid

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -26,7 +39,6 @@ md-progress-linear {
   }
 }
 
-.kd-deploy-input-row {
-  margin-bottom: $baseline-grid;
-  margin-top: $baseline-grid;
+.progress-or-message-container {
```
bryk: Hmmm... I dont know why we neeed this. For me the progress bar is not jittering. What browser do you use?

bryk: How can I reproduce the problem?

src/test/frontend/deploy/deployfromsettings_controller_test.js
```diff
@@ -336,4 +336,39 @@ describe('DeployFromSettings controller', () => {
...
+
+  /**
+   * Test for the name validation pattern (namePattern)
+   *
+   * The pattern matches strings that start and end with a lowercase letter
+   * and contains only lowercase letters and hyphens.
+   * Full pattern used for validation:
+   *
+   * ^[a-z]+(-[a-z]+)*$
+   */
```
bryk: This test is awesome :)

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -200,23 +219,22 @@
   </kd-help-section>
 
   <kd-help-section>
-    <md-switch ng-model="ctrl.runAsPrivileged" class="md-primary">
-      Run as privileged
-    </md-switch>
+    <div>
+      <md-checkbox ng-model="ctrl.runAsPrivileged"
+                   class="md-primary"
+                   layout="row">
```
bryk: Why do you need `layout=row` here? Layouts only apply to child elements, whose this checkbox does not have

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,18 +17,21 @@
...
+           ng-model-options="{ updateOn: 'default blur', debounce: { 'default': 500, 'blur': 0 } }" kd-unique-name>
+    <div class="progress-or-message-container">
+      <md-progress-linear class="kd-deploy-form-progress" md-mode="indeterminate" ng-show="ctrl.form.name.$pending">
+      </md-progress-linear>
+      <ng-messages for="ctrl.form.name.$error"
+                   role="alert">
+        <ng-message when="required">Application name is required.</ng-message>
+        <ng-message when=="uniqueName">Application with this name
+          already exists within namespace <i>{{ctrl.namespace}}</i>.</ng-message>
+        <ng-message when="pattern">lowercase letters only, '-' between words</ng-message>
```
bryk: Name validation does not work for me. Something is broken here. Please open with gulp serve:prod and verify.

bryk:
I'm looking at the two buttons solution for progressive disclosure and it looks weird. We should definitely stay with only one button. But where to place it? @floreks @maciaszczykm @cheld Opinions?

bryk:
PTAL

src/app/frontend/deploy/deployfromsettings.html
cheld: @bryk I understand the code now, but not the reason. Why do we make the behavior inconsistent? Why does sync / async make a difference?

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,18 +17,21 @@
...
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
-    </ng-messages>
+    <input ng-model="ctrl.name" name="name" namespace="ctrl.namespace" required ng-pattern="ctrl.namePattern"
+           ng-model-options="{ updateOn: 'default blur', debounce: { 'default': 500, 'blur': 0 } }" kd-unique-name>
+    <div class="progress-or-message-container">
+      <md-progress-linear class="kd-deploy-form-progress" md-mode="indeterminate" ng-show="ctrl.form.name.$pending">
+      </md-progress-linear>
+      <ng-messages for="ctrl.form.name.$error"
```
cheld: You missed out multiple... Did you remove by intention? Multiple is supposed to show multiple errors at once. I think I have seen this in a demo. However, it seems not working in our code. Still it is used at many places. Does anyone know?

BTW: does anyone know what role=alert is supposed to do? Has no visible effect for me...

maciaszczykm:
@bryk @digitalfishpond I agree, that we should stay with one button. In my opinion we should place it at the bottom of the page, not in the middle. It's most popular way I think.

src/test/frontend/deploy/deployfromsettings_controller_test.js
cheld: I don't agree here!!

I have doubts that the implementation is correct. Even if it is correct, I have doubts that the next developer will change it correctly. Tests should expose (specify) the business logic. If it is not possible in test code, than it should be expressed in comment of the test. 

In this case the business logic is quite complex. The name field is used to create the following resources:
1. Pod name (with generated suffix, I think )
2. label which has some special prefix rules. Have you tried label prefixes?
3. replication controller 
4. service, but only optionally. Have you clicked expose button?

Create test cases that express the business logic properly. e.g. something like "This is the pattern for service, pod, label, rc. This is my pattern and it will match them all".

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,18 +17,21 @@
...
-    <md-progress-linear class="kd-deploy-form-progress" md-mode="indeterminate"
-        ng-show="ctrl.form.name.$pending">
-    </md-progress-linear>
-    <ng-messages for="ctrl.form.name.$error" role="alert" multiple>
-      <ng-message when="required">Application name is required.</ng-message>
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
-    </ng-messages>
+    <input ng-model="ctrl.name" name="name" namespace="ctrl.namespace" required ng-pattern="ctrl.namePattern"
```
cheld: could you please add maxlength? You already investigated the pattern logic. So, I assume it is only little additional effort. Please add test cases for max length as in this case it is non-trivial.

src/app/frontend/deploy/deployfromsettings.html
bryk: Role alert is for tools like screen readers or tools that try to understand the webpage. It's good practice to have it, but our UI is not yet accessible, so this gives little benefit.

src/test/frontend/deploy/deployfromsettings_controller_test.js
bryk: Thanks @cheld for digging into this. I've verified that the regex we have here is invalid. Basically we should follow DNS_LABEL spec from here: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L48
Which is slightly different from what we have now. 

@digitalfishpond Can you fix?

digitalfishpond: I have actually changed my test to this but not updated the test comment yet, my bad, I apologise (this back-end regex gives the same results as my shorter one, but I adopted the original one for consistency). I will update the comment of course. However, I think @cheld is suggesting that the input from this field is used in 4 different places on the back end, each with different sets of requirements. This current RegEx matches the most restrictive of the four, (only lowercase and hyphens) with the outcome that, for example, prefixes with "/" cannot be used. This is not a new restriction as names with anything other than lowercase and hyphens will currently be rejected from the backend (doesn't match the most restrictive regex)

cheld: @digitalfishpond 
1st: I just would like to see a test-case that proves your claim. :) Make it visible for others.
2nd: I would like to see a test-case that visualizes the restrictions. Just for everyone to understand what kind of limitations the design causes.

src/app/frontend/deploy/deployfromsettings.scss
digitalfishpond: Ha! yes, I see it's buttery smooth on Chrome :) i see the glitch in Firefox where the progress bar appears about 5 px below the input line (I've not managed to change this behaviour - but I did find this which has clearly been fixed for Chrome: https://github.com/angular/material/issues/2540)

src/app/frontend/deploy/deployfromsettings.scss
digitalfishpond: For width (left/right) I can see how this makes sense, but for height (top/bottom)? Am I missing something?

src/app/frontend/deploy/deployfromsettings.scss
bryk: Oh, I see! I'll check out this on Safari and Firefox.

Can you extract the variable `$bar-height: 2px;` to be in the file scope and reuse it here?

bryk:
PTAL

digitalfishpond:
PTAL

src/app/frontend/_variables.scss
```diff
@@ -50,3 +50,4 @@ $bold-font-weight: 500;
 $layout-breakpoint-lg: 1280px;
 $whiteframe-shadow-1dp: 0 1px 3px 0 rgba(0, 0, 0, .2), 0 1px 1px 0 rgba(0, 0, 0, .14), 0 2px 1px -1px rgba(0, 0, 0, .12);
 $baseline-grid: 8px;
+$bar-height: 2px;
```
bryk: Move $bar-height: 2px; to deployfromsettings.scss. This variable is used only there.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -38,9 +43,11 @@
 <kd-help-section>
   <md-input-container class="md-block">
     <label>Container image</label>
-    <input ng-model="ctrl.containerImage" name="containerImage" required>
+    <input ng-model="ctrl.containerImage"
+           name="containerImage"
+           required>
     <ng-messages for="ctrl.form.containerImage.$error" role="alert" multiple>
-      <ng-message when="required">Container image is required.</ng-message>
+      <ng-message when="required">Container image is required.></ng-message>
```
bryk: Error in this line too much `>`.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,18 +17,23 @@
...
+      </md-progress-linear>
+    </div>
     <ng-messages for="ctrl.form.name.$error" role="alert" multiple>
       <ng-message when="required">Application name is required.</ng-message>
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
+      <ng-message when=="uniqueName">Application with this name
+        already exists within namespace <i>{{ctrl.namespace}}</i>.</ng-message>
+      <ng-message when="pattern">lowercase letters only, '-' between words</ng-message>
```
bryk: Can you make the error message more like a formal sentence? I.e., `Application should ...`

bryk: Also, this is not only letters, but alphanumeric chars too

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -200,23 +209,25 @@
   </kd-help-section>
 
   <kd-help-section>
-    <md-switch ng-model="ctrl.runAsPrivileged" class="md-primary">
-      Run as privileged
-    </md-switch>
+    <div>
```
bryk: Hmm... Why this change?

digitalfishpond: requested by Ken (UX Designer) 25/01/2016

bryk: Ok, I know.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -17,18 +17,23 @@
...
+    <div class="kd-deploy-progress-container">
+      <md-progress-linear class="kd-deploy-form-progress" md-mode="indeterminate" ng-show="ctrl.form.name.$pending">
+      </md-progress-linear>
+    </div>
     <ng-messages for="ctrl.form.name.$error" role="alert" multiple>
       <ng-message when="required">Application name is required.</ng-message>
-      <ng-message when="uniqueName">
-        Application with this name already exists within namespace <i>{{ctrl.namespace}}</i>.
-      </ng-message>
+      <ng-message when=="uniqueName">Application with this name
```
digitalfishpond: just spotted this ==

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -14,10 +14,17 @@
 
 @import '../variables';
 
+md-input-container {
+  margin-bottom: 5px;
```
bryk: This changes margins of all input containers for entire application. Basically, you should never have CSS rule match `md-*` selector globally. You need to create a CSS class of your own and apply it to desired elements.

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -14,10 +14,17 @@
...
+  margin-top: 10px;
+}
+
+.mdblock {
+  margin-bottom: 20px;
```
bryk: The same.

src/app/frontend/deploy/helpsection/helpsection.scss
```diff
@@ -17,6 +17,7 @@
 // This rule is meant to align items only in the current layout (kd-help-section).
 .kd-help-section > * {
   flex: 1;
+  font-size: 14px;
```
bryk: Instead of doing this, just apply md-body-1 class to the entire deploy form page. E.g., to its root element.

src/test/frontend/deploy/deployfromsettings_controller_test.js
```diff
@@ -336,4 +336,85 @@ describe('DeployFromSettings controller', () => {
...
+   * The data from the App Name field of the deploy form is used implicitly for the creation of
+   * Service Name, Label Name and RC name. Pod names are truncated by the api server and therefore
+   * ignored.
+   * Remark: The maximum characters number should match all three, thereby excluding
+   * service names of more than 63 chars via this form, while it is possible to create service names
+   * of 253 chars manually. This is a restriction of the current design.
+   *
+   * ctrl.maxNameLength = 63
+   */
+
```
bryk: New line not needed here?

bryk:
PTAL

src/app/frontend/deploy/deployfromsettings.html
bryk: Yeah, sorry for confusion. Had to read back again PR description.

src/test/frontend/deploy/deployfromsettings_controller_test.js
```diff
@@ -336,4 +336,85 @@ describe('DeployFromSettings controller', () => {
...
+    // given the pattern used by the App Name field in the UI
+    let namePattern = ctrl.namePattern;
+    // given the patterns of all the names that are implicitly created
+    let allPatterns = {
+      servicePattern: "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*",
+      labelPattern: "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?",
+      rcPattern: "[a-z0-9]([-a-z0-9]*[a-z0-9])?",
+      appNamePattern: namePattern,
+    };
+
```
digitalfishpond: i will remove new line here also

src/test/frontend/deploy/deployfromsettings_controller_test.js
```diff
@@ -336,4 +336,85 @@ describe('DeployFromSettings controller', () => {
...
+   * Remark: The app-name pattern excludes some service names and label values, which could be
+   * created manually. This is a restriction of the current design.
+   */
+  it('should reject names that fail to conform to appNamePattern', () => {
+
+    // given
+    let appNamePattern = ctrl.namePattern;
+
+    // then the following valid service names will be rejected
+    expect("validservice.com".match(appNamePattern)).toBeNull();
```
bryk: Btw, use single quotes everywhere `'`. This is a pattern in the project, and generally in JavaScript. `"` gets bad when you do this in HTML.

digitalfishpond:
PTAL

digitalfishpond:
PTAL

digitalfishpond:
PTAL

bryk:
@floreks Can you take the review?

floreks:
@bryk I'll try. I've been pretty busy with my task and I have to catch up on dashboard :) I hope I'll finish soon.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -15,20 +15,23 @@
 -->
 
 <kd-help-section>
-  <md-input-container class="md-block" md-is-error="ctrl.isNameError()">
+  <md-input-container class="kd-deploy-block" md-is-error="ctrl.isNameError()">
```
floreks: This class is currently ignored anyway. I think we should stick with md-block. 

zrzut ekranu z 2016-02-09 13-21-08

Ps. For margins/paddings etc. use multipliers or divisors of $baseline-grid variable.

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -36,9 +39,11 @@
 </kd-help-section>
 
 <kd-help-section>
-  <md-input-container class="md-block">
+  <md-input-container class="kd-deploy-block">
     <label>Container image</label>
-    <input ng-model="ctrl.containerImage" name="containerImage" required>
+    <input ng-model="ctrl.containerImage"
+           name="containerImage"
+           required>
```
floreks: If it fits in 1 line the leave it be :) I think our cap is 100 chars per line.

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: @floreks: kd-deploy-block is used everywhere else in the form. The CSS for it alters the margins values. If it is currently ignored, wouldn't it be a better idea to keep it consistent with the rest of the form? Potential (however unlikely) alterations to the surroundings of this field will not require alterations to this field's code...?
Re. $baseline-grid: this makes sense for width, but not height? @bryk made the same comment, but I don't have a final answer on this...

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: Sorry, I didn't spot this from earlier alterations. I will rectify now :)

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -165,7 +172,8 @@
           <ng-message when="min">CPU requirement must be given as a positive number.</ng-message>
         </ng-messages>
       </md-input-container>
-      <md-input-container flex>
+      <div flex="5"></div>
```
floreks: Just to be consistent. Use the same spacing for `Environment variables`

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -13,11 +13,20 @@
...
+
+.kd-deploy-input-row {
+  margin: 0;
+}
+
+.kd-deploy-block {
+  @extend %md-block;
+  margin-bottom: 5px;
+  margin-top: 10px;
+}
```
floreks: As mentioned above. Not needed. It is ignored :)

src/app/frontend/deploy/deployfromsettings.html
floreks: If this is the only place where it is ignored then adjust this one places and check for other places. It can't be ignored in any place where it is used.

src/app/frontend/deploy/deployfromsettings.scss
```diff
@@ -26,7 +35,14 @@ md-progress-linear {
...
 
-.kd-deploy-input-row {
-  margin-bottom: $baseline-grid;
-  margin-top: $baseline-grid;
+.kd-user-help {
+  padding-top: 15px;
+}
+
+.kd-label-title {
+  margin: 0 0 5px $baseline-grid / 2;
```
floreks: Like mentioned above. Try to stick with $baseline-grid idea.

floreks:
Few small comments. PTAL

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: I will update css for this to reflect margin changes.

digitalfishpond:
PTAL

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -81,37 +86,38 @@
...
+               ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }">
     Expose service externally
   </md-checkbox>
 </kd-help-section>
 
+<!-- advanced options -->
 <div ng-show="ctrl.isMoreOptionsEnabled()">
   <kd-help-section>
-    <md-input-container>
+    <md-input-container class="md-block kd-deploy-block">
```
cheld: I would suggest to rename to kd-input-block

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -81,37 +86,38 @@
...
       <label>Description</label>
       <textarea ng-model="ctrl.description"></textarea>
     </md-input-container>
-    <kd-user-help>
+    <kd-user-help class="kd-extra-padding-top">
```
cheld: I would suggest to remove. We should aim for a consistent layout concept. I would only make things inconsistent in case a ui expert asks for it.

src/app/frontend/deploy/deployfromsettings.html
floreks: Actually if we want to follow the pattern it should be something like `kd-deploy-form-input-block` or `kd-deploy-input-block`. Normally we also add name of the page where css is applied or put it in chrome.css without this name if it's something that can be globally used.

src/app/frontend/deploy/deployfromsettings.html
digitalfishpond: done

src/app/frontend/deploy/deployfromsettings.html
```diff
@@ -157,7 +163,7 @@
 
   <kd-help-section>
     <div layout="row">
-      <md-input-container flex>
+      <md-input-container class="kd-no-margins" flex>
```
cheld: hmmm...I agree it looks nicer with no margins, but again it is just fiddling. I would rather like to see some consistent concepts. I am just worried that things get messed up as the application ages

src/app/frontend/deploy/deployfromsettings.html
cheld: don't we want to use this style potentially on another page? If a ui designer wishes to distinguish between pages, it would still be possible with something like "deploypage .kd-input-block"

cheld:
I added a few comments. IMHO, a consistent layout makes styling and branding by UI experts easy. But they can work around as well. So, I leave it up to you and the others.

src/app/frontend/deploy/deployfromsettings.html
cheld: IMHO, an input field is a global concept and should be styled globally. Of course, currently we only have one page. So, I agree `kd-deploy-block` is fine, but please refactor later...

bryk:
There's no gap between "Environment variables" and its children. 
selection_025

src/app/frontend/deploy/deployfromsettings.html
bryk: All styles that are defined in this page should be `kd-deploy-form-FOO`. If we decide to reuse them in the future, then we can extract them and rename. Such cases happened in the past. But until this happen, I'd be reluctant to over-generalize for the future. Think abut today and future you've planned. If there's only use case for the class here, make it local here.
Does this make sense?

cheld: Yes, it is the right approach.

bryk:
I've removed all CSS styles that you've added in this pull request, here are the results. Try to find out which is before and which is after.
As already @cheld pointed out, we should not be changing angular material default behaviors, unless we have a very good reason to do so. I think that the CSS styles you've written add maintenance burden and no value for users, because the page looks almost the same.

Can we have this PR for validation and bug fixes only? :) With that we would merge it quickly. And then move styling discussions somewhere else. 

selection_026
selection_027

digitalfishpond:
PTAL

digitalfishpond:
@bryk, so, you mean get rid of kd-deploy-form-block and... well everything?

digitalfishpond:
Ok, I've removed all new CSS, and I've doubled the original margin under Environment Variables title. PTAL

digitalfishpond:
@bryk PTAL

src/app/frontend/deploy/environmentvariables.html
```diff
@@ -25,6 +25,7 @@
         <ng-message when="pattern">Variable name must be a valid C identifier.</ng-message>
       </ng-messages>
     </md-input-container>
+    <div flex="5";></div>
```
bryk: Remove the ';' from here.

bryk:
Last comment and let's merge.

floreks:
@bryk there is a problem with travis here. On fork travis build is green.

https://travis-ci.org/digitalfishpond/dashboard/builds/108511032

Should this PR be closed and reopened to fix it?

bryk:
Can you link to travis error?

digitalfishpond:
https://travis-ci.org/kubernetes/dashboard/builds/108523983

bryk:
Ok, this is my fault. Sauce labs plugin does not work on forks. I'll fix this ASAP and merge this PR in meantime.

PR #298, opened 2016-01-27
Style update

maciaszczykm:
Style update:

- modified `Edit pod count`, `Create namespace` and `Delete replica set` dialogs margins and paddings, because they were too high
- modified deploy button (on replica set list page) colors, as we won't use default palette anymore
- changed `nesting-depth` to value `3`, because previous point couldn't be completed without it
- removed duplicated `Replica Set` text from replica set details page, it seems, that something was wrongly merged

### DIFF
--- a/src/app/frontend/deploy/createnamespace.html
+++ b/src/app/frontend/deploy/createnamespace.html
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<md-dialog aria-label="Create a new namespace" layout="column" layout-padding>
-  <md-content layout-padding>
+<md-dialog aria-label="Create a new namespace" layout="column">
+  <md-dialog-content layout-padding>
     <h4 class="md-title" >Create a new namespace</h4>
-    <p>The new namespace will be added to the cluster.</p>
+    <div>The new namespace will be added to the cluster.</div>
     <!-- TODO(maciaszczykm): Missing detailed cluster info. -->
     <form name="ctrl.namespaceForm" ng-submit="ctrl.createNamespace()"
           novalidate>
@@ -38,5 +38,5 @@ limitations under the License.
         <md-button ng-click="ctrl.cancel()">Cancel</md-button>
       </md-dialog-actions>
     </form>
-  </md-content>
+  </md-dialog-content>
 </md-dialog>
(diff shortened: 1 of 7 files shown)


src/app/frontend/deploy/createnamespace.scss
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@import '../variables';
+
+.kd-createnamespace-title,
+.kd-createnamespace-message  {
```
bryk: Hmm.... I see the same code in the other scss file. Can you move it to a common scss style here? Maybe a separate file? Or in deploy.scss?

maciaszczykm: I've moved it to `chrome.scss`, because these classes are useful for all of dialogs. Is it okay?

maciaszczykm:
PTAL

src/app/frontend/deploy/createnamespace.html
```diff
@@ -16,12 +16,12 @@
 
 <md-dialog aria-label="Create a new namespace" layout="column" layout-padding>
   <md-content layout-padding>
```
bryk: Btw, I did a little research and it happens that we're using the md-dialog directive in a wrong way. Take a look at the documentation: https://material.angularjs.org/latest/api/service/$mdDialog

Basically, there should be md-dialog-content here, not md-content. Also, when you remove layout-padding from line 17 and 18 and remove the CSS styles you've added in this CL, everything starts looking normal.

Please test and tell me what you think :)

bryk:
PTAL

src/app/frontend/deploy/createnamespace.html
maciaszczykm: That explains a lot! I've changed `md-content` into `md-dialog-content` in all dialogs. `layout-padding` is used now only fot `md-dialog-content`, no longer for `md-dialog`. I've reverted my CSS changes. Also used `<div>` instead of `<p>` to avoid additional paddings.

maciaszczykm:
PTAL

bryk:
LGTM

PR #299, opened 2016-01-27
Display metrics usage and connect to in-cluster heapster

bryk:
Now the data is correctly fetched from heapster if you run inside the cluster. 

TODO that is left: format CPU usage. Now raw number is displayed.

### DIFF
--- a/src/app/backend/heapsterclient.go
+++ b/src/app/backend/heapsterclient.go
@@ -29,17 +29,17 @@ import (
 func CreateHeapsterRESTClient(heapsterHost string, apiclient *client.Client) (
 	*unversioned.RESTClient, error) {
 
-	cfg := client.Config{
-		Host: heapsterHost,
-	}
+	cfg := client.Config{}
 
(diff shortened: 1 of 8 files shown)


bryk:
@maciaszczykm Can you review?

maciaszczykm:
@bryk Yes, of course.

maciaszczykm:
I've some troubles with heapster, because it seems to be not starting with `gulp local-up-cluster` anymore. Do you know something about it?

bryk:
It does work for me. You have to wait a while, however. To heapster to go up. 

If this does not help, do you have any stacktrace? Try accessing http://localhost:8082

maciaszczykm:
> It does work for me. You have to wait a while, however. To heapster to go up.

>If this does not help, do you have any stacktrace? Try accessing http://localhost:8082

I've removed my comment, because it has already started.

maciaszczykm:
image

LGTM

PR #300, opened 2016-01-27
Changed condition when to show full pod status (desired, running, created, etc.)

floreks:
On replica set list page:

* only pods running will be displayed if running == desired

On replica set detail page:

* same as on replica set list page + pod status part will be not displayed

Otherwise it is the same as it was before (running != desired)

PS. I need to add tests.

### DIFF
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -71,14 +71,11 @@ a {
 }
 
 .kd-comma-separated-item {
-  &:after {
-    content: ',';
-  }
-
-  &:last-child {
+  &:not(:last-child) {
     &:after {
-      content: '';
+      content: ',';
     }
+
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/app/frontend/replicasetdetail/replicasetdetail_controller.js
```diff
@@ -178,6 +178,14 @@ export default class ReplicaSetDetailController {
   }
 
   /**
+   * @return {boolean}
+   * @export
+   */
+  areDesiredPodsRunning() {
```
maciaszczykm: Missing comment.

floreks: It is self explanatory

src/app/frontend/chrome/chrome.scss
```diff
@@ -71,14 +71,11 @@ a {
 }
 
 .kd-comma-separated-item {
-  &:after {
-    content: ',';
-  }
-
-  &:last-child {
+  &:not(:last-child) {
```
bryk: Niiice :)

floreks: Finally I've had opportunity to check if `:not` selector is working properly.

bryk:
Tests? And LGTM

floreks:
I will add them shortly.

floreks:
Added tests.

bryk:
LGTM

Issue #303, opened 2016-01-27
Toolbar hides when there's a popup opened in the app


bryk:
@maciaszczykm Try opening image pull secret creation popup to reproduce. Maybe adding `z-index` can help?

maciaszczykm:
Okay, I'll check it.

maciaszczykm:
Okay, I got it. It was caused by Angular behavior, which sets `position: fixed` to `<body>` after clicking option picker. Setting `top: 0` on `kd-toolbar` fixes the issue. Created pull request #304, after it will be merged this issue can be closed.

maciaszczykm:
@bryk It's fixed now.

PR #304, opened 2016-01-28
Style fix

maciaszczykm:
Fixed styles:

- updated `Create a new image pull secret` dialog to use `md-dialog-content` instead of `md-content`
- fixed issue #303, please check it for more details

### DIFF
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -24,6 +24,7 @@
   box-shadow: $whiteframe-shadow-1dp;
   height: $toolbar-height-size-base;
   position: fixed;
+  top: 0;
 }
 
 body {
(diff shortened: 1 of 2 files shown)


bryk:
This does not fix the problem. Try scrolling down on replica sets list page and clicking logs menu.

maciaszczykm:
@bryk I'll check. At least it seems to fix issue with option pickers.

bryk:
Try looking at `z-index`. The catch-all container for popups seems to be at z-index: 50.

maciaszczykm:
> This does not fix the problem. Try scrolling down on replica sets list page and clicking logs menu.

Could you send a screenshot of problem? It seems to work for me:

image

bryk:
You're right. Sorry for the mess. I've probably opened wrong instance. Too many dashboards at the same time...

bryk:
LGTM

maciaszczykm:
Okay, then we shouldn't set `z-index`, because it could move toolbar in front of shadow which is created during dialog opening. Situation with too many dashboards happen to us, so for you it must be even worse :)

PR #305, opened 2016-01-28
Fix event filter issue

f-higashi:
When there is no warning events and you change filter to Warning in ReplicaSet Detail page,
 "No events were found" message is displayed and filter form is not shown.

### DIFF
--- a/src/app/frontend/replicasetdetail/replicasetdetail.html
+++ b/src/app/frontend/replicasetdetail/replicasetdetail.html
@@ -256,7 +256,7 @@ limitations under the License.
       </md-content>
     </md-tab>
     <md-tab label="Events">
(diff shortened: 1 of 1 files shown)


maciaszczykm:
It looks good, we can enhance it even more and update `hasEvents()` method to check if `replicaSetEvents.events` (it contains all events) is undefined or empty, instead of checking `events` (contains events after fitlering). Then with updated method there is no need to display filter at all if there are no events of both types.

bryk:
LGTM. I think we can stay with current solution at the UI controls will always stay the same, and only the table content will change. This is good from UX point of view, since you always have the same controls in place.

Can I merge?

cheld:
Cool Isao. LGTM

PR #306, opened 2016-01-28
Log view styling

maciaszczykm:
Fixed some issues from log view:

- controls are no longer turning black after usage (had to use `*` selector in `kd-logs-toolbar-select` to fix it, is it acceptable?)
- modified toolbar labels and font size
- increased log content horizontal paddings (to `16 px`)

### DIFF
--- a/src/app/frontend/logs/logs.scss
+++ b/src/app/frontend/logs/logs.scss
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+@import '../variables';
+
 $logs-color-black: #000;
 $logs-color-white: #fff;
 
@@ -26,6 +28,6 @@ $logs-color-white: #fff;
 }
 
 .kd-logs-element {
-  padding: 0;
+  padding: 0 (2 * $baseline-grid) 0 (2 * $baseline-grid);
   word-wrap: break-word;
 }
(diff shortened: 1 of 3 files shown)


bryk:
> controls are no longer turning black after usage (had to use * selector in kd-logs-toolbar-select to fix it, is it acceptable?)

Can you explain why is this happening in the first place? Why is the font turning black?

maciaszczykm:
I've already searched for the cause, but didn't find anything useful. I'll spend some more time on it.

f-higashi:
> Can you explain why is this happening in the first place? Why is the font turning black? 

It is caused by the following style.
```
md-select:not([disabled]):focus .md-select-value {
  border-bottom-color: rrg(50, 109,230);
  color: rgba(0,0,0,0.87);
}
```
You can confirm it by using developer tool.
* select a md-slect element and add :focus to it 
* select a md-select-value element
* check Styles view



I didn't find where this style comes from.
But I modified style and it works.

```
.kd-logs-toolbar-select {
  margin-bottom: 2 * $baseline-grid;
  margin-left: auto;
  &:not([disabled]) {
    &:focus {
      .md-select-value {
        color: $logs-toolbar-color-white;
      }
    }
  }
}
```

maciaszczykm:
@f-higashi Thanks for helping with this one! So we have to handle with next Angular Material bug.

maciaszczykm:
PTAL

bryk:
LGTM

PR #309, opened 2016-01-29
Display CPU metrics and hide metrics when heapster not running

bryk:


bryk:
@floreks Can you review?

floreks:
Sure, will do it shortly

floreks:
Maybe it would be good to also show `-` if pod does not provide any metrics. Just to have similar look like for other fields. I see that current backend is not designed to allow that easly. Maybe just assume when both metrics are empty that pod does not provide them or check pod status. What do you think?

screenshot from 2016-01-29 13-25-37

bryk:
I can display `-`s easily - just slight change in the backend. 

Whether to show `-` instead of hiding metrics: My reasoning is that an administrator installs this thing, and there comes user. User should not care whether there's heapster installed or not. Maybe some admins would like to run this UI without heapster at all. In that case we would bother users with columns and dashes that would always be there.
What do you think?

floreks:
I didn't actually mean to remove hiding functionality. Just in case there is a pod on a list without metrics display '-' instead of 0. When they are all missing metrics it is ok to hide them. Does that make sense?

bryk:
That's actually current behavior. You can verify it. When the UI says that 0 mCPU is used, this is what it meant.

bryk:
Individual pods can get dashes to show empty metrics. The code is: `ng-if=::ctrl.replicaSetDetail.hasMetrics`

bryk:
PTAL

floreks:
On the screen you can see that even Pending pod has metrics displayed and when I'm checking directly heapster API, metrics are empty (they do not exist). Still backend transforms empty to {0, 0} and that is displayed in UI.

zrzut ekranu z 2016-02-01 11-34-13

zrzut ekranu z 2016-02-01 11-34-31

zrzut ekranu z 2016-02-01 11-34-36

Is that intended?

bryk:
All right, now I know. I've made it possible to show `-` for metrics partially, e.g., when CPU is missing. PTAL

bryk:
PTAL

floreks:
Awesome! Just 2 small things. Go test needs fix and please add tests for new JS methods. :) 
Rest LGTM.

bryk:
Thanks for the review. Will merge soon.

PR #310, opened 2016-01-29
Have help displayed in separate browser tab

GertiPoppel:
Added target="_blank" to the <a href> links for references to help topics so that documentation is displayed in a new browser tab.


bryk:
Awesome, thanks!

PR #316, opened 2016-02-01
moved development docu to sub-folder

cheld:
I moved the development documentation to the doc subfolder.

More and more end-users come across the project and try build the project in order to try it out.

We get a lot of issues with users failing on this attempt. However, the documentation is not intended for end-users so I moved it to a sub folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,115 +7,20 @@ itself.
 
 ## Usage
 
-The Dashboard is currently under active development and is not ready for production use (yet!).
+The Dashboard is currently under active development and is not ready for production use (yet!). To run the latest _unstable_ version execute the following command:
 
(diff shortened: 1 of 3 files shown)


bryk:
Yes, this is very good change.

bryk:
How about adding a link from the front page to the development docs? It is hard to find otherwise.

README.md
```diff
@@ -9,114 +9,6 @@ itself.
 
 The Dashboard is currently under active development and is not ready for production use (yet!).
 
-## Architecture Overview
```
luxas: Maybe there should be a link to getting started?

bryk:
Now looks really nice. I'm merging. LGTM

Issue #314, opened 2016-01-31
Cluster sticks with zerostate even if pod is created


cheld:
Prerequisites:
1. Running, empty cluster
2. Go to zero state

Steps:
1. kubectl run test --image="tutum/hello-world"
2. Reload browser

=> nothing happens. Cluster seems to be empty

bryk:
Fixed?

cheld:
Yes

PR #317, opened 2016-02-01
Fixed state transition when on zerostate and replica set is added

floreks:
Fix for issue #314 

This way user won't be able to manually switch to zerostate. He will be transitioned only if cluster is empty.


cheld:
That was quick :) I manually tested the fix and it seems to work.

bryk:
LGTM :)

PR #319, opened 2016-02-01
Added replica set card status messages

floreks:
This is the current look of the card status. I've also done some refactoring during backend implementation.

I've had a problem with cluster started with `gulp local-up-cluster`. Events returned by it has empty `Type` property. No idea why. It works fine on my local cluster started from kubernetes repo fork.

zrzut ekranu z 2016-02-01 14-03-33

**Important**: Need help in finding `Pending` icon. :)

@maciaszczykm @bryk PTAL

### DIFF
--- a/src/app/backend/events.go
+++ b/src/app/backend/events.go
@@ -137,19 +137,22 @@ func GetReplicationControllerPodsEvents(client *client.Client, namespace, replic
 		return nil, err
 	}
 
+	events, err := GetPodsEvents(client, pods)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return events, nil
+}
+
+// Gets events associated to given list of pods
+// TODO(floreks): refactor this to make single API call instead of N api calls
+func GetPodsEvents(client *client.Client, pods *api.PodList) ([]api.Event, error) {
 	events := make([]api.Event, 0, 0)
 
 	for _, pod := range pods.Items {
-		fieldSelector, err := fields.ParseSelector("involvedObject.name=" + pod.Name)
-
-		if err != nil {
-			return nil, err
-		}
-
-		list, err := client.Events(namespace).List(unversioned.ListOptions{
-			LabelSelector: unversioned.LabelSelector{labels.Everything()},
-			FieldSelector: unversioned.FieldSelector{fieldSelector},
-		})
+		list, err := GetPodEvents(client, pod)
 
 		if err != nil {
 			return nil, err
@@ -162,6 +165,26 @@ func GetReplicationControllerPodsEvents(client *client.Client, namespace, replic
 	}
 
 	return events, nil
+}
+
+// Gets events associated to given pod
+func GetPodEvents(client client.Interface, pod api.Pod) (*api.EventList, error) {
+	fieldSelector, err := fields.ParseSelector("involvedObject.name=" + pod.Name)
+
+	if err != nil {
+		return nil, err
+	}
+
+	list, err := client.Events(pod.Namespace).List(unversioned.ListOptions{
+		LabelSelector: unversioned.LabelSelector{labels.Everything()},
+		FieldSelector: unversioned.FieldSelector{fieldSelector},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return list, nil
 }
 
 // Appends events from source slice to target events representation.
--- a/src/app/backend/eventscommon.go
+++ b/src/app/backend/eventscommon.go
@@ -1,0 +1,152 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"k8s.io/kubernetes/pkg/api"
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"log"
+	"strings"
+)
+
+// Partial string to correctly filter warning events.
+// Has to be lower case for correct case insensitive comparison.
+const FAILED_REASON_PARTIAL = "failed"
+
+// Contains basic information about event related to a pod
+type PodEvent struct {
+	// Short, machine understandable string that gives the reason
+	// for this event being generated.
+	Reason string `json:"reason"`
+
+	// A human-readable description of the status of related pod.
+	Message string `json:"message"`
+}
+
+// Returns warning pod events based on given list of pods.
+// TODO(floreks) : Import and use Set instead of custom function to get rid of duplicates
+func GetPodsEventWarnings(client client.Interface, pods []api.Pod) (result []PodEvent, err error) {
+	for _, pod := range pods {
+		if !isRunningOrSucceeded(pod) {
+			log.Printf("Getting warning events from pod: %s", pod.Name)
+			events, err := GetPodEvents(client, pod)
+
+			if err != nil {
+				return nil, err
+			}
+
+			result = getPodsEventWarnings(events)
+		}
+	}
+
+	return removeDuplicates(result), nil
+}
+
+// Returns list of Pod Event model objects based on kubernetes API event list object
+// Event list object is filtered to get only warning events.
+func getPodsEventWarnings(eventList *api.EventList) []PodEvent {
+	result := make([]PodEvent, 0)
+
+	var events []api.Event
+	if isTypeFilled(eventList.Items) {
+		events = filterEventsByType(eventList.Items, api.EventTypeWarning)
+	} else {
+		events = filterEventsByReason(eventList.Items, FAILED_REASON_PARTIAL)
+	}
+
+	for _, event := range events {
+		result = append(result, PodEvent{
+			Message: event.Message,
+			Reason:  event.Reason,
+		})
+	}
+
+	return result
+}
+
+// Filters kubernetes API event objects based on event type.
+// Empty string will return all events.
+func filterEventsByType(events []api.Event, eventType string) []api.Event {
+	if len(eventType) == 0 || len(events) == 0 {
+		return events
+	}
+
+	result := make([]api.Event, 0)
+	for _, event := range events {
+		if event.Type == eventType {
+			result = append(result, event)
+		}
+	}
+
+	return result
+}
+
+// Filters kubernetes API event objects based on reason property.
+// Empty string will return all events.
+func filterEventsByReason(events []api.Event, partial string) []api.Event {
+	if len(partial) == 0 || len(events) == 0 {
+		return events
+	}
+
+	result := make([]api.Event, 0)
+	for _, event := range events {
+		if strings.Contains(strings.ToLower(event.Reason), partial) {
+			result = append(result, event)
+		}
+	}
+
+	return result
+}
+
+// Returns true if all given events type is filled, false otherwise.
+// This is needed as some older versions of kubernetes do not have Type property filled.
+func isTypeFilled(events []api.Event) bool {
+	if len(events) == 0 {
+		return false
+	}
+
+	for _, event := range events {
+		if len(event.Type) == 0 {
+			return false
+		}
+	}
+
+	return true
+}
+
+// Removes duplicate strings from the slice
+func removeDuplicates(slice []PodEvent) []PodEvent {
+	visited := make(map[string]bool, 0)
+	result := make([]PodEvent, 0)
+
+	for _, elem := range slice {
+		if !visited[elem.Reason] {
+			visited[elem.Reason] = true
+			result = append(result, elem)
+		}
+	}
+
+	return result
+}
+
+// Returns true if given pod is in state running or succeeded, false otherwise
+func isRunningOrSucceeded(pod api.Pod) bool {
+	switch pod.Status.Phase {
+	case api.PodRunning, api.PodSucceeded:
+		return true
+	}
+
+	return false
+}
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -159,7 +159,7 @@ func DeployApp(spec *AppDeploymentSpec, client client.Interface) error {
 		Containers: []api.Container{containerSpec},
 	}
 	if spec.ImagePullSecret != nil {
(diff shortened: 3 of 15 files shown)


src/app/frontend/replicasetlist/replicasetcard.html
```diff
@@ -19,11 +19,21 @@
...
+            {{::ctrl.replicaSet.name}}
+          </span>
+        </a>
+        <md-icon class="material-icons error kd-replicaset-card-status-icon"
+                 ng-if="ctrl.hasErrors()">
+          error
+        </md-icon>
+        <md-icon ng-if="ctrl.isPending()" class="material-icons kd-replicaset-card-status-icon">
+          <!--TODO find the right icon!-->
+          warning
```
floreks: Whoever knows this icon name. Please reply :)
zrzut ekranu z 2016-02-01 14-08-44

bryk:
Re icon: anything worth using on material icons? E.g., https://design.google.com/icons/#ic_timelapse ?
Can you look at others?

bryk:
Or something from the notification list? https://design.google.com/icons/

bryk:
Re performance: It seems that for every card we query for events. I think it was supposed to query for events only when not all pods are running. Am I correct?

Sample log:
```
2016/02/01 16:44:25 Incoming HTTP/1.1 GET /api/replicasets request from [::1]:48812
2016/02/01 16:44:25 Getting list of all replica sets in the cluster
2016/02/01 16:44:25 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Getting list of pods error events
2016/02/01 16:44:26 Outcoming response to [::1]:48812 with 201 status code
```

bryk:
Should this be an error? It is not when I test it 

selection_015

src/app/backend/events.go
```diff
@@ -164,6 +166,26 @@ func GetReplicaSetPodsEvents(client *client.Client, namespace, replicaSetName st
...
 
+// Gets events associated to given pod
+func GetPodEvents(client *client.Client, pod api.Pod) (*api.EventList, error) {
+	fieldSelector, err := fields.ParseSelector("involvedObject.name=" + pod.Name)
+
+	if err != nil {
+		return nil, err
+	}
+
+	list, err := client.Events(pod.Namespace).List(unversioned.ListOptions{
```
bryk: Hey, now instead of one api call we have N api calls. Please fix :)

src/app/backend/replicasetcommon.go
```diff
@@ -100,3 +114,78 @@ func getReplicaSetPodInfo(replicaSet *api.ReplicationController, pods []api.Pod)
 
 	return result
 }
+
+// Returns Pod Events of type warning from given list of pods
+// TODO(floreks) : use some set to get only unique error messages
```
bryk: Isn't this done by having `removeDuplicates`?

src/app/backend/replicasetcommon.go
```diff
@@ -100,3 +114,78 @@ func getReplicaSetPodInfo(replicaSet *api.ReplicationController, pods []api.Pod)
...
+	for _, event := range events {
+		if event.Type == eventType {
+			result = append(result, event)
+		}
+	}
+
+	return result
+}
+
+// Returns list of Pod Event model objects based on kubernetes API event objects
```
bryk: Is this correct documentation?

src/app/backend/replicasetcommon.go
```diff
@@ -100,3 +114,78 @@ func getReplicaSetPodInfo(replicaSet *api.ReplicationController, pods []api.Pod)
...
+func getPodsEventErrors(events *api.EventList) []PodEvent {
```
bryk: Signature says errors, implementation does warnings. So, what is this?

src/app/backend/events.go
floreks: There are N api calls even now. I've only refactored it and extracted function.

https://github.com/kubernetes/dashboard/blob/master/src/app/backend/events.go#L149

src/app/backend/replicasetcommon.go
floreks: I've actually wanted to import `Set` instead of writing workaround function. Godeps are hard so for now I've just left this function. :)

src/app/backend/replicasetcommon.go
floreks: That's confusing for me actually, because api does not name it as errors. There are only events of type warning even if they are actually errors. I'll change everything to warning, but still card is marked as in error state.

floreks:
> Should this be an error? It is not when I test it

Check my initial comment :)
> I've had a problem with cluster started with gulp local-up-cluster. Events returned by it has empty Type property. No idea why. It works fine on my local cluster started from kubernetes repo fork.

You can check if they are empty on replica set detail events tab. If there will be no warning icons then it means that event types are empty and no errors can be recognized.

floreks:
> Re performance: It seems that for every card we query for events. I think it was supposed to query for events only when not all pods are running. Am I correct?

This may be misleading, because log info is on the start of function and then in the function itself is check whether pod is in right state to get events.

src/app/backend/replicasetcommon.go
floreks: You are right. It's actually based on kubernetes API EventList object. I'll change.

floreks:
> Re icon: anything worth using on material icons? E.g., https://design.google.com/icons/#ic_timelapse ?
Can you look at others?

I couldn't find anything better than timelapse, so I've used it.

src/app/backend/replicasetcommon.go
```diff
@@ -100,3 +114,77 @@ func getReplicaSetPodInfo(replicaSet *api.ReplicationController, pods []api.Pod)
...
+
+			result = getPodsEventWarnings(events)
+		}
+	}
+
+	return removeDuplicates(result), nil
+}
+
+// Filters kubernetes API event objects based on event type. Empty string will return all events
+func FilterEventsByType(events []api.Event, eventType string) []api.Event {
```
floreks: Should it be public?

floreks:
I've changed some comments and renamed functions. PTAL

src/app/backend/events.go
bryk: Oh, yes. Sorry. Can you add a todo so that we know what to fix?

src/app/backend/replicasetcommon.go
bryk: No need IMO

src/app/backend/replicasetcommon.go
bryk: Can you change the todo to be specific then?

src/app/backend/replicasetcommon.go
```diff
@@ -100,3 +114,78 @@ func getReplicaSetPodInfo(replicaSet *api.ReplicationController, pods []api.Pod)
...
+		}
+	}
+
+	return result
+}
+
+// Returns true if given pod is in state running or succeeded, false otherwise
+func isRunningOrSucceeded(pod api.Pod) bool {
+	switch pod.Status.Phase {
+	case api.PodFailed, api.PodPending, api.PodUnknown:
```
bryk: Hmm.... I'd reverse this statement. The reason is that if new status is added, the function will be broken. If you list running and success here, then everything would be fine.

src/app/backend/events.go
floreks: Sure. This PR is quite big so i will do that in next PR.

src/app/frontend/replicasetlist/replicasetcard.html
```diff
@@ -19,11 +19,20 @@
     <div layout="column">
       <div flex layout="row" layout-align="space-between center"
           class="kd-replicaset-card-title-row">
-          <a ng-href="{{::ctrl.getReplicaSetDetailHref()}}" class="kd-replicaset-card-name" flex>
-            <h3 class="md-title kd-replicaset-card-title">
-              {{::ctrl.replicaSet.name}}
-            </h3>
-          </a>
+        <span layout="row" flex>
+        <a ng-href="{{::ctrl.getReplicaSetDetailHref()}}" class="kd-replicaset-card-name">
```
bryk: Indent lines.

src/app/backend/replicasetcommon.go
floreks: Makes sense.

src/app/frontend/replicasetlist/replicasetcard.html
```diff
@@ -19,11 +19,20 @@
...
+          <span class="md-title kd-replicaset-card-title">
+            {{::ctrl.replicaSet.name}}
+          </span>
+        </a>
+        <md-icon class="material-icons error kd-replicaset-card-status-icon"
+                 ng-if="ctrl.hasWarnings()">
```
bryk: Use one-time bindings `::` as everywhere in this file.

src/app/frontend/chrome/chrome.scss
```diff
@@ -84,3 +84,9 @@ a {
 [role='tabpanel'] {
   transition: none;
 }
+
+.material-icons {
+  &.error {
```
bryk: Should this be .error or .kd-error?

floreks: I wasn't sure. I'll make it kd-error.

bryk:
The icons need some kind of tooltip that explain what they mean. I.e., the timelapse icon should have description that says that some pods are pending.

bryk:
>> Should this be an error? It is not when I test it

> Check my initial comment :)

Hmm.. I'm running on a "real" cluster and I still see this as pending. We should fix this, IMO. How about some kind of heuristics? I.e., when there is no event type, then check whether reason has `Failed` substring. This should cover most cases, as I'm looking at various failure reasons. What do you think?

bryk:
PTAL

floreks:
> Hmm.. I'm running on a "real" cluster and I still see this as pending. We should fix this, IMO. How about some kind of heuristics? I.e., when there is no event type, then check whether reason has Failed substring. This should cover most cases, as I'm looking at various failure reasons. What do you think?

This should work. I've discussed with @maciaszczykm, because he also implemented events tab on details page and this 'bug' is related to kubernetes. Till some version `Type` is just not filled. I'll check your suggestion.

bryk:
> This should work. I've discussed with @maciaszczykm, because he also implemented events tab on details page and this 'bug' is related to kubernetes. Till some version Type is just not filled. I'll check your suggestion.

That's sad... Try fixing this somehow. If this gets too hacky, then leave it alone.

floreks:
I need to adjust some comments only, but you can take a look.

src/app/backend/eventscommon.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	return result
+}
+
+// Returns true if type of event is filled, false otherwise.
+// This is needed as some older versions of kubernetes do not have Type property filled.
+func isTypeFilled(events []api.Event) bool {
+	for _, event := range events {
+		return len(event.Type) != 0
```
bryk: Hmm... This is equivalent to checking for only first event's type. How about for every event you check whether it has type or not. This is the way we should do this, because, I imagine, that some events have type and some not.

bryk:
Btw, I tested this on my instance. The tool we're building looks better and better! :)

src/app/backend/eventscommon.go
floreks: I think that if api will fill one event type then it will fill it for all events, but just to be extra safe I'll change.

bryk:
We need some more tests on the backend and everything else looks good.

floreks:
This PR has grown quite big :). PTAL

src/app/frontend/chrome/chrome.scss
```diff
@@ -84,3 +84,9 @@ a {
 [role='tabpanel'] {
   transition: none;
 }
+
+.material-icons {
+  &.kd-error {
+    color: $secondary;
```
bryk: I dont think this should be `$secondary`. It is a coincidence that our secondary color is red. It was pink not so long ago. 

This should have color of {{warn-A700}} as defined, e.g., here: https://github.com/angular/material/blob/1f997951b5d480d04ea7d356107837e279d6a271/src/components/input/input-theme.scss#L70

src/app/frontend/replicasetlist/replicasetcard.html
```diff
@@ -19,11 +19,22 @@
...
+          <a ng-href="{{::ctrl.getReplicaSetDetailHref()}}" class="kd-replicaset-card-name">
+            <span class="md-title kd-replicaset-card-title">
               {{::ctrl.replicaSet.name}}
-            </h3>
+            </span>
           </a>
+          <md-icon class="material-icons kd-error kd-replicaset-card-status-icon"
+                   ng-if="::ctrl.hasWarnings()">
+            error
+            <md-tooltip>One or more pods have failed scheduling</md-tooltip>
```
bryk: Have failed scheduling? Maybe just say that have error events? This is because it is not always failed scheduling, it can be other errors.

bryk:
Last two comments and LGTM.

bryk:
Should we merge this after replica set renaming? I think yes.

floreks:
> Last two comments and LGTM.

Fixed

> Should we merge this after replica set renaming? I think yes.

Unfortunately yes. Conflicts are going to kill me for sure :)

src/app/frontend/replicasetlist/replicasetcard.html
```diff
@@ -19,11 +19,22 @@
...
+          <a ng-href="{{::ctrl.getReplicaSetDetailHref()}}" class="kd-replicaset-card-name">
+            <span class="md-title kd-replicaset-card-title">
               {{::ctrl.replicaSet.name}}
-            </h3>
+            </span>
           </a>
+          <md-icon class="material-icons md-warn kd-replicaset-card-status-icon"
+                   ng-if="::ctrl.hasWarnings()">
+            error
+            <md-tooltip>One or more pods have errors</md-tooltip>
```
floreks: Is it ok?

bryk:
Ok, so LGTM on this. Let's wait for the rename to be done, and you'll rebase. Hopefully, without many problems :)

src/app/frontend/replicasetlist/replicasetcard.html
bryk: Yes.

bryk:
Rebase please :)

floreks:
I really hope I've merged it correctly :)

bryk:
LGTM, nice job! Tested on prod, works like a charm. Merging.

Issue #313, opened 2016-01-31
External link not displayed without load balance


cheld:
Prerequisites:
1. Run local cluster
2. Open Dashboard

Steps:

1. Open deploy dialog
2. Enter data
3. click expose service externally
4. Click deploy

=> no external link is shown

The current implementation expects a load-balancer. However, load-balancer is optional and it should work without. A link with nodeport should be displayed in case load balancer is not available.

I would suggest the following implementation:

--- Expose service externally not checked: ClusterIP should be used. No nodeport created. (current implementation creates nodeport)

--- Expose service externally checked: Either nodeport or loadbalancer. Do we know what the platform supports? If not, I would suggest to create loadbalancer and show nodeport until the loadbalancer ip is returned.

Needs to be fixed before release

bryk:
> --- Expose service externally not checked: ClusterIP should be used. No nodeport created. (current implementation creates nodeport)

LGTM

bryk:
> --- Expose service externally checked: Either nodeport or loadbalancer. Do we know what the platform supports? If not, I would suggest to create loadbalancer and show nodeport until the loadbalancer ip is returned.

If we can check it during creation time, then LGTM. But this needs investigation?

bryk:
Can anyone take this?

cheld:
@floreks @maciaszczykm @batikanu Can anyone have a look?

maciaszczykm:
I'll check it.

maciaszczykm:
This issue is fixed now, #369 is continuation of it.

maciaszczykm:
@cheld I believe it's already fixed. Can we close it?

cheld:
Yes we can close

PR #320, opened 2016-02-01
External links handling

maciaszczykm:
Connected to #313 issue.

When `Expose service externally` option is not selected then `api.ServiceTypeClusterIP` is used instead of `api.ServiceTypeNodePort`. Cluster IP is displayed in UI.

@cheld Does it match your expectations? Do you have any suggestions?

Please do not merge yet. Needs some refactoring and implementation of logic for `Expose service externally` checked.

### DIFF
--- a/src/app/backend/replicationcontrollerdetail.go
+++ b/src/app/backend/replicationcontrollerdetail.go
@@ -158,12 +158,19 @@ func GetReplicationControllerDetail(client client.Interface, heapsterClient Heap
 
 	matchingServices := getMatchingServices(services.Items, replicationController)
 
+	// Anonymous callback function to get nodes by their names.
+	getNodeFn := func(nodeName string) (*api.Node, error) {
+		return client.Nodes().Get(nodeName)
+	}
+
 	for _, service := range matchingServices {
-		replicationControllerDetail.Services = append(replicationControllerDetail.Services, getServiceDetail(service))
+		replicationControllerDetail.Services = append(replicationControllerDetail.Services,
+			getServiceDetail(service, *replicationController, pods.Items, getNodeFn))
 	}
 
 	for _, container := range replicationController.Spec.Template.Spec.Containers {
-		replicationControllerDetail.ContainerImages = append(replicationControllerDetail.ContainerImages, container.Image)
+		replicationControllerDetail.ContainerImages = append(replicationControllerDetail.ContainerImages,
+			container.Image)
 	}
 
 	for _, pod := range pods.Items {
@@ -277,22 +284,15 @@ func UpdateReplicasCount(client client.Interface, namespace, name string,
 }
 
 // Returns detailed information about service from given service
-func getServiceDetail(service api.Service) ServiceDetail {
-	var externalEndpoints []Endpoint
-	for _, externalIp := range service.Status.LoadBalancer.Ingress {
-		externalEndpoints = append(externalEndpoints,
-			getExternalEndpoint(externalIp, service.Spec.Ports))
-	}
-
-	serviceDetail := ServiceDetail{
+func getServiceDetail(service api.Service, replicationController api.ReplicationController,
+	pods []api.Pod, getNodeFn GetNodeFunc) ServiceDetail {
+	return ServiceDetail{
 		Name: service.ObjectMeta.Name,
 		InternalEndpoint: getInternalEndpoint(service.Name, service.Namespace,
 			service.Spec.Ports),
-		ExternalEndpoints: externalEndpoints,
+		ExternalEndpoints: getExternalEndpoints(replicationController, pods, service, getNodeFn),
 		Selector:          service.Spec.Selector,
 	}
-
-	return serviceDetail
 }
 
 // Gets restart count of given pod (total number of its containers restarts).
@@ -320,6 +320,82 @@ func getInternalEndpoint(serviceName, namespace string, ports []api.ServicePort)
 		Host:  name,
 		Ports: getServicePorts(ports),
 	}
+}
+
+// Returns array of external endpoints for a replication controller.
+func getExternalEndpoints(replicationController api.ReplicationController, pods []api.Pod,
+	service api.Service, getNodeFn GetNodeFunc) []Endpoint {
+	var externalEndpoints []Endpoint
+	replicationControllerPods := filterReplicationControllerPods(replicationController, pods)
+
+	if service.Spec.Type == api.ServiceTypeNodePort {
+		externalEndpoints = getNodePortEndpoints(replicationControllerPods, service, getNodeFn)
+	}
+
+	if service.Spec.Type == api.ServiceTypeLoadBalancer {
+		for _, ingress := range service.Status.LoadBalancer.Ingress {
+			externalEndpoints = append(externalEndpoints, getExternalEndpoint(ingress,
+				service.Spec.Ports))
+		}
+
+		if len(externalEndpoints) == 0 {
+			externalEndpoints = getNodePortEndpoints(replicationControllerPods, service, getNodeFn)
+		}
+	}
+
+	return externalEndpoints
+}
+
+// Returns pods that belong to specified replication controller.
+func filterReplicationControllerPods(replicationController api.ReplicationController,
+	allPods []api.Pod) []api.Pod {
+	var pods []api.Pod
+	for _, pod := range allPods {
+		if isLabelSelectorMatching(replicationController.Spec.Selector, pod.Labels) {
+			pods = append(pods, pod)
+		}
+	}
+	return pods
+}
+
+// Returns array of external endpoints for specified pods.
+func getNodePortEndpoints(pods []api.Pod, service api.Service, getNodeFn GetNodeFunc) []Endpoint {
+	var externalEndpoints []Endpoint
+	var externalIPs []string
+	for _, pod := range pods {
+		node, err := getNodeFn(pod.Spec.NodeName)
+		if err != nil {
+			continue
+		}
+		for _, adress := range node.Status.Addresses {
+			if adress.Type == api.NodeExternalIP && len(adress.Address) > 0 &&
+				isExternalIPUniqe(externalIPs, adress.Address) {
+				externalIPs = append(externalIPs, adress.Address)
+				for _, port := range service.Spec.Ports {
+					externalEndpoints = append(externalEndpoints, Endpoint{
+						Host: adress.Address,
+						Ports: []ServicePort{
+							{
+								Protocol: port.Protocol,
+								Port:     port.NodePort,
+							},
+						},
+					})
+				}
+			}
+		}
+	}
+	return externalEndpoints
+}
+
+// Returns true if given external IP is not part of given array.
+func isExternalIPUniqe(externalIPs []string, externalIP string) bool {
+	for _, h := range externalIPs {
+		if h == externalIP {
+			return false
+		}
+	}
+	return true
 }
 
 // Returns external endpoint name for the given service properties.
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -208,7 +208,7 @@ func DeployApp(spec *AppDeploymentSpec, client client.Interface) error {
 		if spec.IsExternal {
 			service.Spec.Type = api.ServiceTypeLoadBalancer
 		} else {
-			service.Spec.Type = api.ServiceTypeNodePort
+			service.Spec.Type = api.ServiceTypeClusterIP
 		}
 
 		for _, portMapping := range spec.PortMappings {
--- a/src/app/backend/replicationcontrollerlist.go
+++ b/src/app/backend/replicationcontrollerlist.go
@@ -27,6 +27,9 @@ import (
 // Callback function in order to get the pod status errors
 type GetPodsEventWarningsFunc func(pods []api.Pod) ([]PodEvent, error)
 
+// Callback function in order to get node by name.
+type GetNodeFunc func(nodeName string) (*api.Node, error)
+
 // ReplicationControllerList contains a list of Replication Controllers in the cluster.
 type ReplicationControllerList struct {
 	// Unordered list of Replication Controllers.
@@ -104,7 +107,13 @@ func GetReplicationControllerList(client *client.Client) (*ReplicationController
 		return errors, nil
 	}
 
-	result, err := getReplicationControllerList(replicationControllers.Items, services.Items, pods.Items, getPodsEventWarningsFn)
+	// Anonymous callback function to get nodes by their names.
+	getNodeFn := func(nodeName string) (*api.Node, error) {
+		return client.Nodes().Get(nodeName)
+	}
+
+	result, err := getReplicationControllerList(replicationControllers.Items, services.Items,
+		pods.Items, getPodsEventWarningsFn, getNodeFn)
 
 	if err != nil {
 		return nil, err
@@ -117,8 +126,8 @@ func GetReplicationControllerList(client *client.Client) (*ReplicationController
 // Replication Controller and Service API objects.
 // The function processes all Replication Controllers API objects and finds matching Services for them.
 func getReplicationControllerList(replicationControllers []api.ReplicationController,
-	services []api.Service, pods []api.Pod, getPodsEventWarningsFn GetPodsEventWarningsFunc) (
-	*ReplicationControllerList, error) {
+	services []api.Service, pods []api.Pod, getPodsEventWarningsFn GetPodsEventWarningsFunc,
+	getNodeFn GetNodeFunc) (*ReplicationControllerList, error) {
 
 	replicationControllerList := &ReplicationControllerList{ReplicationControllers: make([]ReplicationController, 0)}
 
@@ -134,10 +143,8 @@ func getReplicationControllerList(replicationControllers []api.ReplicationContro
 		for _, service := range matchingServices {
 			internalEndpoints = append(internalEndpoints,
 				getInternalEndpoint(service.Name, service.Namespace, service.Spec.Ports))
-			for _, externalIP := range service.Status.LoadBalancer.Ingress {
-				externalEndpoints = append(externalEndpoints,
-					getExternalEndpoint(externalIP, service.Spec.Ports))
-			}
+			externalEndpoints = getExternalEndpoints(replicationController, pods, service,
+				getNodeFn)
 		}
 
 		matchingPods := make([]api.Pod, 0)
@@ -156,17 +163,18 @@ func getReplicationControllerList(replicationControllers []api.ReplicationContro
 
 		podInfo.Warnings = podErrors
 
-		replicationControllerList.ReplicationControllers = append(replicationControllerList.ReplicationControllers, ReplicationController{
-			Name:              replicationController.ObjectMeta.Name,
-			Namespace:         replicationController.ObjectMeta.Namespace,
-			Description:       replicationController.Annotations[DescriptionAnnotationKey],
-			Labels:            replicationController.ObjectMeta.Labels,
-			Pods:              podInfo,
-			ContainerImages:   containerImages,
-			CreationTime:      replicationController.ObjectMeta.CreationTimestamp,
-			InternalEndpoints: internalEndpoints,
-			ExternalEndpoints: externalEndpoints,
-		})
+		replicationControllerList.ReplicationControllers = append(replicationControllerList.ReplicationControllers,
+			ReplicationController{
+				Name:              replicationController.ObjectMeta.Name,
+				Namespace:         replicationController.ObjectMeta.Namespace,
+				Description:       replicationController.Annotations[DescriptionAnnotationKey],
+				Labels:            replicationController.ObjectMeta.Labels,
+				Pods:              podInfo,
+				ContainerImages:   containerImages,
+				CreationTime:      replicationController.ObjectMeta.CreationTimestamp,
+				InternalEndpoints: internalEndpoints,
+				ExternalEndpoints: externalEndpoints,
+			})
 	}
 
 	return replicationControllerList, nil
--- a/src/test/backend/replicationcontrollerdetail_test.go
+++ b/src/test/backend/replicationcontrollerdetail_test.go
@@ -236,6 +236,263 @@ func TestGetExternalEndpoint(t *testing.T) {
 	}
 }
 
+func TestIsExternalIPUniqe(t *testing.T) {
+	cases := []struct {
+		externalIPs []string
+		externalIP  string
+		expected    bool
+	}{
+		{
+			[]string{"127.0.0.1", "192.168.1.1"},
+			"172.0.0.1",
+			true,
+		},
+		{
+			[]string{"127.0.0.1", "192.168.1.1", "172.0.0.1"},
+			"172.0.0.1",
+			false,
+		},
+		{
+			[]string{},
+			"172.0.0.1",
+			true,
+		},
+	}
+	for _, c := range cases {
+		actual := isExternalIPUniqe(c.externalIPs, c.externalIP)
+		if !reflect.DeepEqual(actual, c.expected) {
+			t.Errorf("isExternalIPUniqe(%+v, %+v) == %+v, expected %+v", c.externalIPs,
+				c.externalIP, actual, c.expected)
+		}
+	}
+}
+
+func TestFilterReplicationControllerPods(t *testing.T) {
+	firstLabelSelectorMap := make(map[string]string)
+	firstLabelSelectorMap["name"] = "app-name-first"
+	secondLabelSelectorMap := make(map[string]string)
+	secondLabelSelectorMap["name"] = "app-name-second"
+	cases := []struct {
+		replicationController api.ReplicationController
+		pods                  []api.Pod
+		expected              []api.Pod
+	}{
+		{
+			api.ReplicationController{
+				Spec: api.ReplicationControllerSpec{
+					Selector: firstLabelSelectorMap,
+				},
+			},
+			[]api.Pod{
+				{
+					ObjectMeta: api.ObjectMeta{
+						Name:   "first-pod-ok",
+						Labels: firstLabelSelectorMap,
+					},
+				},
+				{
+					ObjectMeta: api.ObjectMeta{
+						Name:   "second-pod-ok",
+						Labels: firstLabelSelectorMap,
+					},
+				},
+				{
+					ObjectMeta: api.ObjectMeta{
+						Name:   "third-pod-wrong",
+						Labels: secondLabelSelectorMap,
+					},
+				},
+			},
+			[]api.Pod{
+				{
+					ObjectMeta: api.ObjectMeta{
+						Name:   "first-pod-ok",
+						Labels: firstLabelSelectorMap,
+					},
+				},
+				{
+					ObjectMeta: api.ObjectMeta{
+						Name:   "second-pod-ok",
+						Labels: firstLabelSelectorMap,
+					},
+				},
+			},
+		},
+	}
+	for _, c := range cases {
+		actual := filterReplicationControllerPods(c.replicationController, c.pods)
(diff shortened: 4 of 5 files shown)


cheld:
No

My expectation on local-cluster (or any other without load-balancer)

Input:
Name,  image

=> no internal and no external link (ok)

Input:
Name, image, port

=> internal link, no external link (nok)

Input
Name, image, port, external

=> internal link, external link (nok)

Input
Name, image, external

=> validation error (nok)

cheld:
PTAL

maciaszczykm:
> --- Expose service externally not checked: ClusterIP should be used. No nodeport created. (current implementation creates nodeport)

So it means, that above requirement needs no more changes than switching service type from `api.ServiceTypeClusterIP` to `api.ServiceTypeClusterIP`. Correct me if I'm wrong please. I wrongly tried to display any external endpoint, while `Expose service externally` wasn't checked.

cheld:
Yes no external link should be created if external is not checked. Also, the link you showed is not externally accessible.

In my first comment I claim that nodeport is created even if external is not checked. I cannot observe this behavior in your code. 

IMHO, the problem is when creating the service we might not know whether to create nodeport or loadbalancer.

maciaszczykm:
Reverted unnecessary changes. Now I'll work on these cases:

> Input
Name, image, port, external

> => internal link, external link (nok)

> Input
Name, image, external

> => validation error (nok)

Regarding this:

> --- Expose service externally checked: Either nodeport or loadbalancer. Do we know what the platform supports? If not, I would suggest to create loadbalancer and show nodeport until the loadbalancer ip is returned.

 You said, that we need to use either one of node port or load balancer, but we don't know what is supported by the platform. I'll do some research, but cannot promise if I'll find anything useful. If not, how we should proceed?

maciaszczykm:
> IMHO, the problem is when creating the service we might not know whether to create nodeport or loadbalancer.

I'll investigate it.

cheld:
We could always create a loadbalancer. This creation is async. When reading the service data from the API server we could check if the loadbalancer-IP is returned. If not use the nodeport.

However, in this case the link might switch at some point in time when the load balancer has been created.

cheld:
BTW: the internal link always shows the dns name even if dns is not supported (like in local-cluster). If it is easy, it would be nice if the IP could be shown instead of DNS in case DNS is not supported. But this has not high priority.

bryk:
> BTW: the internal link always shows the dns name even if dns is not supported (like in local-cluster). If it is easy, it would be nice if the IP could be shown instead of DNS in case DNS is not supported. But this has not high priority.

Yes. We can check for the presence of cluster-dns addon. We should do this, but it's not high priority.

bryk:
I'll create a bug for this.

luxas:
Talking about DNS, it's quite easy to spin that up too in `hyperkube.sh`...

bryk:
Yes, but we cannot guarantee that DNS addon is going to run on user's clusters.

luxas:
No of course not. But I meant that we could make it easier for dev clusters...

bryk:
@cheld I LGTM this change. We can perhaps merge is and deal with loadbalancers later.

cheld:
@luxas I would appreciate a PR that enables DNS (if you know how to do quickly). I don't see any negative effect, is there?

maciaszczykm:
> Input
Name, image, port, external

> => internal link, external link (nok)

Is implemented now. PTAL

cheld:
I get the following link:

tcp-80-80-h4sm2:31975 

Port is correct, but host is not. I guess some kind of ID. Any minion IP will do. I assume that also master IP is ok, but I am not sure. 

The external link is also visible on details page. Must be fixed as well

I assume some tests missing

maciaszczykm:
@cheld I'll add details page and tests tommorow. IP is already changed to same as in `kubectl describe svc <service-name>`. For example:

```
Name:			fff1
Namespace:		default
Labels:			app=fff1,version=v3
Selector:		app=fff1,version=v3
Type:			LoadBalancer
IP:			10.0.0.182
Port:			tcp-80-8080-cx0f5	80/TCP
NodePort:		tcp-80-8080-cx0f5	30365/TCP
Endpoints:		172.17.0.10:8080,172.17.0.11:8080
Port:			tcp-79-7979-5o4ko	79/TCP
NodePort:		tcp-79-7979-5o4ko	30260/TCP
Endpoints:		172.17.0.10:7979,172.17.0.11:7979
Session Affinity:	None
No events.
```

In UI looks following:

image

Shouldn't minions IP should be used?

cheld:
10.0.0.182 is not reachable from the outside. It is a virtual IP. The minion IPs are potential reachable. I think the master IP would be better as it is certainly reachable, but I am not sure if the proxy service is running on master.

luxas:
@cheld It's not hard to setup, but I haven't time to do it atm, sorry

BTW, what does the `Expose service...` switch do in the `deploy` form? There should be documentation about that, since when running on bare-metal there is no `LoadBalancer`

cheld:
@luxas  The idea of the button is to make the deployed service accessible from the outside. Normally, all the IPs of pods and services are created inside the virtual flannel network and are not accessible from a host outside of the k8s cluster. 
My suggestion is that clicking on this button without loadbalancer creates a port-forwarding (proxy) on the physical host. This is called nodeport. 
See: http://kubernetes.io/v1.0/docs/user-guide/services.html#external-services

cheld:
@luxas Its not hard for you...I tried already for an hour or so and failed :)

maciaszczykm:
> 10.0.0.182 is not reachable from the outside. It is a virtual IP. The minion IPs are potential reachable. I think the master IP would be better as it is certainly reachable, but I am not sure if the proxy service is running on master.

Okay. If we aren't convinced about proxy service I guess we shouldn't go for master IP. In this case we should display minions IP. Should we diplay all of them or only one of them?

bryk:
External endpoints should be something that is reachable from outside of the cluster. So, if the service is NodePort, then nodeIP+port. If it is load balancer, then load-balancer-ip+port. What do you think?

maciaszczykm:
I agree, my concern at the moment is number of external endpoints. In case of 4 pods with 3 ports mapping it would give us 12 external endpoints if I'm not wrong.

cheld:
Well...we could check documentation or create a real cluster and check if kubeproxy is running on master and test if the service is reachable on masterIP + nodeport....or we just use nodeIP + nodeport.

I would show only one link and pick a random nodeIP. It is just a convenience function. If a user is realy  interested than he can find out himself with kubectl.

The number of external endpoints for each rc is: #nodes * #ports of service
Total number is: #rc * #nodes * #ports of service

maciaszczykm:
Updated external enpoints hosts and details page. Tests still to be updated, but please take a look if this is okay.

maciaszczykm:
Added test.

@bryk @cheld PTAL

cheld:
@maciaszczykm Now I understand. Sorry, still wrong. service-IP and pod-IP is virtual. node-ip is the IP of the physical host (what you get with ifconfig eth0). You cannot ping the service-IP or pod-IP from a remote host (because the IPs are not in the correct subnet) - just try it. However, you can curl <ifconfig eth0>:nodeport. 

BTW: "kubectl get nodes" returns 127.0.0.1 as NAME on my local cluster. On a real cluster it will return the correct IP, if I remember correctly. So, I guess this is the information you are looking for. Using 127.0.0.1 as IP for external link is fine (on a local cluster) in case it is the only information you have.

luxas:
BTW, `--config=/etc/kubernetes/manifests` should be swapped for `--config=/etc/kubernetes/manifests-multi` to the local cluster is accessible from outside. Now apiserver is listening on `localhost` which makes all outside access (maybe nodeport is an exception idk) blocked.

maciaszczykm:
@cheld I'll switch to node IP then, because previously I made it to work as following command:

```
maciaszczykm:~/workspace/dashboard/src/test/backend$ kubectl describe svc qqq
Name:			qqq
Namespace:		default
Labels:			app=qqq,version=v3
Selector:		app=qqq,version=v3
Type:			LoadBalancer
IP:			10.0.0.204
Port:			tcp-80-8080-ut7c7	80/TCP
NodePort:		tcp-80-8080-ut7c7	32443/TCP
Endpoints:		172.17.0.13:8080,172.17.0.14:8080
Port:			tcp-78-7878-nwgir	78/TCP
NodePort:		tcp-78-7878-nwgir	30222/TCP
Endpoints:		172.17.0.13:7878,172.17.0.14:7878
Port:			tcp-66-6666-4oc1h	66/TCP
NodePort:		tcp-66-6666-4oc1h	31130/TCP
Endpoints:		172.17.0.13:6666,172.17.0.14:6666
Session Affinity:	None
No events.
```

maciaszczykm:
@bryk @cheld Changed `PodIP` to `HostIP`. PTAL

src/app/backend/replicationcontrollerdetail.go
```diff
@@ -159,11 +159,13 @@ func GetReplicationControllerDetail(client client.Interface, heapsterClient Heap
 	matchingServices := getMatchingServices(services.Items, replicationController)
 
 	for _, service := range matchingServices {
-		replicationControllerDetail.Services = append(replicationControllerDetail.Services, getServiceDetail(service))
+		replicationControllerDetail.Services = append(replicationControllerDetail.Services, getServiceDetail(service,
```
bryk: Strange line wrapping here. Try to wrap at 100 chars.

src/app/backend/replicationcontrollerdetail.go
```diff
@@ -322,6 +317,37 @@ func getInternalEndpoint(serviceName, namespace string, ports []api.ServicePort)
...
 
+// Returns array of external endpoints for a replication controller.
+func getExternalEndpoints(replicationController api.ReplicationController, pods []api.Pod,
+	service api.Service) []Endpoint {
+	var externalEndpoints []Endpoint
+	for _, externalIP := range service.Status.LoadBalancer.Ingress {
+		externalEndpoints = append(externalEndpoints, getExternalEndpoint(externalIP, service.Spec.Ports))
+	}
+	// For type node ports or if load balancer is unavailable display node ports.
+	if (len(externalEndpoints) <= 0 && service.Spec.Type == api.ServiceTypeLoadBalancer) ||
```
bryk: How can array size be less than zero? Maybe check for equals zero?

src/app/backend/replicationcontrollerdetail.go
```diff
@@ -322,6 +317,37 @@ func getInternalEndpoint(serviceName, namespace string, ports []api.ServicePort)
...
+						Host: pod.Status.HostIP,
+						Ports: []ServicePort{
+							{
+								Protocol: port.Protocol,
+								Port:     port.NodePort,
+							},
+						},
+					})
+				}
+				// Returning external endpoints for only one pod.
```
bryk: Hmmm.... Should we do this? The UI displays all endpoints for other cases. I think this should stay the same here. @cheld @maciaszczykm

bryk:
PTAL

src/app/backend/replicationcontrollerdetail.go
cheld: In the design mockups we have one loadbalancer link for each port:
https://github.com/kubernetes/dashboard/blob/master/docs/design/mockups/11-11-2015/16_card%20variations.png

In this PR we also have one nodeport link for each port. 

Did I miss something?

cheld:
I have tested with local cluster only. For a short moment the UI shows a link only with the port, like ":31345". After refresh the correct link is shown. 
Maybe you can check if the HostIP is non-empty?

src/app/backend/replicationcontrollerdetail.go
bryk: The mocks don't cover all cases, that's the reason. They just show the idea.

maciaszczykm:
Refactored code, added tests and check for non-empty host IP. PTAL

src/app/backend/replicationcontrollerdetail.go
maciaszczykm: Actually the list of external enpoints is not that long shen using node IP instead of pod IP, so I think we should display all of them. It's already done, please chceck code/test how it works for you.

src/app/backend/replicationcontrollerdetail.go
maciaszczykm: My mistake, I always tend to use `=<` instead of `==` but in this case it makes no sense. Fixed.

src/app/backend/replicationcontrollerdetail.go
maciaszczykm: Done.

bryk:
Hmm.... We need to talk about this. External IP shows me a thing that is only accessible from within a cluster. This should not be like that. I'm running on Google Container Engine. 
selection_021

Can you re-elaborate why we're changing this? What's the problem that we solve here?

maciaszczykm:
CC @cheld

cheld:
The idea is that Dashboard should be usable without a load-balancer. e.g. standard openstack does not have a load-balancer. In the current situation you must use kubectl to get the external link. IMHO, Dashboard is not attractive if one has to fall back to kubectl for even simple use cases. Also, for a simple demo like on KubeCon I would have to cheat and patch the source code...

My suggestion is to show links with node-ports instead of load-balancer links. Marcin should test on GCE. However, Marcin stated above that we have a fundamental problem, because we don't know if the platform supports a load-balancer or not.

maciaszczykm:
@cheld Regarding technical side of the problem, I'm up for tests on GCE, but still, I'm not convinced what should be displayed here. It needs to be exactly clarified before any further steps. There are a lot of possible scenarios and detailed requirements are needed to implement it. All technical issues need to be taken into consideration (no information about load balancer support, available API etc.). We should think once more if using node name of pod is good. At the moment I've used host IP of pod, but  it seems to be incorrect like previous concepts.

@bryk Could you take a look on this discussion and changes made and tell what is your opinion about it? There are a lot of things to clarify and your opinion would help a lot.

Moreover, the more I'm thinking about `Expose service externally` switch the more I'm for adding options like `Cluster IP`, `Node port` and `Load balancer` instead of it. Does it make any sense to you? Perhaps. it can solve problems with identifying what does current platform support.

PTAL

bryk:
Ok, I see the problem. 

1. For me kubectl describe on a NodePort service shows 
```
kubectl describe service default-http-backend --namespace=kube-system
Name:			default-http-backend
Namespace:		kube-system
Labels:			k8s-app=glbc,kubernetes.io/cluster-service=true,kubernetes.io/name=GLBCDefaultBackend
Selector:		k8s-app=glbc
Type:			NodePort
IP:			10.167.252.53
Port:			http	80/TCP
NodePort:		http	30295/TCP
Endpoints:		10.164.2.8:8080
Session Affinity:	None
No events.
```
Which is different from what the UI shows. The port differs. 30295 vs 8080. Why is that? Should we change this? Just asking. 

2. What's algorithm do you propose? For type cluster ip => no external endpoints. For type nodeport => all nodes + ports. For type load balancer: show load balancer ingress, fall back do node port if ingress is empty. Correct? If yes, then LGTM

bryk:
Re what ip to display:
Can you reach your service with current solution? Is the IP external? HostIP does not seem to be reachable from outside of the cluster. At least to me. You could get external ip of a node from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L2018
Opinions?

luxas:
Trying to clarify some things here:
`SERVICE_CLUSTER_IP_RANGE`: `10.0.0.1/24` (default)
`FLANNEL_SUBNET`: `10.1.0.0/16` (default)
When a node (e.g. `node3`) joins the cluster, it allocates a `10.1.x.y` subnet e.g. `10.1.54.y`

`Pod IP`: When a pod is created on `node3`, it gets an `Pod IP` e.g. `10.1.54.178`
The overlay network (e.g. `flannel`) makes this IP routable from within the cluster
Let's say the pod exposes port `80`.
Then a `Endpoint` is created for the `Service` at: `10.1.54.178:80`

`Service IP`, `ClusterIP`: Services get an internal, fixed IP that's allocated as long as the service exists. Can be specified in `yaml`. Otherwise randomly generated by `apiserver`. `SERVICE_CLUSTER_IP_RANGE` is the range for `ClusterIP`s. E.g. `10.0.0.156`
`kube-proxy` runs on every node and forwards `10.0.0.156:80` => `10.1.54.178:80`. So if the pod goes down and is re-created on another node, it might get `Pod IP` `10.1.67.85`. Then `kube-proxy` will forward `10.0.0.156` to `10.1.67.85`.

`NodePort`: When creating a service, one can use `spec.type == nodePort` to expose the ports on **all** nodes (a `ClusterIP` is created too). In this example `apiserver` chose port `32547` for our service. That means `kube-proxy` will forward both `10.0.0.156:80` => `10.1.54.178:80` and `{NODEIP}:32547` => `10.1.54.178:80`.
This makes our service accessible from outside (if we can reach the nodes)

`apiserver proxy`: apiserver automatically exposes all services via this url: `http://{MASTERIP}:8080/api/v1/proxy/namespaces/{NAMESPACE}/services/{SERVICE}[:{SERVICENAME}]`

`LoadBalancer`: When using some cloud providers, there's a built-in loadbalancer. The cloud provider forwards: `CLOUD_PROVIDER_SPECIFIC_IP` => `10.0.0.156:80`

It took me a while to understand all these concepts, but I think I get them now. Hope it helps!

cheld:
I think I have a solution: we can add an environment variable to the deployment.yaml that describes if loadbalancer is available or not. This would be similar to the guestbook-example. 

https://github.com/kubernetes/kubernetes/blob/master/examples/guestbook/all-in-one/guestbook-all-in-one.yaml

In the yaml one can specify if the cluster supports DNS. We could do similar, but with loadbalancer.
``

env:
        - name: GET_HOSTS_FROM
          value: dns
          # If your cluster config does not include a dns service, then to
          # instead access an environment variable to find the master
          # service's host, comment out the 'value: dns' line above, and
          # uncomment the line below.
          # value: env
``

The rest should be just coding and testing with real clusters to make sure we take the right data.

@maciaszczykm let's have a chat today to make sure everything is clear

cheld:
Algorithm:
We have 3 service types (ClusterIP, NodePort, LoadBalancer see: http://kubernetes.io/v1.0/docs/user-guide/services.html#external-services) that must be matched to a switch (expose service externally) that has 2 states (on & off).

So, I suggest the following (assuming Marcin is correct with the analysis that support of LoadBalancer cannot be detected)

off: ClusterIP
on + no-environment-variable-set (default): LoadBalancer
on + environment-variable-set: NodePort

"gulp serve" should use NodePort

bryk:
> We have 3 service types (ClusterIP, NodePort, LoadBalancer see: http://kubernetes.io/v1.0/docs/user-guide/services.html#external-services) that must be matched to a switch (expose service externally) that has 2 states (on & off).
So, I suggest the following (assuming Marcin is correct with the analysis that support of LoadBalancer cannot be detected)
off: ClusterIP
on + no-environment-variable-set (default): LoadBalancer
on + environment-variable-set: NodePort
"gulp serve" should use NodePort

LoadBalancer falls back to NodePort when the environment does not support it. So I think there is no need to change the behavior of the switch.

bryk:
Re current algorithm: I like it, except for how we get node IPs. If we can get its external IP, that would be awesome.

bryk:
@luxas Thanks for the explanation, btw!

cheld:
We might not have to change the way the service is created. We can leave it on type loadbalancer. It does not harm in case it is not supported, but when displaying it, we don't know if we ever get a loadbalancer IP. This is my current understanding.

I assume the current implementation will display on GCE first the NodePort and after 5 minutes will switch to LoadBalancer (when it is ready). Not sure if this is nice...

cheld:
> Which is different from what the UI shows. The port differs. 30295 vs 8080. Why is that? Should we change this? Just asking.

The UI is correct and should not be changed, but I admit it will be a bit confusing for beginners. Let's take the following input:

App name: 
   test
Container Image:
   tutum/hello-world
Port:       Target Port:
   8000           80
External:
   X

Now, which port is the correct one to connect to the container?  Well it is either 30xxx or 80..., but not 8000 :)

bryk:
> I assume the current implementation will display on GCE first the NodePort and after 5 minutes will switch to LoadBalancer (when it is ready). Not sure if this is nice...

Yeah, not perfect. But I think this is just good enough for now.

maciaszczykm:
> What's algorithm do you propose? For type cluster ip => no external endpoints. For type nodeport => all nodes + ports. For type load balancer: show load balancer ingress, fall back do node port if ingress is empty. Correct? If yes, then LGTM

@bryk Current implementation matches your description.

>>Which is different from what the UI shows. The port differs. 30295 vs 8080. Why is that? Should we change this? Just asking.

>The UI is correct and should not be changed, but I admit it will be a bit confusing for beginners. Let's take the following input:

@bryk @cheld I'm not sure about it and at the moment I'll leave it like it's done now and change node IP to correct one.

> Can you reach your service with current solution? Is the IP external? HostIP does not seem to be reachable from outside of the cluster. At least to me. You could get external ip of a node from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L2018
Opinions?

Looks nice, I'll investigate it.

@luxas Thank you for explanation, it's not that easy at the beginning...

Thank you all for explanations and help with it!

floreks:
@luxas That is an awesome explanation. I didn't fully understand that before but now it's clear. Thanks!

Currently our port mapping is setting only in cluster configuration.

```
Name:			dashboard
Namespace:		default
Labels:			app=dashboard,version=canary
Selector:		app=dashboard,version=canary
Type:			LoadBalancer
IP:			    10.0.0.138
Port:			tcp-80-9090-xlh1h	80/TCP
NodePort:		tcp-80-9090-xlh1h	30587/TCP
Endpoints:		172.17.0.2:9090
```

Dashboard app is exposed on port 9090, so I wanted to expose it on port 80 providing port mapping `{Port: 80, TargetPort: 9090}` This way dashboard can be accessed from within the cluster using service VIP `10.0.0.138` on port `80`. In order to access it from outside, I have to actually use NodePort `30587` which is currently auto-generated because we are not setting it. My IP is f.e., `172.26.33.39` so Marcin was able to access dashboard through this IP on port `30587`. 

Endpoints are actually pod IP's and can be accessed also only within the cluster. Based on that I think that when expose externally is checked we should actually set `NodePort` and leave `Port` to be auto-generated and when it is not checked then we can set `Port`.

What do you think?

maciaszczykm:
>>Can you reach your service with current solution? Is the IP external? HostIP does not seem to be reachable from outside of the cluster. At least to me. You could get external ip of a node from https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L2018
Opinions?

>@bryk Looks nice, I'll investigate it.

On local cluster it returns following values:

```
[{LegacyHostIP 127.0.0.1} {InternalIP 127.0.0.1}]
```

I've got to test it on GCE.

bryk:
> On local cluster it returns following values:

Skip InternalIPs and LegacyIPs. Only use the external IP. 

Btw, can you easily get the details of the node that pod is running on? If this is too much hassle and requests, let's don't do it.

cheld:
> Endpoints are actually pod IP's and can be accessed also only within the cluster. Based on that I think that when expose externally is checked we should actually set NodePort and leave Port to be auto-generated and when it is not checked then we can set Port.

NodePort should be auto-generated, otherwise you can end up with port conflicts on the node.
The ports should be set manually, otherwise the internal inter-pod traffic routing will not work. 
Current, design is fine.

I believe Dashboard will be less confusing, once we have working links.

bryk:
@cheld Is correct. Node ports must be autogenerated.

maciaszczykm:
>Btw, can you easily get the details of the node that pod is running on? If this is too much hassle and requests, let's don't do it.

It's not that hard:

```
node, err := client.Nodes().Get(pod.Spec.NodeName)
```

However, at the moment I've passed `client` object through all of methods, which is not nice. I'm thinking about workaround. 

> Skip InternalIPs and LegacyIPs. Only use the external IP.

For local cluster there is no external IP. I will check on GCE (after I'll figure out how it works exacly).

floreks:
Then current implementation is fine and we need to only display Node IP.

bryk:
Checklist for this PR:

1. Can you access your services in your open stack installations through "external endpoint"?

2. Do we display loadbalancer ingress when available? (e.g., on public clouds)?


If answers are yes, looks good. Let's merge this.

maciaszczykm:
Current state:

- backend tests need update
- internal and legacy IPs are skipped
- used https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L2018 to get node external IP
- not tested on GCE and non-local clouds yet

maciaszczykm:
Okay, so we can test it now. Currently node external IP adress should be used, if it's not available then no other IP will be returned instead (legacy IP etc.).

@cheld @bryk PTAL

cheld:
Ok, I will test on GCE. But this will take some time.

bryk:
I'll test it on Monday on my public cloud cluster. Have a nice weekend!

cheld:
So my result:
 - GCE: first the NodePort link is shown. Then after a few minutes the loadbalancer link is shown.
  The IP is a correct IP from the host. 
  However, the NodePort link is not accessible, because of default firewall rules. After changing the firewall, I could access.

I assume this is not acceptable. We could fall back to my suggest environment variable...


- Local cluster: I admit local-cluster is not a valid use-case, but an external link is never shown.

maciaszczykm:
> GCE: first the NodePort link is shown. Then after a few minutes the loadbalancer link is shown. The IP is a correct IP from the host.

I believe this was expected result :)

> However, the NodePort link is not accessible, because of default firewall rules. After changing the firewall, I could access.

Shouldn't it stay like that? It's normal, that firewall will block access.

> Local cluster: I admit local-cluster is not a valid use-case, but an external link is never shown.

It won't be shown because there is no external IP for it. There are only internal and legacy IPs and we decided not to use them.

bryk:
LGTM. Let's wait for @cheld now.

cheld:
Ok, I think the prio 1 objective is solved.

I am not fully happy with leaving the link empty on local cluster. What about putting "localhost:nodeport" into the link as a fallback if no IP is available? 
First, I would help us for writing click-tests, because using external link would make the test work on local cluster (CI) and real cluster.
Second, the link is an important UI feature, it shouldn't be missing on local cluster.

What do you think? Is it too dirty?

maciaszczykm:
@bryk What do you think about @cheld's proposal? Perhaps we can merge this and continue to work on endpoints in another pull requests?

bryk:
I'm for merging this and reiterating on local cluster case in a next PR.

bryk:
@maciaszczykm Can you do a test on how @cheld's proposal works? On local cluster, GCE and openstack.

maciaszczykm:
@bryk I can test, should it be done before merging this PR? If no, then I'll do it after I'll finish my current task.

bryk:
I'd say in another PR.

maciaszczykm:
Okay, then I think we can merge it and continue to develop it in next PR. @cheld Do you agree?

cheld:
Yea, perfect

src/app/backend/replicationcontrollerdetail.go
```diff
@@ -322,6 +322,82 @@ func getInternalEndpoint(serviceName, namespace string, ports []api.ServicePort)
...
+	var pods []api.Pod
+	for _, pod := range allPods {
+		if isLabelSelectorMatching(replicationController.Spec.Selector, pod.Labels) {
+			pods = append(pods, pod)
+		}
+	}
+	return pods
+}
+
+// Returns array of external endpoints for a replication controller.
```
bryk: Update this documentation. It is false now.

bryk:
Last comment and LGTM and merge :)

maciaszczykm:
Done. PTAL

bryk:
Merging.

Issue #318, opened 2016-02-01
Backend test does not clean up after execution


floreks:
During writing go tests I've encountered that this test:

https://github.com/kubernetes/dashboard/blob/master/src/test/backend/deploy_test.go#L163

is creating namespace `development` during execution. Second execution of test results in failure as namespace already exists and is not deleted afterwards.

bryk:
Ouch, that's true. The test talks to real local cluster. 

@batikanu Can you fix this?

batikanu:
Ok I will check it

batikanu:
It is solved with #321

PR #321, opened 2016-02-01
Update deploy file test

batikanu:
Solution for issue #318 is implemented. The created namespace is deleted at the end of the execution.

### DIFF
--- a/src/app/backend/deploy.go
+++ b/src/app/backend/deploy.go
@@ -243,8 +243,16 @@ func getLabelsMap(labels []Label) map[string]string {
 	return result
 }
 
+type createObjectFromInfo func(info *kubectlResource.Info) error
+
+// Implementation of createObjectFromInfo
+var CreateObjectFromInfoFn = func(info *kubectlResource.Info) error {
+	_, err := kubectlResource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
+	return err
+}
+
 // Deploys an app based on the given yaml or json file.
-func DeployAppFromFile(spec *AppDeploymentFromFileSpec) error {
+func DeployAppFromFile(spec *AppDeploymentFromFileSpec, createObjectFromInfoFn createObjectFromInfo) error {
 	const (
 		validate      = true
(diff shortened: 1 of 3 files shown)


src/test/backend/deploy_test.go
```diff
@@ -177,6 +183,13 @@ func TestDeployAppFromFileWithValidContent(t *testing.T) {
 	if err != nil {
 		t.Errorf("Expected return value to be %#v but got %#v", nil, err)
 	}
+
+	//cleaning the created namespace
+	factory := cmdutil.NewFactory(nil)
```
bryk: If the test fail for some reason or I stop it in the middle of the work, the namespace will still be there. I think that the only solution is to create an interface for the dashboard <-> kubectl boundary.  For example, you could create an interface for the line:
`_, err = kubectlResource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)`

In tests you'd pass a verifying callback. On prod you'd call kubectl implementation callback.

bryk: Like, `createObjectFromInfo`.

batikanu: Good idea! I will check it.

batikanu:
PTAL

src/app/backend/deploy.go
```diff
@@ -243,8 +243,20 @@ func getLabelsMap(labels []Label) map[string]string {
 	return result
 }
 
+type ObjectFromInfo interface {
+	createObjectFromInfo(info *kubectlResource.Info) error
+}
+
+type ResourceFromInfo string
```
bryk: You dont need to do this. I.e., creating fake struct that fulfills an interface. You can just have a callback function as the argument. @floreks did this recently in PR https://github.com/kubernetes/dashboard/pull/319 , you can talk with him.

bryk:
Looks very nice. One comment on types. PTAL

batikanu:
I've implemented a callback function. PTAL

bryk:
Awesome, thanks!

PR #328, opened 2016-02-03
Set first release version and fix gulp release pipeline

bryk:
I propose that we start with version 0.1 and use semantic versioning.

### DIFF
--- a/build/conf.js
+++ b/build/conf.js
@@ -59,7 +59,7 @@ export default {
     /**
      * The release version of the image.
      */
-    versionRelease: '0.0.1',
+    versionRelease: 'v0.1.0',
     /**
      * The canary version name of the image. Canary is an image that is frequently published,
      * and has no release schedule.
--- a/build/deploy.js
+++ b/build/deploy.js
@@ -105,7 +105,7 @@ gulp.task('push-to-gcr:canary', ['docker-image:canary'], function(doneFn) {
  * Pushes release image to GCR.
  */
 gulp.task('push-to-gcr:release', ['docker-image:release'], function(doneFn) {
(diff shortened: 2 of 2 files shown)


bryk:
@cheld @floreks Can you take a look?

floreks:
Starting version is fine. As for the semantic versioning I think more common format is:

`0.1.0-alpha.1` or `0.1.0-alpha-1`

I like first better. What do you think?

lavvy:
I think I prefer the second option

bryk:
I didn't want to add hyphen, because kubernetes addons have a requirement to not have it in the version (yeah, that's sad). So how about going with just `0.1.0` for start? This what http://semver.org/ advises for initial pre-release development.

floreks:
Yeah, that's disappointing. I'd go with clean `0.1.0` then.

BTW. Does `0.1.0.alpha.1` look that bad?

bryk:
> BTW. Does 0.1.0.alpha.1 look that bad?

It does not, but semver needs a dash between `FOO.BAR.BAZ` and `alpha.1`

floreks:
Oh yeah. I see now. Due to our proxy I could not open semver page correctly...

luxas:
The best version would be `v0.1.0-alpha.1` as Kubernetes core (e.g. `v1.2.0-alpha.7`), but `v0.1.0` is fine too
Are you OK with `v` in the beginning?

floreks:
Fine by me

lavvy:
Cool

bryk:
Updated. PTAL.

floreks:
LGTM

Issue #323, opened 2016-02-01
Clean up service when deleting a rc


cheld:
Steps:

1. Go to deploy page
2. Select upload yaml
3. Select guestbook-all-in-one.yaml
4. Deploy
5. Delete all RCs
6. Repeat steps 2-4.

=> Error: master service already exists

This is not nice. Especially, in case a yaml is selected, because the names cannot be changed easily.

I think we discussed some kind of clean-up. 

Please someone has a look at this bug. Please fix before release.

bryk:
Yeah. A checkbox and prompt on what will be deleted in the delete form would be nice. 

I cannot take it ATM. This is high priority. Anyone can take this?

floreks:
I can work on this one. I'll prepare some concept.

floreks:
It's done so we can close.

PR #329, opened 2016-02-03
Added option to delete services related to replication controller while deleting RC

floreks:
Related to #323 

I've added option to delete services related to replication controller with some tooltip.

## Concept

Let's take simple label selector `{app: test}`

I'm checking if this selector targets only 1 replication controller. This is the most important check, because if there are more replication controllers targeted by this selector then we can not delete any related services.

If there is only 1 replication controller then services targeted by this selector can be deleted.

zrzut ekranu z 2016-02-03 15-27-02

Is tooltip message clear enough?


src/app/frontend/replicationcontrollerdetail/deletereplicationcontroller.html
```diff
@@ -21,6 +21,14 @@ <h4 class="md-title">Delete Replication Controller</h4>
       Delete replication controller {{::ctrl.replicationController}} in namespace {{::ctrl.namespace}}.<br>
       Pods managed by the replication controller will be also deleted.
     </div>
+    <md-checkbox ng-model="ctrl.deleteServices" class="kd-deletedialog-services-checkbox">
+      Delete related services
+      <md-icon class="material-icons kd-deletedialog-info-icon">
+        info
+        <md-tooltip>Services with label selector matching only this replication controller
+          will be deleted</md-tooltip>
```
floreks: Are tooltip and checkbox message ok?

floreks:
@maciaszczykm @bryk PTAL

maciaszczykm:
LGTM

src/app/backend/apihandler.go
```diff
@@ -297,11 +297,24 @@ func (apiHandler *ApiHandler) handleUpdateReplicasCount(
...
 
 	namespace := request.PathParameter("namespace")
 	replicationController := request.PathParameter("replicationController")
+	deleteServices, err := strconv.ParseBool(request.QueryParameter("deleteServices"))
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	if deleteServices {
```
bryk: Apihandlers should be logic-less. Can you move this to DeleteReplicationController and pass the boolean to it? The idea is to have zero logic in this file and just pass arguments to implementation functions.

floreks: Sure.

bryk:
Looks really nice. One comment. PTAL

floreks:
Moves logic away from api handler. Is function name still ok?

PTAL

bryk:
I think not. How about naming it DeleteReplicationController and then just documenting its behavior in docs?

floreks:
Ye, I think that's the best option

bryk:
Comment PTAL when you're ready.

floreks:
PTAL

bryk:
LGTM :)

PR #332, opened 2016-02-04
Added middle ellipsis directive and fixed long names formatting

floreks:
I've fixed long names handling. Most of names were just overlaping other elements.

zrzut ekranu z 2016-02-04 13-41-03

zrzut ekranu z 2016-02-04 13-40-54

@maciaszczykm @bryk PTAL


src/app/frontend/common/components/middleellipsis/middleellipsis_controller.js
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @final
+ */
+export default class MiddleEllipsisController {
```
bryk: This looks awesome, thnaks!

By the way, when we have time we can improve this even more. The idea is to compute width of the container and do binary search until the ellipsized text fits. Kind of similar to what we do for column layout. But this is for the future :)

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
```diff
@@ -40,7 +42,9 @@ <h1 flex class="md-title kd-replicationcontrollerdetail-app-name">
       <div flex layout="column" class="kd-replicationcontrollerdetail-sidebar-info">
         <span class="kd-replicationcontrollerdetail-sidebar-line">Name</span>
         <span class="kd-replicationcontrollerdetail-sidebar-subline">
-          {{::ctrl.replicationControllerDetail.name}}
+          <kd-middle-ellipsis display-string="{{::ctrl.replicationControllerDetail.name}}"
+                              max-length="{{::ctrl.nameMaxLength}}">
```
bryk: Hmm... I see plenty of free space left here. Should we use different max length?

src/app/frontend/common/components/middleellipsis/middleellipsis_controller.js
floreks: Definitely. I was also thinking about passing only string and calculating the rest. For now that was the fastest solution. I'll prepare something later when we'll be almost done with polishing everything. :)

bryk:
Re truncating labels on card: I think this is too restrictive. How about truncating a thing only when it is too long to fit a line. See this where labels are truncated, but they as well could be left untouched:
selection_022

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
floreks: I've adjusted this to widest possible name I could find that contained only `m` letters. We are not using monospaced font so all letters have different width.

bryk:
How about also increasing limit for the details page. The left nav resizes on demand, so we actually want to block only things that are extremely long. What do you think? 

selection_023

floreks:
I've increased limit for labels, but without actually calculating max length it shouldn't be increased more. I'm testing this for widest possible strings (all `m`). Labels are also displayed in multiple views so they should look ok in all of them.

RC name on details page can't really be increased further without name overlapping table. Sidebar still hax max size of 395px.

zrzut ekranu z 2016-02-04 15-58-21

bryk:
This week we have feature freeze. This is just a better way of displaying labels => this is not a feature :) If you can do this using dynamic calculation, I'd actually encourage you to do this :) Take your time.

floreks:
Fine by me ;) I'll work on dynamic calculation then.

bryk:
If you need any help, just ask.

floreks:
Of course, thanks!

floreks:
Closing for now.

PR #335, opened 2016-02-05
Implement environment variables input in deploy form

bryk:
As per spec:
https://raw.githubusercontent.com/kubernetes/dashboard/master/docs/design/mockups/11-11-2015/14_deploy%20more.png

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -85,6 +85,12 @@ export default class DeployFromSettingsController {
      */
     this.portMappings;
 
+    /**
+     * Initialized from the template.
+     * @export {!Array<!backendApi.EnvironmentVariable>}
+     */
+    this.variables;
+
     /** @export {boolean} */
     this.isExternal = false;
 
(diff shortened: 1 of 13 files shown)


bryk:
@maciaszczykm Can you review?

bryk:
@floreks Or you? :)

maciaszczykm:
@bryk I'll take a look.

src/app/frontend/deploy/deployfromsettings_controller.js
```diff
@@ -308,6 +315,14 @@ export default class DeployFromSettingsController {
   isPortMappingFilled_(portMapping) { return !!portMapping.port && !!portMapping.targetPort; }
 
   /**
+   * Returns true when the given environment variable is filled by the user, i.e., is not empty.
+   * @param {!backendApi.EnvironmentVariable} variable
+   * @return {boolean}
+   * @private
+   */
+  isVariableFilled_(variable) { return !!variable.name; }
```
maciaszczykm: Using `!!` in checks could be tricky. It transforms `0` to `false` and it is potential issue. Here it's not a problem since validation doesn't allow non-C identifiers, but for example in port mapping case when one of the ports is `0` it is the problem, because new field doesn't show and the user doesn't know why. We should change it for port mappings I think :)

maciaszczykm:
LGTM :)

src/app/frontend/deploy/deployfromsettings_controller.js
bryk: I use !! here to convert string to a boolean. I want to filter out empty strings and nulls. If I didn't use !! then the returned value wouldn't be {boolean}. With !! it is. 

The unknown is whether variable.name is always string. In this case I'm pretty sure it is. I'll double check however. 

In the case of port mappings, you're right. We should fix it. I'll create a bug.

bryk:
Verified. It works with 0. Merging, thanks for review :)

PR #339, opened 2016-02-08
Fix port range validation

### LINKED ISSUE #336: Port mapping input fails when 0 value is entered
When you enter `0` value for port mapping value it gets ignored. 

Thanks @maciaszczykm for finding this problem.

maciaszczykm:
Fixed #336 issue. Added range `1 - 65535` for port numbers.


bryk:
LGTM

PR #342, opened 2016-02-09
Branding Proposal

batikanu:
## GUI-Branding Proposal (DO NOT MERGE)

Currently we consider branding options by which we can patch the theme colors, styles, titles and logos with least effort for dashboard project. In this example gulp templates are used to patch the code by using the configuration values in productive build. Unfortunately, it doesn't work for development mode because the source code is served as it is without any conversion process. Do you have a better idea or any suggestion to support branding in dashboard project?

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -25,6 +25,7 @@ import gulpUseref from 'gulp-useref';
 import gulpRev from 'gulp-rev';
 import gulpRevReplace from 'gulp-rev-replace';
 import uglifySaveLicense from 'uglify-save-license';
+import template from 'gulp-template';
 import path from 'path';
 
 import conf from './conf';
@@ -75,6 +76,7 @@ gulp.task('build-frontend', ['assets', 'index:prod', 'clean-dist'], function() {
         conservativeCollapse: true,
       }))
       .pipe(htmlFilter.restore)
+      .pipe(template(conf.branding))
       .pipe(gulp.dest(conf.paths.distPublic));
 });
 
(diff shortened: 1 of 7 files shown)


bryk:
I'm worried that with this change the HTML and JS cannot be run directly, without preprocessing. 


What do you think about creating angular module + JS interface + angular service for theming. Clients that need theming would just ngInject them and use as needed. This is similar how angular-material theming works.

bryk:
@floreks @cheld @maciaszczykm

floreks:
> I'm worried that with this change the HTML and JS cannot be run directly, without preprocessing.

I've had the same feeling when I saw that.

> What do you think about creating angular module + JS interface + angular service for theming? Clients that need theming would just ngInject them and use as needed. This is similar how angular-material theming works.

Angular way of doing this sounds good to me. We should investigate this.

maciaszczykm:
If there is a possibility to make it work without pre-processing and it's not much more complicated I agree with @bryk and @floreks.

bryk:
Just to sketch out my idea: We have an angular service named, like, brandingService. The service can be configured in the application bootstrap phase (see $mdThemingProvider). Then, whenever you want to use a branded thing, you just ngInject the service and use it.

batikanu:
Sounds very good for theming. What about logos, titles and texts? Can we use some kind of element id or css classes to replace content of those elements in custom css stylesheets?

bryk:
You can use angular service for this, e.g., to change logo, you'd use:
```html
 <md-icon md-svg-icon="assets/images/{{::ctrl.branding.logo}}" class="kd-toolbar-logo"></md-icon>
```

Does this make sense?

batikanu:
Thank you! I think this sounds also good then it seems that index page has to have a controller. We will try it later.

bryk:
Awesome, if you have any questions, just ping me

Issue #334, opened 2016-02-04
Details page is really wide. Make it work on smaller screens.


bryk:
When you open the details page on, e.g., 13 inch laptop, it does not feel good. Even worse on mobile or tablet. Can we make something to fit all details on one screen? Maybe hide left nav into a tab?

bryk:
@floreks @maciaszczykm Opinions?

floreks:
I agree. We could just show small tab with arrow back and button that would expand the sidebar over the current view. Clicking outside of it would hide it again. What do you think?

maciaszczykm:
I think, that adding possibility to hide sidebar would be very nice feature. We can hide details into another tab and like @floreks mentioned, we can add `hide/show` switch. Second option however won't solve all problems on mobiles, because with expanded sidebar there will be no difference to current state.

bryk:
Re show/hide on the details on the left: the problem with this is that on the left hand side we'll likely have a navigation menu. So having two hide/show left menus would not look great. Opinions?

maciaszczykm:
@bryk Menu and the sidebar on the same side would be too much. I'm for changing sidebar into tab of existing `md-tabs` then.

maciaszczykm:
@bryk Should we change sidebar into new `md-tab`? I can create PR for this.

bryk:
Hmm.... I dont know. Do you want to just move details to a tab, or hide/show it based on screen width?

maciaszczykm:
Yes, moving it to tab in all cases seems to be not the best idea... Maybe it should be done only in specific cases. But it means **a lot** of refactoring. I'll play with it and check if I'll have any good idea.

floreks:
I was thinking about effect similar to this: When width is too small then sidebar can be shown like that with some button to show/hide.

http://gfycat.com/PracticalZanyEasternglasslizard

bryk:
@floreks Yeah, this is cool, but I'm pretty sure we should reserve this kind of behavior for application wide menu. Do you agree?

floreks:
That's a good point. I agree.

maciaszczykm:
I'm working on this. It'll take some time though.

bryk:
@maciaszczykm Can you describe the behavior you're working on? Or are you just experimenting?

maciaszczykm:
[...]

PR #343, opened 2016-02-09
Hide details page sidebar on narrow screens into tab

maciaszczykm:
Connected to #334 issue.

~~With these changes sidebar will be hidden when page inner width is below `MIN_CONTENT_WIDTH = 800`.~~

~~@bryk `MIN_CONTENT_WIDTH` value can be changed. I saw, that in `replicationcontrollerlistcontainer.js ` there is ineresting usage of `mdMedia` service. Perhaps it should be used also here?~~

Checklist:
- [x] updated UI and created new directive for sidebar content
- [x] mdMedia usage
- [x] updated tests

Screenshots:
image
image

Please do not merge before all above points will be checked. I've created this pull request now to start discussion about possible improvements. If you have any sugesstions please let me know.

### DIFF
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
@@ -15,10 +15,12 @@ limitations under the License.
 -->
 
 <md-content layout="row" layout-fill flex>
-  <div class="kd-replicationcontrollerdetail-sidebar" layout="column" flex="noshrink">
+  <div class="kd-replicationcontrollerdetail-sidebar" layout="column" flex="noshrink"
+       ng-show="ctrl.isSidebarVisible()">
     <div class="kd-replicationcontrollerdetail-sidebar-header-wrapper kd-replicationcontrollerdetail-sidebar-item"
         flex="nogrow">
-      <div class="kd-replicationcontrollerdetail-sidebar-header" layout-align="start center" layout="row">
+      <div class="kd-replicationcontrollerdetail-sidebar-header" layout-align="start center"
+           layout="row">
         <md-button class="md-icon-button" ui-sref="replicationcontrollers">
           <md-icon md-font-library="material-icons">arrow_back</md-icon>
         </md-button>
@@ -27,103 +29,20 @@ limitations under the License.
         </h1>
       </div>
     </div>
-    <div flex layout="column" class="kd-replicationcontrollerdetail-sidebar-item">
-      <div layout="row" flex="nogrow">
-        <md-button class="md-primary" ng-click="ctrl.handleDeleteReplicationControllerDialog()">
-          <md-icon class="material-icons">delete</md-icon>
-          DELETE REPLICATION CONTROLLER
-        </md-button>
-      </div>
-      <span class="kd-replicationcontrollerdetail-sidebar-title kd-replicationcontrollerdetail-sidebar-info">
-        Replication controller
-      </span>
-      <div flex layout="column" class="kd-replicationcontrollerdetail-sidebar-info">
-        <span class="kd-replicationcontrollerdetail-sidebar-line">Name</span>
-        <span class="kd-replicationcontrollerdetail-sidebar-subline">
-          {{::ctrl.replicationControllerDetail.name}}
-        </span>
-        <span class="kd-replicationcontrollerdetail-sidebar-line">Namespace</span>
-        <span class="kd-replicationcontrollerdetail-sidebar-subline">
-          {{::ctrl.replicationControllerDetail.namespace}}
-        </span>
-        <div layout="row" layout-align="start end">
-          <div flex="nogrow">
-            <span class="kd-replicationcontrollerdetail-sidebar-line">Pods</span>
-            <span class="kd-replicationcontrollerdetail-sidebar-subline" ng-if="!ctrl.areDesiredPodsRunning()">
-              {{::ctrl.replicationControllerDetail.podInfo.current}} created,
-              {{::ctrl.replicationControllerDetail.podInfo.desired}} desired
-            </span>
-            <span class="kd-replicationcontrollerdetail-sidebar-subline" ng-if="ctrl.areDesiredPodsRunning()">
-              {{::ctrl.replicationControllerDetail.podInfo.running}} running
-            </span>
-          </div>
-          <div flex="nogrow">
-            <md-button  class="md-icon-button md-primary">
-              <md-icon class="material-icons"
-                  ng-click="ctrl.handleUpdateReplicasDialog()">
-                mode_edit
-              </md-icon>
-            </md-button>
-          </div>
-        </div>
-        <div ng-if="!ctrl.areDesiredPodsRunning()">
-          <span class="kd-replicationcontrollerdetail-sidebar-line">Pods status</span>
-          <span class="kd-replicationcontrollerdetail-sidebar-subline">
-            <span ng-if="::ctrl.replicationControllerDetail.podInfo.pending" class="kd-comma-separated-item">
-              {{::ctrl.replicationControllerDetail.podInfo.pending}} pending<!-- Collapse whitespace
-            --></span>
-            <span ng-if="::ctrl.replicationControllerDetail.podInfo.failed" class="kd-comma-separated-item">
-              {{::ctrl.replicationControllerDetail.podInfo.failed}} failed<!-- Collapse whitespace
-            --></span>
-            <span ng-if="::ctrl.replicationControllerDetail.podInfo.running" class="kd-comma-separated-item">
-              {{::ctrl.replicationControllerDetail.podInfo.running}} running
-            </span>
-          </span>
-        </div>
-        <span class="kd-replicationcontrollerdetail-sidebar-line">Label selector</span>
-        <span class="kd-replicationcontrollerdetail-sidebar-subline kd-has-labels">
-          <kd-labels labels="::ctrl.replicationControllerDetail.labelSelector"></kd-labels>
-        </span>
-        <span class="kd-replicationcontrollerdetail-sidebar-line">Labels</span>
-        <span class="kd-replicationcontrollerdetail-sidebar-subline kd-has-labels">
-          <kd-labels labels="::ctrl.replicationControllerDetail.labels"></kd-labels>
-        </span>
-        <span class="kd-replicationcontrollerdetail-sidebar-line">Images</span>
-        <span class="kd-replicationcontrollerdetail-sidebar-subline"
-              ng-repeat="image in ctrl.replicationControllerDetail.containerImages">
-          {{image}}
-        </span>
-        <div ng-if="::ctrl.replicationControllerDetail.services" layout="column">
-          <span class="kd-replicationcontrollerdetail-sidebar-title">Services</span>
-          <div ng-repeat="service in ::ctrl.replicationControllerDetail.services">
-            <span class="kd-replicationcontrollerdetail-sidebar-line">Name</span>
-            <span class="kd-replicationcontrollerdetail-sidebar-subline">
-              {{::service.name}}
-            </span>
-            <span class="kd-replicationcontrollerdetail-sidebar-line">Label selector</span>
-            <span class="kd-replicationcontrollerdetail-sidebar-subline kd-has-labels">
-              <kd-labels labels="::service.selector"></kd-labels>
-            </span>
-            <span class="kd-replicationcontrollerdetail-sidebar-line">Internal endpoint</span>
-            <div class="kd-replicationcontrollerdetail-sidebar-subline">
-              <div ng-if="::service.internalEndpoint">
-                <kd-service-endpoint endpoint="::service.internalEndpoint"></kd-service-endpoint>
-              </div>
-              <span ng-hide="service.internalEndpoint">none</span>
-            </div>
-            <span class="kd-replicationcontrollerdetail-sidebar-line">External endpoint</span>
-            <div class="kd-replicationcontrollerdetail-sidebar-subline">
-              <div ng-repeat="endpoint in service.externalEndpoints">
-                <kd-service-endpoint endpoint="::endpoint"></kd-service-endpoint>
-              </div>
-              <span class="kd-replicationcontrollerdetail-sidebar-subline"
-                  ng-hide="::service.externalEndpoints">
-                none
-              </span>
-            </div>
-          </div>
-        </div>
-      </div>
+    <kd-replication-controller-info details="::ctrl.replicationControllerDetail">
+    </kd-replication-controller-info>
+  </div>
+  <div flex>
+  <div class="kd-replicationcontrollerdetail-sidebar-header-wrapper kd-replicationcontrollerdetail-sidebar-item"
+       flex="nogrow" ng-show="!ctrl.isSidebarVisible()">
+    <div class="kd-replicationcontrollerdetail-sidebar-header" layout-align="start center"
+         layout="row">
+      <md-button class="md-icon-button" ui-sref="replicationcontrollers">
+        <md-icon md-font-library="material-icons">arrow_back</md-icon>
+      </md-button>
+      <h1 flex class="md-title kd-replicationcontrollerdetail-app-name">
+        {{::ctrl.replicationControllerDetail.name}}
+      </h1>
     </div>
   </div>
   <md-tabs flex="grow" md-border-bottom md-dynamic-height>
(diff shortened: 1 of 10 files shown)


bryk:
> @bryk MIN_CONTENT_WIDTH value can be changed. I saw, that in replicationcontrollerlistcontainer.js there is ineresting usage of mdMedia service. Perhaps it should be used also here?

Yes, use mdMedia for that kind of bahavior. It is documented on angular material site.

maciaszczykm:
>@bryk MIN_CONTENT_WIDTH value can be changed. I saw, that in replicationcontrollerlistcontainer.js there is ineresting usage of mdMedia service. Perhaps it should be used also here?

>Yes, use mdMedia for that kind of bahavior. It is documented on angular material site.

Yeah, it's very simple and works exactly as we need :)

bryk:
Also, please elaborate on whether details tab should be first/last/middle.

maciaszczykm:
> Also, please elaborate on whether details tab should be first/last/middle.

At the moment it's last tab, because there is problem with tab order when it's different. I think goal should be making it first tab and solve problem with tab order when resizing browser window.

bryk:
What's the problem?

maciaszczykm:
> What's the problem?

If we make details tab first - when window size is reduced and first visible tab is active (for example pods), then it will be switched to details tab (which is really first tab). I think there is some kind of pointer on currently active index and it's not user friendly to switch tab like this.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
```diff
@@ -108,6 +106,18 @@ export default class ReplicationControllerDetailController {
...
+    /**
+     * If true then sidebar will be displayed, otherwise it won't.
+     * @export {boolean}
+     */
+    this.displaySidebar = this.mdMedia('gt-sm');
+
+    /**
+     * Binding resize listener, which switches displaySidebar boolean variable.
+     */
+    angular.element(this.window)
```
batikanu: Why don't you prefer doing all those element and resize staff in the directive link function?

maciaszczykm: Could you explain a little bit more, because I don't understand? Are you saing, that `this.displaySidebar = this.mdMedia('gt-sm');` and `angular.element(this.window).bind('resize', () => { this.displaySidebar = this.mdMedia('gt-sm');});` should be moved to directive?

maciaszczykm:
Updated tests, build is passing right now. Problem with logo placement seems to not be connected to this pull request, so I've created the #352 issue. 

Please give any feedback about details tab placement and content (should we switch to horizontal layout?).

PTAL

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
batikanu: Yes like here https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/deploy/filereader_directive.js#L32
But it requires also to encapsulate md-tab in your directive to make it work.

maciaszczykm: It would at least double number of lines in the code, so is it really needed? I mean, that I would rather keep it simple. Moving `md-tab` into my new directive is not possible, because it's used in two places and one of them is sidebar. We would need to create new directive for it I think, if I understood correctly.

In your case you had pretty much logic, and here I'm using only `mdMedia`.

maciaszczykm: Solved it in even better way. There is no need for additional logic in controller, because we can use `mdMedia` directly in HTML.

batikanu: Yes I see the duplication. It looks much better I think now.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
```diff
@@ -39,13 +38,17 @@ export default class ReplicationControllerDetailController {
...
-   * kdReplicationControllerService
    * @ngInject
    */
   constructor(
-      $mdDialog, $stateParams, $state, $resource, $log, replicationControllerDetail,
-      replicationControllerEvents, kdReplicationControllerService) {
+      $window, $mdMedia, $mdDialog, $stateParams, $state, $resource, $log,
+      replicationControllerDetail, replicationControllerEvents) {
+    /** @export {!angular.$window} */
+    this.window = $window;
```
batikanu: Do you still use it?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
batikanu: One step further might be to put ctrl.mdMedia("gt-sm") into a method in controller like 
hasMinimumSize() {
    return this.mdMedia('gt-sm');
}
Because the logic is doubled in the html code.

maciaszczykm: @batikanu I think we shouldn't create method just to call another method, if there would be more logic, but only thing in common is `mdMedia` size. We need moethod call anyways. Moreover, we should test all of ours methods, and doing it for `hasMinimumSize()` has not much sense.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
maciaszczykm: Good point, I'll remove it. It's old code leftover.

maciaszczykm:
PTAL

bryk:
There is no back button when the screen goes small.

maciaszczykm:
@bryk Where should we place it? At the moment clicking Kubernetes logo redirects to replication controller list page.

bryk:
Hmm... How about having a bar between toolbar and tabs. The bar would have RC name and back link.

maciaszczykm:
@bryk Okay, good idea.

maciaszczykm:
Done. PTAL

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+            mode_edit
+          </md-icon>
+        </md-button>
+      </div>
+    </div>
+    <div ng-if="!infoCtrl.areDesiredPodsRunning()">
+      <span class="kd-replicationcontrollerinfo-line">Pods status</span>
+          <span class="kd-replicationcontrollerinfo-subline">
+            <span ng-if="::infoCtrl.details.podInfo.pending" class="kd-comma-separated-item">
+              {{::infoCtrl.details.podInfo.pending}} pending<!-- Collapse whitespace-->
```
bryk: This does not collapse whitespace. You need to end comment on the next element.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+            </span>
+            <span ng-if="::infoCtrl.details.podInfo.failed" class="kd-comma-separated-item">
+              {{::infoCtrl.details.podInfo.failed}} failed<!-- Collapse whitespace-->
```
bryk: This does not collapse whitespace. You need to end comment on the next element.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+            </span>
+            <span ng-if="::infoCtrl.details.podInfo.running" class="kd-comma-separated-item">
+              {{::infoCtrl.details.podInfo.running}} running
+            </span>
+          </span>
+    </div>
+    <span class="kd-replicationcontrollerinfo-line">Label selector</span>
+        <span class="kd-replicationcontrollerinfo-subline kd-has-labels">
```
bryk: Formatting here?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+          <kd-labels labels="::infoCtrl.details.labelSelector"></kd-labels>
+        </span>
+    <span class="kd-replicationcontrollerinfo-line">Labels</span>
+        <span class="kd-replicationcontrollerinfo-subline kd-has-labels">
```
bryk: Formatting here?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+          <kd-labels labels="::infoCtrl.details.labels"></kd-labels>
+        </span>
+    <span class="kd-replicationcontrollerinfo-line">Images</span>
+        <span class="kd-replicationcontrollerinfo-subline"
```
bryk: Formatting here?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+              ng-repeat="image in infoCtrl.details.containerImages">
+          {{image}}
+        </span>
+  </div>
+  <div flex layout="column" class="kd-replicationcontrollerinfo-info">
+    <div ng-if="::infoCtrl.details.services" layout="column">
+      <span class="kd-replicationcontrollerinfo-title">Services</span>
+      <div ng-repeat="service in ::infoCtrl.details.services">
+        <span class="kd-replicationcontrollerinfo-line">Name</span>
+            <span class="kd-replicationcontrollerinfo-subline">
```
bryk: Formatting here?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
```diff
@@ -0,0 +1,119 @@
...
+            </kd-service-endpoint>
+          </div>
+          <span ng-hide="service.internalEndpoint">none</span>
+        </div>
+        <span class="kd-replicationcontrollerinfo-line">External endpoint</span>
+        <div class="kd-replicationcontrollerinfo-subline">
+          <div ng-repeat="endpoint in service.externalEndpoints">
+            <kd-service-endpoint endpoint="::endpoint"></kd-service-endpoint>
+          </div>
+              <span class="kd-replicationcontrollerinfo-subline"
```
bryk: Formatting here?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo_controller.js
```diff
@@ -0,0 +1,95 @@
...
+  stateName as replicationcontrollers,
+} from 'replicationcontrollerlist/replicationcontrollerlist_state';
+
+/**
+ * Controller for the replication controller info directive.
+ * @final
+ */
+export default class ReplicationControllerInfoController {
+  /**
+   *
```
bryk: Delete empty line?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo_controller.js
```diff
@@ -0,0 +1,95 @@
...
+  handleDeleteReplicationControllerDialog() {
+    this.kdReplicationControllerService_
+        .showDeleteDialog(this.stateParams_.namespace, this.stateParams_.replicationController)
+        .then(this.onReplicationControllerDeleteSuccess_.bind(this));
+  }
+
+  /**
+   * Callbacks used after clicking dialog confirmation button in order to delete replication
+   * controller
+   * or log unsuccessful operation error.
```
bryk: Format?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo_controller.js
```diff
@@ -0,0 +1,95 @@
...
+    this.kdReplicationControllerService_ = kdReplicationControllerService;
+
+    /** @private {!ui.router.$state} */
+    this.state_ = $state;
+
+    /** @private {!angular.$log} */
+    this.log_ = $log;
+
+    /**
+     * Replication controller details, contains all the information about it (two-way binding).
```
bryk: Is this two way? I'd say this is "Initialized from the scope" and that's all. You don publish any changes back, don't you?

bryk:
Can you make the incremental coverage 100%?

bryk:
A few style comments. PTAL

maciaszczykm:
> Can you make the incremental coverage 100%?

To do it I would need to test following function:

```
onReplicationControllerDeleteSuccess_() {
    this.log_.info('Replication controller successfully deleted.');
    this.state_.go(replicationcontrollers);
  }
```

https://codecov.io/github/kubernetes/dashboard/compare/8b856f688e5150dd4c9522d0c6537e170f9dd7b1...0bb38d1a4e00b5e3d758b1c268637880a13c5666

Is it needed?

I've applied all formatting changes. PTAL

bryk:
Yeah, you should test that when a replication controller is deleted, then state transitions to `replicationcontrollers` :)

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
```diff
@@ -15,10 +15,12 @@
 -->
 
 <md-content layout="row" layout-fill flex>
-  <div class="kd-replicationcontrollerdetail-sidebar" layout="column" flex="noshrink">
+  <div class="kd-replicationcontrollerdetail-sidebar" layout="column" flex="noshrink"
+       ng-show="ctrl.mdMedia('gt-sm')">
```
bryk: Btw, move all `ctrl.mdMedia('gt-sm')` to the ctrl. Otherwise this will most likely be broken when somebody changes the line here, but not 37.

maciaszczykm:
Done! PTAL

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
```diff
@@ -27,103 +29,20 @@ <h1 flex class="md-title kd-replicationcontrollerdetail-app-name">
...
-                none
-              </span>
-            </div>
-          </div>
-        </div>
-      </div>
+    <kd-replication-controller-info details="::ctrl.replicationControllerDetail">
+    </kd-replication-controller-info>
+  </div>
+  <div flex>
```
bryk: Indent next line.

bryk: Btw, can you add html formatter to the project? 
https://www.npmjs.com/package/gulp-prettify

This would save us time on formatting disputes. To me, it already works awesome for JavaScript, so it should for HTML.

maciaszczykm: @bryk Good idea, we should have it. I'm using IntelliJ, but it's not always working as expected.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
```diff
@@ -353,16 +267,22 @@ <h1 flex class="md-title kd-replicationcontrollerdetail-app-name">
...
-
-	<div class="kd-replicationcontrollerdetail-no-events" ng-if="!ctrl.hasEvents()">
-          <md-icon class="material-icons kd-replicationcontrollerdetail-no-events-icon">info_outline</md-icon>
-          <div class="kd-replicationcontrollerdetail-no-events-text">No events were found</div>
-          <span class="kd-replicationcontrollerdetail-no-events-info">
-            There are no events to display. It's possible that all of them have expired.
-            <a href="">Learn more</a>
-          </span>
-	</div>
+	        <div class="kd-replicationcontrollerdetail-no-events" ng-if="!ctrl.hasEvents()">
```
bryk: Yeah, IntelliJ has some problems with formatting. Please do a double check.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
```diff
@@ -111,6 +107,12 @@ export default class ReplicationControllerDetailController {
   }
 
   /**
+   * Returns true if sidebar is visible, false if it is hidden.
+   * @returns {boolean}
```
bryk: `@export` needed. Please test on prod.

bryk:
Last comments and let's merge. PTAL.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
maciaszczykm: I'm always forgetting about it, sorry!

maciaszczykm:
Added missing `@export` and tested on `serve:prod`. I'll reformat all HTML code in my next pull request, which will resolve #371 issue.

PTAL

PR #344, opened 2016-02-09
Fix code to make it run and test on IE/FF/Chrome

bryk:
This is in preparation for brand-new CI setup where we run karma and
integration test against all major browsers (IE/FF/Chrome). With this
change the tests pass.

1. for ... of does not work quite well with IE + Karma with browserify. Removed it.
2. Array.prototype.find is not present on IE
3. IE has some errors on the console from WebDriver framework.

### DIFF
--- a/src/app/frontend/logs/logstoolbar/logstoolbar_controller.js
+++ b/src/app/frontend/logs/logstoolbar/logstoolbar_controller.js
@@ -130,7 +130,14 @@ export default class LogsToolbarController {
    * @return {!backendApi.ReplicationControllerPodWithContainers|undefined}
    * @private
    */
-  findPodByName_(array, name) { return array.find((element) => element.name === name); }
+  findPodByName_(array, name) {
+    for (let i = 0; i < array.length; i++) {
+      if (array[i].name === name) {
+        return array[i];
+      }
+    }
+    return undefined;
+  }
 
   /**
    * Find Container by name.
@@ -141,7 +148,13 @@ export default class LogsToolbarController {
    * @private
    */
   initializeContainer_(array, name) {
-    let container = array.find((element) => element.name === name);
+    let container = undefined;
+    for (let i = 0; i < array.length; i++) {
+      if (array[i].name === name) {
+        container = array[i];
+        break;
+      }
+    }
     if (!container) {
       container = array[0];
     }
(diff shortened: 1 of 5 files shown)


bryk:
@maciaszczykm Can you review?

maciaszczykm:
I'll do it tommorow.

src/app/frontend/logs/logstoolbar/logstoolbar_controller.js
```diff
@@ -130,7 +130,14 @@ export default class LogsToolbarController {
    * @return {!backendApi.ReplicationControllerPodWithContainers|undefined}
    * @private
    */
-  findPodByName_(array, name) { return array.find((element) => element.name === name); }
+  findPodByName_(array, name) {
```
maciaszczykm: What about creating some set of common JavaScript functions like find etc.?

maciaszczykm:
One question. LGTM

src/app/frontend/logs/logstoolbar/logstoolbar_controller.js
floreks: Can we maybe extend array object prototype with this functions manually? It would allow use to use it in the old way.

bryk: Modifying built in classes is a controversial topic, and, personally, I'd avoid it at any cost. You can find many articles on it on the net, e.g., https://www.nczonline.net/blog/2010/03/02/maintainable-javascript-dont-modify-objects-you-down-own/

If this incompatibility hurts us, we can fall back to closure library wrappers (e.g., goog.array.find) Or use closure compiler to polyfil missing function (e.g., Array.prototype.find). But, if possible, I'd like us to run with Babel on all major browsers (ie, chrome, ff).
What do you think?

bryk:
Thanks for the review. Will merge soon.

src/app/frontend/logs/logstoolbar/logstoolbar_controller.js
floreks: That's a very nice article. I've read it and it has very good point. Overriding is dangerous. It's safer to rely on some 3rd party libraries. Polyfill with closure compiler is also good idea. We should think of that.

PR #354, opened 2016-02-10
Fix heapster client config to make it work when master moves

bryk:
Previously the heapster client would always talk to the same IP, even
when the master moved somewhere. Now it reuses apiserver client and
abstracts service proxy. This makes is always work when the apiserver
client is correct.

### DIFF
--- a/src/app/backend/heapsterclient.go
+++ b/src/app/backend/heapsterclient.go
@@ -15,7 +15,6 @@
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


bryk:
@maciaszczykm Can you review?

maciaszczykm:
@bryk Yes.

maciaszczykm:
LGTM

Issue #352, opened 2016-02-10
Logo placement bug


maciaszczykm:
We've an issue with Kubernetes logo placement if page is too narrow.

Standard case, where everything is okay (window is 1200 px wide):

![image](https://cloud.githubusercontent.com/assets/2823399/12948706/ff5fae78-d003-11e5-8bb1-b66cdb8ae300.png)

Bug (window is 800 px wide):

![image](https://cloud.githubusercontent.com/assets/2823399/12948725/349a01ba-d004-11e5-93d5-7c67e2658f63.png)

I can work on it later, but not yet. Feel free to take it.

bryk:
Thanks for reporting. Just comment here if you start working on this.

digitalfishpond:
Working now

digitalfishpond:
It comes from Chrome.css > .kd-toolbar-tools  This class is only used for this logo. Is there any reason anyone knows about why it's set to "0 auto" rather than just "auto" (which solves the issue)?

bryk:
I probably did the `0 auto`. If just `auto` works, then perfect, go ahead.

PR #362, opened 2016-02-11
fix logo position

digitalfishpond:
#352 fixed


bryk:
Thanks for the fix! LGTM, merging.

Issue #348, opened 2016-02-10
Add namespace selector for replication-controller-list page


cheld:
When deploying an app, it is possible to select/create a namespace for it. However, in the replication-controller-list page it is not possible to utilize this information. Also, by default all the add-on containers are shown. IMHO, this is not so nice. As a user I would say I do not want the add on containers mixed with my application containers.

I would suggest to add some kind of namespace selector (e.g. combo-box) in the header that shows the default-namespace by default.

```
-------------------------------
| default                v    |
-------------------------------
 my-namespace
 kube-system (?)
 all
```

There might be better ways. Piotr mentioned that the mockups contain a suggested namespace handling concept. However, I could not find it.

bryk:
Now I remember. This was in some of Ken's early mocks. Take a look: 
![namespaces](https://cloud.githubusercontent.com/assets/1159999/12946666/5d6bf0f2-cff6-11e5-9f43-ef719d29ec8d.png)

cheld:
Nice.

From user point of view, I think in most cases I only want to see one namespace at a time.

So, my suggestion:
- add a selector at the top
- show the cards like in the mockup if 'all' is selected

bryk:
@floreks @maciaszczykm Inviting to discussion re: "All namespaces or namespace selector".

maciaszczykm:
This early mock looks nice, but I agree, that we should be available to filter displayed namespaces. I would list them all and create checklist:

- [x] default
- [x] namespace-1
- [ ] namespace-2

In this case we should display replication controllers only from default and namespace-1 namespaces.

Moreover, some switch to hide groups (current state) would be nice.

floreks:
I've have a little different idea of handling that. How about making expandable rows that will hold card containers. This represents the idea: http://jsfiddle.net/Pixic/6Texj/

On click all other expanded namespaces would be collapsed again. What do you think?

bryk:
@floreks So you say to show all namespaces, but make the containers that contain them collapsable?

floreks:
Yes, then on click we can async load data related to namespace and collapse others. One namespace would be shown at a time. Normally we would start with default namespace.

cheld:
Hmm...the namespace concept is a bit fuzzy. So lets define some scenarios and use cases first.

If I remember correctly, the namespaces have been requested by RedHat for OpenShift. I have never checked myself, but I assume that OpenShift uses namespaces to separate applications
 
=> Use Case 1: one namespace for each application

[...]

PR #363, opened 2016-02-11
Add namespace restriction option

titilambert:
Hello,
I just add a new option to dashboard backend to be able to limit queries to a specified namespace.
With this option we can provide a pod of kubernetes dashboard with all new k8s namespace.
Kuberntes Dashboard can be the user GUI for each namespaces.

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/363)

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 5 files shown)


bryk:
Hey @titilambert, thanks for the contribution! Hold on a while with the implementation. We're discussing this issue at the moment and are thinking about best solution. Please join the discussion at https://github.com/kubernetes/dashboard/issues/348

titilambert:
@bryk Hello ! Any news on this one ?

bryk:
Yes, I'm here - just returned from vacation.

bryk:
First thought: cannot you achieve this by just configuring your installation? I.e., when you deploy Dashboard UI to namespace `foo`, then it has access only to resources in namespace `foo`? This would be much better solution that doing this on the UI side. 
What do you think?

titilambert:
> when you deploy Dashboard UI to namespace foo, then it has access only to resources in namespace foo ?

How can I do that ? (Because the objective of this PR is to to that !)
Maybe I missed something, but, for now, It has access to resources in all namespaces.
Also, I want to limit users to be able to edit/delete/create resources in other namespace.

bryk:
@titilambert Yes, you can definitely do this. It is much better solution than baking it into UI. Check out this: http://kubernetes.io/docs/admin/authorization/. You can basically say that pods (or only the UI) in namespace `foo` can only access resources in namespace `foo`. 

Does this solve your problem?

@kubernetes/sig-auth

erictune:
For some use cases, @bryk's solution will work.  However, it is definitely possible to write authorization policies that are incompatible with the current Dashboard design (in the sense that the Dashboard will allow users to do things they could not do otherwise).

erictune:
https://github.com/kubernetes/kubernetes/issues/23201

bryk:
@titilambert I think this problem is solved. Please reopen if you need any help.

PR #365, opened 2016-02-11
build: include URL scheme in host configuration

joeatwork:
This change allows passing empty strings to the dashboard backend
process, which in turn allows developers to run the dashboard using a
heapster service proxy.

I'm proposing it so that I can more easily develop against Heapster while running a dashboard that isn't inside of my cluster, bu setting heapsterServerHost to the empty string in my build.


bryk:
LGTM, thanks for this! :)

PR #369, opened 2016-02-12
Add external endpoint for cluster IP services

maciaszczykm:
Connected to discussion from pull request #320 and issue #313:

> I am not fully happy with leaving the link empty on local cluster. What about putting "localhost:nodeport" into the link as a fallback if no IP is available? First, I would help us for writing click-tests, because using external link would make the test work on local cluster (CI) and real cluster. Second, the link is an important UI feature, it shouldn't be missing on local cluster

I've set `localhost:targetport` as a external endpoint for cluster IP services. I've used target port instead of nodeport, because it was always 0 during tests. Is it correct?

Please take a look and say your opinion about it. Change is not big, so I can quickly adapt it (for example use node port instead of target port).

CC @cheld @bryk

### DIFF
--- a/src/app/backend/replicationcontrollerdetail.go
+++ b/src/app/backend/replicationcontrollerdetail.go
@@ -330,9 +330,7 @@ func getExternalEndpoints(replicationController api.ReplicationController, pods 
 
 	if service.Spec.Type == api.ServiceTypeNodePort {
 		externalEndpoints = getNodePortEndpoints(replicationControllerPods, service, getNodeFn)
-	}
-
-	if service.Spec.Type == api.ServiceTypeLoadBalancer {
+	} else if service.Spec.Type == api.ServiceTypeLoadBalancer {
 		for _, ingress := range service.Status.LoadBalancer.Ingress {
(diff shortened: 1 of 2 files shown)


bryk:
I like the size and scope of this PR, btw. "one thing at a time".

cheld:
I can't see a link...are you sure it is working?

bryk:
Wasn't this supposed to return localhost addresses only for nodeports/loadbalancers that fail to resolve any other meaningful address. Am I right?

CC @cheld

cheld:
Yes exactly..I will try it once again.

maciaszczykm:
Okay, so this need to be refactored. I'll let you know when it will be ready for review.

maciaszczykm:
@cheld @bryk Updated. PTAL

cheld:
I tested for local cluster. It works exactly as intended. I have not tested on vagrant or gce, yet.

cheld:
LGTM

maciaszczykm:
@bryk PTAL

bryk:
LGTM :)

PR #370, opened 2016-02-12
Scope current API as v1 API.

bryk:
This is in preparation for release. After we do this, incompatible
changes to the api will be forbidden.

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -61,7 +61,7 @@ func CreateHttpApiHandler(client *client.Client, heapsterClient HeapsterClient) 
 
 	deployWs := new(restful.WebService)
 	deployWs.Filter(wsLogger)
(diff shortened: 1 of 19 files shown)


bryk:
@floreks Can you take a look?

floreks:
LGTM

bryk:
Travis is down. Merging without it.

joeatwork:
Ah! I'm a bit late for this - I was hoping to get metrics as time series in before the API freeze.

bryk:
Don't worry. This is presentation layer API. We scoped it with v1 to avoid future conflicts. We plan to increment api versions fast and often - whenever this is needed. Just think of it as a way to make sure the UI works during a rolling update.

bryk:
Just let me know if you need any help :)

joeatwork:
Thanks! I'll ask questions as they come up, and I'll make sure not to break any existing APIs.

PR #375, opened 2016-02-15
Fixed logs menu api call

floreks:
Typo fix. Logs menu was not working due to this wrong api call path.


bryk:
LGTM

Sorry for my mistake. We should have a way to catch such issues. Maybe integration tests?

floreks:
I was also looking at that and didn't catch this typo :) Integration tests would for sure catch that. I guess even unit tests with httpBackend would catch this kind of mistakes.

PR #376, opened 2016-02-15
manual test for deploy-from-file

cheld:
I created a test sheet for the deploy from file dialog. This test-sheet is rather simple. It is intended for initial discussion and to serve as a template for further test-sheets. I simplified the document as much as possible.

TBD: 
* which format should the test-sheet have? xls, openoffice, google docs
* where to store the documents?

@maciaszczykm please have a first review

### DIFF
--- a/docs/test/deploy-from-file.md
+++ b/docs/test/deploy-from-file.md
@@ -1,0 +1,29 @@
+# Deploy from file
+
+## Prerequisites
+Open deploy dialog (+) and select "Upload a YAML or JSON file"
+
+## Expected result at execution
+* In positive case the "Deploy" button must lead to the creation of a running application
+* Validation errors affecting particular fields must be displayed beneath these fields. In case the deployment  fails due to an exception, then a pop dialog is shown
+
+## Specific factors
+
+| Factor       | 1                            | 2                      | 3                                           | 4                | 5            | 6                             | 7                       | Comment |
+|--------------|------------------------------|------------------------|---------------------------------------------|------------------|--------------|-------------------------------|-------------------------|---------|
+| File name    | Empty                        | Short                  | Max-length                                  |                  |              |                               |                         |         |
+| File content | Correct YAML, e.g. Guestbook | Correct JSON           | Binary, e.g. PNG                            | Validation error | Syntax error | no RC, e.g. Only pod, svc,... | Special chars, e.g. XSS |         |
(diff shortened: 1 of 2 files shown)


cheld:
Any factor missing?

bryk:
Hmm.... Maybe we can use plain old markdown? You can convert your excel to it: https://github.com/mplewis/csvtomd

maciaszczykm:
What I like about this file is, that it's simple and testing it should take minutes/hours instead of days. It should also help us to not forget about any of standard use cases. 

Regarding format I would also try to use markdown, because we can view all files directly on GitHub. 

Regarding test cases, is there any `?` tooltip on deploy page? It works on mouse over, there is no need to click. Moreover, at the moment we do not have any success message for validation. Proof, that deploy succeded is presence of application on replication controller list I think.

cheld:
Ok, I convert to markdown.

About factors:
The last four lines are more like reminders, not real factors. These are the same for all sheets. I suggest to copy and paste them to all sheets. I change the wording a bit to make clear.

We don't have a success message - yes, but this is a bug :) 

> Proof, that deploy succeeded is presence of application on replication controller list I think.

What about if the yaml just contains a namespace? In this case no feedback at all...

maciaszczykm:
> What about if the yaml just contains a namespace? In this case no feedback at all...

That's true. It seems to be another bug.

cheld:
PTAL

maciaszczykm:
LGTM, markdown will be good when it will come to merging these files.

cheld:
Right. I was wondering if there is a good way to create big tables in md. Any idea?

bryk:
Re editing markdown tables: http://www.tablesgenerator.com/markdown_tables

bryk:
Can you remove trailing empty columns? You can use http://www.tablesgenerator.com/markdown_tables#

bryk:
Other than that, LGTM :) Very nice

cheld:
1. I cleaned up the tables
2. I moved the documents from src/test/manual to docs/test

PTAL

maciaszczykm:
@cheld It seems, that trailing empty columns are still there.

cheld:
Ah, now I understand...ok

cheld:
Sorry, once again. I left the comments field for the special factors, because we used this field frequently in the past. So everyone is aware of the field and feels tempted to enter something.

cheld:
PTAL

bryk:
LGTM :)

Issue #381, opened 2016-02-16
Crash loudly when Dashboard cannot connect to master on startup


bryk:
We should crash loudly when Dashboard cannot connect to the master on startup, instead of continuing silently.

PR #383, opened 2016-02-16
Fail and quit the binary when cannot connect to K8s master on startup

bryk:
This is to prevent false positives when admins think that the UI works
correctly, while it is not.

Fixes: #381

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -43,7 +43,7 @@ func main() {
 
 	apiserverClient, err := CreateApiserverClient(*argApiserverHost, new(ClientFactoryImpl))
 	if err != nil {
-		log.Print(err)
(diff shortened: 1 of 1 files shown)


bryk:
@floreks Can you review?

floreks:
I'm trying to reproduce a case where it will fail and I can't.

I've killed my local cluster and still `apiserverClient` is created on `localhost:8080` even when it is not reachable. I've looked shortly in the code and can't find place where it tries to ping master to check if it's actually alive.

What's the expected behavior here?

bryk:
It'll fail on `build/hyperkube.sh` cluster. Try it out. Basically, it fails, e.g., when is on a cluster with no serviceaccounts.

floreks:
So basically this will work only for in-cluster dashboard setup.

LGTM

bryk:
Yes. We can later improve it further. To test the connection and get, e.g., apiversion. If this fails, then crash.

PR #386, opened 2016-02-17
API endpoint for CPU and Memory usage history

joeatwork:
This pull request proposes an additive change to the front end API, that will allow us to present sparkline-style mini-charts along with CPU and Memory usage statistics.  It includes a change to the now-versioned API. The change won't break clients that accept and/or ignore new fields in expected JSON objects (like go or browser JSON parsers), but please comment if you're worried that these changes might break other clients.

### DIFF
--- a/src/app/backend/replicationcontrollerpodsmetrics.go
+++ b/src/app/backend/replicationcontrollerpodsmetrics.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	heapster "k8s.io/heapster/api/v1/types"
 	"k8s.io/kubernetes/pkg/api"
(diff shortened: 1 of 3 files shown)


bryk:
> It includes a change to the now-versioned API. The change won't break clients that accept and/or ignore new fields in expected JSON objects (like go or browser JSON parsers), but please comment if you're worried that these changes might break other clients.


Exactly. This is a versioned private presentation layer API for the UI. The requirement we have is that N versions of the API and its client should be compatible on the wire. So adding new fields is thing we can be doing. Removing/renaming likely not. N will be defined later in the project.

bryk:
@maciaszczykm Can you review?

maciaszczykm:
@bryk Yes.

maciaszczykm:
Code looks very good and works on `serve:prod` without any issues. I'm already looking forward to merge of sparklines! LGTM

bryk:
LGTM :) You need to rebase so that we can merge this.

joeatwork:
Rebased! Thanks for all of your help!

Issue #366, opened 2016-02-12
Display usage metrics for multiple resources


bryk:
Dashboard should provide kubedash-like functionality to allow for monitoring Kubernetes resources, like seeing resource usage of a pod over time.

bryk:
As of now, mini-graphs of resource usage are worked on, see mocks: https://raw.githubusercontent.com/kubernetes/dashboard/master/docs/design/mockups/11-11-2015/17_job%20details.png

When this is done, adding new graphs in other places will be easy, as the infrastructure will be there.

bryk:
CC @bgrant0607-cc @aronchick @joeatwork

vishh:
cc @vishh

joeatwork:
I'm in the process of adding this fn now - the API freeze might make it more difficult, though.

aronchick:
That's great to hear Joe! Let me know if there's anything we can do between
now and then.

maciaszczykm:
I think it's almost finished. Only thing missing seems to be adding sparklines to the replication controller list cards. Am I right? Should we close this issue and create another for missing enhancement?

bryk:
I think that this is not that close to be finished. We need more metrics displayed on more resources (e.g., containers or nodes).

maciaszczykm:
Yes, I didn't consider containers and nodes. And there will be even more resources soon.

floreks:
This is done already.

/close

PR #387, opened 2016-02-17
Cpu and memory sparklines

joeatwork:
The changes proposed here depend on the new API information in #386 - comment on this pull request can probably wait until those changes are revised and accepted or declined.

This pull request proposes adding a sparkline-like visualization of the resource usage history for individual pods in a replication controller detail screen, fixing #366

kubernetes dashboard - google chrome_053

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/387)

### DIFF
--- a/src/app/frontend/common/components/sparkline/sparkline.html
+++ b/src/app/frontend/common/components/sparkline/sparkline.html
@@ -1,0 +1,23 @@
+<!--
+Copyright 2015 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<svg viewBox="0,0 1,1"
+     preserveAspectRatio="none"
+     class="kd-sparkline">
+  <polygon
+     ng-attr-points="0,1 {{::sparklineCtrl.polygonPoints()}} 1,1"
+     class="kd-sparkline-series"/>
+</svg>
--- a/src/app/frontend/common/components/sparkline/sparkline_controller.js
+++ b/src/app/frontend/common/components/sparkline/sparkline_controller.js
@@ -1,0 +1,51 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @final
+ */
+export default class SparklineController {
+  /**
+   * Constructs sparkline controller.
+   * @ngInject
+   */
+  constructor() {
+    /**
+     * An array of {backendAPI.MetricResult} objects. The timestamp
+     * values of each object must be unique, and value must be greater
+     * than or equal to zero.
+     * @export {!Array<!backendApi.MetricResult>} Initialized from the scope.
+     */
+    this.timeseries;
+  }
+
+  /**
+   * Formats the underlying series suitable for display as an SVG polygon.
+   * @return string
+   * @export
+   */
+  polygonPoints() {
+    const series = this.timeseries.map(({timestamp, value}) => [Date.parse(timestamp), value]);
+    const sorted = series.slice().sort((a, b) => a[0] - b[0]);
+    const xShift = Math.min(...sorted.map((pt) => pt[0]));
+    const shifted = sorted.map(([x, y]) => [x - xShift, y]);
+    const xScale = Math.max(...shifted.map((pt) => pt[0])) || 1;
+    const yScale = Math.max(...shifted.map((pt) => pt[1])) || 1;
+    const scaled = shifted.map(([x, y]) => [x / xScale, y / yScale]);
+
+    // Invert Y because SVG Y=0 is at the top, and we want low values
+    // of Y to be closer to the bottom of the graphic
+    return scaled.map(([x, y]) => `${x},${(1 - y)}`).join(' ');
+  }
+}
--- a/src/app/frontend/common/components/sparkline/sparkline_directive.js
+++ b/src/app/frontend/common/components/sparkline/sparkline_directive.js
@@ -1,0 +1,30 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SparklineController from './sparkline_controller';
+
+/**
+ * Returns directive definition for sparkline.
+ * @return {!angular.Directive}
+ */
+export default function sparklineDirective() {
+  return {
+    controller: SparklineController,
+    controllerAs: 'sparklineCtrl',
+    templateUrl: 'common/components/sparkline/sparkline.html',
+    templateNamespace: 'svg',
+    scope: {},
+    bindToController: {'timeseries': '='},
+  };
+}
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
@@ -135,21 +135,31 @@ limitations under the License.
                   -
                 </span>
               </td>
-              <td kd-responsive-header="CPU" class="kd-replicationcontrollerdetail-table-cell"
-                  ng-if="::ctrl.replicationControllerDetail.hasMetrics">
-                <span ng-if="::ctrl.hasCpuUsage(pod)">
-                  {{::(pod.metrics.cpuUsage | kdCores)}}
+              <!-- width=1 is a hack to shrink-wrap the column around its content -->
+              <td kd-responsive-header="CPU"
+                  class="kd-replicationcontrollerdetail-table-cell kd-replicationcontrollerdetail-sparkline-table-cell"
+                  ng-if="::ctrl.replicationControllerDetail.hasMetrics"
+                  ng-attr-width="{{ctrl.shouldShrinkSparklineCells()}}">
+                <span ng-if="::ctrl.hasCpuUsage(pod)" class="kd-replicationcontrollerdetail-labeled-sparkline">
+                  <span class="kd-replicationcontrollerdetail-labeled-sparkline-label"
+                       >{{::(pod.metrics.cpuUsage | kdCores)}}</span>
+                  <kd-sparkline timeseries="pod.metrics.cpuUsageHistory"></kd-sparkline>
                 </span>
-                <span ng-if="::!ctrl.hasCpuUsage(pod)">
+                <span ng-if="::(!ctrl.hasCpuUsage(pod))">
                   -
                 </span>
               </td>
-              <td kd-responsive-header="Memory" class="kd-replicationcontrollerdetail-table-cell"
-                  ng-if="::ctrl.replicationControllerDetail.hasMetrics">
-                <span ng-if="::ctrl.hasMemoryUsage(pod)">
-                  {{::(pod.metrics.memoryUsage | kdMemory)}}
+              <!-- width=1 is a hack to shrink-wrap the column around its content -->
+              <td kd-responsive-header="Memory"
+                  class="kd-replicationcontrollerdetail-table-cell kd-replicationcontrollerdetail-sparkline-table-cell"
+                  ng-if="::ctrl.replicationControllerDetail.hasMetrics"
+                  ng-attr-width="{{ctrl.shouldShrinkSparklineCells()}}">
+                <span ng-if="::ctrl.hasMemoryUsage(pod)" class="kd-replicationcontrollerdetail-labeled-sparkline">
+                  <span class="kd-replicationcontrollerdetail-labeled-sparkline-label"
+                       >{{::(pod.metrics.memoryUsage | kdMemory)}}</span>
+                  <kd-sparkline timeseries="pod.metrics.memoryUsageHistory"></kd-sparkline>
                 </span>
-                <span ng-if="::!ctrl.hasMemoryUsage(pod)">
+                <span ng-if="::(!ctrl.hasMemoryUsage(pod))">
                   -
                 </span>
               </td>
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
@@ -117,3 +117,45 @@ md-icon {
   color: $delicate;
   font-size: $subhead-font-size-base;
 }
+
+.kd-replicationcontrollerdetail-labeled-sparkline-label {
+  white-space: nowrap;
+  word-spacing: normal;
+}
+
+.kd-sparkline {
+  background-color: $body;
+  height: 2.5 * $baseline-grid;
+  vertical-align: -$baseline-grid / 2; // 20% below baseline, to flow with text
+  width: 12.5 * $baseline-grid;
+
+  @media only screen and (min-width: $layout-breakpoint-lg) {
+    margin-left: $baseline-grid;
+  }
+
+  @media only screen and (min-width: $layout-breakpoint-xs) and (max-width: $layout-breakpoint-lg) {
+    margin-top: $baseline-grid / 2;
+  }
+}
+
+.kd-sparkline-series {
+  fill: $chart-1;
+}
+
+.kd-replicationcontrollerdetail-labeled-sparkline {
+  @media only screen and (min-width: $layout-breakpoint-lg) {
+    display: block;
+    text-align: right;
+    white-space: nowrap;
+  }
+
+  @media only screen and (min-width: $layout-breakpoint-xs) and (max-width: $layout-breakpoint-lg) {
+    display: block;
+    padding-bottom: $baseline-grid;
+    padding-top: $baseline-grid;
+  }
+
+  @media only screen and (max-width: $layout-breakpoint-xs) {
+    word-spacing: $baseline-grid;
+  }
+}
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
@@ -103,7 +103,7 @@ export default class ReplicationControllerDetailController {
    * @returns {boolean}
    * @export
    */
-  isSidebarVisible() { return this.mdMedia('gt-sm'); }
+  isSidebarVisible() { return this.mdMedia('gt-md'); }
 
   /**
    * Returns true if event is a warning.
@@ -183,7 +183,7 @@ export default class ReplicationControllerDetailController {
    * @export
    */
   hasCpuUsage(pod) {
-    return !!pod.metrics && (!!pod.metrics.cpuUsage || pod.metrics.cpuUsage === 0);
+    return !!pod.metrics && !!pod.metrics.cpuUsageHistory && pod.metrics.cpuUsageHistory.length > 0;
   }
 
   /**
@@ -192,6 +192,19 @@ export default class ReplicationControllerDetailController {
    * @export
    */
   hasMemoryUsage(pod) {
-    return !!pod.metrics && (!!pod.metrics.memoryUsage || pod.metrics.memoryUsage === 0);
+    return !!pod.metrics && !!pod.metrics.memoryUsageHistory &&
+        pod.metrics.memoryUsageHistory.length > 0;
   }
+
+  /**
+   * Returns either 1 (if the table cells containing sparklines should
+   * shrink around their contents) or undefined (if those table cells
+   * should obey regular layout rules). The idiosyncratic return
+   * protocol is for compatibility with ng-attr's behavior - we want
+   * to generate either "width=1" or nothing at all.
+   *
+   * @return {(number|undefined)}
+   * @export
+   */
+  shouldShrinkSparklineCells() { return (this.mdMedia('gt-xs') || undefined) && 1; }
 }
--- a/src/app/frontend/_variables.scss
+++ b/src/app/frontend/_variables.scss
@@ -38,6 +38,7 @@ $hover-secondary: #ff1c19;
 $body: #eee;
 $emphasis: #000;
 $content-background: #fff;
+$chart-1: #00c752;
 
 // TODO(bryk): Get those variables from Angular Material scss files.
 $foreground-1: rgba(0, 0, 0, .87);
--- a/src/app/frontend/common/components/components_module.js
+++ b/src/app/frontend/common/components/components_module.js
@@ -15,6 +15,7 @@
 import filtersModule from '../filters/filters_module';
 import labelsDirective from './labels/labels_directive';
 import middleEllipsisDirective from './middleellipsis/middleellipsis_directive';
+import sparklineDirective from './sparkline/sparkline_directive';
 
 /**
  * Module containing common components for the application.
@@ -27,4 +28,5 @@ export default angular
           filtersModule.name,
         ])
     .directive('kdLabels', labelsDirective)
-    .directive('kdMiddleEllipsis', middleEllipsisDirective);
+    .directive('kdMiddleEllipsis', middleEllipsisDirective)
+    .directive('kdSparkline', sparklineDirective);
--- a/src/test/frontend/common/components/sparkline/sparkline_controller_test.js
+++ b/src/test/frontend/common/components/sparkline/sparkline_controller_test.js
@@ -1,0 +1,81 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SparklineController from 'common/components/sparkline/sparkline_controller';
+import componentsModule from 'common/components/components_module';
+
+describe('Sparkline controller', () => {
+  /**
+   * @type {!SparklineController}
+   */
+  let ctrl;
+
+  beforeEach(() => {
+    angular.mock.module(componentsModule.name);
+
+    angular.mock.inject(($controller) => { ctrl = $controller(SparklineController); });
+  });
+
+  it('should produce no points with an empty series', () => {
+    // given
+    ctrl.timeseries = [];
+
+    // then
+    expect(ctrl.polygonPoints()).toBe('');
+  });
+
+  it('should shift and scale times such that the minimum time is zero and the maximum time is 1',
+     () => {
+       // given
+       ctrl.timeseries = [
+         {timestamp: '1976-01-15T00:00:00Z', value: 1},
+         {timestamp: '2026-01-15T00:00:00Z', value: 1},
+       ];
+
+       // then
+       expect(ctrl.polygonPoints()).toBe('0,0 1,0');
+     });
+
+  it('should handle zero values and times without throwing an exception', () => {
+    // given
+    ctrl.timeseries = [
+      {timestamp: '1970-01-01T00:00:00Z', value: 0},
+    ];
+
+    // then
+    expect(ctrl.polygonPoints()).toBe('0,1');
+  });
+
+  it('should scale values to <= 1 and invert them', () => {
+    // given
+    ctrl.timeseries = [
+      {timestamp: '1976-01-15T10:00:00Z', value: 10},
+      {timestamp: '1976-01-15T10:00:10Z', value: 1},
+    ];
+
+    // then
+    expect(ctrl.polygonPoints()).toBe('0,0 1,0.9');
+  });
+
+  it('should sort a time series by time', () => {
+    // given
+    ctrl.timeseries = [
+      {timestamp: '1976-01-15T10:00:10Z', value: 1},
+      {timestamp: '1976-01-15T10:00:00Z', value: 10},
+    ];
+
+    // then
+    expect(ctrl.polygonPoints()).toBe('0,0 1,0.9');
+  });
+});
--- a/src/test/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller_test.js
+++ b/src/test/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller_test.js
@@ -111,21 +111,23 @@ describe('Replication Controller Detail controller', () => {
   it('should show/hide cpu and memory metrics for pods', () => {
     expect(ctrl.hasMemoryUsage({})).toBe(false);
     expect(ctrl.hasMemoryUsage({metrics: {}})).toBe(false);
-    expect(ctrl.hasMemoryUsage({metrics: {memoryUsage: 0}})).toBe(true);
-    expect(ctrl.hasMemoryUsage({metrics: {memoryUsage: 1}})).toBe(true);
-    expect(ctrl.hasMemoryUsage({metrics: {memoryUsage: null}})).toBe(false);
-    expect(ctrl.hasMemoryUsage({metrics: {memoryUsage: undefined}})).toBe(false);
-    expect(ctrl.hasMemoryUsage({metrics: {cpuUsage: 1}})).toBe(false);
+    expect(ctrl.hasMemoryUsage({metrics: {memoryUsageHistory: []}})).toBe(false);
+    expect(ctrl.hasMemoryUsage({metrics: {memoryUsageHistory: [0]}})).toBe(true);
+    expect(ctrl.hasMemoryUsage({metrics: {memoryUsageHistory: [1]}})).toBe(true);
+    expect(ctrl.hasMemoryUsage({metrics: {memoryUsageHistory: null}})).toBe(false);
+    expect(ctrl.hasMemoryUsage({metrics: {memoryUsageHistory: undefined}})).toBe(false);
+    expect(ctrl.hasMemoryUsage({metrics: {cpuUsageHistory: [1]}})).toBe(false);
 
     expect(ctrl.hasCpuUsage({})).toBe(false);
     expect(ctrl.hasCpuUsage({metrics: {}})).toBe(false);
-    expect(ctrl.hasCpuUsage({metrics: {memoryUsage: 0}})).toBe(false);
-    expect(ctrl.hasCpuUsage({metrics: {memoryUsage: 1}})).toBe(false);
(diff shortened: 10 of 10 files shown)


joeatwork:
There are a few product questions associated with this feature, in addition to issues with the implementation -

- Right now, the scale of each chart is independent of the others - the maximum point in history will always be at the top of the chart, and even if the chart is showing only one data point the line will cross the whole chart. It's possible that these charts should all share a common scale, either on a single page or in general - if so, please comment here.

- As proposed here, the charts are always rendered, no matter how wide or narrow the window is, and are never wrapped below the text measures. Should they disappear or wrap at some point?

bryk:
Just tested. The UI always shows graphs in the same shape. Did you have this problem? Tested on `gulp serve` and in-cluster heapster. 
selection_031

bryk:
selection_032

bryk:
I'll investigate this tomorrow.

joeatwork:
I see those shapes for very young processes - it looks like your CPU only has two data points, and your memory history only has three. I'd love to see what you're getting back from the API. It's possible that our Heapsters are configured differently (and if so, I should probably account for that before showing the charts)

bryk:
@floreks Can you help with review?

src/app/frontend/common/components/sparkline/sparkline.html
```diff
@@ -0,0 +1,8 @@
+<svg viewBox="0,0 1,1"
```
bryk: Nit: please add the ` Copyright ...` boilerplate to html files. See how it is done in others. 

No worries, it is stripped and minified during build process.

src/app/frontend/common/components/sparkline/sparkline.html
```diff
@@ -0,0 +1,8 @@
...
+     preserveAspectRatio="none"
+     class="kd-sparkline">
+  <polygon
+     ng-if="sparklineCtrl.series"
```
bryk: Can we maybe ng-if entire SVG?

bryk: And bind once? Since points are already bound once.

src/app/frontend/common/components/sparkline/sparkline_controller.js
```diff
@@ -0,0 +1,86 @@
...
+   * Constructs sparkline controller.
+   * @ngInject
+   */
+  constructor($scope) {
+    /**
+     * An array of [x,y] pairs. The first values of each pair must be
+     * unique, and the second value must be greater than or equal to
+     * zero. Exactly one of series or timeseries should be specified by
+     * the scope.
+     * @export {!Array<!Array<number>>} Initialized from the scope.
```
bryk: Actually everything this directive needs as an input is this.timeseries. Am I correct? If yes, then how about design like this. The directive takes only one input parameter: timeseries. Then it has one `@exported`, argumentless function `polygonPoints`. The function computes polygon points based on its only state which is this.timeseries. The benefit is that there are no `$watch`ers and the class is (almost) stateless. This means that is is easier to read and reason about the code here. There should be no performance downgrade, because the binding is one time. If the binding was not one time, then we can talk. 

What do you think?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
```diff
@@ -85,6 +85,15 @@ $table-cell-height-half: $table-cell-height / 2;
...
     word-break: break-all;
   }
+
+  &.kd-replicationcontrollerdetail-metric-table-cell {
+    text-align: right;
+    white-space: nowrap;
+  }
+
+  &.kd-replicationcontrollerdetail-sparkline-table-cell {
+    padding-left: 1.25 * $baseline-grid;
```
bryk: Can we avoid 1.25? We tend to have $baseline-grid everywhere, unless justified otherwise.

src/app/frontend/common/components/sparkline/sparkline_controller.js
```diff
@@ -0,0 +1,86 @@
...
+   */
+  onSeriesUpdate_() {
+    if (this.timeseries) {
+      this.series = this.timeseries.map(({timestamp, value}) => [Date.parse(timestamp), value]);
+    }
+
+    const sorted = this.series.slice().sort((a,b) => a[0] - b[0]);
+    if (sorted.find(([_,y]) => y < 0)) {
+      throw new Error(
+        "sparkline doesn't support negative y values"
```
bryk: Nit: use `'` or `\`` for strings quoting. This is to be consistent across codebase.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
```diff
@@ -120,3 +129,15 @@ md-icon {
   color: $delicate;
   font-size: $subhead-font-size-base;
 }
+
+.kd-sparkline {
+  width: 100px;
+  height: 20px;
+  vertical-align: -5px;
+  margin-left: 4px;
```
bryk: 4px here? Without it looks equally good.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
```diff
@@ -120,3 +129,15 @@ md-icon {
   color: $delicate;
   font-size: $subhead-font-size-base;
 }
+
+.kd-sparkline {
+  width: 100px;
+  height: 20px;
+  vertical-align: -5px;
```
bryk: Do you want to center this in a cell? Remove all margins and set line-height: 0. With that you wont have to hardcode anything.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
```diff
@@ -120,3 +129,15 @@ md-icon {
   color: $delicate;
   font-size: $subhead-font-size-base;
 }
+
+.kd-sparkline {
+  width: 100px;
+  height: 20px;
```
bryk: Can you make this a multiplier of $baseline-grid?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
```diff
@@ -120,3 +129,15 @@ md-icon {
   color: $delicate;
   font-size: $subhead-font-size-base;
 }
+
+.kd-sparkline {
+  width: 100px;
```
bryk: Can you make this a multiplier of $baseline-grid?

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
```diff
@@ -192,6 +192,22 @@ export default class ReplicationControllerDetailController {
...
+    return !!pod.metrics && !!pod.metrics.memoryUsageHistory && pod.metrics.memoryUsageHistory.length > 0;
+  }
+
+  /**
+   * Produces the last value of the last element of an array of MetricsResults.
+   * @param {!Array<!backendApi.MetricResult>} ts
+   * @return {number}
+   * @export
+   */
+  latestTimeseriesValue(ts) {
```
bryk: How about to rely on what backend returns in the cpuUsage field in JSON. Let the backend decide what to display and frontend only care how to display it.

bryk: What do you think?

bryk:
This is awesome @joeatwork :) I'm really happy with the shape of this PR. A few comments, PTAL

src/app/frontend/common/components/sparkline/sparkline.html
joeatwork: Fixed in 3062264c2fbf17752cc940fcab0e11e778fb9301

src/app/frontend/common/components/sparkline/sparkline.html
joeatwork: Fixed in a6b3d3ba377ac2f85691ba534766202d6d20dcc8 (but the conditional goes away in bcf427d86d62322dd0673e2bf6962658d00440d4 )

src/app/frontend/common/components/sparkline/sparkline_controller.js
joeatwork: Less code is better! Fixed in bcf427d86d62322dd0673e2bf6962658d00440d4

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
joeatwork: Fixed in becde9ed735aa9274c1578b0ec70f8a2f9acc812 - however, I've added more goofy coefficients in there to size the charts (based on counting pixels in the mock) - I'm happy to relax those as well, but I don't want to do any design-on-the-fly, or step on your designer's toes.

src/app/frontend/common/components/sparkline/sparkline_controller.js
joeatwork: Thanks! I'll try to avoid double quotes in the future! (This line obsoleted in bcf427d86d62322dd0673e2bf6962658d00440d4)

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
joeatwork: Indeed! Cut in becde9ed735aa9274c1578b0ec70f8a2f9acc812

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
joeatwork: It turns out to look fine without the baseline (this was a vestige of an earlier layout strategy) - removed in becde9ed735aa9274c1578b0ec70f8a2f9acc812

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
joeatwork: Changed in becde9ed735aa9274c1578b0ec70f8a2f9acc812, but with screwy coefficients. Let me know if I can round these out :)

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
joeatwork: Changed in becde9ed735aa9274c1578b0ec70f8a2f9acc812

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
joeatwork: This was an attempt to allow for deprecation of the old values, but less code is better! Fixed in d94f58ce5d174241ea7749af349ae8f435ae0834

joeatwork:
Thanks for all of your help! I believe I've addressed your comments - please take another look when you get the time.

floreks:
> @floreks Can you help with review?

I wanted to this today morning but I see you've done it all :) I'll take a look anyway.

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
```diff
@@ -138,22 +140,32 @@ <h1 flex class="md-title kd-replicationcontrollerdetail-app-name">
...
+              <!-- width=1 is a hack to shrink-wrap the column around its content -->
+              <td class="kd-replicationcontrollerdetail-table-cell kd-replicationcontrollerdetail-metric-table-cell"
+                  ng-if="::ctrl.replicationControllerDetail.hasMetrics"
+                  width="1">
                 <span ng-if="::ctrl.hasCpuUsage(pod)">
                   {{::(pod.metrics.cpuUsage | kdCores)}}
                 </span>
-                <span ng-if="::!ctrl.hasCpuUsage(pod)">
-                  -
-                </span>
```
floreks: Without that when there are no metrics to display columns are empty.

zrzut ekranu z 2016-02-19 10-01-10
zrzut ekranu z 2016-02-19 10-00-10

This doesn't look very good. Sometimes we have to wait a bit for metrics to appear so i guess this is needed.

src/app/frontend/common/components/sparkline/sparkline_directive.js
```diff
@@ -0,0 +1,33 @@
...
+export default function sparklineDirective() {
+  return {
+    controller: SparklineController,
+    controllerAs: 'sparklineCtrl',
+    templateUrl: 'common/components/sparkline/sparkline.html',
+    templateNamespace: 'svg',
+    scope: {},
+    bindToController: {
+      'series': '=?',
+      'timeseries': '=?'
```
floreks: Comma missing ;) Run gulp format-javascript and gulp-check it will tell you what needs to be changed.

src/app/frontend/common/components/sparkline/sparkline_controller.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @final
+ */
+export default class sparklineController {
+  /**
+   * Constructs sparkline controller.
+   * @ngInject
+   */
+  constructor($scope) {
```
floreks: Unused

src/app/frontend/common/components/sparkline/sparkline_directive.js
bryk: Yeah, `gulp lint`, `gulp check` are your friends when sending a PR.

bryk:
Reviewed 7 of 12 files at r1, 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 29 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmTtC-KuihVjF8MSQ:-KAtmTtDZ-_8poN7GkZj:810033481) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L29)):*
You dont need this input param anymore. 

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmZJLnWidKk89dBNA:-KAtmZJM9NQgpB6-iYU8:-2094990651) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
Is timeseries optional? What will happen if this is not defined? I think the Javascript will fail.  Mark this as non-optional? 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 158 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmqktukPuOzVEkFWQ:-KAtmqktukPuOzVEkFWR:-1527555220) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L158)):*
Hmm... How about we put this in the same column as current usage? So that the sparkline wraps as needed. The reason is that the table is already pretty wide and forcing it to be even wider worries me. 

@floreks ? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

bryk:
(bear in mind that I'm testing reviewable.io, so might miss something)

floreks:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmZJLnWidKk89dBNA:-KAtucVV03qt0lDdHKP9:-888901762) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
Do we actually need 2 way binding here? Is '@' not sufficient?

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 158 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmqktukPuOzVEkFWQ:-KAtuxBTXCSbq-PKdFnY:975240291) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L158)):*
I agree. I'm also worried how introducing additional column would affect row to table mapping on smaller screens. Same column for both sounds good.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

bryk:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmZJLnWidKk89dBNA:-KAu3lP3L9a4P73TuB4m:-945283009) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
Timeseries is an object. You usually dont want to bind objects with `@` which converts everything to a string. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

bryk:
Btw, there's also a `gulp format-javascript` task.

src/app/frontend/common/components/sparkline/sparkline_directive.js
joeatwork: Fixed by `gulp format-javascript` in a7564b27a685f

src/app/frontend/common/components/sparkline/sparkline_controller.js
joeatwork: Fixed in a7564b27a685f

joeatwork:
Review status: 4 of 9 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 29 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmTtC-KuihVjF8MSQ:-KAui-3VOKxNj6l7wJ9H:487744423) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L29)):*
Fixed in 0ae5a0a

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmZJLnWidKk89dBNA:-KAujR1twXYm70gmEKDv:1306575208) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
I've made timeseries non-optional in 205a21507071b1802e968bc78d817a5dcb0d90b2 - for now, I'm leaving the '=' binding, if that's ok.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 158 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmqktukPuOzVEkFWQ:-KAvAThXuKVpI5q2nx3-:417086295) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L158)):*
We should probably explicitly wrap the column at certain widths with media-queries. Otherwise it will be really tough to get a column that looks like the mock - the mock right now is right justified, but the cell shrinks to so that the left edge of the longest cell in the column lines up with the header (I've included an image with what I think is the desired behavior - the content of the widest cell lines up on the left with the header, but all of the other cells are right-aligned, so their charts line up)



Getting this effect requires that the cells be as narrow as they can be, or the left edge of the longest content will wander off and leave the header looking strange.



A fix (and the only one I can think of) is to make the sparkline cells as narrow as possible. Unfortunately, this means they'll *always* wrap, since a wrapped cell is narrower than an unwrapped cell. We could change this to prevent the wrap, but this will (in the absence of media queries) prevent wrapping in all cases, which doesn't get us anywhere.



Is there an example of media-querying, or responding in layout to size, elsewhere in the code? It seems like the natural thing to do might be to explicitly hide or wrap the charts when the viewport is narrower than some threshold, rather than trying to get the natural text wrapping to work. Alternately, we could just left-justify the content, or compromise on the mock, if that seems appropriate. Do you guys have any thoughts about this?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

floreks:
Review status: 4 of 9 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 158 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmqktukPuOzVEkFWQ:-KB7byveQm3mXNb3V_BF:-2036180687) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L158)):*
Display of CPU usage units should be changed soon to more user friendly format (2.5 CPU). This way there won't be so big number s used and we can style it better. For memory I think we can also round it to 1 decimal place. 

Still for display we can use 1 table column but wrap it in 2 divs or something and align text to the left. I'd love to help with styling that but for now I've problem with heapster on my local cluster and can't make him produce any metrics. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

bryk:
Reviewed 4 of 5 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmZJLnWidKk89dBNA:-KB7xBGl5oshOjUE1VkQ:-557631074) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
Nice, thanks! 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 158 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmqktukPuOzVEkFWQ:-KB8M_I8UXwOO2yD7xw7:1528216514) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L158)):*
We have media queries examples in the code, e.g., `./app/frontend/chrome/chrome.scss:89:@media only screen and (max-width: $layout-breakpoint-xs) {`. There's even media queries in JS code using `$mdMedia` service. See `_variables.scss` file for breakpoints or take a breakpoint from https://github.com/angular/material/blob/HEAD/src/core/style/variables.scss. 

I say: go for media queries. Display underneath and next to depending on the size. Please verify that this works from 1920px width to phone size :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

bryk:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KB8SjpMY9V51ZNciVCL:-KB8SjpN04STxjN9K89K:-1477215031) (no related file):*
Btw, this PR needs rebase. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
```diff
@@ -117,3 +117,45 @@ md-icon {
...
+
+.kd-replicationcontrollerdetail-labeled-sparkline-label {
+  white-space: nowrap;
+  word-spacing: normal;
+}
+
+.kd-sparkline {
+  background-color: $body;
+  height: 2.5 * $baseline-grid;
+  vertical-align: -$baseline-grid / 2; // 20% below baseline, to flow with text
```
joeatwork: @bryk - I've added the vertical-align back in. Without it, the graph visually looks too high, like it's a step up from the label text. If I change the enclosing line-height to zero, it moves the graph but also the label text, so the graph still looks higher than the mock. (I might not understand your line-height suggestion) - please let me know if this is ok.

joeatwork:
I've changed the look of empty fields (That is, when you are expecting metrics from the server but get an empty list of metrics) to match the other fields, displaying a left-aligned dash in that case:



---

Review status: 5 of 9 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/app/frontend/common/components/sparkline/sparkline_controller.js, line 23 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtlOLstXn85B6kEKub-r2-23:-KAuhXuRlNAEurI9tjf7:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_controller.js#L23)):*
Done.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 29 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmTtC-KuihVjF8MSQ:-KAvqbjSeFFULQ1brGM8:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L29)):*
Done.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtlOLstXn85B6kEKuc-r2-30:-KAvqbO5_l3kO2JzvdMo:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
Done.

---

*[src/app/frontend/common/components/sparkline/sparkline_directive.js, line 30 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmZJLnWidKk89dBNA:-KAvqc0IUsHVABBmh1fO:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/common/components/sparkline/sparkline_directive.js#L30)):*
Done.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 158 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtmqktukPuOzVEkFWQ:-KBKb_IYkrMRBYNP5k89:-858937258) ([raw file](https://github.com/kubernetes/dashboard/blob/d94f58ce5d174241ea7749af349ae8f435ae0834/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L158)):*
I've added media queries - I've also moved the breaking point that hides the sidebar from gt-sm to gt-md, since the new columns expand the width of the table. I've made some choices about the padding, spacing of the sparkline that are probably worth taking a look at- here are some screenshots of the different sized layouts after the proposed patch:

"Natural"  XL width:



Middle width - no sidebar, wrapping sparkline. I've padded the top and bottom of the cell in this case by $baseline-grid



Standard "phone" width - there is enough room for the sparkline to display next to the label for the most part



Most narrow - the sparkline wraps naturally (the sparkline will also wrap if the label is very long, which seems reasonable)



---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-:-KBKb2y2gvSOasLLYRM6:1197909230)*

src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
joeatwork: fixed in 3074c8ff and 937893dcc2af22b784681a24c9f1896b22c03854

joeatwork:
@floreks, @bryk I think I've addressed your concerns with wrapping and single-table-cell layouts in the recent commits - I haven't changed the formats or metrics for the CPU or Memory, but those can be changed globally whenever you'd like since I'm using the same filters to format the values that are used elsewhere. Please take another look and let me know if these changes work for you! The biggest things to consider (and maybe ask me to change) are

- I've changed the breakpoint that makes the left column disappear from sm to md, since the new columns were breaking out of the viewport much earlier

- I've added a baseline adjustment back to the sparkline (this time in terms of $baseline-grid) because otherwise the baseline of the text and the bottom of the chart want to line up in a funny looking way

- I've added some padding and spacing at various viewport widths that isn't in any mock, which means I may be doing goofy un-designerly things. You can see the results in the screenshots attached to the PRQ.

Thanks again for all of your help with this!

bryk:
Nice! My last comments about style and some tests and let's merge. I cant wait to have this on HEAD 

---

Reviewed 2 of 4 files at r5.
Review status: 7 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/app/frontend/common/components/sparkline/sparkline_controller.js, line 18 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KBMm4nvLgSXqkTb2CXj:-KBMm4nvLgSXqkTb2CXk:-1996729719) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/common/components/sparkline/sparkline_controller.js#L18)):*
Btw, classes start should with Capital Letter. 

---

*[src/app/frontend/common/components/sparkline/sparkline_controller.js, line 23 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KBMm2IflnEwGFJcAY1R:-KBMm2IflnEwGFJcAY1S:-1608042257) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/common/components/sparkline/sparkline_controller.js#L23)):*
Can you test this class? 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 156 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KBMlN7VcNm8mLMq7oWm:-KBMlN7VcNm8mLMq7oWn:-162652394) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L156)):*
Can you move this and the other `{{(ctrl.mdMedia('gt-xs') || undefined) && 1}}` to a controller's method? We don't want to be putting logic into templates, just method calls and field reads. In controller, we'll be able also to test this. 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss, line 129 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KAtlOLstXn85B6kEKuX-r5-129:-KBMgPdmcBYAjo5Ail3_:825380998) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss#L129)):*
Sounds good. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-:-KBMq2ghY0H6h6Z8B-sX:-124257173)*

joeatwork:
Review status: 5 of 10 files reviewed at latest revision, 6 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KB8SjpMY9V51ZNciVCL:-KBKdh8ZiD26zGiy2Do0:1999988409) (no related file):*
Rebased against current master. Thanks!

---

*[src/app/frontend/common/components/sparkline/sparkline_controller.js, line 18 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KBMm4nvLgSXqkTb2CXj:-KBPpwyicry34lQldD_a:-2116011021) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/common/components/sparkline/sparkline_controller.js#L18)):*
Fixed in 7ed2a98ec5f0f2

---

*[src/app/frontend/common/components/sparkline/sparkline_controller.js, line 23 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KBMm2IflnEwGFJcAY1R:-KBPpskfft3iFR7YLYW1:1546554755) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/common/components/sparkline/sparkline_controller.js#L23)):*
Added tests in 1427bef40c53d

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html, line 156 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-KBMlN7VcNm8mLMq7oWm:-KBPpkF2rR0MbJhuelIV:656050373) ([raw file](https://github.com/kubernetes/dashboard/blob/937893dcc2af22b784681a24c9f1896b22c03854/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html#L156)):*
Fixed in 9d996908b47

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387)*

joeatwork:
Thanks for your continuing help with this! I think your last comments have been addressed.

cheld:
@joeatwork 
1. The code seems to have some formatting issues. You must run `gulp format-javascript` before committing.
2. Normally we squash our commits

@bryk I also see only three data points when using local cluster

bryk:
@cheld Update your heapster. `0.18.x` version does not quite work. 

@joeatwork squash and `gulp format-javascript` and we're ready to merge :) :lgtm: 

---

Reviewed 1 of 4 files at r5, 4 of 4 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-:-KBh2RWkfse8pJczD7LS:-677959618)*

joeatwork:
Linted and squashed!

bryk:
Awesome! Merging.  If you need any help from us, just ping. 

---

Reviewed 2 of 2 files at r7.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/387#-:-KBi-OkvGCyVF2thgh6K:-152925377)*

joeatwork:
Thanks, guys!

cheld:
I played around with heapster 0.19.0 and the parameters -stats_resolution -model_resolution -model_frequency. 

I experimented a lot, but I was not able to create pretty charts. Can you suggest proper settings?

bryk:
@joeatwork What is your heapster setup?

cheld:
ok, I tried 2.0-alpha. This seems to work much better. However, even the default values produce something that is acceptable.

bryk:
.>ok, I tried [heapster] 2.0-alpha. This seems to work much better. Unfortunately, it crashes if one changes the default flags. However, even with the default values, this version produce something that is acceptable.

CC @mwielgus

cheld:
@mwielgus 

Unfortunately, the crashes are not only related to flags. The version also crashes randomly. I created an issue.

joeatwork:
I've been running with gcr.io/google_containers/heapster:v0.20.0-alpha5 - here's my heapster replication controller in detail:

```yaml
apiVersion: v1
kind: ReplicationController
metadata: 
  labels: 
    k8s-app: heapster
    kubernetes.io/cluster-service: "true"
    version: v10
  name: heapster-v10
  namespace: kube-system
spec: 
  replicas: 1
  selector: 
    k8s-app: heapster
    version: v10
  template: 
    metadata: 
      creationTimestamp: null
      labels: 
        k8s-app: heapster
        kubernetes.io/cluster-service: "true"
        version: v10
    spec: 
      containers: 
        - command: 
            - /heapster
            - "--source=kubernetes:https://kubernetes.default"
            - "--sink=influxdb:http://monitoring-influxdb:8086"
          image: gcr.io/google_containers/heapster:v0.20.0-alpha5
          imagePullPolicy: IfNotPresent
          name: heapster
          resources: 
            limits: 
              cpu: 100m
              memory: 224Mi
            requests: 
              cpu: 100m
              memory: 224Mi
          terminationMessagePath: /dev/termination-log
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      terminationGracePeriodSeconds: 30
```

Issue #380, opened 2016-02-16
Event warning icon not displayed in events table


floreks:
When using `local-up-cluster` or some older version of kubernetes master event type data is empty and warning icon is not displayed together with error messages.

![zrzut ekranu z 2016-02-16 10-36-46](https://cloud.githubusercontent.com/assets/2285385/13072485/ee483660-d499-11e5-852f-dc6ce17a6e9e.png)

I'll work on that as I've already solved this issue while implementing card statuses.

bryk:
Thanks!

PR #389, opened 2016-02-17
Fixed event warning icon not displayed in events table

floreks:
Related to #380 

I've changed approach to empty event Type issue a little. When it is not filled now, I'm adding the type based on event reason message. Also I've deleted PodEvent structure as it's not really needed. Events struct can be used instead and there is no need for additional structure.

@maciaszczykm @bryk PTAL

### DIFF
--- a/src/app/backend/replicationcontrollerlist.go
+++ b/src/app/backend/replicationcontrollerlist.go
@@ -25,7 +25,7 @@ import (
 )
 
 // Callback function in order to get the pod status errors
-type GetPodsEventWarningsFunc func(pods []api.Pod) ([]PodEvent, error)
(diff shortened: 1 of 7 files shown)


src/app/backend/replicationcontrollerlist.go
```diff
@@ -51,7 +51,7 @@ type ReplicationController struct {
 	// Label of this Replication Controller.
 	Labels map[string]string `json:"labels"`
 
-	// Aggregate information about pods belonging to this repolica set.
+	// Aggregate information about pods belonging to this replica set.
```
maciaszczykm: At the moment it's `replication controller`. Unfortunately, not all of the comments were updated during rename.

maciaszczykm:
LGTM

bryk:
LGTM :)

PR #390, opened 2016-02-17
Add kd-responsive-table style for details page

### LINKED ISSUE #334: Details page is really wide. Make it work on smaller screens.

maciaszczykm:
It makes details page more responsive, used it for both tables visible here. Fixes #334. Directive doesn't support tooltips, but do we need them on such small screens?

### DIFF
--- a/src/app/frontend/_variables.scss
+++ b/src/app/frontend/_variables.scss
@@ -31,6 +31,7 @@ $primary: #326de6;
 $secondary: #f51200;
 $warning: #ffa500;
 $delicate: #aaa;
+$border: #ddd;
 $muted: #888;
 $hover-primary: #1254df;
 $hover-secondary: #ff1c19;
@@ -48,5 +49,8 @@ $regular-font-weight: 400;
 $bold-font-weight: 500;
 
 $layout-breakpoint-lg: 1280px;
+$layout-breakpoint-xs: 600px;
 $whiteframe-shadow-1dp: 0 1px 3px 0 rgba(0, 0, 0, .2), 0 1px 1px 0 rgba(0, 0, 0, .14), 0 2px 1px -1px rgba(0, 0, 0, .12);
 $baseline-grid: 8px;
+$table-cell-height: 6 * $baseline-grid;
+$table-cell-height-half: $table-cell-height / 2;
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -84,3 +84,53 @@ a {
 [role='tabpanel'] {
   transition: none;
 }
+
+// Responsive table style.
+@media only screen and (max-width: $layout-breakpoint-xs) {
+  .kd-responsive-table {
+    width: 100%;
+
+    // Switches table to blocks.
+    thead,
+    tbody,
+    tr,
+    th {
+      display: block;
+    }
+
+    // Removes table headers.
+    thead {
+      tr,
+      th {
+        display: none;
+      }
+    }
+
+    // Adds border between records.
+    tr {
+      &:not(:last-child) {
+        border-bottom: $baseline-grid / 2 solid $border;
+      }
+    }
+
+    // Row data style.
+    td {
+      border-bottom: 1px solid $border;
+      display: block;
+      height: auto;
+      line-height: $table-cell-height;
+      padding-left: 50%;
+      position: relative;
+
+      // Row title style.
+      &::before {
+        content: attr(kd-responsive-header);
+        font-weight: $bold-font-weight;
+        left: $baseline-grid * 2;
+        padding-right: $baseline-grid * 2;
+        position: absolute;
+        word-wrap: break-word;
+      }
+    }
+  }
+}
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
@@ -49,7 +49,7 @@ limitations under the License.
   <md-tabs flex="grow" md-border-bottom md-dynamic-height>
     <md-tab label="Pods">
       <md-content>
-        <table class="kd-replicationcontrollerdetail-table" cellspacing="0">
+        <table class="kd-replicationcontrollerdetail-table kd-responsive-table" cellspacing="0">
           <thead>
             <tr>
               <th class="kd-replicationcontrollerdetail-table-header">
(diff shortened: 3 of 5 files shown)


src/app/frontend/chrome/chrome.scss
```diff
@@ -84,3 +84,27 @@ a {
 [role='tabpanel'] {
   transition: none;
 }
+
+// Overrides wt-responsive-table directive styles. More details at
+// https://github.com/awerlang/angular-responsive-tables.
+@media only screen and (max-width: 800px) {
+  .responsive {
+    tr {
+      &:not(:last-child) {
```
maciaszczykm: Not sure why, but it requires two empty lines if I'm putting it after `border: 0;`. That's why I've moved it to the beggining.

bryk:
Hmm... I looked at the angular-responsive-table code. I'm worried about it. It is trivial in implementation and does not have any flexibility. There's `800px` value hardcoded there and that's all. It is not compatible with AngularMaterial, which uses diferent values for breakpoints. How about pages where breakpoint is at 600px and others where it is at 960px?

bryk:
What if we do this in our code? I believe this can be done with a little bit of css. You can inspire on the original CSS

bryk:
Generally, I'm very worried about including any UI components that are not angular material. This is because of compatibility and their maturity. Please talk with me if you think my worries are wrong.

maciaszczykm:
Ok, it shouldn't take too long to create our own responsive table directive.

maciaszczykm:
@bryk Done, I've reused compile function from previously used directive and added our own stylesheets, which are way more simple. PTAL

src/app/frontend/common/components/responsivetable/responsivetable_directive.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Returns directive definition for responsive table.
+ *
+ * @return {!angular.Directive}
+ */
+export default function responsiveTableDirective() {
+  return {
+    restrict: 'A',
+    compile: function(element, attrs) {
```
bryk: Hey, you see what this code does? It just adds a class and sets data-title attributes. I dont think we need this at all. Wouldn't CSS just suffice?

src/app/frontend/common/components/responsivetable/responsivetable.scss
```diff
@@ -0,0 +1,64 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@import '../../../variables';
+
+@media only screen and (max-width: $layout-breakpoint-md) {
+  .responsive {
```
bryk: Prefix classes introduced in this project with kd- :)

src/app/frontend/_variables.scss
```diff
@@ -31,6 +31,7 @@ $primary: #326de6;
 $secondary: #f51200;
 $warning: #ffa500;
 $delicate: #aaa;
+$border: #ddd;
```
bryk: Isn't border the same as foreground-4? How about using foreground-4 color instead of $border? General idea is the less variables the better.

bryk: And foreground-X are favored unless other color is really better.

bryk:
Nice, PTAL

src/app/frontend/_variables.scss
maciaszczykm: All of `foreground-X` have alpha factor, which can cause overlays in few scenarios. To make borders color always the same I've added new variable.

src/app/frontend/common/components/responsivetable/responsivetable_directive.js
maciaszczykm: I see and I think it's pretty nice way to avoid adding labels directly for each table like author of https://css-tricks.com/responsive-data-tables/ does:

```css
td:nth-of-type(1):before { content: "First Name"; }
td:nth-of-type(2):before { content: "Last Name"; }
td:nth-of-type(3):before { content: "Job Title"; }
...
td:nth-of-type(10):before { content: "Arbitrary Data"; }
```

Instead of this we got just `content: attr(data-title);` in stylesheet. Do you see any other generic solution for it?

Regarding code, it's on MIT license, but anyways, I forgot to put an information about author in source code. Sorry!

src/app/frontend/common/components/responsivetable/responsivetable.scss
maciaszczykm: Okay.

maciaszczykm:
@bryk PTAL

bryk:
@maciaszczykm You can give title attributes to `td`s by hand. We have only a few columns, so this should not be tedious. E.g. kd-my-title="Nodes". And then in CSS use `content: attr(kd-my-content);`

Makes sense?

maciaszczykm:
@bryk Yes, it does. Will change it in a moment.

maciaszczykm:
It's a little bit more complex when adding new table, but result is the same. 

@bryk PTAL

src/app/frontend/_variables.scss
bryk: Ok, sounds good.

src/app/frontend/common/components/responsivetable/responsivetable_directive.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Returns directive definition for responsive table.
+ *
+ * @return {!angular.Directive}
+ */
+export default function responsiveTableDirective() {
+  return {
+    restrict: 'A',
+    compile: function(tElement, tAttrs) { tAttrs.$addClass('kd-responsive-table'); },
```
bryk: Well you don't need this now :) Just add a class to the element.

bryk:
I'm thinking about breakpoint... Should it maybe be lower? W're going straight from max layout to responsive table. There is no point we have normal table and details in a tab. Just thinking... Your opinion?

maciaszczykm:
@bryk Yes, directive is no more needed. Updated. PTAL

maciaszczykm:
> I'm thinking about breakpoint... Should it maybe be lower? W're going straight from max layout to responsive table. There is no point we have normal table and details in a tab. Just thinking... Your opinion?

I will try to scale it till moment when horizintall scrollbar appears.

bryk:
Use `$layout-breakpoint-xs`?

bryk:
Tested it. Looks quite nice.

maciaszczykm:
`$layout-breakpoint-xs` seems to be just fine. PTAL

bryk:
LGTM :) Thanks

Issue #338, opened 2016-02-08
apiserver-host incorrectly handles trailing slashes 


therc:
Setting --apisever-host to http://a.b.c.d:x results in broken requests like

http://a.b.c.d:x/v1/replicationcontrollers

I'm not sure exactly how that's happening. Perhaps APIPath should be set in the client config?

bryk:
How do you run the Dashboard? I run it like this and it works.

```
[bryk@bryk dist (master)]$ ./dashboard --apiserver-host=http://localhost:8080
2016/02/08 11:27:30 Starting HTTP server on port 9090
2016/02/08 11:27:30 Creating API client for http://localhost:8080
2016/02/08 11:27:30 Creating Heapster REST client for http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/heapster/api
```

therc:
This is the setup:
```
        - --apiserver-host=http://a.b.c.d:8080/
        - --port=9090
```
Your log message doesn't guarantee anything, because I get that far, too:

```
2016/02/05 15:52:18 Starting HTTP server on port 9090
2016/02/05 15:52:18 Creating API client for http://a.b.c.d:8080/
2016/02/05 15:52:18 Creating Heapster REST client for http://a.b.c.d:8080/api/v1/proxy/namespaces/kube-system/services/heapster/api
```

It's when you actually try to open the dashboard that you find it's broken:

```
2016/02/05 16:12:30 Incoming HTTP/1.1 GET /api/replicationcontrollers request from x:y
2016/02/05 16:12:30 Getting list of all replication controllers in the cluster
2016/02/05 16:12:30 the server could not find the requested resource (get replicationControllers)
```

There's no way to log what it's doing, so with the help of tcpdump I found that it's trying to hit `/v1/replicationcontrollers` on the apiserver, without the leading `/api`.

therc:
Now I'm wondering if this actually a bug in the client itself, not in the dashboard.

bryk:
You have trailing slash at apiserver host `http://a.b.c.d:8080/`. Maybe removing it can fix this?

therc:
Thanks... it's a bit crazy that the trailing slash makes a difference, but it does. I'm assuming I'll be the first of many.

bryk:
Thanks for reporting this! :) We will fix the problem or document the behavior.

bryk:
This is fixed. Apiserver host works with trailing slashes.

PR #382, opened 2016-02-16
Update godeps to latest HEAD

bryk:

PR #391, opened 2016-02-17
Added hostname args normalization for dashboard backend

floreks:
Related to #338 

Provided hostnames will be normalized to the form accepted by the k8s master.

Fixes problem with trailing slashes by removing them from end of the string. Also replaces any backslashes with slashes if there are any.

@maciaszczykm @bryk PTAL

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -23,6 +23,29 @@ import (
 	"github.com/spf13/pflag"
 )
 
+// Represents hostname string i.e. http://localhost:8080
+type Hostname string
+
+// Normalizes hostname by replacing any backslashes with slashes and removing
+// any trailing slashes
+func (host Hostname) normalize() string {
+	hostStr := string(host)
+
+	// Replace any backslashes with slashes
+	for i, char := range hostStr {
+		if char == '\\' {
+			hostStr = hostStr[:i] + string('/') + hostStr[i+1:]
+		}
+	}
+
+	// Remove any trailing slashes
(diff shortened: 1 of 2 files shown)


bryk:
I actually did a test on this a few days ago and `http://localhost:8080/` and `http://localhost:8080` work just fine. Maybe we've updated go client library in the meantime?

bryk:
Also, I think we want to keep the path in the host. E.g., with this change when somebody has master running at http://foo.bar:8080/my/api/prefix they cannot point to it. I could do this a few days ago and it worked. Try using kubectl proxy --api-prefix=/foo/bar/baz

floreks:
I'm not sure how this is related. What this will do is when someone provides `apiserver-host` as `http://a.b.c.d:8080/` it will change it to `http://a.b.c.d:8080` and then create apiserverClient based on trailed string. Basically everything that worked before should be working now + case that @therc described with `http://a.b.c.d:8080/` will also work.

bryk:
Yeah, but I've tested `http://a.b.c.d:8080/` and `http://a.b.c.d:8080` and they both work correctly on my machine. Even without this change.

floreks:
If this is confirmed then we can close issue and I will close this PR.

bryk:
Yeah, please verify this.

floreks:
I've checked how dashboard handle api hostname with trailing slashes and it works now. I guess #382 fixed this issue.

Closing.

bryk:
Awesome, thank you for investigation!

Issue #373, opened 2016-02-13
App name validation in deploy flow causes 500s


codegoblins:
<img width="1049" alt="screen shot 2016-02-12 at 3 29 09 pm" src="https://cloud.githubusercontent.com/assets/1609602/13024291/66a95436-d1a7-11e5-9cb3-8db2d4d06979.png">

Console output: 
```
2016/02/12 15:27:55 Validating runcmdtest34239i23 application name in jefsweeney namespace
2016/02/12 15:27:55 the server does not allow access to the requested resource (get replicationControllers runcmdtest34239i23)
2016/02/12 15:27:55 Outcoming response to 127.0.0.1:51717 with 500 status code
```
- This causes the form to be invalidated
- This happens in any namespace
- It seems only application names that start with "run" cause this, but I haven't tried everything

floreks:
What version of kubernetes are you running? Do you have newest version of dashboard deployed? 
Maybe you have a service with that name in this namespace already? Validation also checks for services. Probably error message could be more clear when existing service blocks creation of app.

I've tried to reproduce it and couldn't. I was able to create apps with same name in different namespaces. Name starting with `run...` was fine.

bryk:
@codegoblins What is your security setup? Are your namespaces restricted?

codegoblins:
@floreks Our cluster is running Kubernetes 1.1. I am on the latest master commit of dashboard. Also no matching services in the namespace.  

@bryk It doesn't look like our namespaces are restricted, I have no problems creating pods in other tenants' namespaces. 

Any suggestions how I can further debug this?

bryk:
I was able to reproduce this. Investigating....

bryk:
@codegoblins This is a bug in `kubectl proxy`. Can you maybe contribute to k8s core to fix this? Should be as easy as changing a regex.

floreks:
Ok, now I know why I couldn't reproduce the problem. I'm not using proxy. We should propose some kind of template for creating issues, so we can reproduce and fix problems faster.

bryk:
Yeah, templates like this:https://github.com/blog/2111-issue-and-pull-request-templates

floreks:
This looks really nice. We should prepare a template at some point.

PR #393, opened 2016-02-18
Assume valid when name validation errors.

bryk:
Addresses #373

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/393)

### DIFF
--- a/src/app/frontend/deploy/uniquename_directive.js
+++ b/src/app/frontend/deploy/uniquename_directive.js
@@ -68,7 +68,11 @@ function validate(name, namespace, resource, q) {
           deferred.reject();
         }
       },
-      () => { deferred.reject(); });
+      () => {
+        // On error assume that the name is valid. If it is not, the error is caught later in the
+        // deploy pipeline.
+        deferred.resolve();
+      });
 
   return deferred.promise;
 }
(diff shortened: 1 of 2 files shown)


bryk:
@codegoblins

bryk:
@codegoblins can you review?

codegoblins:
With this patch I was able to create a service & RC from the deploy flow called 'runtest':

screen shot 2016-02-18 at 1 28 14 pm

However, I am unable to do anything else from the dashboard e.g. edit pod count, view details:
screen shot 2016-02-18 at 1 30 15 pm

It seems like [kubernetes/kubernetes/21464](https://github.com/kubernetes/kubernetes/issues/21464) is the real culprit here. If we want to fix this from the frontend side, we're just pushing the problem further down the deployment/administration flow.

bryk:
> However, I am unable to do anything else from the dashboard e.g. edit pod count, view details:

True. You'll be able to access it when running the Dashboard inside a cluster.

bryk:
> It seems like kubernetes/kubernetes/21464 is the real culprit here. If we want to fix this from the frontend side, we're just pushing the problem further down the deployment/administration flow.

You're correct. The linked issue is the real problem. This PR, however, addresses problem with our UI where we assume that name is invalid when we cannot determine its correctness. I believe we should assume it is valid because potential problems will be caught later.

codegoblins:
Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/393)*

codegoblins:
Confirmed that the problem is nonexistant when running from within the cluster.

maciaszczykm:
Should it be merged?

bryk:
I say yes :) Merging.

PR #398, opened 2016-02-18
Fix for layout compatibility issues with Microsoft Edge

taimir:
So this is the first batch of changes I came up with to make the layout work in MS Edge. The biggest change was replacing `flex` with `flex="auto"` everywhere, because otherwise the parent elements were "collapsed" and had height 0 at a number of places throughout dashboard. AFAICS this does not influence the layout in Chrome.

Also there was some trouble with the list of replicaset cards and the "auto-inference" of the best height so that a grid can be enforced (cf. algorithm at *replicationcontrollerlistcontainer.js*).  The algorithm works fine, but it seems that Edge fails to adapt the  height of some the direct children of the rc-list container, even if the content enclosed by them is actually bigger in height (I guess some of the grand-children elements are floating). Therefore the cards were cut-off at the bottom, because the algorithm finds appropriate height for the direct children, which is actually not sufficient (since some parents are "shorter" than their content).

My approach to this was to introduce a small directive that expands the parent it's put in, so that the parent covers all children even if they are floating. Tell me what you think about this? 

** Side note: the algorithm that finds the appropriate height to enforce a "wrap" at the *replicationcontrollerlist* works, but seems a bit hacky to me? Maybe in the future we could actually use `md-grid-list` for this? I just stumbled across the code :)

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/398)

### DIFF
--- a/src/app/frontend/chrome/chrome.html
+++ b/src/app/frontend/chrome/chrome.html
@@ -23,10 +23,10 @@ limitations under the License.
       <h2>
         <span>kubernetes</span>
       </h2>
-      <div flex ui-view="toolbar"></div>
+      <div flex="auto" ui-view="toolbar"></div>
     </div>
   </md-toolbar>
-  <div class="kd-app-content-wrapper" ng-switch="ctrl.showLoadingSpinner">
+  <div class="kd-app-content-wrapper" ng-switch="ctrl.showLoadingSpinner" flex="auto" >
     <div ng-switch-when="true" class="kd-center-fixed">
       <md-progress-circular md-mode="indeterminate" md-diameter="96">
       </md-progress-circular>
--- a/src/app/frontend/deploy/createsecret.scss
+++ b/src/app/frontend/deploy/createsecret.scss
@@ -14,6 +14,10 @@
 
 @import '../variables';
 
+.kd-create-secret-dialog {
+  min-width: $layout-breakpoint-xs;
+}
+
 .kd-secret-data {
   overflow-y: scroll;
 }
--- a/src/app/frontend/deploy/deploy.scss
+++ b/src/app/frontend/deploy/deploy.scss
@@ -20,6 +20,7 @@
   // TODO(bryk): Find a better, application wide constant for the min width.
   max-width: 960px;
   min-width: 600px;
+  width: 50%;
 }
 
 .kd-deploy-form-title {
--- a/src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
+++ b/src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
@@ -17,7 +17,7 @@ limitations under the License.
 <md-card class="kd-replicationcontroller-card">
   <md-card-content>
     <div layout="column">
-      <div flex layout="row" layout-align="space-between center"
+      <div flex="auto" layout="row" layout-align="space-between center"
             class="kd-replicationcontroller-card-title-row">
         <span layout="row" flex="85">
           <md-icon class="material-icons md-warn kd-replicationcontroller-card-status-icon"
@@ -73,8 +73,8 @@ limitations under the License.
       <div layout="row" layout-wrap>
         <div ng-if="::ctrl.replicationController.description" flex="100" layout="column"
             class="kd-replicationcontroller-card-section">
-          <span flex class="kd-replicationcontroller-card-section-title">Description</span>
-          <div flex class="kd-replicationcontroller-card-description">{{::ctrl.replicationController.description}}</div>
+          <span flex="auto" class="kd-replicationcontroller-card-section-title">Description</span>
+          <div flex="auto" class="kd-replicationcontroller-card-description">{{::ctrl.replicationController.description}}</div>
         </div>
 
         <div flex="60" layout="column" class="kd-replicationcontroller-card-section">
@@ -97,7 +97,7 @@ limitations under the License.
 
         <div flex="60" layout="column" class="kd-replicationcontroller-card-section">
           <span flex="initial" class="kd-replicationcontroller-card-section-title">Internal endpoint</span>
-          <div flex>
+          <div flex="auto" >
             <div ng-repeat="endpoint in ::ctrl.replicationController.internalEndpoints track by $index">
               <kd-service-endpoint endpoint="::endpoint"></kd-service-endpoint>
             </div>
@@ -109,7 +109,7 @@ limitations under the License.
 
         <div flex="40" layout="column" class="kd-replicationcontroller-card-section">
           <span flex="initial" class="kd-replicationcontroller-card-section-title">External endpoint</span>
-          <div flex>
+          <div flex="auto" >
             <div ng-repeat="endpoint in ::ctrl.replicationController.externalEndpoints track by $index">
               <kd-service-endpoint endpoint="::endpoint"></kd-service-endpoint>
             </div>
@@ -122,7 +122,7 @@ limitations under the License.
       <div layout="column" ng-if="::ctrl.hasWarnings()">
         <md-divider class="kd-replicationcontroller-card-divider"></md-divider>
         <div layout-wrap>
-          <div flex class="kd-replicationcontroller-card-section"
+          <div flex="auto" class="kd-replicationcontroller-card-section"
                ng-repeat="warning in ::ctrl.replicationController.pods.warnings">
             <span class="kd-replicationcontroller-card-error">{{::warning.message}}</span>
           </div>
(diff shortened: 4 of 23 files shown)


taimir:
@bryk PTAL

bryk:
> ** Side note: the algorithm that finds the appropriate height to enforce a "wrap" at the replicationcontrollerlist works, but seems a bit hacky to me? Maybe in the future we could actually use md-grid-list for this? I just stumbled across the code :)

I'd be very happy to use md-grid-list, but it is not possible for out use case, unfortunately :( Any grid list requires width/height to be a multiplier of some X. In our case heights are variable.

src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
```diff
@@ -131,3 +131,4 @@
     </div>
   </md-card-content>
 </md-card>
+<kd-expand-parent></kd-expand-parent>
```
bryk: Can you just style the md-card to clear:both instead of introducing a new common component directive? This would be a much easier and lightweight :)

bryk:
@maciaszczykm Can you help with review?

bryk:
> So this is the first batch of changes I came up with to make the layout work in MS Edge. The biggest change was replacing flex with flex="auto" everywhere, because otherwise the parent elements were "collapsed" and had height 0 at a number of places throughout dashboard. AFAICS this does not influence the layout in Chrome.

What did this cause? The fact that elements were collapsed? Do you say we should be using flex=auto always?

bryk:
> Also there was some trouble with the list of replicaset cards and the "auto-inference" of the best height so that a grid can be enforced (cf. algorithm at replicationcontrollerlistcontainer.js). The algorithm works fine, but it seems that Edge fails to adapt the height of some the direct children of the rc-list container, even if the content enclosed by them is actually bigger in height (I guess some of the grand-children elements are floating). Therefore the cards were cut-off at the bottom, because the algorithm finds appropriate height for the direct children, which is actually not sufficient (since some parents are "shorter" than their content).

My approach to this was to introduce a small directive that expands the parent it's put in, so that the parent covers all children even if they are floating. Tell me what you think about this?

Another solution may be to add display:block and/or clear:both styles to elements that are too low in height.

bryk:
PTAL, thanks a lot for working on this :)

maciaszczykm:
> @maciaszczykm Can you help with review?

Yes, I will start soon.

src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
taimir: Yes, sure. I just thought we might want to actually reuse something like `clear: both` on a number of occasions where Edge fails to expand the parent, but you are right that a css class will do the same job and still be self-explanatory :) thanks

taimir:
>  What did this cause? The fact that elements were collapsed? Do you say we should be using flex=auto always? 

Here's what I figured out yesterday: when you specify a `flex` in any tag, angular automatically applies flexbox styles to the item, the default attributes for just `flex` are: `flex-grow: 1, flex-shrink: 1, flex-basis: 0`. Angular additionally applies css styles with `ms-` prefix for IE compatibility, which correspond to the ones above and have the same value.
With `flex-basis: 0` however, the flexbox container elements get shrunk in Edge / IE and are not expanded to cover their contents. If you specify `flex-basis: auto`, this is no longer the case and Edge displays everything correctly. So by setting `flex="auto"` in the html tags, I avoid this issue.

Note that this deduction results from empirical testing, since I do not know how Edge / IE actually implement the flexbox CSS standard.

Long story short: from what I've seen `flex="auto"` can be used throughout instead just `flex` to avoid issues with Edge.

bryk:
> Long story short: from what I've seen flex="auto" can be used throughout instead of just flex to avoid issues with Edge.

Thanks a lot for investigating this! I'll remember this while doing code reviews.

src/app/frontend/chrome/chrome.html
```diff
@@ -14,7 +14,7 @@
 limitations under the License.
 -->
 
-<md-content class="kd-content">
+<md-content class="kd-content" flex="auto" >
```
bryk: Flex wasnt here before. Can you verify that it is really needed here?

src/app/frontend/deploy/deploy.scss
```diff
@@ -20,6 +20,7 @@
   // TODO(bryk): Find a better, application wide constant for the min width.
   max-width: 960px;
   min-width: 600px;
+  width: 50%;
```
bryk: Why do we need this?

src/app/frontend/deploy/createsecret.scss
```diff
@@ -14,6 +14,10 @@
 
 @import '../variables';
 
+.kd-create-secret-dialog {
+  min-width: 600px;
```
bryk: Why do we need this? What was the problem? 

If we really need this can you use the 600px variable from variables.scss? the $layout-break-sm (or similar).

bryk:
Awesome. A few questions. PTAL

src/app/frontend/replicationcontrollerlist/zerostate/zerostate.scss
```diff
@@ -30,6 +30,7 @@
 
 .kd-zerostate-card-logo {
   height: 80px;
+  min-width: 80px;
```
maciaszczykm: Could you introduce variable for this `80px`?

maciaszczykm:
Tested on Windows 10 with Microsoft Edge 20.10240.16384.0 and it looks fine.

maciaszczykm:
However, there is some error on Travis: https://travis-ci.org/kubernetes/dashboard/builds/110365680

src/app/frontend/chrome/chrome.html
taimir: Let me see, maybe It's a leftover from my experiments with flex...

src/app/frontend/deploy/createsecret.scss
taimir: Regarding both of the `width` styles: MS-Edge and Chrome seem to have different behavior when it comes to the size of DOM elements. Chrome makes them wider, while Edge as narrow as possible. So leaving the `width` unset (e.g. in the two cases above) leads to very different representations in the two browsers (and in Edge it looks terrible #nosurprise).

taimir:
>background-color: $content-background;

The tests on travis hick-up here and say `$content-background` is not recognized. Am I missing something, I'm sure the variable exists?

maciaszczykm:
Hmm, I have tested also on current master and I can't see too much issues here. Are you sure all of these changes are needed?

taimir:
Let me see how it looks on my machine. I am positive without the changes it was a mess yesterday hah.

maciaszczykm:
image

image

image

But, of course, I might miss something.

taimir:
Well I am testing with IE 11, Document Mode Edge (I think that's how they call their compatib. setting in the developer console) ... and it looks like this with the current master:

taimir:
Uups, sorry guys ... misclicked the close button

maciaszczykm:
I think you should test with browser, not the plugin. You can download it here, like I did: https://dev.windows.com/en-us/microsoft-edge/tools/vms/windows/.

taimir:
O.K., I'll grab the IE 11 image from there as well ... but what's with those download speeds :(

maciaszczykm:
@taimir Yes, I know...

And still, this probably will be useful for IE11. I've tested on Edge, because it was only browser mentioned here.

taimir:
Mhm, thanks ... I didn't know there was a big difference between Edge and the Edge Document mode.

bryk:
We need to test both on IE and Edge. I expect Edge to behave more like chrome (problem-less) and IE should have problems.

taimir:
O.K, I can confirm that the problems seen in my images are indeed there in IE11 before my commit is applied, and that the commit resolves them. I tested with the image from the link @maciaszczykm referenced :)

maciaszczykm:
LGTM

bryk:
Reviewed 18 of 22 files at r1, 5 of 5 files at r2.

---

*[src/app/frontend/deploy/createsecret.scss, line 18 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/398#-KAtlNynsaxFFgd7JGVH-r1-18:-KBDX_jU8Kv9rX733Pa1:1388198179) ([raw file](https://github.com/kubernetes/dashboard/blob/22ea69c767417b045715497950f59c321e70d4aa/src/app/frontend/deploy/createsecret.scss#L18)):*
`#nosuprise` :) 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/398)*

PR #399, opened 2016-02-19
Link for external endpoints

cheld:
In demos, I want to show the running application after deployment in an easy way. I think this is also a natural behavior for users to deploy and then check the running application. 

Now, the user could copy & paste the text from the endpoint, but even this is not possible any longer as the text is potentially shortened. 

To solve that, I created a http link for external endpoints. It is also indicated by the mock-ups. 

After thinking about it for a while I realized that this might not be correct all the time.  Nodeports could be anything, e.g. a Minecraft server - not necessarily a HTTP server. In case of GCE load-balancer I am not sure. I assume it is only intended for HTTP...

After contemplating again, I still think I should add such a link. The advantages outweight the disadvantages. Maybe we style it a little bit differently. I am tempted to just add an icon next to the link, like in the mockup.

Any opinion?

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/399)


src/app/frontend/replicationcontrollerdetail/serviceendpoint_directive.js
```diff
@@ -22,6 +22,7 @@ export default function serviceEndpointDirective() {
     templateUrl: 'replicationcontrollerdetail/serviceendpoint.html',
     scope: {
       'endpoint': '=',
+      'url': '@',
```
bryk: When url param is '@' you'll show it as a link when the value is `foo` `true` `false` or `x`. How about changing it to `=`?

bryk: This parameter should have a more descriptive name. Url does not convey what it actually is. You wouldn't know that this is a boolean value to display the endpoint as a link. How about `is-http` or `is-external-http` or something else to say that this is an externally available http endpoint. Using this info we can display this as a link.

floreks: I was thinking about some heuristic that would actually establish if `endpoint` has something that is worth showing as a link. Checking common ports (HTTP/HTTPS) and maybe trying to actually check if this link points to something that browser can display.

Checking this as true in code would make link available for everything and this is not desired I guess.

What do you think?

bryk:
We need `open_in_new` icon next to the endpoints. Like this: 
selection_033

src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
```diff
@@ -111,7 +111,7 @@
           <span flex="initial" class="kd-replicationcontroller-card-section-title">External endpoint</span>
           <div flex>
             <div ng-repeat="endpoint in ::ctrl.replicationController.externalEndpoints track by $index">
-              <kd-service-endpoint endpoint="::endpoint"></kd-service-endpoint>
+              <kd-service-endpoint endpoint="::endpoint" url="true"></kd-service-endpoint>
```
bryk: Add an implementation note here that says why we always show external endpoints as links. Basically the same as you did in the PR comment. Use `<!-- .... --!>`

src/app/frontend/replicationcontrollerdetail/serviceendpoint_directive.js
bryk: Can you document what the param means? A short note would suffice.

bryk:
This is really helpful. Small change that brings a lot. Thanks! :)

src/app/frontend/replicationcontrollerdetail/serviceendpoint_directive.js
cheld: @floreks I think this is not feasible. We would have to ping the server. This would be slow and might do harm to it.

cheld:
what about separating the directive "kd-service-endpoint" in two directives instead of the boolean flag? e.g.:

kd-external-endpoint
kd-internal-endpoint

PTAL

bryk:
> what about separating the directive "kd-service-endpoint" in two directives instead of the boolean flag? e.g.:

Yes, this makes a lot of sense. I expect that we'll write a more sophisticated logic to detect whether to link-ify an endpoint and kd-external-endpoint directive would be a place for this. Please do this :)

cheld:
PTAL

src/app/frontend/replicationcontrollerdetail/internalendpoint.html
```diff
@@ -14,9 +14,10 @@
 limitations under the License.
 -->
 
+
```
floreks: Not needed here.

src/app/frontend/replicationcontrollerdetail/internalendpoint.html
```diff
@@ -14,9 +14,10 @@
...
 <div>
   <span ng-repeat="port in ::endpoint.ports">
-    <kd-middle-ellipsis display-string="{{::endpoint.host}}:{{::port.port}} {{::port.protocol}}">
-    </kd-middle-ellipsis>
+      <kd-middle-ellipsis display-string="{{::endpoint.host}}:{{::port.port}} {{::port.protocol}}">
+      </kd-middle-ellipsis>
```
floreks: Indent was already there. This is not needed.

cheld: Thanks

cheld:
PTAL

bryk:
PTAL, a few style comments. 

---

Reviewed 6 of 9 files at r2, 1 of 1 files at r3.
Review status: 6 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/app/frontend/replicationcontrollerdetail/externalendpoint_directive.js, line 20 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/399#-KBDa1x6ye-gPSl2c48N:-KBDa1x6ye-gPSl2c48O:-1113192730) ([raw file](https://github.com/kubernetes/dashboard/blob/799b8861aea5c122f9cbcd0be814e2ec657c557d/src/app/frontend/replicationcontrollerdetail/externalendpoint_directive.js#L20)):*
Rename to externalEndpoint...

---

*[src/app/frontend/replicationcontrollerdetail/internalendpoint_directive.js, line 20 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/399#-KBDa-Ut7yJfJtNKXXNm:-KBDa-UuytGg5IihKDSx:1184676020) ([raw file](https://github.com/kubernetes/dashboard/blob/799b8861aea5c122f9cbcd0be814e2ec657c557d/src/app/frontend/replicationcontrollerdetail/internalendpoint_directive.js#L20)):*
Rename to internalEndpoint...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/399#-:-KBDa6-iytdmlTFEzCC5:1897098372)*

cheld:
Thanks, PTAL

bryk:
:lgtm: 

---

Reviewed 1 of 9 files at r2, 4 of 4 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/399#-:-KBIRfMwlloaAJJcxKlB:980077423)*

PR #400, opened 2016-02-19
Create factor table for deploy from settings

f-higashi:
I created test factors for 'deploy from settings'.
Now I don't write specific conditions;ex. app name: most 63 characters, matching regex [a-z0-9]([-a-z0-9]*[a-z0-9])?).
Should I need to write them for all fields?

@cheld

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/400)


cheld:
I will review it

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Expected result at execution
+* In positive case the "Deploy" button must lead to the creation of a running application
+* Validation errors affecting particular fields must be displayed beneath these fields. In case the deployment  fails due to an exception, then a pop dialog is shown
+
+## Specific factors
+
+|Factor                      |1                 |2                                 |3               |4               |5                        |6                    |7               |Comment|
+|----------------------------|------------------|----------------------------------|----------------|----------------|-------------------------|---------------------|----------------|-------|
+|App name                    |Empty             |Correct value                     |Max-length      |Validation error|Same name app exists     |                     |                |       |
```
cheld: App name has more complexity:

1. The field is used to create a pod, RC, SVC and label. The pod has an auto-generated suffix, but SVC and RC get the name of this field. Existence should also check that part e.g. service exists, but RC not.. Think about if there are more factors
2. The app name is used as a label (think about if this could create some factors) 
3. used in the help text which does have some impact on layouting - I would add this as a comment

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Container image             |Empty             |Correct value                     |Max-length      |With version    |With hostname of registry|With port of registry|Validation error|       |
```
cheld: This field also creates a label. Think about if this creates more factors, e.g. are versions possible that are not correct label values? I would add a note in the comment field, at least.

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Number of pods              |0                 |Positive integer                  |Negative integer|Not integer     |Not a number             |                     |                |       |
```
cheld: Maybe we can reduce the text a bit, e.g. positive integer -> positive or. >1. Not integrer and not a number is almost the same, drop not integer.

max-value is missing. of course, this will kill your cluster. There should be some warning dialog, I guess.

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Port                        |0                 |Positive integer                  |Negative integer|Not integer     |Not a number             |                     |                |       |
```
cheld: I am wondering about border cases: port 1 or port 65535. I am also wondering about if the same port is mapped to different target ports (with two mapping rules)

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Target Port                 |0                 |Positive integer                  |Negative integer|Not integer     |Not a number             |                     |                |       |
```
cheld: I would suggest to remove this and the following line and collapse into one factor.

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Protocol                    |TCP               |UDP                               |                |                |                         |                     |                |       |
+|Expose service externally   |Checked           |Not checked                       |                |                |                         |                     |                |       |
```
cheld: local | vagrant or OpenStack | GCE

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Description                 |Empty             |Correct value                     |Max-length      |Validation error|                         |                     |                |       |
```
cheld: hmm..how is a description mapped to k8s resources? I could not retrieve it with kubectl. Can you investigate? Could be interesting to create factors

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Key(Labels)                 |Empty             |Correct value                     |Max-length      |Validation error|Existing key             |                     |                |       |
```
cheld: Labels have a strange prefix handling. Add it please

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Value(Labels)               |Empty             |Correct value                     |Max-length      |Validation error|                         |                     |                |       |
+|Namespace                   |Existing namespace|Select "Create a new namespace..."|                |                |                         |                     |                |       |
```
cheld: Can you include the subdialog. We will not make a separate sheet for that

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Image Pull Secret           |Existing secret   |Select "Create a new secret..."   |                |                |                         |                     |                |       |
```
cheld: Also please include

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|CPU requirement(cores)      |0                 |Positive integer                  |Negative integer|Not integer     |Not a number             |                     |                |       |
+|Memory requirement(MB)      |0                 |Positive integer                  |Negative integer|Not integer     |Not a number             |                     |                |       |
+|Run command                 |Empty             |Correct value                     |Max-length      |Validation error|                         |                     |                |       |
+|Run command arguments       |Empty             |Correct value                     |Max-length      |Validation error|                         |                     |                |       |
+|Run as privileged           |Checked           |Not checked                       |                |                |                         |                     |                |       |
```
cheld: Investigate details and add factors

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Key(Environment variables)  |Empty             |Correct value                     |Max-length      |Validation error|Existing key             |                     |                |       |
```
cheld: We should test if the environment variable is really set in the container (at least once). Add a comment

BTW: env variables are not displayed on the details page, maybe they should

cheld:
@f-higashi I think we should adapt our style of FBT to OpenSource. I think, the FBT sheet should be less systematic and more interesting to read. Currently, the sheet is somewhat boring to read. So, I would suggest to simplify the sheet. Reduce validation factors and add more business logic factors. We can think about dropping "normal input" and focus about unusual input. This way the tables would get smaller.

PTAL

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,47 @@
...
+|Value(Environment variables)|Empty             |Correct value                     |Max-length      |Validation error|                         |                     |                |       |
+|Action                      |Deploy            |Cancel                            |                |                |                         |                     |                |       |
+
```
cheld: I think we should add concurrency explicitly.

docs/test/deploy-from-settings.md
f-higashi: I used 'Not integer' such as 1.3.
but it is not good word. I use  'Floating point number'.

f-higashi:
@cheld 
I reduced systematic factors.
I don't finish research for some topic( Run command, Run as privileged).

But please check other topics  and give me your opinion.

docs/test/deploy-from-settings.md
```diff
@@ -0,0 +1,46 @@
...
+|Port                      |Border cases(1, 65535)                |Out of range(<0, >65535)           |Floating point number                                    |Not a number                       |Same port is mapped to different target ports|                                                                                |
+|Protocol                  |TCP                                   |UDP                                |                                                         |                                   |                                             |                                                                                |
+|Exporse service externally|lcoal                                 |vagrant or OpenStack               |GCE                                                      |                                   |                                             |                                                                                |
```
cheld: Sorry, I forgot...better this way

local | without load-balancer (vagrant) | with load-balancer (GCE)

cheld:
I think it is a good starting point. One small issue. Otherwise LGTM

f-higashi:
Thanks for your review.
I fixed the small issue.

bryk:
:lgtm: 

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/400#-:-KBIZxPvgwYwz9zkytCl:980077423)*

bryk:
Merging.

Issue #350, opened 2016-02-10
Manual testing


cheld:
We want to create a manual testing document. A process we have had best results in the past is what we call 'FBT' (Factor Based Testing).

In this approach test factors are defined. e.g.

* Browsers
* Input border cases
* Concurrency

Then we define variations of each factor (we call this level)
```
|-----------------|----1----------------|-----2------------------|
| browser         | IE                  | Firefox                |
| appname         | empty               | max length ...         | 
| concurrency     | create concurrently | delete concurrently ...|

....
```
Important to understand:
* the document does not state the expected outcome. This is intentional. It is up to the tester to decide if the behavior is a bug or not
* the tester is expected to do some multi-factor testing, e.g. IE + delete concurrently. The combination is up to his choice.

This is a sample document:
https://drive.google.com/open?id=1OJYM9gcR92msnSqxQ7ueCzpI5Hgdf4C0kEM2n0KoeMI

bryk:
I dont quite understand how I use the document to execute the test. Can you explain a little bit?

cheld:
I suggest we make a hangout call on friday. I am not in office tomorrow.

bryk:
OK

cheld:
Following the template in PR

https://github.com/kubernetes/dashboard/pull/376

As discussed with Piotr, we should create 1-2 additional sheets as a starting point

I would suggest:
- Deploy from settings
- ReplicationController list page

cc @digitalfishpond , @maciaszczykm , @floreks , @f-higashi 

Please pick if you have time

maciaszczykm:
I'll prepare tests for replication controller list page.

f-higashi:
I'll create 'Deploy from settings' sheet.

cheld:
Manual testing day!

So, we have the following test areas:
- Deploy from file. Covered by FBT
- Deploy from settings. Covered by FBT
- RC list. Covered by FBT
- RC details. Not yet covered by FBT.
- Logs. Not yet covered by FBT. 
(- ZeroState. Not yet covered by FBT.)
(- Deployment into different environments - actually this should be a factor  )

We can divide the work the following way:
- Fujitsu style: every developer picks one test area and test all factors, including all browsers (browsersync will help?)
- Google style: every developer tests all areas, but only one browser

How should we divide the work, any suggestion?

@bryk, @maciaszczykm, @digitalfishpond, @floreks, @f-higashi, @batikanu, @taimir

bryk:
[...]

PR #401, opened 2016-02-19
Add FBT for replication controller list

maciaszczykm:
Added FBT for replication controller list page. Connected to #350 issue.

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/401)

### DIFF
--- a/docs/test/replication-controller-list.md
+++ b/docs/test/replication-controller-list.md
@@ -1,0 +1,42 @@
+# Replication controller list
+
+## Prerequisites
+Cluster with few deployed applications, where: 
+
+* at least one is valid
+* at least one is invalid (for example with wrong image) 
+
+## Expected result
+Page displays correctly, especially:
+
+* cards are displayed correctly even when page is scaled down/up
+* application details are valid and matching data from `kubectl` command
+* logs menu display correct pods information
+* all application actions are available
+* invalid applications have warning symbols and error messages
+
+## Specific factors
+
+| Factor                               | 1                             | 2                                   | 3              | 4            | 5               | 6                         | 7          | 8      | Comment                                                                                     |
+|--------------------------------------|-------------------------------|-------------------------------------|----------------|--------------|-----------------|---------------------------|------------|--------|---------------------------------------------------------------------------------------------|
+| Replication controller count on page | 1                             | odd                                 | even           | 20-30        |                 |                           |            |        |                                                                                             |
+| Screen size                          | normal                        | mobile                              | narrow         | wide         | active resizing |                           |            |        |                                                                                             |
+| Replication controller status        | valid state                   | error state                         |                |              |                 |                           |            |        | error increases height of the card, error state can be caused by setting non-existing image |
+| Replication controller fields        | normal                        | max length                          | high pod count |              |                 |                           |            |        |                                                                                             |
+| Replication controller labels        | normal                        | long text (253 prefix and 63 label) | 1              | odd          | even            | short and long text mixed | 20-30      |        | labels increase height of the card                                                          |
+| Actions                              | view details                  | edit pod count                      | delete         | display logs |                 |                           |            |        |                                                                                             |
+| Edit pod count                       | negative                      | 0                                   | 1              | less than 10 | more than 30    | empty                     | scale down | cancel |                                                                                             |
+| Delete replication controller action | normal                        | with checkbox checked               | cancel         |              |                 |                           |            |        |                                                                                             |
+| Concurrency                          | delete replication controller | edit pod count                      |                |              |                 |                           |            |        |                                                                                             |
+
+## Common Factors
+
+| Factor          | 1                          | 2                                | 3  | 4      |
+|-----------------|----------------------------|----------------------------------|----|--------|
+| Browser         | Chrome                     | Firefox                          | IE | Safari |
+| Form            | Desktop                    | Mobile                           |    |        |
+| Help            | Hover over every `?`       | Click on every `Learn more` link |    |        |
+| Log entries     | check log entry on success | Check log entry on failure       |    |        |
+| Success message | Shown for every action     |                                  |    |        |
+| Failure message | Proper sentence            |                                  |    |        |
+| Concurrency     | UI refreshes after error   |                                  |    |        |


cheld:
@cheld I will review

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Specific factors
+
+| Factor                               | 1                                                | 2                                             | 3                                        | 4                                       | 5                                                                 | 6                                                                   | Comment |
+|--------------------------------------|--------------------------------------------------|-----------------------------------------------|------------------------------------------|-----------------------------------------|-------------------------------------------------------------------|---------------------------------------------------------------------|---------|
+| Replication controller count on page | 1 replication controller on wide page            | 1 replication controller on narrow page       | 3 replication controller on on wide page | 3 replication controller on narrow page | Any other positive number of replication controllers on wide page | Any other positive number of replication controllers on narrow page |         |
```
cheld: Layout | normal | mobile | narrow | wide | active resizing

RC count | 0 (?) | 1 | odd | even | 20-30

cheld: I would suggest to keep layout and rc count separated as two factors. Easy for tester to combine

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,39 @@
...
+| Replication controller status        | Valid                                            | Invalid (for example with not-existing image) |                                          |                                         |                                                                   |                                                                     |         |
```
cheld: RC status  | valid | error    | Comment: error increases height of card

Less text is better, I think

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,39 @@
...
+| Replication controller data          | Image string length (short or too long for card) | Age                                           | Logs (with links)                        | Internal and external endpoints         | Number of running and desired pods                                | Labels                                                              |         |
```
cheld: RC fields | normal | max-lenght | high pod count
Label  | normal | long-text (253 prefix + 63 label) | 1 | odd | even | short-text & long-text mixed | 20 -30       Comment: labels increase the height of the card

- I split normal fields and label, because label has much more complexity. 
- I droped age, because I did not see enough complexity
- Pod count has some dependency to logs combo box. e.g. it seems that only the first 10 logs are shown even if the pod count is higher. A bug? Because of this dependency, I kept the field

cheld: Ok, I just noticed you have pod count as a separate factor below. Maybe, remove it here.

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,39 @@
...
+| Actions                              | View details                                     | Edit pod count                                | Delete                                   | Logs                                    |                                                                   |                                                                     |         |
```
cheld: hmm...a bit boring - remove?

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,39 @@
...
+| Edit pod count action                | Non-positive value                               | 1                                             | Value less than 10                       | Value bigger than 30                    | String value                                                      | Empty field                                                         |         |
```
cheld: Edit pod count | 1 | negative | < 10 | > 30 | empty | scale-down

String value is not possible to enter - so remove

BTW: empty case is a bug

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,39 @@
...
+| Delete replication controller action | Okay button                                      | Cancel button                                 | Delete service checkbox checked          | Delete service checkbox not checked     |                                                                   |                                                                     |         |
```
cheld: Ok, I would shorten the text a bit. Also, there are much more test cases about this check box. Investigate a bit of the algorithm and find more levels

cheld:
I gave some comments to give you some ideas. I would shorten the text - unless it is not understandable. 

I am not sure if we always should state normal cases like '1', 'short', 'ok'. Maybe we should remove them. They are obvious anyway. We add them just for completeness. But I am on the fence and undecided. 
'Cancel' seems superfluous too, but because they are never tested it happens from time to time that they do not work. So, 'cancel' I would rather leave in the sheet

cheld:
I am not sure about concurrency. Maybe we should de-normalize and add it as a specific factor. Delete concurrently, scale pods up & down concurrently...

I think we should do that otherwise it is forgotten

cheld:
BTW: found 3 bugs during review :)

docs/test/replication-controller-list.md
maciaszczykm: Good point!

maciaszczykm:
@cheld Thank you for review, I've applied modifications. PTAL

cheld:
I like it. 

LGTM

docs/test/replication-controller-list.md
```diff
@@ -0,0 +1,42 @@
...
+| Actions                              | view details                 | edit pod count                      | delete          | display logs |                 |                           |            |        |                                                                                             |
+| Edit pod count                       | negative                     | 0                                   | 1               | less than 10 | more than 30    | empty                     | scale down | cancel |                                                                                             |
+| Delete replication controller action | normal                       | with checkbox checked               | cancel          |              |                 |                           |            |        |                                                                                             |
```
cheld: I don't know the algorithm in details, but maybe a deployment is possible (using kubectl) that will produce something like the following:
* the wrong service is deleted
* no service is deleted
* too much is deleted
* error
* nothing happens although the service should be deleted

IMHO, the factor does not encourage deep enough testing. Where should be put this tests? How to describe the factors? 

BTW: The checkbox is also shown if no service exits - I would say this is a bug.

cheld:
One last comment. Otherwise, I like it

docs/test/replication-controller-list.md
maciaszczykm: Yes, but if we don't want to combine factors then there are only two possibilities - checked or not.  We must remember, that this factor will be combined with other ones. There is a possibility of adding factor telling if replication controllers have services or not, but it's not visible from this view... It's very problematic from my point of view.

cheld: Ok, let's postpone it. We can add it later.

cheld:
LGTM

maciaszczykm:
@bryk PTAL

bryk:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[docs/test/replication-controller-list.md, line 25 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/401#-KB8TcKTdly1Jk59L3I-:-KB8TcKTdly1Jk59L3I0:-496683965) ([raw file](https://github.com/kubernetes/dashboard/blob/f352943cd14ddfc0857fa27b4e18a78859d43f5a/docs/test/replication-controller-list.md#L25)):*
Typo: "hight"

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/401)*

maciaszczykm:
@bryk Fixed typos. PTAL

bryk:
LGTM

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/401#-:-KBC-KoYnYACyFacSyEW:676834233)*

PR #403, opened 2016-02-19
Fix to parse containerImage to get version

f-higashi:
When I use private.registry:5000/nginx for container image, version label is set '5000/nginx' and it has error.
Fixed it.

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/403)

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -19,6 +19,7 @@ import {
   stateName as replicationcontrollerliststate,
 } from 'replicationcontrollerlist/replicationcontrollerlist_state';
 import {uniqueNameValidationKey} from './uniquename_directive';
+import DockerImageReference from '../common/docker/dockerimagereference';
 
 // Label keys for predefined labels
 const APP_LABEL_KEY = 'app';
@@ -345,25 +346,7 @@ export default class DeployFromSettingsController {
    * @return {string}
    * @private
    */
-  getContainerImageVersion_() {
-    /** @type {number} */
-    let index = (this.containerImage || '').lastIndexOf('/');
-
-    let imagename;
-    if (index > -1) {
-      imagename = this.containerImage.substring(index + 1);
-    } else {
-      imagename = this.containerImage;
-    }
-
-    index = (imagename || '').lastIndexOf(':');
-
-    if (index > -1) {
-      return imagename.substring(index + 1);
-    }
-
-    return '';
-  }
+  getContainerImageVersion_() { return new DockerImageReference(this.containerImage).tag(); }
 
   /**
    * Returns application name.
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -351,9 +351,9 @@ export default class DeployFromSettingsController {
 
     let imagename;
     if (index > -1) {
-	imagename = this.containerImage.substring(index + 1);
+      imagename = this.containerImage.substring(index + 1);
     } else {
-	imagename = this.containerImage;
+      imagename = this.containerImage;
     }
 
     index = (imagename || '').lastIndexOf(':');
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -347,10 +347,19 @@ export default class DeployFromSettingsController {
    */
   getContainerImageVersion_() {
     /** @type {number} */
-    let index = (this.containerImage || '').lastIndexOf(':');
+    let index = (this.containerImage || '').lastIndexOf('/');
+
+    let imagename;
+    if (index > -1) {
+	imagename = this.containerImage.substring(index + 1);
+    } else {
+	imagename = this.containerImage;
+    }
+
+    index = (imagename || '').lastIndexOf(':');
 
     if (index > -1) {
-      return this.containerImage.substring(index + 1);
+      return imagename.substring(index + 1);
     }
 
     return '';
--- a/src/app/frontend/common/docker/dockerimagereference.js
+++ b/src/app/frontend/common/docker/dockerimagereference.js
@@ -55,6 +55,6 @@ export default class DockerImageIdentifier {
     if (index > -1) {
       return last_block.substring(index + 1);
     }
-    return ''
+    return '';
   }
 }
--- a/src/app/frontend/common/docker/dockerimagereference.js
+++ b/src/app/frontend/common/docker/dockerimagereference.js
@@ -1,0 +1,60 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Parser to get TAG from reference of Docker container image
+ *
+ * Docker Container image can be set with the following value:
+ *   [REGISTRY_HOST[:REGISTRY_PORT]/]NAME[:TAG]
+ * REGISTRY can be omitted and NAME can contain '/' as delimiters for NAMESPACE.
+ *
+ * So only TAG is specified.
+ *
+ */
+
+export default class DockerImageIdentifier {
+
+  /**
+   * Constructs DockerUri object.
+   *
+   * @param {string} identifier
+   */
+  constructor(identifier = '') { this.identifier = identifier; }
+
+  /**
+   * Returns tag
+   *
+   * TAG cannot contain ':' and '/'.
+   * Not if TAG is omitted, last block text(separated by '/') contains TAG.
+   * @return {string}
+   */
+  tag() {
+    /** @type {number} **/
+    let index = this.identifier.lastIndexOf('/');
+
+    /** @type {string} **/
+    let last_block;
+    if (index > -1) {
+      last_block = this.identifier.substring(index + 1);
+    } else {
+      last_block = this.identifier
+    }
+
+    index = last_block.lastIndexOf(':');
+    if (index > -1) {
+      return last_block.substring(index + 1);
+    }
+    return ''
+  }
+}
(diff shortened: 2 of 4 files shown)


src/app/frontend/deploy/deployfromsettings_controller.js
```diff
@@ -347,10 +347,19 @@ export default class DeployFromSettingsController {
    */
   getContainerImageVersion_() {
     /** @type {number} */
-    let index = (this.containerImage || '').lastIndexOf(':');
+    let index = (this.containerImage || '').lastIndexOf('/');
```
bryk: You know what...? I think we'll never be fully correct when doing manual URI parsing. Can we maybe use library https://github.com/derek-watson/jsUri to do the URI parsing for us? You can include the library in bower and use it straight away.

bryk: Or am I wrong? Can you convince me and other code readers that this algorithm works?

floreks: Second test actually shows that even if version is provided in the link it will be parsed as empty string, so potential version information is lost. Moreover checking only for `:` or `/` or some other char will be measured only for certains cases and not fully generic.

This library is a good idea and I'd go with that.

cheld: In our case there is special semantic attached to the URI. From my experience, parsers are often designed for simple http urls and are buggy for other use cases. But, we should give it a try. In any case I would be most convinced by explicit test cases. The tests in this PR are very nice, but not comprehensive to cover every possible URI

bryk: Yes, can we have more test cases for this feature?

f-higashi:
Okay, 
I'll try this library and add more test cases.

bryk:
Awesome. Just comment when you're ready for rereview.

f-higashi:
I checked jsUri and docker command.
But I think my idea is still simpler and easier to understand because using jsUri seems to cause more complicated condition.



'Container image' can be set with the following value:
[REGISTRY_HOST[:REGISTRY_PORT]/]NAME[:TAG]
cf) https://github.com/docker/docker/blob/master/docs/reference/commandline/pull.md

REGISTRY can be omitted and NAME can contain '/' as delimiters for namespace.

ex)
1. 'Container image' contains registry url, namespace and tag
registry:5000/namespace/test:1
a result of jsUri is:

      host: registry
      port: 5000
      path: namespace/test:1

We can get 1 by parsing path   
2. 'Container image' contains namespace and tag but does not contain registry url
namespace/test:1

    host: namespace
    path: test:1

We can get 1 by parsing path. But host property is not corrected. 
It misleads developers.
3. 'Container image' contains tag but does not contain registry and namespace
test:1

    host: test
    port: 1
    path: (empty)

In this case, we need to use existing way.


TAG cannot contain ':' and '/'. So it is no problem to get last '/'.

cf) https://github.com/docker/docker/blob/master/image/spec/v1.md

What do you think?

cheld:
I would suggest to move the uri-logic to its own class, something like dockeruri.js Add meaningful methods in context of docker. If you use a library for parsing is an implementation detail, then.

bryk:
Yeah, I like what Christoph said. Move the parsing logic to a separate file and do not use the library. In your code add implementation comments that explain why the code is correct (i.e., the way you explained this to us now). Add a few more tests and let's submit this. 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-:-KBCC7G43ctBGjHP1HC5:746657075)*

f-higashi:
PTAL

I moved logic to separate file.
A format of Image name is not URI. I cannot find the name in docker documents.
So I named DockerImageReference.

If you have suitable name, please tell me.

f-higashi:
In my environment, 'gulp format' changes many files. (Now I modified format manually..)

For example)
```
--- a/src/app/frontend/common/filters/filters_module.js
+++ b/src/app/frontend/common/filters/filters_module.js
@@ -20,12 +20,11 @@ import relativeTimeFilter from './relativetime_filter';
 /**
  * Module containing common filters for the application.
  */
-export default angular
-    .module(
-        'kubernetesDashboard.common.filters',
-        [
-          'ngMaterial',
-        ])
+export default angular.module(
+                          'kubernetesDashboard.common.filters',
+                          [
+                            'ngMaterial',
+                          ])
     .filter('middleEllipsis', middleEllipsisFilter)
     .filter('kdMemory', memoryFilter)
     .filter('kdCores', coresFilter)
```

It seems to be affected by other setting. 

My environment:
- 14.04.1-Ubuntu


Where should I check?
Do you have any idea?

bryk:
There may be a new formatter release. If yes, I'll quickly fix this. 

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-:-KBHsMU9OQQHxPrAI6Qd:985506018)*

f-higashi:
The above problem was solved.
After deleting files which is written in .gitignore, it works.

PTAL

bryk:
A few style comments and let's merge :)

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/app/frontend/common/docker/dockerimagereference.js, line 22 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHtslfDmbJOtARlzAJ:-KBHtslfDmbJOtARlzAK:-1447719176) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/app/frontend/common/docker/dockerimagereference.js#L22)):*
Maybe delete this line? I cannot understand it. 

---

*[src/app/frontend/common/docker/dockerimagereference.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHu2M8sm89pDyYlqBn:-KBHu2M8sm89pDyYlqBo:7761806) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/app/frontend/common/docker/dockerimagereference.js#L32)):*
Should this have default value? I think not - it is not needed :) 

---

*[src/app/frontend/common/docker/dockerimagereference.js, line 38 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHuROCTUaGk-9ObSCu:-KBHuROCTUaGk-9ObSCv:1739966372) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/app/frontend/common/docker/dockerimagereference.js#L38)):*
Typos: "If TAG is not omitted"

---

*[src/test/frontend/common/docker/dockerimagereference_test.js, line 17 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHumJP0tM87TnFJg4A:-KBHumJP0tM87TnFJg4B:1362895675) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/test/frontend/common/docker/dockerimagereference_test.js#L17)):*
How about "testImage:" case? What will it return? And ":"? And "::"? And "/"? 

---

*[src/test/frontend/common/docker/dockerimagereference_test.js, line 65 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHsifbb4y3p4iZmjCU:-KBHsifbb4y3p4iZmjCV:1324461051) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/test/frontend/common/docker/dockerimagereference_test.js#L65)):*
This is repeated test. Identical as previous. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-:-KBHv8pa0TDqcyI0x_9p:-1697847062)*

f-higashi:
I fixed it. 

PTAL

bryk:
Last comment PTAL. Then we can merge :)

---

Reviewed 1 of 2 files at r1, 1 of 3 files at r2, 1 of 2 files at r5.
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/common/docker/dockerimagereference.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHu2M8sm89pDyYlqBn:-KBMWkBp19I2ZJp7EbmC:1792042598) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/app/frontend/common/docker/dockerimagereference.js#L32)):*
You need to add `@type` information to the this.identifier field. It should also be private (should end with trailing _). Like this. 
```
/**
 * @private {string}
 */
this.identifier_ = identifier;
```
PTAL :) 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-:-KBMXK9tgpsIQ6i7CNSj:-1687566691)*

f-higashi:
I added @type information.
And identifier is old name. I use reference in test code. 
But I missed to change some part.  Fix it too.

---

Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/common/docker/dockerimagereference.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-KBHu2M8sm89pDyYlqBn:-KBMd3_n-zi8pUlt-bqb:-2016192950) ([raw file](https://github.com/kubernetes/dashboard/blob/98878ea1674a86c9001dc2a54606dc8d47da45ef/src/app/frontend/common/docker/dockerimagereference.js#L32)):*
I added @type information.
And identifier is old name. I use reference in test code. 
But I missed to change some part.  Fix it too.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-:-KBMf9iy0uoMRQI5GFhg:-1589059136)*

f-higashi:
I found one issue.
Newer docker reference can has digest information.

```
[REGISTRY_HOST[:REGISTRY_PORT]/]NAME[:TAG][@DIGEST]
```

Is the issue fixed in this pull request?
Or this pull request merge and create new issue?

What do you think?

bryk:
> Is the issue fixed in this pull request?
Or this pull request merge and create new issue?
What do you think?

I say merge this and solve the problem in a new pull request.

bryk:
Reviewed 1 of 1 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403)*

bryk:
:lgtm: 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/403#-:-KBNv_tmAHxF8HcQ7qrK:980077423)*

PR #413, opened 2016-02-22
Add Rolling Update endpoint for RC

titilambert:
[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/413)

### DIFF
--- a/src/app/backend/apihandler.go
+++ b/src/app/backend/apihandler.go
@@ -117,6 +117,10 @@ func CreateHttpApiHandler(client *client.Client, heapsterClient HeapsterClient,
 			To(apiHandler.handleUpdateReplicasCount).
 			Reads(ReplicationControllerSpec{}))
 	replicationControllerWs.Route(
+		replicationControllerWs.POST("/{namespace}/{replicationController}/rolling-update").
+			To(apiHandler.handleRollingUpdate).
+			Reads(AppDeploymentFromFileSpec{}))
+	replicationControllerWs.Route(
 		replicationControllerWs.DELETE("/{namespace}/{replicationController}").
 			To(apiHandler.handleDeleteReplicationController))
 	replicationControllerWs.Route(
@@ -305,6 +309,26 @@ func (apiHandler *ApiHandler) handleUpdateReplicasCount(
 
 	if err := UpdateReplicasCount(apiHandler.client, namespace, replicationControllerName,
 		replicationControllerSpec); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	response.WriteHeader(http.StatusAccepted)
+}
+
+// Handles Rolling update of Replication Controller.
+func (apiHandler *ApiHandler) handleRollingUpdate(
+	request *restful.Request, response *restful.Response) {
+
+	oldRcName := request.PathParameter("replicationController")
+	namespace := request.PathParameter("namespace")
+
+	deploymentSpec := new(AppRollingUpdateFromFileSpec)
+	if err := request.ReadEntity(deploymentSpec); err != nil {
+		handleInternalError(response, err)
+		return
+	}
+	if err := RollingUpdateReplicationController(apiHandler.client, namespace, oldRcName, deploymentSpec); err != nil {
 		handleInternalError(response, err)
 		return
 	}
--- a/src/app/backend/replicationcontrollerupdate.go
+++ b/src/app/backend/replicationcontrollerupdate.go
@@ -1,0 +1,153 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"errors"
+	"strings"
+	"time"
+
+	"k8s.io/kubernetes/pkg/api"
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/kubectl"
+	cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
+	kubectlResource "k8s.io/kubernetes/pkg/kubectl/resource"
+	"k8s.io/kubernetes/pkg/util/intstr"
+)
+
+// Specification for deployment from file
+type AppRollingUpdateFromFileSpec struct {
+	// Name of the file
+	Name string `json:"name"`
+
+	// File content
+	Content string `json:"content"`
+
+	// Period
+	UpdatePeriod int64 `json:"update-period"`
+
+	// Interval
+	PollInterval int64 `json:"poll-interval"`
+
+	// Period
+	Timeout int64 `json:"timeout"`
+}
+
+/*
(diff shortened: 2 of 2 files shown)


bryk:
cc @cheld @floreks @maciaszczykm

bryk:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/backend/replicationcontrollerupdate.go, line 147 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/413#-KBDZI_aCHIlIKu1cwJI:-KBDZI_bOqlbmeFmKWoI:1422504310) ([raw file](https://github.com/kubernetes/dashboard/blob/b5d63345979ab0f0dd15a613d8b2d92d458cd4f3/src/app/backend/replicationcontrollerupdate.go#L147)):*
Is this a long lasting operation? Does this return immediately and execute in the background? 

cc @floreks

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/413)*

floreks:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/backend/replicationcontrollerupdate.go, line 147 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/413#-KBDZI_aCHIlIKu1cwJI:-KBD_qBxXaY-BeG4v6DY:1480130494) ([raw file](https://github.com/kubernetes/dashboard/blob/b5d63345979ab0f0dd15a613d8b2d92d458cd4f3/src/app/backend/replicationcontrollerupdate.go#L147)):*
Looking at the code: https://github.com/kubernetes/contrib/blob/master/Ingress/Godeps/_workspace/src/k8s.io/kubernetes/pkg/kubectl/rolling_updater.go#L155
I'd say that this is a long lasting operation. I'm not sure if we should run it in parallel. Error handling would be hard.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/413)*

titilambert:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/backend/replicationcontrollerupdate.go, line 147 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/413#-KBDZI_aCHIlIKu1cwJI:-KBDhbPl3t3tn0Nc8Toc:-1896447515) ([raw file](https://github.com/kubernetes/dashboard/blob/b5d63345979ab0f0dd15a613d8b2d92d458cd4f3/src/app/backend/replicationcontrollerupdate.go#L147)):*
This function is blocking. Do you prefer a none blocking function ?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/413)*

bryk:
@titilambert Give us some time. This week is really busy. Please ping if you feel this PR is jammed.

titilambert:
@bryk Thanks for your reply !

titilambert:
@bryk Hello ! Any news on this one ?

floreks:
@bryk will be unavailable for the next few days. We'll check and prioritize all issues and PR once he's back. You've got to wait few more days :)

titilambert:
@floreks ok, thanks !

bryk:
@titilambert Thanks for your patience. Let's move forward with this.

bryk:
So, this PR would be a breeze, except for one problem that we haven't solved yet. This is; how do we handle long-lasting operations? E.g., rolling update that may take 10+ minutes.

How do we handle this in the UI? There are a few options.

1. One is to ignore this fact and just block the UI. But is is like in 90s where you had these modals saying "DO NOT REFRESH THIS PAGE WHEN SPINNER IS ON".
1. The next option is to implement a framework for the UI to handle the operations and save them somewhere and then handle this on the client somehow. This is a large and non-trivial task.
1. The last option I see is to avoid doing this on the UI side and just call what API allows us to call. In this case we'd support rolling updates only for deployments: https://github.com/kubernetes/kubernetes/blob/release-1.2/docs/proposals/deployment.md. This would simplify *a lot* of things on the UI side, but will allow for rolling updates only for deployments. 

@olekzabl @floreks @cheld @maciaszczykm @titilambert I need your opinions on this.

bryk:
CC @bgrant0607-cc @kubernetes/sig-api-machinery Rolling updates in UI: Embed kubectl? Implement long operations handling? Or use what apiserver provides? See previous comment

titilambert:
@bryk Why just launch rolling update in a go routine server-side ?
The UI could just list all running jobs.

bryk:
> @bryk Why just launch rolling update in a go routine server-side ?
The UI could just list all running jobs.

We could do this, sure. But such design has serious flaws, which we should not be ignoring. Basically, everything boils down to the fact that the UI is a stateless distributed system. For example, you can have two replicas of the UI. Then one would have the running goroutine and the other would not. How would you sync this? Other thing: what to do when the UI crashes/is killed in the middle of a rolling update? This happens often in K8s world. 
We can, of course, solve all these problems. This has been done many times in the past. But this requires a lot of work, which I'm not sure we should be doing.

titilambert:
@bryk I'agree with you, but it's kubectl issue not UI.
In my opinion, the UI acts as a kubectl (with a beautiful web UI :)) ( but maybe I should not to ? )
So if a rolling update crashes, you just have to relaunch the rolling update, like with kubectl.

And of course, I'm agree with you, all these issues will be fixed with deployment stuff, but this PR could be the first rolling update implementation before switching to the server-side one...

lavalamp:
Why not use deployment for this?

lavalamp:
We need to move more items to be done server-side for exactly the reason you want. But we can't just simply run kubectl from a server. Deployment solves the particular use case here, so I'd actually prefer not to have rolling-update as a primitive at all any more.

bgrant0607:
Yes, we should move the UI to Deployment.

I didn't want to block on that for the initial version, because Deployment wasn't ready, but now it is. Rolling update should only be supported via Deployment.

cc @janetkuo @pwittrock

bgrant0607:
How does the UI handle multiple replication controllers for the same service?

FYI, we've moved `kubectl run` to Deployment by default in 1.2.

Also, something that may or may not be useful for other things:
https://github.com/rancher/kubectld

bgrant0607:
FYI: https://github.com/kubernetes/kubernetes/issues/12143

bryk:
@kubernetes/dashboard-maintainers Comments?

floreks:
In my opinion we should avoid adding functionalities that would be removed soon anyway. It's best to work on good solution which rolling update via deployments seems to be. This PR is really nice but it introduces some major problems with UI handling which we're trying to avoid.

maciaszczykm:
I agree with you, first and second option seem to be risky, so we should go for rolling update via deployments. Now the question is how we will support them and other kind of objects?

titilambert:
Oki, I'm agree with all of you.
I just wanted this PR to be able to do rolling updates from UI.
So the actual stuff to do, is to implement deployment object handler. So do have any comments before I starts this stuff ?

bgrant0607:
@titilambert 

@janetkuo just updated the Deployment documentation for 1.2:
http://kubernetes.io/docs/user-guide/deployments/

Let us know if you have questions.

cheld:
We could put the rolling update in a job ;-). If we use deployments, what happens if dashboard is deployed in a old cluster? (There are a lot of old clusters, because Kubernetes upgrades are not easy, yet)

bgrant0607:
kubectl run uses the discovery API to determine whether Deployment is supported by the apiserver or not.

bgrant0607:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/run.go#L162

bryk:
> So, the actual stuff to do is to implement deployment object handler. So do have any comments before I start this stuff ?

We need to come up with a proposal on how to scale the UX to handle many new types of resources to display (e.g., deployments, daemon sets, etc.). We should have a proposal next week. Can you wait till then?

titilambert:
@bryk Yeah ! of course ! :) I would love see the next UI/UX designs :)

bryk:
So the final outcome is: rolling update will be implemented for deployments. Deployments support will be added soon to the UI.

Closing. Please comment if this needs further discussion.

PR #461, opened 2016-02-29
Fix logs toolbar fields placement

### LINKED ISSUE #455: Header field name and value are not aligned in log page.
#### Issue details

Header field name and value are not aligned in logs page.

##### Environment

Dashboard version: cffd34a90ae4a2334db4f062202f3a00fdf73dc6 ( I used gulp serve:product)
Kubernetes version: v1.1.2
Operating system: Ubuntu 14.0.4
Node.js version: v5.2.0
Go version: 1.5.2
Browser: IE

##### Steps to reproduce

1. Open Log page

##### Observed result

ds_log1

##### Expected result

Header field name and value are aligned in log page.

ds_log2

maciaszczykm:
Fixes #455 issue.

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/461)

### DIFF
--- a/src/app/frontend/logs/logstoolbar/logstoolbar.scss
+++ b/src/app/frontend/logs/logstoolbar/logstoolbar.scss
@@ -51,4 +51,12 @@ md-toolbar {
       }
     }
   }
+
+  // Fixes pickers placement issue on Internet Explorer. It's caused by Angular Material
+  // stylesheets, which apply 'transform: translate3d(0, 2px, 0)' on 'md-select-value' elements.
+  .md-select-value {
+    .kd-logs-toolbar-text {
+      -ms-transform: translateY(-1px);
+    }
+  }
 }
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -50,7 +50,11 @@ rules:
   no-transition-all: 2
   # Disabled because the project includes external fonts.
   no-url-protocols: 0
-  no-vendor-prefixes: 2
+  # Excluded Internet Explorer prefix (ms), because some of bug fixes require it.
+  no-vendor-prefixes:
+      - 2
+      - additional-identifiers: []
+        excluded-identifiers: [ms]
   no-warn: 2
   property-units: 2
 


src/app/frontend/logs/logstoolbar/logstoolbar.scss
```diff
@@ -51,4 +51,11 @@ md-toolbar {
       }
     }
   }
+
+  // Fixes pickers placement issue on Internet Explorer.
+  .md-select-value {
+    *:first-child{
+      -ms-transform: translateY(0.75px);
```
maciaszczykm: You will probably ask, why exactly `0.75px` and honestly I'm not sure. I know it's only value, which makes it look good. I don't understand IE, and why translate works different here.

maciaszczykm:
Travis fails due to following error:

```
src/app/frontend/logs/logstoolbar/logstoolbar.scss
  49:7  warning  Nesting depth 3 greater than max of 2  nesting-depth
  58:7  error    Vendor prefixes should not be used     no-vendor-prefixes

✖ 2 problems (1 error, 1 warning)
```

We can fix this issue by using vendor prefix for IE. Other solution, which doesn't look so nice is to set transform to `none` for all vendors.

maciaszczykm:
Current state:

image

I had to change `no-vendor-prefixes` rule to `0`. Is it acceptable or should I search for another workaround?

bryk:
Hmm... Do you want to add something only on IE? I'm asking because vendor prefixes are automatically added. 

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/logs/logstoolbar/logstoolbar.scss, line 58 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-KBhHA8vfelgfFAJC4Nz-r1-58:-KBhchQu-zQzIkdaLtBn:2018049459) ([raw file](https://github.com/kubernetes/dashboard/blob/81a0c6db125f0ffc2993d1640e6ac9760579bd04/src/app/frontend/logs/logstoolbar/logstoolbar.scss#L58)):*
Hmm... Why translatingY? Changing display property does not help? Line height? Or that kind of things? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBhcuP_LOTCASsw3Ope:-942695002)*

maciaszczykm:
Yes, this issue is caused by `transform` set on `md-select-value` by default in Angular Material. Behavior of `translate` is different for IE than for other browsers. To fix it we can set same `transform` for all browsers or just for IE. Second options looks better, because if we want to change value for all browsers it has to be changed to one value (I assume `0` looks acceptable, but still not the best).

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/logs/logstoolbar/logstoolbar.scss, line 58 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-KBhHA8vfelgfFAJC4Nz-r1-58:-KBhgaIOtMiIiQUT15Zq:469394728) ([raw file](https://github.com/kubernetes/dashboard/blob/81a0c6db125f0ffc2993d1640e6ac9760579bd04/src/app/frontend/logs/logstoolbar/logstoolbar.scss#L58)):*
Unfortunately not... We need to override `transform` to fix it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBhflfUiiITu3AOYtQ6:1677581112)*

maciaszczykm:
My explanation aboive is a bit messy... sorry for it! If you need more details, just ask :)

bryk:
Can you send me a link to the lines in angularmaterial that cause problems? 

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBi8Ko5X0FRWT-58C-4:800282727)*

bryk:
Isnt this already fixed by https://github.com/angular/material/pull/6741/files ?

Can you verify the behavior on newest angular material? 

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBi9nqMcbXuUNIka8rH:1589213192)*

maciaszczykm:
I'' verify on newest Material, but after looking into code it seems, that bug is still there:

https://github.com/angular/material/blob/78ccef6354cbe8898bfe3c489e3b7ea1d43bbffc/src/components/select/select.scss#L116

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBlFBlMFIdaYyBEbUyL:-1630161792)*

bryk:
OK, if the issue is still there. Then go with current solution and disable the lint check in only one file: `// scss-lint:disable BorderZero`. https://github.com/brigade/scss-lint

Add an implementation note about why this was done.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBlPFDoN6csukhKPOMH:-775003970)*

maciaszczykm:
Issue is still there . Unfortunately, we cannot use this one line comment to disable check in only one place, because we are using `sass-lint` plugin, not `scss-lint`. Names are quite similar, but configuration is a bit different:

http://sasstools.github.io/make-sass-lint-config/

Anyways, I've come up with another solution, which is enabling vendor prefixes only for Internet Explorer as it's our main issue at the moment. Other browsers seem to be working similar to each other, but we might have more problems with Internet Explorer and this will allow us to use `-ms` prefix in stylesheets to fix them.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBly-cGR4d981qYP1fP:-491259123)*

bryk:
OK, thanks for this fix. This PR is very high quality :) :lgtm: 

---

Reviewed 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/461#-:-KBm6U6pFZfL-_K4tz4a:485394804)*

PR #470, opened 2016-03-01
Make delete RC button less verbose and inline with RC title.

### LINKED ISSUE #465: "Delete replication controller" button too verbose
The button is very verbose. Perhaps we should rename it to "delete" and better indicate that this page is about RCs (if this confuses users). 

selection_036

@floreks @cheld @maciaszczykm

bryk:
Fixes #465 
selection_038

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/470)

### DIFF
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerinfo.html
@@ -15,21 +15,27 @@ limitations under the License.
 -->
 
 <div flex="nogrow" class="kd-replicationcontrollerinfo-item">
-  <md-button class="md-primary" ng-click="infoCtrl.handleDeleteReplicationControllerDialog()">
-    <md-icon class="material-icons">delete</md-icon>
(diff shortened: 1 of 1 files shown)


bryk:

floreks:
I think @maciaszczykm will not like this solution ;)

maciaszczykm:
@bryk How about removing delete label too and leaving only icon? Currently it's incosistent with edit pencil.

floreks:
I agree with @maciaszczykm

bryk:
selection_039

I like your suggestion. PTAL

bryk:
We can alternatively move this to the `Name` section. Opinions?

floreks:
Sounds good to me. It will be event more specific, just like `edit`.

maciaszczykm:
@bryk I was going to make such proposal, LGTM.

bryk:
Now it looks like this, can you take another look?
selection_041

maciaszczykm:
LGTM :+1:

PR #487, opened 2016-03-03
[WIP] Namespace separation in RC list view

bryk:
Do not review - for discussion only

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/487)

### DIFF
--- a/src/app/backend/replicationcontrollerlist.go
+++ b/src/app/backend/replicationcontrollerlist.go
@@ -32,6 +32,11 @@ type GetNodeFunc func(nodeName string) (*api.Node, error)
 
 // ReplicationControllerList contains a list of Replication Controllers in the cluster.
 type ReplicationControllerList struct {
+	Namespaces []ReplicationControllerNamespace `json:"namespaces"`
+}
+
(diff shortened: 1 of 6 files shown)


bryk:
@floreks @cheld @f-higashi I did a demo on how RC list view can be separated by namespaces.
Can you look at it? To do this, all you need is:
```
git checkout -b bryk-namespaces-show master
git pull https://github.com/bryk/dashboard.git namespaces-show
```

bryk:
I have mixed feelings about this and I'm leaning towards namespace selector.

floreks:
I've just briefly checked it and I'm starting to think that selector is a better way to go. Clearly this is not so user friendly when there are many apps deployed and can quickly cause performance issues.

cheld:
In general I like it. Visually, it could be improved I guess - but, I am not sure what.

cheld:
If you prefer the selector, then I would suggest to keep the implementation and add an all option

f-higashi:
I like it.
And I would like RC list page of each namespaces and to open deploy RC dialog which is set same namespace as default from this page.

Issue #483, opened 2016-03-02
Deploy Page - styling issue when shrinking page on IE


floreks:
##### Environment

```
Dashboard version: HEAD
Kubernetes version: HEAD
Operating system: Windows 7 x64
Browser: IE 11.0.9600.18161
```

##### Steps to reproduce

1. Go to deploy page.
2. Resize browser so width is very small.

##### Observed result

After shrinking the browser windows all elements are collapsed and overlap each other.

##### Comments
![image](https://cloud.githubusercontent.com/assets/2285385/13463442/8b53950a-e08c-11e5-81c3-5dd7345db4fc.png)

bryk:
I assume fixed.

PR #488, opened 2016-03-03
Small layout fix for IE11

taimir:
As a follow-up of issue https://github.com/kubernetes/dashboard/issues/483

@floreks can you confirm this resolves it?

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/488)

### DIFF
--- a/src/app/frontend/deploy/helpsection/helpsection.scss
+++ b/src/app/frontend/deploy/helpsection/helpsection.scss
@@ -17,8 +17,10 @@
 // This rule is meant to align items only in the current layout (kd-help-section).
 .kd-help-section > * {
   flex: 1;
+  flex-basis: auto;
   margin: $baseline-grid 0;
   padding-right: 4 * $baseline-grid;
+  width: 100%;
 
   &:last-child {
     padding-right: 0;


floreks:
Partially. Now it looks good on small screen but has some problems on bigger screen.

image

taimir:
Ok, let me see ...

taimir:
OK, I don't see a clear solution to this right now - IE11 shrinks `flex-basis` of children which are wrapped to 0 after the wrap takes place, even though I have specified it to be 1 explicitly. When the `flex-basis` is `auto` (of the children in the help section), we see the behaviour in the pic above: when things are wrapped they look fine, but not when in a bigger window.

taimir:
@floreks I couldn't do it without forcing a `width: 100%` style. I know we try to avoid this, but here I do not see another solution. Can you check it now?

floreks:
:lgtm: 

---

Reviewed 1 of 1 files at r2.

---

*[src/app/frontend/deploy/helpsection/helpsection.scss, line 23 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/488#-KBwoEVjzfw53VgL52gK:-KBwoEVjzfw53VgL52gL:-947251157) ([raw file](https://github.com/kubernetes/dashboard/blob/3eee1f3bb887151d230027a60d9e804f46bc82b3/src/app/frontend/deploy/helpsection/helpsection.scss#L23)):*
I'd say that width 100% is not that bad and this may be our only option. IE has major problems with flexbox and proper calculation of elements size.

It looks good on all browsers now.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/488#-:-KBwoFf-MrRUJfwWsjxS:980077423)*

taimir:
@bryk PTAL

bryk:
:lgtm: 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/488#-:-KBwu3Rn0F3diJIE8aPr:980077423)*

PR #491, opened 2016-03-03
Fixed styling issues on logs page and made it more responsive

floreks:
I've changed logs toolbar on smaller screens to menu bar in order to fit it in available space.

There is 1 major issue with `md-menu-bar` that I've had to solve using `!important` property in css.

They've solved this issue: https://github.com/angular/material/issues/6049 by changing style of parent `md-toolbar` to `position: relative` in here: https://github.com/angular/material/blob/b58343c20ac4bd3418629c29abddfe3ac3840eb5/src/components/menuBar/js/menuBarController.js#L90

This is changed by code so pure css is not enough to handle this better and in our case we need toolbar to have `position: fixed` that's why I've added `!important`.

Current view:
zrzut ekranu z 2016-03-03 16-06-54

@bryk what do you think?

Btw. I'm out of office tomorrow and on monday so any comments I'll resolve on tuesday.

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/491)

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -41,7 +41,7 @@ rules:
   no-empty-rulesets: 2
   no-extends: 0
   no-ids: 2
-  no-important: 2
+  no-important: 0
   no-invalid-hex: 2
   no-mergeable-selectors: 2
   no-misspelled-properties: 2
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -23,7 +23,7 @@
 .kd-toolbar {
   box-shadow: $whiteframe-shadow-1dp;
   height: $toolbar-height-size-base;
-  position: fixed;
+  position: fixed !important;
   top: 0;
 }
 
--- a/src/app/frontend/logs/logstoolbar/logstoolbar.html
+++ b/src/app/frontend/logs/logstoolbar/logstoolbar.html
@@ -14,28 +14,27 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div layout="row" layout-wrap layout-margin layout-align="center center">
-  <span flex="10"></span>
+<div layout="row" hide show-gt-sm layout-margin layout-align="center center"
+     class="kd-logs-toolbar">
+  <span class="kd-logs-toolbar-text">Pod:</span>
+  <md-select class="kd-logs-toolbar-select" aria-label="Logs from pod" ng-model="ctrl.pod"
+             md-on-close="ctrl.onPodChange(ctrl.pod.name)"
+             required>
+    <md-option ng-repeat="item in ctrl.pods" ng-value="item">
+      <span class="kd-logs-toolbar-text">{{item.name}}</span>
+    </md-option>
+  </md-select>
 
-  <div layout="row" layout-wrap layout-margin layout-align="center center">
-    <span class="kd-logs-toolbar-text">Logs from pod:</span>
-    <md-select class="kd-logs-toolbar-select" aria-label="Logs from pod" ng-model="ctrl.pod"
-               md-on-close="ctrl.onPodChange(ctrl.pod.name)"
-               required>
-      <md-option ng-repeat="item in ctrl.pods" ng-value="item">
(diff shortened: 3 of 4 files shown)


bryk:
Unfortunately, this does not fix the problem entirely:

selection_042

bryk:
Re "logs menu": how about changing this name to something more descriptive. I.e., users already see that this is a "menu" so this information is not needed. How about "logs source"?

floreks:
> Unfortunately, this does not fix the problem entirely:

On which browser did you check? I've checked on 3 different browsers (IE, Chrome, FF) and on chrome in mobile mode and it looks good. There's no wrapping. As soon as it wants to wrap, view changes to menu mode.

> Re "logs menu": how about changing this name to something more descriptive. I.e., users already see that this is a "menu" so this information is not needed. How about "logs source"?

Sure. I've just used first name that came to my mind. "Logs source" sound better.

floreks:
Ok I know where the problem is. Pod and container names may be very long and width of container is extended. I need to make sure that it changes to 'menu style' as soon as width extends some limit.

floreks:
PTAL

bryk:
@maciaszczykm Can you take this?

maciaszczykm:
Yes, I will do it soon.

maciaszczykm:
image

Seems to be working as expected. Cannot reproduce @bryk's problem on Chrome, Firefox nor IE. Maybe it's due to operatin system?

My only concern is styling :)

- Could you move pickers slightly down? It's like 2 or 3 pixels to align it with Kubernetes logo.
- Could you make arrow downwards a bit smaller? It seems to big at the moment.
- Perhaps application and container names should not be transfromed to uppercase? I know, that's Angular Material's default, but you should consider it. Not neccessary however.

floreks:
PTAL

maciaszczykm:
:lgtm:

---

Reviewed 2 of 4 files at r1, 1 of 2 files at r3, 1 of 1 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/491#-:-KCLiYiaWGyoNeQl5pU2:1417088721)*

bryk:
Thanks for doing this :)

bryk:
Re `position: fixed !important;`: I'm pretty persistent on this and the opinion that we should never ever use it :) There is always a way to avoid !important unless you write a framework. Can we have this fixed and no-important rule reenabled? I'm happy to discuss this

floreks:
I couldn't find other way to overload css that is set by the code. We would need our own code that overloads already overloaded css :) This seemed like too much to me, that's why I've decided to use `!important`. What about enabling `no-important`rule and only disabling it in css with comment? If you have any better solution for that then it would be great.

bryk:
> What about enabling no-importantrule and only disabling it in css with comment? 

This sounds very nice then. Say why we need important in this very place and enable the lint rule then :)

floreks:
Unfortunately this feature for `sass-lint` is still under development: https://github.com/sasstools/sass-lint/issues/70

We'd have to migrate to `scss-lint` to have option to disable particular css rule check. What do you think?

bryk:
What problem does the `!important` fix, btw? I've tested it now, and `!important` does not change the behavior and no other rule wants to override position.

floreks:
Remove `!important` then go to logs page. You need more logs so the page can be scrolled down. Click logs menu button. Toolbar will disappear because style will be changed by code :)

bryk:
Now I can reproduce this. I see that this is the menu bar that overrides the position: https://github.com/angular/material/blame/b58343c20ac4bd3418629c29abddfe3ac3840eb5/src/components/menuBar/js/menuBarController.js

I also see that single menus dont nee menu bars. How about removing the menu bar directive? https://material.angularjs.org/latest/demo/menu

floreks:
That would require making 2 separate menus right? I'm not sure if they would fit in the available space for  small devices.

bryk:
Hmm... I see and dont like this :) We have an issue logged, so let's move forward and see that maybe material design fixes this.

PR #493, opened 2016-03-04
Move metric units to table header to save space and disambiguate

### LINKED ISSUE #405: Confusing measure for CPU usage
The unit for measuring the CPU usage - "mCPU" - hard to understand. Our K8s Ops as well as myself could not guess the meaning until @cheld let us know it's "milli CPU".
mili
Some options:

1. Enhance the tool-tip
2. Change the label to "‰ CPU"

bryk:
With that we'll not have `kcores` (kilo-cores) neither `TCPU`
(tera-CPU). This also saves space since all numbers in a column have the
same unit.

Fixes #405 

Now it looks like this (using heapster with bad data):
selection_043

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/493)

### DIFF
--- a/src/app/frontend/common/filters/cores_filter.js
+++ b/src/app/frontend/common/filters/cores_filter.js
@@ -16,7 +16,7 @@
 const base = 1000;
 
 /** Names of the suffixes where I-th name is for base^I suffix. */
-const powerSuffixes = ['m', '', 'k', 'M', 'G'];
+const powerSuffixes = ['', 'k', 'M', 'G', 'T'];
 
 /**
  * Returns filter function that formats cores usage.
@@ -31,6 +31,9 @@ export default function coresFilter(numberFilter) {
    * @return {string}
    */
   let formatCores = function(value) {
+    // Convert millicores to cores.
+    value = value / 1000;
+
     let divider = 1;
     let power = 0;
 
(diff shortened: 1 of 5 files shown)


bryk:
@joeatwork @cheld Can you review?

cheld:
Very nice. Thank you. 

BTW: If no data is received from there is a dash in the UI. Maybe you want to add a test case for no-data too.

cheld:
LGTM

src/app/frontend/common/filters/cores_filter.js
```diff
@@ -39,7 +42,7 @@ export default function coresFilter(numberFilter) {
       power += 1;
     }
     let formatted = numberFilter(value / divider);
-    return `${formatted} ${powerSuffixes[power]}CPU`;
+    return `${formatted} ${powerSuffixes[power]}`;
```
joeatwork: The trailing space in the case of `{{1| kdCores}}` might surprise people using the filter - would you consider the slightly uglier

```js
    let suffix = powerSuffixes[power];
    return  suffix ? `${formatted} ${suffix} : formatted;
```

joeatwork:
LGTM

src/app/frontend/common/filters/cores_filter.js
bryk: Yeah, this is a good suggestion. Thanks :)

PR #512, opened 2016-03-09
Add FBT for replication controller details

### LINKED ISSUE #503: Finish remaining FBT sheets
#### Issue details
We should complete the remaining FBT sheets.

Missing:
- Replication Details
- Logs

Any comments on the process? Should we adapt or change something?

@floreks , @digitalfishpond @maciaszczykm @f-higashi 
please pick

maciaszczykm:
Fixes #503 issue.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/512)

### DIFF
--- a/docs/test/replication-controller-details.md
+++ b/docs/test/replication-controller-details.md
@@ -1,0 +1,46 @@
+# Replication controller details
+
+## Prerequisites
+Cluster with few deployed applications, where: 
+
+* at least one is valid
+* at least one is invalid (for example with wrong image) 
+
+## Expected result
+Page displays correctly, especially:
+
+* tables with pods and events are available
+* sidebar is available
+* application details are valid and matching data from `kubectl` command
+* all application actions are available
+* warning events have warning icon nearby to their messages
+
+## Specific factors
+
+| Factor                        | 1                               | 2                       | 3                 | 4             | 5            | 6               | 7          | 8      | Comment                                                                               |
+|-------------------------------|---------------------------------|-------------------------|-------------------|---------------|--------------|-----------------|------------|--------|---------------------------------------------------------------------------------------|
+| Screen size                   | normal                          | mobile                  | narrow            | middle size   | wide         | active resizing |            |        |                                                                                       |
+| Heapster                      | installed and running           | not installed           |                   |               |              |                 |            |        |                                                                                       |
+| Pods count                    | 1                               | 1-10                    | more than 100     |               |              |                 |            |        |                                                                                       |
+| Events count                  | 0                               | 0-10                    | more than 1000    |               |              |                 |            |        | events number can be changed for example by scaling replication controller pods count |
+| Replication Controller status | valid state                     | invalid error state     |                   |               |              |                 |            |        |                                                                                       |
+| Replication Controller age    | recently created                | few hours               | at least 24 hours |               |              |                 |            |        |                                                                                       |
+| Replication Controller fields | normal                          | max length              | high pod count    |               |              |                 |            |        |                                                                                       |
+| Table sorting                 | default                         | different columns       | upwards           | downwards     |              |                 |            |        | arrow near column header indicates sort order                                         |
+| Events filtering              | default (all)                   | warning                 |                   |               |              |                 |            |        |                                                                                       |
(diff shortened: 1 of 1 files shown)


maciaszczykm:
@cheld Could you take a look?

docs/test/replication-controller-details.md
```diff
@@ -0,0 +1,44 @@
...
+* sidebar is available
+* application details are valid and matching data from `kubectl` command
+* all application actions are available
+* warning events have warning icon nearby to their messages
+
+## Specific factors
+
+| Factor                        | 1                               | 2                       | 3              | 4             | 5            | 6               | 7          | 8      | Comment                                                                               |
+|-------------------------------|---------------------------------|-------------------------|----------------|---------------|--------------|-----------------|------------|--------|---------------------------------------------------------------------------------------|
+| Screen size                   | normal                          | mobile                  | narrow         | middle size   | wide         | active resizing |            |        |                                                                                       |
```
cheld: Gain, I would suggest to drop and leave only the general factor screen size

docs/test/replication-controller-details.md
```diff
@@ -0,0 +1,44 @@
...
+| Pods count                    | 1                               | odd                     | even           | 20-30         |              |                 |            |        |                                                                                       |
```
cheld: 1 | 1-10 | > 100

docs/test/replication-controller-details.md
```diff
@@ -0,0 +1,44 @@
...
+| Events count                  | 0                               | 0-10                    | more than 10   |               |              |                 |            |        | events number can be changed for example by scaling replication controller pods count |
```
cheld: 0 (thats a good case!! but how to test? ) 
0-10
>1000  (the number of events increases with the number of pods)

Is there an upper limit of events? How many events are likely to occur for each pod?

docs/test/replication-controller-details.md
```diff
@@ -0,0 +1,44 @@
...
+| Replication Controller status | valid state                     | invalid error state     |                |               |              |                 |            |        |                                                                                       |
```
cheld: The card status is not relevant - I think. The status of each single event is marked. There are a lot of bugs around highlighting events. Please investigate and add factors to test.

docs/test/replication-controller-details.md
maciaszczykm: In this case it's even more important than on replication controller list page. We have three page breaks here. One with full view, second with sidebar moved into tab and last one with responsive table. I'd leave it like it's done at the moment.

docs/test/replication-controller-details.md
maciaszczykm: > 0 (thats a good case!! but how to test? ) 

With time some of events disappear, but I don't know if it's a rule...

Is there an upper limit of events? How many events are likely to occur for each pod?

Would have to check. 

Problematic thing with this factor is we cannot directly change events count...

docs/test/replication-controller-details.md
maciaszczykm: Yes, but like I said before, we cannot directly change events count nor their message (warning or normal). That's why I've decided to put here factor which have impact on it and we can easily change it. To make replication controller invalid we just need to specify wrong image for example.

maciaszczykm:
PTAL

docs/test/replication-controller-details.md
maciaszczykm: Done.

docs/test/replication-controller-details.md
cheld: ok, I see

docs/test/replication-controller-details.md
cheld: Yes, but on the page it is a event in the table that is marked with a warning. So, I would rather make the factor. something like 

Event warnings | image does not exist | not enough resources ...

But I would like to see more semantics than in my upper examle

cheld:
Also add factor about heapster:

not installed | ...

How much data is shown? 24h? Can we test that? Maybe we can leave it running over night, at least

maciaszczykm:
@cheld Applied yours suggestions.

PTAL

cheld:
LGTM

PR #508, opened 2016-03-08
Fixed test flakiness on application deploy in IE

floreks:
Fixes #494

I've changed some button queries to use xpath and moved expect block to be executed after click promise is resolved.

[Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/508)

PR #515, opened 2016-03-10
Fix test of fileupload

f-higashi:
I added FileDetector setting to test file upload in saucelabs.
FileDetector locates selenium-webdriver/remote/index.js.
But the file path is different between npm2 and npm3.
https://docs.npmjs.com/how-npm-works/npm3-nondet

npm in travis is v2.
If we want to use v3 in travis, we need to install npm3.

.travis.yaml
```yaml
before_install: if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
```

Which version is better?
Do you have a good idea to support both versions?

@bryk  @floreks  @maciaszczykm

---
This change is [Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/515)

### DIFF
--- a/src/test/integration/stories/deploy_from_invalid_file_test.js
+++ b/src/test/integration/stories/deploy_from_invalid_file_test.js
@@ -13,17 +13,18 @@
 // limitations under the License.
 
 import path from 'path';
+import remote from 'selenium-webdriver/remote';
 
 import DeployFromFilePageObject from '../deploy/deployfromfile_po';
 
 // Test assumes, that there are no replication controllers in the cluster at the beginning.
-// TODO(#494): Reenable this test when fixed.
-xdescribe('Deploy from invalid file user story test', () => {
+describe('Deploy from invalid file user story test', () => {
 
   /** @type {!DeployFromFilePageObject} */
   let deployFromFilePage;
 
   beforeAll(() => {
+    browser.driver.setFileDetector(new remote.FileDetector());
     deployFromFilePage = new DeployFromFilePageObject();
     browser.get('#/deploy');
     // switches to deploy from file
@@ -36,7 +37,8 @@ xdescribe('Deploy from invalid file user story test', () => {
     let absolutePath = path.resolve(__dirname, fileToUpload);
 
     // when
-    deployFromFilePage.filePicker.sendKeys(absolutePath);
+    deployFromFilePage.makeInputVisible();
+    deployFromFilePage.setFile(absolutePath);
     deployFromFilePage.deployButton.click();
 
     // then
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ cache:
 # Use Node.js as primary language because Gulp is the build system used in the project.
 language: node_js
 node_js:
-  - 4.2.2
+  - 5.1.1
 
 addons:
   # Run tests that require a browser on SauceLabs CI provider. Learn more at: https://saucelabs.com
--- a/docs/devel/getting-started.md
+++ b/docs/devel/getting-started.md
@@ -14,8 +14,8 @@ logic and fetches raw data from the various Kubernetes APIs.
 Make sure the following software is installed and added to the `$PATH` variable:
 * Docker (1.3+)
 * go (1.5+)
-* nodejs (4.2.2+)
-* npm (1.3+)
+* nodejs (5.1.1+)
+* npm (3+)
 * java (7+)
 * gulp (3.9+)
 
--- a/src/app/frontend/replicationcontrollerdetail/deletereplicationcontroller.html
+++ b/src/app/frontend/replicationcontrollerdetail/deletereplicationcontroller.html
@@ -30,8 +30,8 @@ limitations under the License.
       </md-icon>
     </md-checkbox>
     <md-dialog-actions>
-      <md-button class="md-primary" ng-click="ctrl.cancel()">Cancel</md-button>
-      <md-button class="md-primary" ng-click="ctrl.remove()">Delete</md-button>
+      <md-button class="md-primary kd-cancel-btn" ng-click="ctrl.cancel()">Cancel</md-button>
+      <md-button class="md-primary kd-delete-btn" ng-click="ctrl.remove()">Delete</md-button>
     </md-dialog-actions>
   </md-dialog-content>
 </md-dialog>
--- a/src/test/integration/deploy/deployfromfile_po.js
+++ b/src/test/integration/deploy/deployfromfile_po.js
@@ -27,9 +27,29 @@ export default class DeployFromFilePageObject {
     this.inputContainer = element(this.inputContainerQuery);
 
     this.filePickerQuery = by.css('.kd-upload-file-picker');
-    this.filePicker = element(this.filePickerQuery);
+    this.filePicker_ = element(this.filePickerQuery);
 
     this.mdDialogQuery = by.tagName('md-dialog');
     this.mdDialog = element(this.mdDialogQuery);
   }
+
+  /**
+   * Make filePicker input field visible
+   * Firefox does not allow sendKeys to invisible input[type=file] element
+   */
+  makeInputVisible() {
+    browser.driver.executeScript(function() {
+      /* global document */
+      let filePickerDomElement = document.getElementsByClassName('kd-upload-file-picker')[0];
+      filePickerDomElement.style.visibility = 'visible';
+      filePickerDomElement.style.height = '1px';
+      filePickerDomElement.style.width = '1px';
+    });
+  }
+
+  /**
+   * Sets filepath on the filePicker input field
+   * @param {string} filePath
+   */
+  setFile(filePath) { this.filePicker_.sendKeys(filePath); }
 }
--- a/src/test/integration/replicationcontrollerdetail/deletereplicationcontroller_po.js
+++ b/src/test/integration/replicationcontrollerdetail/deletereplicationcontroller_po.js
@@ -14,15 +14,10 @@
 
 export default class DeleteReplicationControllerDialogObject {
   constructor() {
-    let deleteDialogXPath = '//md-dialog[@aria-label=\'Delete Replication Controller\']';
-    this.deleteDialogQuery = by.xpath(deleteDialogXPath);
-    this.deleteDialog = element(this.deleteDialogQuery);
-
-    this.deleteAppButtonQuery =
-        by.xpath(`${deleteDialogXPath}//md-dialog-actions//span[text() = 'Delete']/..`);
+    this.deleteAppButtonQuery = by.css('.kd-delete-btn');
     this.deleteAppButton = element(this.deleteAppButtonQuery);
 
-    this.deleteServicesCheckboxQuery = by.xpath(`${deleteDialogXPath}//md-checkbox`);
+    this.deleteServicesCheckboxQuery = by.css('.kd-deletedialog-services-checkbox');
     this.deleteServicesCheckbox = element(this.deleteServicesCheckboxQuery);
   }
 }
--- a/src/test/integration/stories/deploy_from_valid_file_test.js
+++ b/src/test/integration/stories/deploy_from_valid_file_test.js
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import path from 'path';
+import remote from 'selenium-webdriver/remote';
 
 import DeployFromFilePageObject from '../deploy/deployfromfile_po';
 import ReplicationControllersPageObject from '../replicationcontrollerslist/replicationcontrollers_po';
@@ -20,8 +21,7 @@ import DeleteReplicationControllerDialogObject from '../replicationcontrollerdet
 import ZeroStatePageObject from '../zerostate/zerostate_po';
 
 // Test assumes, that there are no replication controllers in the cluster at the beginning.
-// TODO(#494): Reenable this test when fixed.
-xdescribe('Deploy from valid file user story test', () => {
+describe('Deploy from valid file user story test', () => {
 
   /** @type {!DeployFromFilePageObject} */
   let deployFromFilePage;
@@ -39,6 +39,7 @@ xdescribe('Deploy from valid file user story test', () => {
   let appName = 'integration-test-valid-rc';
 
   beforeAll(() => {
+    browser.driver.setFileDetector(new remote.FileDetector());
     deployFromFilePage = new DeployFromFilePageObject();
     replicationControllersPage = new ReplicationControllersPageObject();
     deleteDialog = new DeleteReplicationControllerDialogObject();
@@ -54,7 +55,8 @@ xdescribe('Deploy from valid file user story test', () => {
     let absolutePath = path.resolve(__dirname, fileToUpload);
 
     // when
-    deployFromFilePage.filePicker.sendKeys(absolutePath);
+    deployFromFilePage.makeInputVisible();
+    deployFromFilePage.setFile(absolutePath);
     deployFromFilePage.deployButton.click();
 
     // then


maciaszczykm:
Could you please rebase commits into one? If you want to split your contribution then best idea is to do it in multiple pull requests.

f-higashi:
It's OK.
But I will do after a decision that we use npm2 or npm3.

If we decide to use npm v2, I remove 1acbe14 easily.

floreks:
NPM 3 is fine for me. As long as we are using `npm install` we should end up with the same dep. tree.

src/test/integration/stories/deploy_from_invalid_file_test.js
```diff
@@ -16,14 +16,17 @@ import path from 'path';
 
 import DeployFromFilePageObject from '../deploy/deployfromfile_po';
 
+import remote from '../../../../node_modules/selenium-webdriver/remote';
```
floreks: Remove newlines and try to sort it in alphabetic order. It's easier to read this way. Also you can just use

```javascript
import remote from 'selenium-webdriver/remote';
```

floreks:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 42 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRHVLfm0GizsKcgpq:-KCVRHVLfm0GizsKcgpr:-1422903844) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L42)):*
Although I don't like usage of document here I don't see any other possible solution for that. @bryk do you have any other idea?

---

*[src/test/integration/deploy/deployfromfile_po.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVTxALWo7y-Ra2nZuq:-KCVTxALWo7y-Ra2nZur:-487795505) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L45)):*
Do we actually need to change height and width from 0 to 1px? Isn't `visible` sufficient?

---

*[src/test/integration/stories/deploy_from_invalid_file_test.js, line 19 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVCwS8JIVHmwUGMNmt:-KCVCwS9gzgVCySJI_bA:350747710) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/stories/deploy_from_invalid_file_test.js#L19)):*
Remove newlines, try to sort it in alphabetic order and separate external imports from test file imports. It's easier to read this way. 

You can just use:

```javascript
import path from 'path';
import remote from 'selenium-webdriver/remote';

import DeployFromFilePageObject from '../deploy/deployfromfile_po';
```

---

*[src/test/integration/stories/deploy_from_valid_file_test.js, line 22 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRXMx_ianj4fic9fQ:-KCVRXMx_ianj4fic9fR:1662569705) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/stories/deploy_from_valid_file_test.js#L22)):*
Same here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

floreks:
This is really nice work. Just few small comments.

f-higashi:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 42 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRHVLfm0GizsKcgpq:-KCnNbirOY97JEgjqgAJ:-948787065) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L42)):*
How about the following method in deployfromfile_po.js.
And this.filePciker changes private variable.

In this case I can remove usage.

  ```
  /**
   * Sets file into filePicker
   * @param {string} filePath
   */
  setFile(filePath) {
    /**
     * Firefox does not allow sendKeys to invisible input[type=file] element.
     */
    let makeInputVisible = function() {
      /* global document */
      let filePickerDomElement = document.getElementsByClassName('kd-upload-file-picker')[0];
      filePickerDomElement.style.visibility = 'visible';
      filePickerDomElement.style.height = '1px';
      filePickerDomElement.style.width = '1px';
    };
    browser.driver.executeScript(makeInputVisible);
    this.filePicker_.sendKeys(filePath);
  }
  ```

In test code

```
    // when
    deployFromFilePage.setFile(absolutePath);
    deployFromFilePage.deployButton.click();
```

---

*[src/test/integration/deploy/deployfromfile_po.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVTxALWo7y-Ra2nZuq:-KCnMvzGGGruR4WCPJ7t:283114412) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L45)):*
I referred bellow page; 
http://stackoverflow.com/questions/6101461/how-to-force-selenium-webdriver-to-click-on-element-which-is-not-currently-visib
It's about button.
In fact out tests did not work when I set only `visible`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

floreks:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 42 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRHVLfm0GizsKcgpq:-KCoQmZ6jxyChjnQggOa:1795316398) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L42)):*
I don't think this is a good idea. This part of code 

```javascript
browser.driver.executeScript(makeInputVisible);
this.filePicker_.sendKeys(filePath);
```
belongs to test and should not be executed in PO file. PO should be only for retrieving elements on page and logic related to that. It should be clear and visible in test that we are executing some script on page and sending input to field.

What do you think?

---

*[src/test/integration/deploy/deployfromfile_po.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVTxALWo7y-Ra2nZuq:-KCoSEsBx2br-ndwVt12:1971846896) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L45)):*
Ok. Thanks for explanation.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

f-higashi:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 42 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRHVLfm0GizsKcgpq:-KCoYffg_6HwGBbuW3Jf:-262242077) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L42)):*
I agree that PO should be only for retrieving elements on page ~.

On the web page we provides `fileName input` and `...  button` instead of `filePicker`.
`fileName.sendsKey` does not work. 
I don't think it is bad that we provide other function `setFile`  instead of `filePicker`.
I could say `setFile` is one of Services in this page such as `typeUsername` of the following example.

https://github.com/SeleniumHQ/selenium/wiki/PageObjects#summary


But `setFile` changes PO status.
It is better to reset css in this fucntion. 

ex)
```
browser.driver.executeScript(makeInputVisible);
this.filePicker_.sendKeys(filePath);
browser.driver.executeScript(restoreVisibility);
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

floreks:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 42 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRHVLfm0GizsKcgpq:-KCp6QBoGL7zyvMQAwEq:145853367) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L42)):*
What about making 2 functions in PO

`setFileName(path)` that will basically do `this.filePickerElement.sendKeys()` and
`makeInputVisible()` that will execute `browser.driver.executeScript(() => {})` with inline function.

Then we can call in test:
```
po.makeInputvisible();
po.setFileName(path);
```

I think restore vibisility is not needed. Style will be restored anyway when page will be reloaded or state changes.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

f-higashi:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 42 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCVRHVLfm0GizsKcgpq:-KCp9PAYKk0HfkSlsUz1:-480927898) ([raw file](https://github.com/kubernetes/dashboard/blob/f9bda22b1505aeabf71445c5784602d48c56c078/src/test/integration/deploy/deployfromfile_po.js#L42)):*
It looks good.
I'll fix it.
Thanks.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

f-higashi:
PTAL

floreks:
:lgtm:

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-:-KCuNTQmroGTAdbmJG22:1417088721)*

floreks:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/test/integration/deploy/deployfromfile_po.js, line 37 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCuOgPearxy0lIAIuLv:-KCuOgPearxy0lIAIuLw:-121875309) ([raw file](https://github.com/kubernetes/dashboard/blob/86ae875c4c3a6244ce239c21804d8266bafa6ec7/src/test/integration/deploy/deployfromfile_po.js#L37)):*
I think it should be `Make filePicker input field visible`

---

*[src/test/integration/deploy/deployfromfile_po.js, line 51 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCuOt11gybwD_EcKjrC:-KCuOt11gybwD_EcKjrD:-1395804973) ([raw file](https://github.com/kubernetes/dashboard/blob/86ae875c4c3a6244ce239c21804d8266bafa6ec7/src/test/integration/deploy/deployfromfile_po.js#L51)):*
Actually it only `Sets filepath on the filePicker input field`

---

*[src/test/integration/stories/deploy_from_invalid_file_test.js, line 21 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KCuOVyyYhJhyJ4dT_Wt:-KCuOVyyYhJhyJ4dT_Wu:-1867958842) ([raw file](https://github.com/kubernetes/dashboard/blob/86ae875c4c3a6244ce239c21804d8266bafa6ec7/src/test/integration/stories/deploy_from_invalid_file_test.js#L21)):*
This comment is no longer needed in both test files.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

floreks:
Last style comments and we can merge

---

Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-:-KCuPAvBaJAnhzcg4pUZ:-959500073)*

f-higashi:
Thanks for your comment.
I fixed them.

PTAL

floreks:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

floreks:
:lgtm:

---

Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-:-KCyqHX4AuL9TjLIYCAW:1417088721)*

floreks:
@bryk could you shortly check that?

bryk:
One comment for simplification. 

---

Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[.travis.yml, line 39 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KD2rACrbpOUmXFYaEqY:-KD2rACsmwwzk_SHgBMg:953960089) ([raw file](https://github.com/kubernetes/dashboard/blob/199c3da043c2b7421daab94a431dd5ebef65fa81/.travis.yml#L39)):*
Instead of this, you can specify node_js version to `5.x.y` which is bundled with npm 3. Do this, it'll avoid this little hack :) 

See https://nodejs.org/en/blog/release/v5.0.0/

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-:-KD2rnBcWZjkBO0qjgBH:-1098245164)*

f-higashi:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[.travis.yml, line 39 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-KD2rACrbpOUmXFYaEqY:-KD2tZGSp0J6hTDwlkPF:-696115259) ([raw file](https://github.com/kubernetes/dashboard/blob/199c3da043c2b7421daab94a431dd5ebef65fa81/.travis.yml#L39)):*
Thanks for your comment.
I'll use 5.1.1 which is latest in travis.

https://docs.travis-ci.com/user/languages/javascript-with-nodejs

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515)*

bryk:
:lgtm:

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-:-KD4DqqJrflkG78Uar7b:1417088721)*

bryk:
Tests fail. Can you fix them?

f-higashi:
Yes.
I check it  on the way.

In my local pc tests success. but in travis test fail because test fail to delete created RC.

bryk:
Can you fix the test?

bryk:
And which test does not delete created RCs?

f-higashi:
I cannot find the solution now.


`deploy_from_valid_file_test` create container and fail to delete the container in after_all.
Next test `Zero state view should do something` fail because the container is existing. 

https://travis-ci.org/kubernetes/dashboard/builds/116599998
```
AfterAll Failed: No element found using locator: By.xpath("//md-dialog[@aria-label='Delete Replication Controller']//md-dialog-actions//span[text() = 'Delete']/..")
```

I tried to test in same version; node, npm, browser and  dependencies library in local PC. 
But tests succeeded in local PC.

And I reverted source which succeeded 3 days ago but tests has same error in travis.
Of course tests succeed in local PC.


Do you have any idea?

floreks:
I believe this is related to #546. There had to be some minor update of some dependency in our project. Since we use `~1.0.0` versioning in bower/package json then it accepts all newest versions up to `1.0.x`. Travis always downloads everything. 

This xpath is now not correct. Anyway as discussed in #508 we should not use xpath and refactor the code that uses it.

zrzut ekranu z 2016-03-18 13-04-31

There is no `<span>` anymore.

f-higashi:
I see. 
I deleted node_module but not bower_components..

I reproduced error in local PC.

Thanks for your help : )

bryk:
:lgtm: Thanks for fixing this issue! Will merge when travis gets green. 

---

Reviewed 2 of 2 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/515#-:-KD8lZP-jVFdC8iFKKaI:-1098095845)*

PR #528, opened 2016-03-14
i18n & l10n via Google getMsg()

taimir:
As a follow up to https://github.com/kubernetes/dashboard/pull/252.

This is an example of a full cycle integration of goog.getMsg() for i18n & l10n into our build process. 
I've described the basic workflow in [this docu entry](https://github.com/taimir/dashboard/blob/i18n-closure-get-msg/docs/design/i18n.md).

@zreigz @bryk PTAL

The translations are available when serving with `gulp serve:prod` only (seems reasonable to me). I used [this chrome extension] (https://chrome.google.com/webstore/detail/quick-language-switcher/pmjbhfmaphnpbehdanbjphdcniaelfie?hl=en) to test. Right now there is an example for a translation in japanese besides english. By default, if there is no translation for a language, english should be shown.

Do you see any issues? Sorry, I know the PR is kind of big, but I wanted to have it in its whole integrity.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/528)


.eslintrc
```diff
@@ -42,8 +42,6 @@
     "no-multiple-empty-lines": [2, {"max": 1}],
     // Disallow the use of ternary operators when a simpler alternative exists.
     "no-unneeded-ternary": 2,
-    // Avoid code that looks like two expressions but is actually one.
-    "no-unexpected-multiline": 2,
```
taimir: This beast seems to contradict the current clang-format that was updated earlier today. We should either leave this out or change the clang-format rules...

bryk:
I like this change a lot :) First round of comments, PTAL 

---

Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions.

---

*[.eslintrc, line 46 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCywhkbLCJ4NMjucZCg:-KCywhkccfa5I4vr6fi9:-2073151435) ([raw file](https://github.com/kubernetes/dashboard/blob/4857aa4c07cf04ab25a13d6775f067031f2b4206/.eslintrc#L46)):*
Ok, this is fine as clang-format wants different. Can you create a new PR with this change? It'll be merged in a breeze. 

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KCyy9Sfi2dN_ujJU_AG:287566182) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
Hmm... I think you can still use `multiDest`, instead of creating `O(outputDirs.length)` buildSteps for each translation. 

---

*[build/build.js, line 135 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyxJlW7xT5FPmvCTim:-KCyxJlXIfzD-EcUXgDp:1086703588) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L135)):*
Why not async.parallel? 

---

*[build/build.js, line 151 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyacg5voKDhejiX-C:-KCyyachYgXtwBbU4EsA:-467660488) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L151)):*
The same here about multi dest. 

---

*[build/conf.js, line 154 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyiwYyn7cAoSyOTCD:-KCyyiwZ1yIe4NXoqYqY:-1815221335) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/conf.js#L154)):*
Isn't the key `en_US`? 

---

*[build/deploy.js, line 156 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyzx-ufv_qpnmIUFu:-KCyyzx-ufv_qpnmIUFv:-722868510) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/deploy.js#L156)):*
Hmm... Why this change? Now it does the same as before, doesn't it? 

---

*[build/i18n.js, line 27 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyz9RRL4H_RFP2omuf:-KCyz9RRL4H_RFP2omug:-791750578) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L27)):*
Can you add `@type` information here and document this? 

---

*[build/i18n.js, line 50 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyz7GsSzGPBHXK-tMK:-KCyz7GsSzGPBHXK-tML:483073784) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L50)):*
New Line here? 

---

*[build/i18n.js, line 53 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzgr-wCe17OuRVQ90:-KCyzgr0bWuy1shh79rz:633826272) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L53)):*
Why dependency on scripts? 

---

*[build/i18n.js, line 55 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzQxZi-nYjHNSDTQO:-KCyzQxZi-nYjHNSDTQP:871788703) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L55)):*
Hmm.... I believe you can do this easier, like:

```js
let promises = conf.translations.map((translation) => extractForLanguage(translation.key));
return q.all(promises); 
```

---

*[build/index.js, line 31 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzn55YdRmTGxBhpaj:-KCyzn55YdRmTGxBhpak:-2084693323) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/index.js#L31)):*
You added `@params` but they arent here. 

---

*[build/index.js, line 56 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzxC9ytapUEcJAmwF:-KCyzxC9ytapUEcJAmwG:1595489322) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/index.js#L56)):*
You can use multiDest here. multiDest is faster than doing this synchronously. Just pass an array of indexPaths. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-:-KCz-vmw90a0iQF9Gwfk:-1914099702)*

bryk:
@cheld @maciaszczykm I need your help with reviewing this :) 

---

Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-:-KCz00Qtg0u-mztEOnzh:-455284441)*

taimir:
Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions.

---

*[.eslintrc, line 46 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCywhkbLCJ4NMjucZCg:-KCz45ilcJLy_7pu_vUM:77705682) ([raw file](https://github.com/kubernetes/dashboard/blob/4857aa4c07cf04ab25a13d6775f067031f2b4206/.eslintrc#L46)):*
Cool, I'll do that.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KCz0flnCjYnkeYEdf5q:-529858853) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
I forgot to add my comments, didn't I :D So, I did this partially because I still had to iterate over the translations and I wanted both things to be consistent (in essence multiarch / localization both produce multiple versions of the code).

Second, I think the current multiDest implementation is wrong (or I misunderstand some part of it), hence why I removed it. 
```
  outputStream.on('data', (data) => outputs.forEach((dest) => { dest.write(data); }));
  outputStream.on('end', () => outputs.forEach((dest) => { dest.end(); }));
```

Isn't that wrong, shouldn't we wait on 'end' event from all `dest` streams, and *then* end the outputStream? The logic seems reversed to me. After all, the `outputStream` should represent the execution of the multiple `dest`. 

If I rewrite it as I said above, I basically reinvent the wheel: [merge-stream](https://github.com/grncdr/merge-stream) does exactly what I described in its source code (merge() method). Since I was already using `async`, I decided not to use `merge-stream` as well (to keep it consistent) and left it all with `async`.

Also, when using `merge-stream` I had trouble with the execution itself. For some reason the `build-frontend` task took 10 minutes to finish? Haven't figured this out yet.

---

*[build/build.js, line 135 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyxJlW7xT5FPmvCTim:-KCz35lWiWsc_eZP7ESE:-1524010540) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L135)):*
Last I check the `scripts:prod` task that calls the closure compiler didn't work well with `async.parallel`. I got a very strange error that `app.js` was missing, even though `app.js` is the actual file created by the task. With `async.series` everything went well, so I went for that.

I think I could try `async.parallel` again though, I was experimenting with `merge-stream` a lot, so it might be that the errors I saw were not caused by `async.parallel`. Will comment on this again to confirm.

---

*[build/build.js, line 151 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyacg5voKDhejiX-C:-KCz3zHhGiyDAu0LDg5o:428223373) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L151)):*
See my comment above (the long one, hah).

---

*[build/conf.js, line 154 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyiwYyn7cAoSyOTCD:-KCz4CDvB93EbYymRS26:-1324921679) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/conf.js#L154)):*
OK, the HTTP [Accept-Language](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) spec. says that the prefix must not necessarily be of the form en_US. It could be just en. So the question is, how do we handle it? Do we really want to support different english versions, for example, etc.

It was simpler to just check for the first part of the Accept-Language, since it always is there (based on specification). That's why I did it like this. 

---

*[build/deploy.js, line 156 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyzx-ufv_qpnmIUFu:-KCz572Myg32m07jlmQ1:-483320343) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/deploy.js#L156)):*
That's related to my mistrust in multiDest, see comments above.

---

*[build/i18n.js, line 53 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzgr-wCe17OuRVQ90:-KCz5e5S2Dqj9B3zd6M5:-823303544) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L53)):*
I extract the languages from `app.js` that's created with `gulp scripts`. This is the merged source code in development.  If you can still remember, I couldn't parse ES6 with the XtbGenerator (to extract the messages). So that's a quick and easy (and a bit hacky) way to get all the transpiled code into 1 place (and parse it with XtbGenerator). 

---

*[build/i18n.js, line 55 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzQxZi-nYjHNSDTQO:-KCz6YBizrj7BEwQa3Hb:-1631497814) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L55)):*
Cool :)  still getting used to all `q` has to offer.

---

*[build/index.js, line 56 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzxC9ytapUEcJAmwF:-KCz6u1N2HRHJCgAFX5K:-1549864681) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/index.js#L56)):*
See multiDest comment?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

maciaszczykm:
Reviewed 8 of 19 files at r1, 4 of 4 files at r3.
Review status: 12 of 20 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/app/backend/localehandler.go, line 29 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCzAg0L6Zf0cBD9RvnM:-KCzAg0L6Zf0cBD9RvnN:-2028262086) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/src/app/backend/localehandler.go#L29)):*
`AL` is probably not the best name for a variable.

---

*[src/app/backend/localehandler.go, line 35 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCzAreJroovg5YYTJl-:-KCzAreJroovg5YYTJl0:349838035) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/src/app/backend/localehandler.go#L35)):*
Make it constant?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 14 unresolved discussions.

---

*[build/build.js, line 135 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyxJlW7xT5FPmvCTim:-KCzThwoO3SveSglGCJC:195842597) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L135)):*
So yes, I can confirm that with `async.parallel` on my machine `scritps:prod` occasionally errors. I do not understand why though, since it works fine with `async.series`... To me it is important that the build is reliable, and with `async.parallel` and `merge-stream` I was not able to achieve this ... Maybe someone knows what the problem with those is?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

bryk:
*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD2yroedZFctg5iRllf:266577509) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
> Second, I think the current multiDest implementation is wrong (or I misunderstand some part of it), hence why I removed it.
  outputStream.on('data', (data) => outputs.forEach((dest) => { dest.write(data); }));
  outputStream.on('end', () => outputs.forEach((dest) => { dest.end(); }));
Isn't that wrong, shouldn't we wait on 'end' event from all dest streams, and then end the outputStream? The logic seems reversed to me. After all, the outputStream should represent the execution of the multiple dest.

The implementation is correct to me. There is only one input stream and it writes to a few output streams. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 12 unresolved discussions.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD2zt9F2WLLwsvMYQHo:1603819234) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
My argument was not that the way the data is passed is wrong. The first line is fine, `outputStream` sends data to the `dest` streams when it gets data, cool. But why do we call the `end()` event on the `dest` streams when `outputStream` receives its end event? I don't see how the second line can be correct. 

E.g. in our build pipeline we have something like: `compile js & styles` -> `multiDest`. What if there is another stream after `multiDest` in the chain, that must be triggered when`multiDest` finishes. It is going to wait on the `end` event of the `outputStream` variable, right? Well, shouldn't outputStream actually wait on all the `dest` to trigger their `end()` events before it triggers its?

That's how the library `merge-stream` merges streams that need to be executed in one step. If you can convince me the second line of `multiDest` above is correct, I do not mind changing the pull request :D

And if my argumentation is correct, why not use something like `merge-stream` in the first place for this? Merge stream uses `through` internally anyways...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

bryk:
Review status: 11 of 20 files reviewed at latest revision, 17 unresolved discussions.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD2z5uWvLVvV9k3y8iU:-321322013) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
How about this implementation:

```js
let streams = conf.translations.map(() {
  let localizedOutputDirs = ...;
  return gulp.src(...)
    .pipe(multiDest(localizedOutputDirs)); 
}

return mergeStream.apply(null, streams); 
```

This has the benefit that everything runs in parallel, thus possibly cuts build time a lot. This is the spirit of gulp where all things are parallel and streamed. Other than that, canonical constructs are used to perform tasks, i.e., when you want to merge streams, you use mergeStream function instead of imperatively coding it.

What do you think? 

---

*[build/build.js, line 135 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyxJlW7xT5FPmvCTim:-KD304Z7bHTjHrh0G14c:-1386363700) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L135)):*
Can you show me the stacktrace? I think this may be because JSCompiler invocations aren't isolated. We should fix this. 

---

*[build/conf.js, line 154 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyiwYyn7cAoSyOTCD:-KD30Lk7DxP7ziWXAcTL:-405833205) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/conf.js#L154)):*
Ok, that's fine. When we need more english langs, we'll add them :) 

---

*[build/deploy.js, line 156 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyzx-ufv_qpnmIUFu:-KD30SWaxn5vSRYxQG_O:-1955001966) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/deploy.js#L156)):*
I believe mulitDest works correctly :) Can you show the case where it is wrong? 

---

*[build/i18n.js, line 53 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyzgr-wCe17OuRVQ90:-KD30obJ7Y1QBzbGztkA:-2078060934) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/i18n.js#L53)):*
Oh, thats interesting. Can you add this comment to the task here? :) For future readers.  

---

*[build/i18n.js, line 31 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD30b81Bo7aBUzYtX-G:-KD30b81Bo7aBUzYtX-H:-2064125034) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/i18n.js#L31)):*
You can just time it as {!Promise}. Promise object is in EcmaScript6 already.

---

*[build/script.js, line 58 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD31MSuo5co9fV38qjq:-KD31MSuo5co9fV38qjr:1205873034) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/script.js#L58)):*
You can include deps.js and base.js by setting overrides in bower.json, like here: https://github.com/AlexYankee/closure-ng/blob/master/bower.json. With this you can remove this change. 

---

*[build/script.js, line 143 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD32Z6Qh4nOVRhRfcKP:-KD32Z6Qh4nOVRhRfcKQ:-1349446317) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/script.js#L143)):*
mergeStream here? :) 

---

*[build/style.js, line 72 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD32ruCQMaYzAPv9jKQ:-KD32ruCQMaYzAPv9jKR:-777455799) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/style.js#L72)):*
MultiDest here? Styles are the same for all translations, right? The benefit of that is the fact that the compliation is done once, and files are just compied to multiple places. This is huge speedup. 

---

*[src/app/backend/localehandler.go, line 26 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD33LcLfu4OGKn6hRcx:-KD33LcLfu4OGKn6hRcy:919771383) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/src/app/backend/localehandler.go#L26)):*
Hmm... I see potential security threats when somebody sets the accept-language as, like, `../../../etc/passwd`. Can we maybe save translation keys in a JSON file and load it on startup to get the config? The json file can be also read from conf.js. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 17 unresolved discussions.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD343zS7K5FqxfYPksC:-356390238) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
Yes, that was my first implementation :) I ditched it because for some reason using it on the `gulp build-frontend` made it reaaaally slow (10 minutes). I did EXACTLY what you have written above :/ Do you have any clue why this could happen?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 17 unresolved discussions.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD35MPAiVUNP-8YavX1:-900409048) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
And again, I am still not convinced about `multiDest`? What is wrong in my logic in my previous comment?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 17 unresolved discussions.

---

*[build/build.js, line 135 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyxJlW7xT5FPmvCTim:-KD36bCuQ94zUtynATN_:1325278095) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L135)):*
That's all I get:

```
[11:26:05] Starting 'angular-templates'...
[11:26:05] Starting 'scripts'...
[11:26:08] Finished 'scripts' after 2.94 s
[11:26:08] Starting 'extract-translations'...
[11:26:08] Finished 'angular-templates' after 3.03 s
[11:26:10] Finished 'extract-translations' after 1.68 s
[11:26:10] Starting 'scripts:prod'...

stream.js:74
      throw er; // Unhandled stream error in pipe.
      ^
Error: ENOENT: no such file or directory, open 'app.js'
    at Error (native)
```

Is there a way to make node more verbose?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/app/backend/localehandler.go, line 26 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD33LcLfu4OGKn6hRcx:-KD37VJdhGUdyXeOGBgy:440402493) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/src/app/backend/localehandler.go#L26)):*
Nice catch, I'll see to it :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[build/deploy.js, line 156 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyyzx-ufv_qpnmIUFu:-KD3EWxog8r_b9T4ccIg:1569992697) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/deploy.js#L156)):*
Ok, sorry for the delay. Can you try out this example gulp task:

```
/**
 * multiDest fail example. It is supposed to copy the contents of dist
 * after the build-frontend:cross is finished.
 */
gulp.task('move-files', ['build-frontend:cross'], function() {
  gulp.src(path.join(conf.paths.dist, '**/*'))
      .pipe(gulp.dest(path.join(conf.paths.tmp, 'example')));
});
```
 It should copy the contents of the `dist` folder, correct? On my machine it copies peanuts (comparing the contents of `dist` and `.tmp/example`...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

bryk:
Review status: 11 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD8TMqwUKrcJg6K9Jxj:746625649) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
>  Do you have any clue why this could happen?

Because of the multidest bug? Or that out pipeline writes temp files that are not parallel? Can you revert to the version with mergeStream & multiDest so that we can debug this. I ask this because I believe this is the best long term solution - because we do the least amount of computation and everything is parallelized. 

---

*[build/build.js, line 135 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyxJlW7xT5FPmvCTim:-KD8j-AvUxkY5pBRDCQG:2080344145) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L135)):*
You should hack the https://github.com/steida/gulp-closure-compiler plugin to show all output of the compiler. Just modify files in node_modules. Can you revert this file to a state where it throws the error? So that we can debug.

---

*[build/build.js, line 133 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD8RRLcEBcgBmAFmRoZ:-KD8RRLdklp1WdXXpUKC:1340280438) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/build.js#L133)):*
Let's start the other thread here:

> My argument was not that the way the data is passed is wrong. The first line is fine, outputStream sends data to the dest streams when it gets data, cool. But why do we call the end() event on the dest streams when outputStream receives its end event? I don't see how the second line can be correct.
E.g. in our build pipeline we have something like: compile js & styles -> multiDest. What if there is another stream after multiDest in the chain, that must be triggered whenmultiDest finishes. It is going to wait on the end event of the outputStream variable, right? Well, shouldn't outputStream actually wait on all the dest to trigger their end() events before it triggers its?

You're correct! :) Can you fix the issue with multiDest in a separate PR? :) 

> That's how the library merge-stream merges streams that need to be executed in one step. If you can convince me the second line of multiDest above is correct, I do not mind changing the pull request :D
And if my argumentation is correct, why not use something like merge-stream in the first place for this? Merge stream uses through internally anyways...

Here we're not merging streams, but spiting them. So we have one stream of compiled stuff that we split into multiple destinations. Sounds right? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

taimir:
Review status: 11 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KD8k24xiaqrE5JXTEEd:-246247145) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
Oh sure, I don't mind investigating this :) I might have to open another pull request so we can check this out.

---

*[build/build.js, line 133 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD8RRLcEBcgBmAFmRoZ:-KD8kHvYUT6E8J00gb4A:1697216753) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/build.js#L133)):*
Haha, correct. I guess it's a matter of language here, we are indeed splitting what comes through the pipe into multiple streams (the `dest` streams). 

What I'm pointing out here is that this "splitting" process is actually a single step in our build pipeline. And it might be (eventually, if needed) followed by another step. So the whole splitting  should be seen as 1 stream by gulp (that's why `multiDest` used `through2` in the first place). Hence why I said "merged" into 1 stream. See the picture below.

The library `merge-stream` is no different than`async.parallel` (I've checked its code), it just uses the newer definitions of the `Stream` class in Node. It basically achieves exactly what we aim for: it allows us to split a pipeline flow into multiple flows (our `dest`), and we see this whole step as 1 single stream (hence why it's called `merge-stream`).

So in essence what we achieve is:
                                
```
			-------------------------------
			|  +--------dest 1 ----end()  |
			|  |			              |
--> src --->   |+-------dest 2 ----end()  |---end() when all
			|  |			              |    dest call their end()
			|  +--------dest 3 ----end()  |
			-------------------------------
				outputStream
				("merged" dest)

```

The encapsulation here is what was wrong in our `multiDest` implementation, because the end() event of the `outputStream` (which represents the overall step of splitting into multiple dirs) didn't wait for all `dest` to finish. So If we would want to add one more step after `multiDest`, it would be wrong (as I showed in my example).

I'll tackle this problem in a separate pull request and try to make it parallel (I know it is much better). 
The problem I had was that `merge-stream` just refused to work as expected, we'll have to investiage.

P.S. : ASCII Art FTW :D

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

bryk:
Review status: 11 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[build/build.js, line 132 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KCyy9Sfi2dN_ujJU_AF:-KDNjg-Tl6VU-faxoiCd:922616853) ([raw file](https://github.com/kubernetes/dashboard/blob/a286bd24c79282596f702b5bf7865add3fc4bc99/build/build.js#L132)):*
Yeah, perfect. Thank you! 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528)*

bryk:
@taimir Are there any action items left for me in this PR? Or are you doing the multiDest fix? 

---

Review status: 11 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[build/build.js, line 133 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KD8RRLcEBcgBmAFmRoZ:-KDNk8qqzq_XwsXI7nlS:-1457197270) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/build.js#L133)):*
Ok, perfect. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-:-KDNkAx5mgB_ewHGHbnu:814532020)*

taimir:
No, I think I do not need any more feedback from you. I am busy with another task (non-dashboard related) at the moment, but I will notify you ASAP when I have fixed this. It must be done in the next few days anyways :)

---

Review status: 11 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-:-KDO8UZPGX8pBkhePcvV:-809572933)*

taimir:
@cheld @bryk I have finally had some time to come back to this topic. I figured out where the problem with the [parallelization of](https://github.com/taimir/dashboard/blob/i18n-closure-get-msg/build/build.js#L135) `build-frontend` is. @bryk, can you check [this issue] (https://github.com/jonkemp/gulp-useref/issues/201). `gulpUseref` (used [here](https://github.com/taimir/dashboard/blob/i18n-closure-get-msg/build/build.js#L112)) fails to work when run in parallel. I can reproduce this bug on my machine. This is why I had trouble with `async.parallel()`, or any other parallel execution I tried... And also why `async.series()` actually worked.

On a second note: I looked deeper into the Node.js stream documentation this time, and it turns out our current `Multidest.js` implementation is O.K. after all. I am now quite sure we can keep it as is, or enhance it a bit if we want to support the "encapsulation" I was concerned about in this PR.

PTAL

bryk:
> @cheld @bryk I have finally had some time to come back to this topic. I figured out where the problem with the parallelization of build-frontend is. @bryk, can you check this issue. gulpUseref (used here) fails to work when run in parallel. I can reproduce this bug on my machine. This is why I had trouble with async.parallel(), or any other parallel execution I tried... And also why async.series() actually worked.

Can you add a TODO with link to the bug in all not-actually-needed occurrences of parallel work? This will help us change this to parallel when useref issue is fixed.

bryk:
Can you rebase? This PR is quite old and many things have changed since last update :)

bryk:
There's a few unresolved comments here. Please take a look at reviewable. 

---

Reviewed 2 of 19 files at r1, 1 of 4 files at r3.
Review status: 13 of 20 files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[.eslintrc, line 45 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZ8AOjJqxWu15LIUs:-KGGZ8APsEOh7tCBqKDT:1988603082) ([raw file](https://github.com/kubernetes/dashboard/blob/4857aa4c07cf04ab25a13d6775f067031f2b4206/.eslintrc#L45)):*
Revert? 

---

*[package.json, line 11 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZAg2R6R5pFKEHlXa:-KGGZAg2R6R5pFKEHlXb:-1358315204) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/package.json#L11)):*
Use ~ versioning. 

---

*[package.json, line 24 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZCOuQDYpr3T-338l:-KGGZCOuQDYpr3T-338m:1600333743) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/package.json#L24)):*
Please use ~versioning as all other deps.

---

*[package.json, line 57 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZFWDifO0kkRDSCnM:-KGGZFWDifO0kkRDSCnN:1600333743) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/package.json#L57)):*
Please use ~versioning as all other deps.

---

*[package.json, line 66 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZG5P8hIlBe0g6vR5:-KGGZG5P8hIlBe0g6vR6:1600333743) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/package.json#L66)):*
Please use ~versioning as all other deps.

---

*[build/build.js, line 135 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZLB2uS0xQGPl5f5d:-KGGZLB2uS0xQGPl5f5e:-142363001) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/build.js#L135)):*
So this is the only part that is done through series?

---

*[build/build.js, line 153 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZXJh6ahOlfSbH7ps:-KGGZXJh6ahOlfSbH7pt:520591754) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/build.js#L153)):*
So how about using multiDest here now? 

---

*[build/deploy.js, line 162 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZd6s_XpQD3PYCiRH:-KGGZd6s_XpQD3PYCiRI:520591754) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/deploy.js#L162)):*
So how about using multiDest here now? 

---

*[build/index.js, line 61 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZoQMaIiJvRMrDlaO:-KGGZoQMaIiJvRMrDlaP:30820126) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/index.js#L61)):*
I like this :) This fixes the issue when styles not always reload. 

---

*[build/index.js, line 74 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-KGGZsfuWSrC1uMa65h-:-KGGZsfuWSrC1uMa65h0:-1893851561) ([raw file](https://github.com/kubernetes/dashboard/blob/ef70abed8dfb5849db50f059bc99e0b2f8309f48/build/index.js#L74)):*
multiDest? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/528#-:-KGGbCyDjJ8WksSi3l-o:1962546957)*

taimir:
@bryk I have a workaround for the `useref` issue that prevented us from making the build concurrent. I'm now fixing up all of the open remarks and I will create a separate pull request with the changes (this one has become rather hairy) for you to review :)

bryk:
@taimir Thanks!

Should I be reviewing this PR? Or wait for the other one?

taimir:
Closing this, it finished in  https://github.com/kubernetes/dashboard/issues/725

PR #531, opened 2016-03-14
Bundle material icons and roboto font into the project so they can work offline

### LINKED ISSUE #529: Wrong display of icons in producion
##### Environment
```
Dashboard version: HEAD
Kubernetes version: HEAD
Operating system: Ubuntu 14.04 LTS
Node.js version: v4.2.3
Go version: 1.5.1
```

##### Steps to reproduce
Checkout HEAD of master branch of dashboard. If needed, run `npm install` to get the recent icons changes. Run `gulp serve:prod` to serve in production.

##### Observed result
The icons (e.g. in the deploy form) are not displayed correctly:

dashboard_issue

##### Expected result
The material design icons should be displayed correctly. The text on the screen should not overlap.

##### Comments
I think this was caused by 949f4c976a042ed9728753cf6d89eb803fd96276. When I check out a previous commit (e.g. fed0c9fed7f1a42a416005a7aa4b866999036f1a), there is no issue.

floreks:
Fixes #529

For the icons I've used this guide for self-hosting icons: http://google.github.io/material-design-icons/#icon-font-for-the-web

I've taken icon files from release 2.2.0: https://github.com/google/material-design-icons/tree/2.2.0/iconfont

For the fonts I've downloaded css font file from: http://fonts.googleapis.com/css?family=Roboto:300,400,500,700,400italic

To get `woff2` files I've downloaded font from https://www.google.com/fonts and converted `ttf` files to `woff2` files, then changed urls in css file to corresponding `woff2` files

Tested and works on server prod.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/531)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ cache:
 # Use Node.js as primary language because Gulp is the build system used in the project.
 language: node_js
 node_js:
-  - 4.2.2
+  - 5.1.1
 
 addons:
   # Run tests that require a browser on SauceLabs CI provider. Learn more at: https://saucelabs.com
--- a/docs/devel/getting-started.md
+++ b/docs/devel/getting-started.md
@@ -14,8 +14,8 @@ logic and fetches raw data from the various Kubernetes APIs.
 Make sure the following software is installed and added to the `$PATH` variable:
 * Docker (1.3+)
 * go (1.5+)
-* nodejs (4.2.2+)
-* npm (1.3+)
+* nodejs (5.1.1+)
+* npm (3+)
 * java (7+)
 * gulp (3.9+)
 
--- a/src/app/frontend/replicationcontrollerdetail/deletereplicationcontroller.html
+++ b/src/app/frontend/replicationcontrollerdetail/deletereplicationcontroller.html
@@ -30,8 +30,8 @@ limitations under the License.
       </md-icon>
     </md-checkbox>
     <md-dialog-actions>
-      <md-button class="md-primary" ng-click="ctrl.cancel()">Cancel</md-button>
-      <md-button class="md-primary" ng-click="ctrl.remove()">Delete</md-button>
+      <md-button class="md-primary kd-cancel-btn" ng-click="ctrl.cancel()">Cancel</md-button>
+      <md-button class="md-primary kd-delete-btn" ng-click="ctrl.remove()">Delete</md-button>
     </md-dialog-actions>
   </md-dialog-content>
 </md-dialog>
--- a/src/test/integration/deploy/deployfromfile_po.js
+++ b/src/test/integration/deploy/deployfromfile_po.js
@@ -27,9 +27,29 @@ export default class DeployFromFilePageObject {
     this.inputContainer = element(this.inputContainerQuery);
 
     this.filePickerQuery = by.css('.kd-upload-file-picker');
-    this.filePicker = element(this.filePickerQuery);
+    this.filePicker_ = element(this.filePickerQuery);
 
     this.mdDialogQuery = by.tagName('md-dialog');
     this.mdDialog = element(this.mdDialogQuery);
   }
+
+  /**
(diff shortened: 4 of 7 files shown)


cheld:
Process question: when reading the list of PR, it is a bit surprising that this PR has priority 0. Well, it is P0 because the previous implementation is broken. So, what about splitting this PR in a fix that reverts the changes and label this PR as P3 or enhancement?

floreks:
Revert PR was already merged. I've used the same labels as on issue but you're right that this PR is not a bug fix itself. I've changed labels.

bryk:
Hmm... Why cannot we use the bower component for this?  What's the problem?

floreks:
The problem was that required assets were not available in prod environment. We can access `assets` in prod build but there are no bower components in path.

Other option would to be to add these dependencies and extend gulp to copy needed files (pretty much the same that I've added to assets) and make them available in prod. They would also have to be available in dev so I'm not sure how we can assure that without just copying needed files to assets.

bryk:
I'd say we should avoid copying into the codebase anything we don't own. This is because the copy will for 100% get out of sync at some point. Also, it is not our responsibility to maintain updates of a lib. If this is possible, I strongly favor changing out build setup so that this problem is solved with use of bower libs.


AFAIK, every bower library gets its CSS merged into vendor.css, so when you add a lib there it should just work. Why doesn't it? Do you know?

floreks:
Probably because CSS itself is not enough and for both icons and font CSS imports some files that are not available in prod build. When url is provided it downloads not only css files but also woff2 files that are imported by related CSS.

zrzut ekranu z 2016-03-16 09-03-33

For icons only this is available as bower component and it is already out of sync: https://github.com/jossef/material-design-icons-iconfont

We may have to find something in the middle. Maybe gulp task that will download and update assets based on newest releases. With bower components there is always possibility that some fork will become outdated.

maciaszczykm:
We can always download from original repository instead of fork, https://github.com/google/material-design-icons. It's much more to download, but at least we'd have version which is always up to date. Rest can be handled from Gulp task.

bryk:
Ok, I think what @maciaszczykm is proposing is best. We can depend on  https://github.com/google/material-design-icons via bower. See how here [0]. Then we can inject all dependencies (woff, css, etc.) using wiredep [1]. The same what we do for CSS and JS from bower. See some hints [2].

With that we'll have only bower dependency that will download all we need. Nothing copied to our repo. 

What do you think? Does this help?

[0] http://stackoverflow.com/questions/16949173/how-to-install-latest-untagged-state-of-a-repo-using-bower
[1] https://www.npmjs.com/package/wiredep
[2] https://github.com/yeoman/generator-webapp/issues/71

floreks:
What about using [bower-installer](https://github.com/blittle/bower-installer) to download only needed files from repo instead of whole repository and directly copy them to `assets`. With 1 command we would have up-to-date icons and font already placed in assets. We could always just download needed files with this and use gulp to manage te deps if this gives some advantage.

What do you think?

bryk:
Yeah, but this is still copying not-our code into `src/`. This is tempting to do, but there's many problems with that. E.g., form my experience, developers will modify the copied code there. Hence, we'll get out of sync forever, or override the changes. Both is bad.

When we import this via bower (or any other external dependency), then there is no way we modify angular mateiral css or fonts, so we're safe.

Does this convince you?

floreks:
Good enough for me. Still, can we use bower-installer to only download needed files but place them in bower components, then use wiredep to inject needed files?

bryk:
bower-installer is yet another tool in our build pipeline. I'd rather avoid using it, unless there is a big gain. For me getting the material icons repo was a few seconds, which is short.

floreks:
Ok then. For now lets keep it simple. It can be added at any point anyway. I've only brought it up because clean npm install already takes 15-20m for me but this is due to low connection speed.

PR #532, opened 2016-03-15
Fix format; "=>'

f-higashi:
clang-format updated to v1.0.36 which checks double quotation.

https://travis-ci.org/kubernetes/dashboard/builds/116048772

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/532)

### DIFF
--- a/build/gocommand.js
+++ b/build/gocommand.js
@@ -108,8 +108,8 @@ function checkGoVersion() {
         if (semver.lt(currentGoVersion, minGoVersion)) {
           deferred.reject(
               new Error(
-                  `The current go version "${currentGoVersion}" is older than ` +
-                  `the minimum required version "${minGoVersion}". ` +
+                  `The current go version '${currentGoVersion}' is older than ` +
+                  `the minimum required version '${minGoVersion}'. ` +
                   `Please upgrade your go version!`));
           return;
         }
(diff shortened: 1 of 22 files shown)


cheld:
Thanks, nice - kind of....So, clang was updated to minor version 1.0.36 (because we use ~ in package.json). As a result our build broke and you fixed it, right? Hmm, what about Greenkeeper? If we would have specified a specific version, than Greenkeeper would have tested and suggested the update, right? 

Anyway, result is nice

f-higashi:
> As a result our build broke and you fixed it, right?

yes.

Currently, we specifies gulp-clang-format which depends on clang-format and does not update recently.
In this case, Greenkeeper does not work maybe.

floreks:
For me npm still download old version of `gulp-clang-format`. Even if I delete it still 1.0.23 is downloaded. Did you change the configuration? I'm using npm 3.8.0

floreks:
I've cleared cache and removed all `clang-format` dependencies. New version is downloaded.

floreks:
:lgtm:

---

Reviewed 22 of 22 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/532#-:-KCtNpgbF6OfSb2ULnCE:1417088721)*

PR #542, opened 2016-03-17
Fix relative time filter

### LINKED ISSUE #427: Time ago filter issue
##### Environment

```
Dashboard version: latest
Kubernetes version: v1.2.0-alpha.8.186+ac644e03bd6acb
Operating system: Ubuntu 14.04.1
Node.js version: v5.5.0
Go version: go1.5.1 linux/amd64
```

##### Steps to reproduce

1. Setup Kubernetes cluster and Dashboard on server with valiad time.
2. On second machine change time, for example by 1 hour ahead.
3. On second machine access Dashboard, create new pod and check time ago filter output on page.

##### Observed result

Difference between current local client machine time and server time (when pod was created) is displayed.

##### Expected result

Difference between server times is displayed.

##### Comments

Server time should be used in all cases. Otherwise we can spot situations like `didn't happen yet` on running pods.

maciaszczykm:
Relative time filter will now use server time as current time (if it will be provided, otherwise it works like before). It fixes issue with differences between client and server times. Set it up for filter on replication controller list and replication controller detail pages.

There is a way to make it work without need to pass time in HTML, but it will cause performance problems. Current implementation require only one time request per page refresh and if we will move whole logic to filter then we would have `n` requests, where `n` is number of cards on replication controller list page or number of pods on replication controller details page. I think in this case performance is more impotrant.

Fixes #427.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/542)

### DIFF
--- a/src/app/backend/confighandler.go
+++ b/src/app/backend/confighandler.go
@@ -1,0 +1,69 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+	"text/template"
+	"time"
+)
+
+// AppHandler is a application handler.
+type AppHandler func(http.ResponseWriter, *http.Request) (int, error)
+
+// AppConfig is a global configuration of application.
+type AppConfig struct {
+	// ServerTime is current server time (milliseconds elapsed since 1 January 1970 00:00:00 UTC).
+	ServerTime int64 `json:"serverTime"`
+}
+
+const (
+	// ConfigTemplateName is a name of config template
+	ConfigTemplateName string = "appConfig"
+	// ConfigTemplate is a template of a config
+	ConfigTemplate string = "var appConfig_DO_NOT_USE_DIRECTLY = {{.}}"
+)
+
+// ServeHTTP serves HTTP endpoint with application configuration.
+func (fn AppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if _, err := fn(w, r); err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError),
+			http.StatusInternalServerError)
+	}
+}
+
+func getAppConfigJSON() string {
+	log.Printf("Getting application global configuration")
+
+	config := &AppConfig{
+		// TODO(maciaszczykm): Get time from API server instead directly from backend.
+		ServerTime: time.Now().UTC().UnixNano() / 1e6,
+	}
+
+	json, _ := json.Marshal(config)
+	log.Printf("Application configuration %s", json)
+	return string(json)
+}
+
+func configHandler(w http.ResponseWriter, r *http.Request) (int, error) {
+	template, err := template.New(ConfigTemplateName).Parse(ConfigTemplate)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return http.StatusInternalServerError, err
+	}
+	return http.StatusOK, template.Execute(w, getAppConfigJSON())
+}
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -61,6 +61,8 @@ func main() {
 	// TODO(bryk): Disable directory listing.
 	http.Handle("/", http.FileServer(http.Dir("./public")))
 	http.Handle("/api/", CreateHttpApiHandler(apiserverClient, heapsterRESTClient, config))
+	// TODO(maciaszczykm): Move to /appConfig.json as it was discussed in #640.
+	http.Handle("/api/appConfig.json", AppHandler(configHandler))
 	log.Print(http.ListenAndServe(fmt.Sprintf(":%d", *argPort), nil))
 }
 
--- a/src/app/frontend/common/filters/relativetime_filter.js
+++ b/src/app/frontend/common/filters/relativetime_filter.js
@@ -14,6 +14,7 @@
 
 /**
  * Unit name constants (singular and plural form), that will be used by the filter.
+ *
  * @enum {!Array<string>}
  */
 const Units = {
@@ -27,6 +28,7 @@ const Units = {
 
 /**
  * Unit conversion constants.
+ *
  * @enum {number}
  */
 const UnitConversions = {
@@ -41,6 +43,7 @@ const UnitConversions = {
 
 /**
  * Time constants.
+ *
  * @enum {string}
  */
 const TimeConstants = {
@@ -50,17 +53,24 @@ const TimeConstants = {
 
 /**
  * Returns filter function to display relative time since given date.
+ *
+ * @param {!../appconfig/appconfig_service.AppConfigService} kdAppConfigService
  * @return {function(string): string}
+ * @ngInject
  */
-export default function relativeTimeFilter() {
+export default function relativeTimeFilter(kdAppConfigService) {
   /**
    * Filter function to display relative time since given date.
+   *
    * @param {string} value Filtered value.
    * @return {string}
    */
   let filterFunction = function(value) {
+    // Current server time.
+    let serverTime = kdAppConfigService.getServerTime();
+
     // Current and given times in miliseconds.
-    let currentTime = (new Date()).getTime();  // TODO(maciaszczykm): Use server time.
+    let currentTime = getCurrentTime(serverTime);
     let givenTime = (new Date(value)).getTime();
 
     // Time differences between current time and given time in specific units.
@@ -92,11 +102,25 @@ export default function relativeTimeFilter() {
       return formatOutputTimeString_(diffInYears, Units.YEAR);
     }
   };
+
   return filterFunction;
 }
 
 /**
+ * Returns current time. If appConfig.serverTime is provided then it will be returned, otherwise current
+ * client time will be used.
+ *
+ * @param {?Date} serverTime
+ * @return {number}
+ * @private
+ */
+function getCurrentTime(serverTime) {
+  return serverTime ? serverTime.getTime() : (new Date()).getTime();
+}
+
+/**
  * Formats relative time string. Sample results look following: 'a year', '2 days' or '14 hours'.
+ *
  * @param {number} timeValue Time value in specified unit.
  * @param {!Array<string>} timeUnit Specified unit.
  * @return {string} Formatted time string.
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -36,6 +36,7 @@ function getFileList() {
   };
 
   return wiredep(wiredepOptions).js.concat([
+    path.join(conf.paths.frontendTest, '**/*.json'),
     path.join(conf.paths.frontendTest, '**/*.js'),
     path.join(conf.paths.frontendSrc, '**/*.js'),
     path.join(conf.paths.frontendSrc, '**/*.html'),
--- a/src/app/backend/heapsterclient.go
+++ b/src/app/backend/heapsterclient.go
@@ -17,8 +17,8 @@ package main
 import (
 	"log"
 
-	client "k8s.io/kubernetes/pkg/client/unversioned"
 	restclient "k8s.io/kubernetes/pkg/client/restclient"
+	client "k8s.io/kubernetes/pkg/client/unversioned"
 )
 
 // HeapsterClient  is a client used to make requests to a Heapster instance.
--- a/src/app/externs/backendapi.js
+++ b/src/app/externs/backendapi.js
@@ -344,3 +344,6 @@ backendApi.SecretSpec;
  * }}
  */
 backendApi.SecretsList;
+
+/** @typedef {{serverTime: number}} */
+const appConfig_DO_NOT_USE_DIRECTLY = {};
--- a/src/app/frontend/common/appconfig/appconfig_module.js
+++ b/src/app/frontend/common/appconfig/appconfig_module.js
@@ -1,3 +1,4 @@
+
 // Copyright 2015 Google Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -12,23 +13,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import coresFilter from './cores_filter';
-import memoryFilter from './memory_filter';
-import middleEllipsisFilter from './middleellipsis_filter';
-import relativeTimeFilter from './relativetime_filter';
-import appConfigModule from '../appconfig/appconfig_module';
+import appConfigServiceProvider from './appconfig_serviceprovider';
 
 /**
- * Module containing common filters for the application.
+ * Angular module containing application configuration.
  */
 export default angular
     .module(
-        'kubernetesDashboard.common.filters',
+        'kubernetesDashboard.appconfig',
         [
           'ngMaterial',
-          appConfigModule.name,
         ])
-    .filter('middleEllipsis', middleEllipsisFilter)
-    .filter('kdMemory', memoryFilter)
-    .filter('kdCores', coresFilter)
-    .filter('relativeTime', relativeTimeFilter);
+    .provider('kdAppConfigService', appConfigServiceProvider)
+    .value('appConfig', appConfig_DO_NOT_USE_DIRECTLY);
--- a/src/app/frontend/common/appconfig/appconfig_service.js
+++ b/src/app/frontend/common/appconfig/appconfig_service.js
@@ -1,0 +1,43 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Application configuration service.
+ *
+ * @final
+ */
+export class AppConfigService {
+  /**
+   * @param {!appConfig_DO_NOT_USE_DIRECTLY} appConfig
+   * @ngInject
+   */
+  constructor(appConfig) {
+    /** @private {!appConfig_DO_NOT_USE_DIRECTLY} */
+    this.appConfig_ = appConfig;
+  }
+
+  /**
+   * Returns current server time.
+   *
+   * @return {?Date}
+   * @export
+   */
+  getServerTime() {
+    if (!isNaN(this.appConfig_.serverTime)) {
+      return new Date(this.appConfig_.serverTime);
+    } else {
+      return null;
+    }
+  }
+}
--- a/src/app/frontend/common/appconfig/appconfig_serviceprovider.js
+++ b/src/app/frontend/common/appconfig/appconfig_serviceprovider.js
@@ -1,0 +1,27 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {AppConfigService} from './appconfig_service';
+
+/**
+ * Application configuration provider.
+ */
+export default class AppConfigServiceProvider {
+  /**
+   * @param {appConfig_DO_NOT_USE_DIRECTLY} appConfig
+   * @ngInject
+   * @export
+   */
+  $get(appConfig) { return new AppConfigService(appConfig || {}); }
+}
--- a/src/app/frontend/index.html
+++ b/src/app/frontend/index.html
@@ -54,10 +54,14 @@ limitations under the License.
     <!-- endbower -->
     <!-- endbuild -->
 
+    <!-- Script with configuration generated at backend. -->
+    <script src="api/appConfig.json"></script>
+
     <!-- build:js static/app.js -->
     <!-- inject:js -->
     <!-- Application JS files are populated here. -->
     <!-- endinject -->
     <!-- endbuild -->
+
   </body>
 </html>
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
@@ -38,7 +38,7 @@ export default class ReplicationControllerDetailController {
    * @ngInject
    */
   constructor(
-      $mdMedia, $stateParams, $state, $resource, $log, replicationControllerDetail,
+      $mdMedia, $stateParams, $state, $log, replicationControllerDetail,
       replicationControllerEvents) {
     /** @export {function(string):boolean} */
     this.mdMedia = $mdMedia;
--- a/src/test/frontend/common/filters/relativetime_filter_test.js
+++ b/src/test/frontend/common/filters/relativetime_filter_test.js
@@ -16,6 +16,10 @@ import filtersModule from 'common/filters/filters_module';
 
 describe('Relative time filter', () => {
 
+  /**
+   * Current time mock.
+   * @type {Date}
+   */
   const currentTime = new Date(
       2015,  // year
       11,    // month
--- a/src/test/frontend/fakeServerConfig.json
+++ b/src/test/frontend/fakeServerConfig.json
@@ -1,0 +1,16 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/** @typedef {{serverTime: number}} */
+const appConfig_DO_NOT_USE_DIRECTLY = {};


floreks:
Reviewed 3 of 16 files at r1.
Review status: 3 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/backend/apihandler.go, line 466 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3YM88GMpYkG3CzfO2:-KD3YM88GMpYkG3CzfO3:-551710422) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/backend/apihandler.go#L466)):*
When you remove error then this won't be needed.

---

*[src/app/backend/time.go, line 41 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3Y7x0I3LhOn2FLfOf:-KD3Y7x1bVpjkoaAFSaD:-813998244) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/backend/time.go#L41)):*
Error here never occurs. It's not needed then.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KD3c2zAjyrp4BKrBEg2:1756720095) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
We should think of a way to avoid duplicating this code. Also if it's possible try to design filter in a way that it won't need any additional code to work on other pages and still have only 1 api call per page load.

Temporary solution would be to export this function and import in other file just to avoid duplicating code.

If there won't be a way to redesign filter, maybe we can make it a directive?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
Updated. PTAL

---

Review status: 3 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/backend/apihandler.go, line 466 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3YM88GMpYkG3CzfO2:-KD3dhdXKuqyw77Ly06h:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/backend/apihandler.go#L466)):*
Done.

---

*[src/app/backend/time.go, line 41 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3Y7x0I3LhOn2FLfOf:-KD3dk_cxTY1e8ba5VNW:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/backend/time.go#L41)):*
Done.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KD3doLEUi8wSp0Wocvp:-1470943013) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
Filter is standalone component and its instances aren't connected in any way, I don't see any other solution for it than passing server time directly to it. Of course current implementation doesn't require it, if it won't be provided then client time will be used. 

Exporting this function would allow to not repeat code, so it's good idea, but where we should place code like this?

Regarding idea with directive, again, I don't see any benefits from it. Still we would need to call API in directive (it means many API calls per page refresh) or pass value to it (so it will be done like now). Maybe I miss something. If you have any ideas, please let me know.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KD3eqzmzqLJTM4tuC37:141676887)*

floreks:
Review status: 1 of 16 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KD3haUlzPlGf6Yg1g1C:-679844635) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
Ok, directive will not help us. What should work is to move this logic to the service. It's singletone so constructor will be called only once per full page refresh. Getter can then just return initialized time value. We can inject service into the filter and this should work.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Review status: 1 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KD83nVMw60xRcjFqxFG:350684323) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
The second important topic you touched in this PR is how we do initialization of the application. Server time can be considered as application configuration constant, since it can be potentially used by many components. And is now used by a global service/filter. 

This is a problem that we should solve in a good and future-proof way. I imagine that over time we'll be getting more similar config objects, e.g., user preferences, selected language, last login time, apiserver version, etc.. If we follow the pattern that every bit of this thing has a separate api call, then we may get slow and complicated, since many components will be asynchronously calling for the same data. This is especially important for things in common/ folder. 

How about we do the configuration for all common services/directives/filters once and for all? That we have one application configuration object and with it we initialize services as needed? 

@maciaszczykm @floreks 

---

*[src/app/backend/time.go, line 38 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD8-S0TBGYJU0rNGjCj:-KD8-S0TBGYJU0rNGjCk:-1862659644) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/time.go#L38)):*
Hmm....? 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KD80rbs2Q7KxYStD2RJ:-1682036868) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
I think we should make the filter API so that it does not expose its implementation to clients. So, it should not require server time to be provided to it. 

How about making it that the filter initializes itself? This can be easily done. In short: on first call it initializes itself via shared service/variable and returns `undefined` until it is intialized. When initialization is done, it returns the real value. E.g., http://stackoverflow.com/questions/19046641/angularjs-asynchronously-initialize-filter

What do you think? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

floreks:
Review status: 1 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KD89pJDj3SNkzz9GpCa:-1409940716) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
I was thinking about async initializing service. So pretty much this logic would be in a service and filter would be responsible only for formatting.

What do you think?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Review status: 1 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KD8pNksxKehbA2pN-O9:-620209099) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
Sounds perfect! 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
Review status: 1 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KD97hPSmM5vJAQ5KCey:2024723883) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
Should it be some kind of configuration service or do you think about other kind of object?

---

*[src/app/backend/time.go, line 38 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD8-S0TBGYJU0rNGjCj:-KD97RzZT34cBCl0IiaY:-966296871) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/time.go#L38)):*
It formats time with following format, output is good:

```
2016/03/18 15:32:27 Getting current server time
2016/03/18 15:32:27 Current server time is 2016-03-18T15:32:27.889139+01:00
```

But I agree, that it could be moved to constant or I should find a better way to format time string :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Review status: 1 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KDNly0k08oH0e79PjOv:1642926742) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
Configuration service sounds just about right. 

What do you think about this? If you want, we can discuss over a chat or video.

---

*[src/app/backend/time.go, line 38 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD8-S0TBGYJU0rNGjCj:-KDNm70EPQ3tOAVEDifo:1467705057) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/time.go#L38)):*
Oh, I imagined that this always logs the same constant. I was wrong :) 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
Review status: 1 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KDSgbEbu6ESAlbxLNuC:-429546862) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
I've updated the code. It's still work in progress, but discussion at this point would be helpful for me. I've problem with getting display value updated. It should work like in http://jsfiddle.net/fJC5Q/3/:

``` javascript
var data = null, serviceInvoked = false;
    
    function realFilter(value) {
        // REAL FILTER LOGIC, DISREGARDING PROMISES
        return data + value;
    }
    
    return function(value) {
        if( data === null ) {
            if( !serviceInvoked ) {
                serviceInvoked = true;
                console.log("SIMULATION OF ASYNC CALL");
                $timeout(function() {
                    var result = "Async: ";
                    data = result;
                }, 2000);
            }
            return "-";
        }
        else return realFilter(value);
    }
```

Using above code in `relativetime_filter.js` produces different results than on fiddle. Do we use different configuration?

Is it okay for you to setup a meeting for today or should we discuss it tommorow when you will be in Lodz?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Review status: 1 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KDTCY4HJy4C6jU2zLdd:-454831225) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
I bet we have one-time binding which stops at the `-` value. 
Set up the meeting for today. Tomorrow we're talking about the UX. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Review status: 1 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KDTF42suNoQDb32T0qA:2133000673) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
Hmm... Did you set up the meeting for tomorrow? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
Review status: 1 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KDTF_PvYzdcbKXkWhj3:-1855576825) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
Ups, it was missclick.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

src/app/backend/confighandler.go
```diff
@@ -0,0 +1,61 @@
...
+			http.StatusInternalServerError)
+	}
+}
+
+func getAppConfig() *AppConfig {
+	log.Printf("Getting application global configuration")
+
+	config := &AppConfig{
+		// TODO(maciaszczykm): Get time from API server instead directly from backend.
+		// TODO(maciaszczykm): Time offset should be calculated to pass this value just once.
```
maciaszczykm: @bryk Do you know if there is a possibility to ensure that template is generated just once, not with every request to the page?

src/app/backend/dashboard.go
```diff
@@ -61,6 +61,7 @@ func main() {
 	// TODO(bryk): Disable directory listing.
 	http.Handle("/", http.FileServer(http.Dir("./public")))
 	http.Handle("/api/", CreateHttpApiHandler(apiserverClient, heapsterRESTClient, config))
+	http.Handle("/api/appConfig.js", AppHandler(configHandler))
```
maciaszczykm: Temporarily using `/api/*` path to avoid problems when running on `serve` instead of `serve:prod`. I'll try to use tip from #640.

src/app/frontend/common/filters/relativetime_filter.js
```diff
@@ -92,11 +97,28 @@ export default function relativeTimeFilter() {
...
 
 /**
+ * Returns current time. If appConfig.serverTime is provided then it will be returned, otherwise current
+ * client time will be used.
+ *
+ * @return {number}
+ * @private
+ */
+function getCurrentTime() {
+  if (appConfig && appConfig.serverTime) {
```
maciaszczykm: @bryk I've issue with making this line pass `gulp-closure-compiler` check. Could you take a look at it? Do you have any suggestions?

maciaszczykm:
Added template generation. Still few issues (please take a look at the comments).

PTAL

bryk:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/backend/confighandler.go, line 46 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KEfq9LZieYEY8-PsCuk-r5-46:-KF9d423AK31vkSw2LPM:-1639121488) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/backend/confighandler.go#L46)):*
I think we *must* generate this on every request. So current implementation is fine. What is the problem that you see? 

---

*[src/app/backend/dashboard.go, line 64 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KEfq9LZieYEY8-PsCuj-r5-64:-KF9dEDCFXSiykVgESGI:1834699543) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/backend/dashboard.go#L64)):*
Change `.js` to `.json` then.  And add a TODO. 

---

*[src/app/frontend/index.html, line 64 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KF9dSlSIO7crJ6cSVOz:-KF9dSlSIO7crJ6cSVP-:-794219072) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/index.html#L64)):*
Hmm... We need this before "Application JS files are populated here. This is because in app files we may already use the config synchronously. 

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 112 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3Hom2nDs9dasK2o7U-r5-112:-KF9dk897yQVJTLLuODz:-1719900732) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/common/filters/relativetime_filter.js#L112)):*
Add new extern definition to backendApi.js. Like this:

```js
/** @typedef {{serverTime: foo}} */
const appConfig; 
```

But what's more important, we should have an angular service for accessing this. So that we do not rely on global static objects. This is erroneous. We should do the same what Angular team did with $location and $window. So, how about creating a service named, .e.g., appConfigService. Placed somewhere in common/ folder. The service can be configured using providers, e.g., 

```js
myModule.provider('appConfigService', AppConfigServiceProvider);

//...

class AppConfigServiceProvider {
  /**
    * @param {appConfig} appConfig 
    * @export @ngInject
    */
  $get(appConfig) {
    return new AppConfigService(appConfig);
  }
}

//...
myModule.value('appConfig', appConfig_DO_NOT_USE_DIRECTLY);
```

If you think that the service is overkill, we can just add the app config to dependency injection by. This would work for now untill we need the service. 
```js
myModule.value('appConfig', appConfig_DO_NOT_USE_DIRECTLY);
```

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 112 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KF9gA33SCo10n_1Se_j:-KF9gA33SCo10n_1Se_k:653249485) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/common/filters/relativetime_filter.js#L112)):*
Other comment: how about we rename appConfig to (something like) `kubernetesDashboardAppConfig_DO_NOT_USE_DIRECTLY`. This would  make it sufficiently verbose to avoid potential conflicts and we need to tell ppl that they should use dependency injection for this, never use directly. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/backend/confighandler.go, line 46 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KEfq9LZieYEY8-PsCuk-r5-46:-KF9tTuLmCF-s29igeCu:331661127) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/backend/confighandler.go#L46)):*
Only performance issues, but at least now we have no problem with asynchronous calls like before.

---

*[src/app/backend/dashboard.go, line 64 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KEfq9LZieYEY8-PsCuj-r5-64:-KF9tJZ6nzQQgJ-AW0aU:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/backend/dashboard.go#L64)):*
Done.

---

*[src/app/frontend/index.html, line 64 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KF9dSlSIO7crJ6cSVOz:-KF9trglTIb0H5JVyFe_:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/index.html#L64)):*
Done.

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 112 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KF9gA33SCo10n_1Se_j:-KF9ui9MgjCeiMvDY-bV:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/common/filters/relativetime_filter.js#L112)):*
Done.

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 112 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3Hom2nDs9dasK2o7U-r5-112:-KF9zLBWD3kCKY3nd-ek:-694759893) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/common/filters/relativetime_filter.js#L112)):*
I'll do it tommorow.

---

*[src/app/backend/apihandler.go, line 191 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD83nVMw60xRcjFqxFF:-KF9zRQnfRJEdyXQNW0N:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/913c2b1c0b4e94b316ef1b281c6461dd94e37152/src/app/backend/apihandler.go#L191)):*
Done.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js, line 45 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3c2zAjyrp4BKrBEg1:-KF9zQIrPp9L0M1-RHAv:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/57504bd12db2d91aa007727ed5feeb39cad7dfa6/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js#L45)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Reply PTAL when ready for rereview. Also, please have all Javascript lines covered here.  

---

Review status: 0 of 7 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFDSPeQtqqOv0Ie23rO:1632621661)*

maciaszczykm:
Added provider everything bar annotations and tests seems to be done at the moment. I may need a bit more help with Closure Compiler. PTAL

---

Review status: 0 of 11 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/app/backend/confighandler.go, line 46 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KEfq9LZieYEY8-PsCuk-r5-46:-KFEpPUrnq9bcb_KSB8a:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/backend/confighandler.go#L46)):*
Done.

---

*[src/app/frontend/common/appconfig/appconfig_module.js, line 28 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFEpaFND3uCIUflEmM6:-KFEpaFND3uCIUflEmM7:-513085767) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_module.js#L28)):*
This one seems to be strange for me, it should see definition from backendapi.js, but it doesn't. Do you have any tips?

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 112 \[r5\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KD3Hom2nDs9dasK2o7U-r5-112:-KFEpLxiS6IDEQ1EDY4G:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/5531186656be9ccc4471f011692ba8294f813f48/src/app/frontend/common/filters/relativetime_filter.js#L112)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFEouXQO5oGUtJyxaGZ:2026737412)*

bryk:
Reviewed 2 of 15 files at r4, 1 of 13 files at r5, 2 of 5 files at r6, 3 of 6 files at r7.
Review status: 6 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/appconfig/appconfig_module.js, line 28 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFEpaFND3uCIUflEmM6:-KFEzuaW7VghFBn2rd43:-659735455) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_module.js#L28)):*
`gulp scripts:prod` work. This is a problem with something else. I think that tests do not have access to this variable, so they fail. See
```
gulp frontend-test
[16:22:55] Requiring external module babel-core/register
[16:22:56] Using gulpfile ~/src/github.com/dashboard/gulpfile.babel.js
[16:22:57] Starting 'frontend-test'...
Chrome 49.0.2623 (Linux 0.0.0) ERROR
  Uncaught ReferenceError: appConfig_DO_NOT_USE_DIRECTLY is not defined
  at /tmp/28f12f92a65dd9083e4f6e80bc066738.browserify:117 <- ../src/app/frontend/common/appconfig/appconfig_module.js:9:324
```


You need to create a file with `const appConfig_FOO = {};`. And place it in `src/test/frontend`. Name it foo.json so that it is not babelified. In karma.conf.js include `
    path.join(conf.paths.frontendTest, '**/*.json'),
    path.join(conf.paths.frontendTest, '**/*.js'),
    path.join(conf.paths.frontendSrc, '**/*.js'),
    path.join(conf.paths.frontendSrc, '**/*.html'),
  ]);`

Should work

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 37 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFF2bGfArubNDX41jev:-KFF2bGgUFuu609jriIi:-1906920314) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_service.js#L37)):*
This.appConfig_ is always nonNull. You don't need to check this. 

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 38 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFF2pnx6qXu598EqO-_:-KFF2pnx6qXu598EqO-a:1055274572) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_service.js#L38)):*
Server time, though. May be null. Please make all configuration params nullable. This is to be safe for the future when API changes. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
@bryk Updated annotations. Your advice was very accurate, thank you! However, now I'm dealing with problems with Travis CI. It fails during integration tests and I don't see connection to my changes. I'll work on it, but you could also take a look in your free time.

Tests remain to be updated, I need successful build to get Codecov report.

---

Review status: 6 of 13 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[src/app/frontend/common/appconfig/appconfig_module.js, line 28 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFEpaFND3uCIUflEmM6:-KFJiHW6trPRxwQCPIXA:1508991519) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_module.js#L28)):*
Done. Thanks!

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 37 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFF2bGfArubNDX41jev:-KFJiJRPu8uYKSSIVfJo:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_service.js#L37)):*
Done.

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 38 \[r7\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFF2pnx6qXu598EqO-_:-KFJiJl-PRw5ZTpJS3uV:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/4430809fe427530af0441c6a9fa9e733ae179c7d/src/app/frontend/common/appconfig/appconfig_service.js#L38)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFJhd2irouNVnLxHePV:189243149)*

bryk:
Re integration test: I had the same problem recently. Please synchronize to latest changes on master and try again. 

---

Review status: 6 of 13 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFNYCkODz35Wz3edK-j:874604926)*

bryk:
Reviewed 3 of 4 files at r8.
Review status: 9 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[src/app/backend/confighandler.go, line 55 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNYVMd9RurIiWBZt4n:-KFNYVMd9RurIiWBZt4o:1071081809) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/backend/confighandler.go#L55)):*
Do not do manual marshalling. Use the same JSON conversion as we use for API calls. Annotate AppConfig struct with field names and just marshal it using json package: https://gobyexample.com/json

---

*[src/app/frontend/common/appconfig/appconfig_module.js, line 28 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNYnStWkQsUtWXcN-b:-KFNYnStWkQsUtWXcN-c:-551605090) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/appconfig/appconfig_module.js#L28)):*
Remove TODO? 

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 20 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNYx17x4_W1rbac-N6:-KFNYx17x4_W1rbac-N7:1888054302) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/appconfig/appconfig_service.js#L20)):*
Remove `default` from here and you'll be able to annotate this in `@type` JSDocs. 

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 36 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZPAEEuwgrrWQt26R:-KFNZPAEEuwgrrWQt26S:-1169756876) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/appconfig/appconfig_service.js#L36)):*
Can you make this return `?Date` instance, not number.  It'll be easier to use in clients of this API 

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 54 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZ2nIXghpKr-2x-Mb:-KFNZ2nIXghpKr-2x-Mc:936213378) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/filters/relativetime_filter.js#L54)):*
You can annotate this when you remove `default` export from the service. Please do this :) 

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 59 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZARMOVU0Tffv5XfW:-KFNZARMOVU0Tffv5XfX:-133968671) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/filters/relativetime_filter.js#L59)):*
I bet we need `ngInject` now. 

---

*[src/test/frontend/foo.json, line 1 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZ_cowd1BG4XkoYUb:-KFNZ_cowd1BG4XkoYUc:-2044112217) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/test/frontend/foo.json#L1)):*
Hmm... How about naming this fakeServerConfig.json ? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
All done, will update tests after Codecov will create a new report.

---

Review status: 9 of 13 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[src/app/backend/confighandler.go, line 55 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNYVMd9RurIiWBZt4n:-KFOKQsRqC-7ar3tmWuF:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/backend/confighandler.go#L55)):*
Done.

---

*[src/app/frontend/common/appconfig/appconfig_module.js, line 28 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNYnStWkQsUtWXcN-b:-KFNi7YsdomWJ0kKcrq8:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/appconfig/appconfig_module.js#L28)):*
Done.

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 20 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNYx17x4_W1rbac-N6:-KFO-yuzuwfGy8bngpIg:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/appconfig/appconfig_service.js#L20)):*
Done.

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 36 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZPAEEuwgrrWQt26R:-KFO9jY4SMJ9ELuVt8hp:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/appconfig/appconfig_service.js#L36)):*
Done.

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 54 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZ2nIXghpKr-2x-Mb:-KFNv7QHJJOj6bSqBl3V:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/filters/relativetime_filter.js#L54)):*
Done.

---

*[src/app/frontend/common/filters/relativetime_filter.js, line 59 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZARMOVU0Tffv5XfW:-KFNhkptIraGU1ONaCIw:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/app/frontend/common/filters/relativetime_filter.js#L59)):*
Done.

---

*[src/test/frontend/foo.json, line 1 \[r8\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFNZ_cowd1BG4XkoYUb:-KFNhlpMg_zqmevdnBH9:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/f414fb7745545b6749c653fba639fb52adc239aa/src/test/frontend/foo.json#L1)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFOLSqtcSWzThgORUPB:-1614790180)*

maciaszczykm:
Tests remain to be fixed.

maciaszczykm:
PTAL

bryk:
:lgtm:

---

Reviewed 7 of 8 files at r10, 1 of 1 files at r11.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFcATO6Cf_6WVBXUAYU:1417088721)*

bryk:
One last thing: 

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 25 \[r11\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFcE_jbxAZSM3WbIrPt:-KFcE_jcF6FaWhpfyJ9y:827735963) ([raw file](https://github.com/kubernetes/dashboard/blob/1f9715d8aecc07ba57c35424595c2ad73544c63e/src/app/frontend/common/appconfig/appconfig_service.js#L25)):*
Hey, one last comment. You should make this service immune to inesistence of appConfig

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFcEFmfcWfT-F_M6e6h:-382130263)*

bryk:

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 25 \[r11\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFcE_jbxAZSM3WbIrPt:-KFcEhwlrsdV2aJ8JFxL:-1723243243) ([raw file](https://github.com/kubernetes/dashboard/blob/1f9715d8aecc07ba57c35424595c2ad73544c63e/src/app/frontend/common/appconfig/appconfig_service.js#L25)):*
This may happen when, e.g., doing a rolling update and there is older version of the UI running. With this `var appConfig_FOO` will be undefined. Can you maybe in the provide do `return new AppConfigService(appConfig || {});`? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

bryk:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/app/backend/confighandler.go, line 65 \[r11\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFcGBMKB9i7yswTXCkP:-KFcGBMLjA9DKhZf7u7n:-1342200380) ([raw file](https://github.com/kubernetes/dashboard/blob/1f9715d8aecc07ba57c35424595c2ad73544c63e/src/app/backend/confighandler.go#L65)):*
Don't panic, but return 500 error to the browser. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542)*

maciaszczykm:
PTAL

---

Review status: 12 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/app/backend/confighandler.go, line 65 \[r11\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFcGBMKB9i7yswTXCkP:-KFcR2wHlGbkloDlLszp:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/1f9715d8aecc07ba57c35424595c2ad73544c63e/src/app/backend/confighandler.go#L65)):*
Done.

---

*[src/app/frontend/common/appconfig/appconfig_service.js, line 25 \[r11\]](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-KFcE_jbxAZSM3WbIrPt:-KFcR3hn1jm0nFazglTD:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/1f9715d8aecc07ba57c35424595c2ad73544c63e/src/app/frontend/common/appconfig/appconfig_service.js#L25)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFcR1PdqdX5ZHgzafQH:676965300)*

bryk:
:lgtm:

---

Reviewed 2 of 2 files at r12.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFcVm8bACsqsQJdHZGA:1417088721)*

bryk:
Fix travis and merge. 

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/542#-:-KFcVo6oXuMMsMvZL1dy:-484494354)*

PR #552, opened 2016-03-21
Fix logs view inconsistency between different k8s versions

### LINKED ISSUE #423: Inconsistent behavior on logs page for different k8s master versions
#### Issue details
I'm using `local-up-cluster` with newest kubernetes HEAD. After deploying some card with wrong `containerImage` pod stays in pending state with errors. Logs page show `Internal error(500)`. On older cluster we can see normal log entry saying that pod is not started and is in state pending.

##### Environment

```
Dashboard version: HEAD
Kubernetes version: v.1.2.0-alpha-8 (Newest HEAD)
```

##### Steps to reproduce

1. Deploy app with not existing `containerImage`.
2. Wait for card to show error status.
3. Go to logs page.
4. Check same page on older kubernetes version.

##### Observed result

Inconsistent behavior for logs page between different kubernetes versions.

##### Comments

It's hard to say what is the expected behavior here. Pod is actually in pending state so the older version is correct to show this entry but there are some errors related to this pod and newer version actually displays internal error with event error message. For me newest behavior seems more convenient.

Here are some screens.

Newest k8s:
zrzut ekranu z 2016-02-25 09-57-39

Older k8s:
zrzut ekranu z 2016-02-25 10-01-48

What do you think? @bryk @maciaszczykm @cheld

floreks:
Fixes #423

With this small change when kubernetes API returns error it will process it and show it as log and not internal error in the dashboard UI.

Before:
zrzut ekranu z 2016-02-25 09-57-39

After:
zrzut ekranu z 2016-03-21 10-07-52

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/552)

### DIFF
--- a/src/app/backend/logs.go
+++ b/src/app/backend/logs.go
@@ -81,7 +81,7 @@ func getRawPodLogs(client *client.Client, namespace, podID string, logOptions *a
 
 	readCloser, err := req.Stream()
 	if err != nil {
-		return "", err
+		return err.Error(), nil
 	}
 
 	defer readCloser.Close()


bryk:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/552#-:-KDNqe0cCAxNLRmofh6g:1417088721)*

Issue #443, opened 2016-02-25
App list UI gets really slow when there is an RC with many failing pods


bryk:
Try deploying app with 10000 replicas. 

This is because we check for errors for every pod. We could stop doing this after first error is found.

cc @floreks

floreks:
What if there will be 2 pods with different errors?

bryk:
Dont we already display "and possibly others"? Which means we trim the list to one?

floreks:
We actually do not. Only actual error messages are displayed. Adding `and possibly others` suffix does not fit every error msg. We would end with:

`pod (frontend-p5mjc) failed to fit in any node fit failure on node (127.0.0.1): Node didn't have enough resource: CPU, requested: 1000, used: 7620, capacity: 8000 and possibly others` or

`Failed to pull image "longname-longname-longname-longname-longname-longname": Error: image library/longname-longname-longname-longname-longname-longname not found and possibly others`

bryk:
Can we get events in one call instead of N calls? This should work according to the API: http://kubernetes.io/third_party/swagger-ui/#!/api%2Fv1/listNamespacedEvent

Am I correct?

floreks:
There is TODO in code for this I think. `getEvents` call should be changed to get events in 1 call.

bryk:
All right.

floreks:
I've already working solution for that. Just need to add some tests. UI speed with many failed pods increased significantly.

cheld:
So the steps are:
1. Go to deploy page
2. Enter app data
3. enter invalid image
4. enter 10000 as pod count

But if I enter a valid image, then the cluster would have much bigger problems, right?

In general the pod count should be limited to much smaller numbers, I think

maciaszczykm:
@cheld We cannot limit pod count if core doesn't do that. IMO we can display confirmation dialog, i.e. `Are you sure? It can slow down your cluster...` (just a template) instead.

floreks:
The only thing that we can do is display a warning asking if user is sure that he wants to create App with so many pods. There may be some users with big clusters that can handle even 10000 pods. For 1000 nodes cluster this is 10 pods per node.

cheld:
We know that kubernetes supports currently 30 pods per node. This could be starting point. 

We could make pop-up (similar to delete) or add a red message to the learn more text on the right side

floreks:
[...]

PR #587, opened 2016-03-29
Make single API call to get events and filter them on backend

### LINKED ISSUE #402: RC card is pending while it should be in error state
It has pending pods that *fail* to schedule. It is pending though.
screen shot 2016-02-19 at 12 34 17

@floreks opinions?

floreks:
This is my proposal to solve issue #443 as discussed with @bryk here: https://github.com/floreks/dashboard/commit/f050cf3d45373b69733a26757dcfac384184f720

This change will improve performance of both replication controllers list page and replication controller detail page. Now instead of making N api calls to get pod events there is only 1 API call that gets all events in given namespace and then filters events based on given list of pods.

This should be extended to all places where N api calls are made based on some resource i.e. RCs.

This also fixes #402. When last pod on the list didn't have error, card state was shown as pending.

@maciaszczykm @bryk @cheld PTAL

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/587)

### DIFF
--- a/src/app/backend/events.go
+++ b/src/app/backend/events.go
@@ -17,7 +17,7 @@ package main
 import (
 	"log"
 
-	api "k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/fields"
@@ -142,54 +142,18 @@ func GetReplicationControllerPodsEvents(client *client.Client, namespace, replic
 		return nil, err
 	}
 
-	events, err := GetPodsEvents(client, pods)
-
-	if err != nil {
-		return nil, err
-	}
-
-	return events, nil
-}
-
-// GetPodsEvents gets events associated to given list of pods
-// TODO(floreks): refactor this to make single API call instead of N api calls
-func GetPodsEvents(client *client.Client, pods *api.PodList) ([]api.Event, error) {
-	events := make([]api.Event, 0, 0)
-
-	for _, pod := range pods.Items {
-		list, err := GetPodEvents(client, pod)
-
-		if err != nil {
-			return nil, err
-		}
-
-		for _, event := range list.Items {
-			events = append(events, event)
-		}
-
-	}
-
-	return events, nil
-}
-
-// GetPodEvents gets events associated to given pod
-func GetPodEvents(client client.Interface, pod api.Pod) (*api.EventList, error) {
-	fieldSelector, err := fields.ParseSelector("involvedObject.name=" + pod.Name)
-
-	if err != nil {
-		return nil, err
-	}
-
-	list, err := client.Events(pod.Namespace).List(api.ListOptions{
+	eventList, err := client.Events(namespace).List(api.ListOptions{
 		LabelSelector: labels.Everything(),
-		FieldSelector: fieldSelector,
+		FieldSelector: fields.Everything(),
 	})
 
 	if err != nil {
 		return nil, err
 	}
 
-	return list, nil
+	events := filterEventsByPodsUID(eventList.Items, pods.Items)
+
+	return events, nil
 }
 
 // AppendEvents appends events from source slice to target events representation.
--- a/src/app/backend/eventscommon.go
+++ b/src/app/backend/eventscommon.go
@@ -16,8 +16,7 @@ package main
 
 import (
 	"k8s.io/kubernetes/pkg/api"
-	client "k8s.io/kubernetes/pkg/client/unversioned"
-	"log"
+	"k8s.io/kubernetes/pkg/types"
 	"strings"
 )
 
@@ -30,36 +29,28 @@ import (
 var FailedReasonPartials = []string{"failed", "err", "exceeded", "invalid", "unhealthy",
 	"mismatch", "insufficient", "conflict", "outof", "nil"}
 
-// GetPodsEventWarnings returns warning pod events based on given list of pods.
+// GetPodsEventWarnings returns warning pod events by filtering out events targeting only given pods
 // TODO(floreks) : Import and use Set instead of custom function to get rid of duplicates
-func GetPodsEventWarnings(client client.Interface, pods []api.Pod) (result []Event, err error) {
+func GetPodsEventWarnings(eventList *api.EventList, pods []api.Pod) []Event {
+	result := make([]Event, 0)
+	if eventList == nil {
+		return result
+	}
+
+	// Filter out only warning events
+	events := getWarningEvents(eventList)
+	failedPods := make([]api.Pod, 0)
+
+	// Filter out only 'failed' pods
 	for _, pod := range pods {
 		if !isRunningOrSucceeded(pod) {
-			log.Printf("Getting warning events from pod: %s", pod.Name)
-			events, err := GetPodEvents(client, pod)
-
-			if err != nil {
-				return nil, err
-			}
-
-			result = getPodsEventWarnings(events)
+			failedPods = append(failedPods, pod)
 		}
 	}
 
-	return removeDuplicates(result), nil
-}
-
-// Returns list of Pod Event model objects based on kubernetes API event list object
-// Event list object is filtered to get only warning events.
-func getPodsEventWarnings(eventList *api.EventList) []Event {
-	result := make([]Event, 0)
-
-	var events []api.Event
-	if !isTypeFilled(eventList.Items) {
-		eventList.Items = fillEventsType(eventList.Items)
-	}
-
-	events = filterEventsByType(eventList.Items, api.EventTypeWarning)
+	// Filter events by failed pods UID
+	events = filterEventsByPodsUID(events, failedPods)
+	events = removeDuplicates(events)
 
 	for _, event := range events {
 		result = append(result, Event{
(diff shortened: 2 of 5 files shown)


bryk:
Can we have one events list request per page refresh? Right now it is O(number or cards). In normal case it takes long time to get all events. I did the timing:
```
2016/03/30 10:57:22 Outcoming response to [::1]:42433 with 201 status code
2016/03/30 10:57:35 Incoming HTTP/1.1 GET /api/v1/replicationcontrollers request from [::1]:42434
2016/03/30 10:57:35 Getting list of all replication controllers in the cluster
2016/03/30 10:57:36 Get took 321.305297ms
2016/03/30 10:57:36 Filter took 1.442698ms
2016/03/30 10:57:36 Get took 184.528953ms
2016/03/30 10:57:36 Filter took 1.551455ms
2016/03/30 10:57:36 Get took 129.390919ms
2016/03/30 10:57:36 Filter took 3.112µs
2016/03/30 10:57:36 Get took 129.93973ms
2016/03/30 10:57:36 Filter took 4.145µs
2016/03/30 10:57:37 Get took 130.349689ms
2016/03/30 10:57:37 Filter took 4.43µs
2016/03/30 10:57:37 Get took 129.520676ms
2016/03/30 10:57:37 Filter took 4.229µs
2016/03/30 10:57:37 Get took 129.582782ms
2016/03/30 10:57:37 Filter took 3.962µs
```

---

Reviewed 3 of 5 files at r1.
Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/backend/eventscommon.go, line 67 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KE5fdppC7DvYAxZaycR:-KE5fdpqclKBL-5j4fyo:321963821) ([raw file](https://github.com/kubernetes/dashboard/blob/86f10ec20c94e2c6008a288256a09865b254ab84/src/app/backend/eventscommon.go#L67)):*
How about filtering by UID rather than the name?  event.InvolvedObject.UID exists. It is more precise than the name.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KE5jhWGULSkV4treu12:-940547514)*

floreks:
It's done this way to narrow number of events that we get for filtering to single namespace.  I was tempted to do single call but also it may be dangerous for big clusters with many namespaces. Filtering takes O(n + m) where n is number of pods and m is number of events we need to filter.

For now we can make single call but keep in mind that for some corner cases it may be actually slower.

---

Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/backend/eventscommon.go, line 67 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KE5fdppC7DvYAxZaycR:-KE5qyxSqIkuHvzYuPqK:1556207734) ([raw file](https://github.com/kubernetes/dashboard/blob/86f10ec20c94e2c6008a288256a09865b254ab84/src/app/backend/eventscommon.go#L67)):*
Sure, that makes sense.  I've used name as it's unique for the namespace and I couldn't think of a case where we get events for list of pods located in different namespaces.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KE5oUsiv2ECH4MkQucw:593502228)*

bryk:
I think this is slower in typical cases (many valid cards) and faster in corner cases (there's a card with many pods).

In the case of RC list we get events from all namespaces anyway, so we could get them in one call. 

Or, how about making one call for each namespace? Instead of a call for each card? 

---

Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KE5s9-Vkh0zkx-Zr3n1:-190458142)*

floreks:
I agree. Ideally we should get events per namespace only once  and then use already retrieved data without making API call again.

---

Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KE5wnsOO_5_u1mRJgap:218718203)*

floreks:
Review status: 2 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/app/backend/replicationcontrollerlist.go, line 101 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KEAtxjbuLbFpwpRTX00:-KEAtxjbuLbFpwpRTX01:-480713854) ([raw file](https://github.com/kubernetes/dashboard/blob/47ba8920b637ec73d1245a234a97ec5f7f96e7ba/src/app/backend/replicationcontrollerlist.go#L101)):*
I'm thinking about solution design for this problem. We need some kind of storage to keep events per namespace at least for the time when they are used in single place. Now responsibility of freeing the memory and invalidating 'cached' events is on dev. We could always make some go routine and set timeout to some short period of time to make sure that it will be cleared. Other possibility is to use some 3rd party or built in caching library for go.

Simple solution that would not need any caching would be to get all events in all namespaces and just filter them.

@maciaszczykm @bryk What do you think?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587)*

bryk:
Review status: 2 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/app/backend/replicationcontrollerlist.go, line 101 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KEAtxjbuLbFpwpRTX00:-KEAvlKAD1gHs1J3EZlT:-460740306) ([raw file](https://github.com/kubernetes/dashboard/blob/47ba8920b637ec73d1245a234a97ec5f7f96e7ba/src/app/backend/replicationcontrollerlist.go#L101)):*
Do not do global caching in the UI code for now. We should keep the backend stateless for as long as possible. When we decide to add state to the backend, this should be carefully designed, not just added. Do you agree?


How about you do caching on per-request basis? I.e., you create a cache (e.g., Map) at the start of the request and then pass it to methods that use/fill it. When the request is done, the cache is gone.

Other way is to get all events for the list page and from one namespace for details page. Basically two implementations.


Does this help? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587)*

maciaszczykm:
Review status: 2 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/app/backend/replicationcontrollerlist.go, line 101 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KEAtxjbuLbFpwpRTX00:-KEAxCtsQoZw31RxKlVy:-748398207) ([raw file](https://github.com/kubernetes/dashboard/blob/47ba8920b637ec73d1245a234a97ec5f7f96e7ba/src/app/backend/replicationcontrollerlist.go#L101)):*
I'd also try to avoid global caching. It can bring many issues and we are planning to do many changes in the application now. Of course it would improve performance, but at the moment I'd use maps or something like that.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587)*

floreks:
Review status: 2 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/app/backend/replicationcontrollerlist.go, line 101 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KEAtxjbuLbFpwpRTX00:-KEAx9un1LupK3KGqFz-:304216938) ([raw file](https://github.com/kubernetes/dashboard/blob/47ba8920b637ec73d1245a234a97ec5f7f96e7ba/src/app/backend/replicationcontrollerlist.go#L101)):*
Creating map at the beginning of the rquest was my first thought but then I'd need method which accepts map, client and namespace. It just isn't very intuitive. I don't like global cache also because it introduces other complex issues. What I need is to keep state of object at least per method call, but what I want is to keep code simple and not force devs to pass map to the method.

Should we just stick with getting all events for list page and ignore namespaces for now?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587)*

floreks:
Getting all events should be sufficient for most use cases. Now for 1000 pods and 650 events it takes roughly ~500us to filter them.

```
2016/03/31 11:45:09 Events: 654
2016/03/31 11:45:09 Pods: 1000
2016/03/31 11:45:09 Filterting took: 421.13µs
2016/03/31 11:45:09 Events: 654
2016/03/31 11:45:09 Pods: 45
2016/03/31 11:45:09 Filterting took: 218.857µs
2016/03/31 11:45:09 Events: 654
2016/03/31 11:45:09 Pods: 88
2016/03/31 11:45:09 Filterting took: 112.969µs
2016/03/31 11:45:09 Events: 654
2016/03/31 11:45:09 Pods: 100
2016/03/31 11:45:09 Filterting took: 121.288µs
2016/03/31 11:45:09 Events: 654
2016/03/31 11:45:09 Pods: 0
2016/03/31 11:45:09 Filterting took: 35.015µs
```

To not overcomplicate things I'll prepare simpler solution.

---

Review status: 2 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KEB2z2bNS20674qNq02:1791410969)*

floreks:
PTAL

bryk:
:lgtm:

Nice! I see huge speedup. Thanks for fixing this!

---

Reviewed 1 of 4 files at r2, 4 of 5 files at r3.

---

*[src/app/backend/replicationcontrollerlist.go, line 101 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KEAtxjbuLbFpwpRTX00:-KEBVMV54I5-B4CO1bu0:-106828166) ([raw file](https://github.com/kubernetes/dashboard/blob/47ba8920b637ec73d1245a234a97ec5f7f96e7ba/src/app/backend/replicationcontrollerlist.go#L101)):*
Sounds really good! 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KEBW-wJJW2kVNV5Qrbt:448217262)*

bryk:
@maciaszczykm  Please also review and if all is OK merge. 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KEBW5oEPktSYj9ky9lS:1318994263)*

floreks:
Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/backend/replicationcontrollerlist.go, line 101 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-KEAtxjbuLbFpwpRTX00:-KEB_ouSir0sBt7Eg0lp:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/47ba8920b637ec73d1245a234a97ec5f7f96e7ba/src/app/backend/replicationcontrollerlist.go#L101)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587)*

bryk:
Reviewed 3 of 3 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587)*

maciaszczykm:
:lgtm:

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kubernetes/dashboard/587#-:-KEBawyRc2pxgJpxIQni:1417088721)*

Issue #584, opened 2016-03-29
Design and implement UX that can support all K8s resources and scale for the future


bryk:
Umbrella issue to design and implement UX that can support all K8s resources and scale for the future.

This effort was already started by one day design sprint that we had on March 23 2015: https://github.com/kubernetes/dashboard/pull/577.

Issue #21, opened 2015-10-23
Console thoughts


bgrant0607:
Copying from https://github.com/kubernetes/kubernetes/issues/8270:

This was discussed at the Kubernetes contributors meeting in December.

First of all, visualization/dashboard is more important than actuation. The CLI (kubectl) is expected to be the main tool for controlling the system and applications running on it. Table-friendly views should also be incorporated into kubectl, not just a GUI. I want to ensure that we maintain a common look and feel and functionality with kubectl for analogous table views and use cases -- the primary use cases and essential information are identical. That's not to say it shouldn't support actuation, but that creates additional issues, such as SSO auth.

With respect to visualization/dashboards,  we need views that are customized to the most common use cases:

1. Understanding the system architecture and/or application topology: diagram view that shows interconnection between objects
2. Deployment status: what images/versions are deployed, what deployments are in progress, what pods deviate, health by deployment version, provenance info
3. App debugging: for devs -- what’s failing most recently / most often, why, how often, logs, events
4. System debugging: dashboard w/ system and node health, uptime, versions, config, status events, change history, surface logs (from builds, containers or container failures), what’s where (node-centric view: example visualization: http://azure.microsoft.com/blog/2014/08/28/hackathon-with-kubernetes-on-azure/)
5. Resource usage analysis: how much resources  (mem, cpu, disk) are being used (current/historical, individual/distribution), usage relative to quota or limits, why did my thing run out of resources, top (sort by decreasing usage)
6. App dashboard / launchpad: status overview, launch links to app-domain, custom app dashboards, links to cadvisor and/or kubelet GUIs, links to elasticsearch and heapster (logging and monitoring dashboards), etc.

Some functionality is probably lacking in order to support all these views at the moment (e.g., stats collection).

Presentation guidelines:
* timeseries graphs: single-column, all same time-scale, for easy correlation
* text:
  * dense (information density)
    * multi-column tables
    * alternate colors of rows/columns rather than introduce whitespace
  * highlight high-value content
    * sort by “most important” information for the view (e.g., priority, scale, recency, eventful)
    * color to highlight important/abnormal details
    * don’t show fields in tables by default that are almost the default/boilerplate values
* flexible grouping (e.g., tree view), filtering, sorting, searching, aggregating
  * by namespace
  * by labels
  * by reverse label lookup
  * by image
  * by phase

Come up with recommended semantic labels/annotation keys and meanings:
* env (dev, test, stage, prod)
* source/containing object/provenance
[...]

PR #589, opened 2016-03-29
Design doc for next version of user experience of the UI

bryk:
Tracking issue: #584

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/589)


bryk:
Please review and CC others @dodwyer @preillyme @joeatwork @bgrant0607-cc

bgrant0607:
cc @pwittrock @janetkuo

janetkuo:
Will you add filters to tabularized resource lists? For example, filter by status, by label, ...etc. It'd be useful when, say, I want to find all "failing" pods with label "app=nginx, track=stable".

janetkuo:
Some resources have hierarchical relationships. For example, a Deployment manages multiple ReplicaSets, each manages multiple Pods. How you plan to present that relationship (parent/children) in the detail page of each resource? 

Does it look like this:

Deployment foo:
 - ReplicaSet: foo-123, foo-456, ... (the children)
 - Pods: foo-123-aaa, foo-456-aaa, ... (the grandchildren?)
 - Service: foo-service

ReplicaSet foo-123:
 - Deployment: foo (the parent)
 - Pods: foo-123-aaa, foo-123-bbb (the children)
 - Service: foo-service

Pod foo-123-aaa:
 - Deployment: (grandparent?) 
 - ReplicaSet: (parent)
 - Service: foo-service

Service foo-service:
 - Deployment: 
 - ReplicaSet:
 - Pods:

janetkuo:
Do we plan to show that hierarchical relationships in the apps view?

bgrant0607:
Ref #21

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+These limitations have been number one source of feature requests and
+user complaints (e.g., #564, #524, #509, #510, #232). The core team also
+considers solving these limitations as a natural evolution of the UI.
+
+## Problem statement
+
+* Scale the UX to support displaying and editing all current and future
+  Kubernetes resources (e.g., replica sets, daemon sets, container, nodes,
+  volumes, secrets, deployments, etc.).
+* Scale for the number of objects displayed (currently: cards are suitable for
```
bgrant0607: Cards are suitable for <= 6 IMO

bgrant0607: We need to keep multiple audiences and scenarios in mind. I outlined some operational scenarios on #21. 

Personally, I think most people running workloads in production won't mutate them with the UI:
http://martinfowler.com/bliki/InfrastructureAsCode.html

However, one scenario I didn't cover is demos. The current functionality is aimed at that use case, IMO (as was kube-ui), and there's more we could do there, such as visualizations of cluster and app topology.

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+  O(20), desired: show any number of objects).
+* Guide users through use-case based views (don't just do API dump).
+* Make the UI aware of namespaces and multi-cluster installations.
+
+# Design
+The design is based on a few novelties to the UX of Kubernetes Dashboard. This
+is:
+* introduction of use-case based navigation menu and views
+* replacement of free-form cards by tabularized resource lists
```
bgrant0607: Total replacement, or alternative formats? For instance, file explorers generally allow users to choose between icons and detailed lists, or in some cases choose automatically based on the number of items.

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+* being explicit about Kubernetes resources displayed in the UI
+* introduction of action bar for performing all actions on resources
+
+## Navigation
+Navigation is realized by left hand side menu. The menu consists of use case
+driven categories that group Kubernetes resources. For example, there may be
+category called "Apps" which deals with Replication Controllers, Pods,
```
bgrant0607: We should talk about how to present the concepts, because I want to align our docs and the UI.

https://github.com/kubernetes/kubernetes/issues/22687

cc @kubernetes/docs

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+  O(20), desired: show any number of objects).
+* Guide users through use-case based views (don't just do API dump).
+* Make the UI aware of namespaces and multi-cluster installations.
+
+# Design
+The design is based on a few novelties to the UX of Kubernetes Dashboard. This
+is:
+* introduction of use-case based navigation menu and views
+* replacement of free-form cards by tabularized resource lists
+* being explicit about Kubernetes resources displayed in the UI
```
bgrant0607: This is a tradeoff. kubectl run and expose deliberately gloss over the underlying resource model. There's a place for that in the getting-started/demo experience.

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+
+![Mobile navigation](mockups/23-03-2016-scale-and-navigation/mobile-navigation.png)
+
+## View templates
+
+This section shows concepts used as templates for all views.
+
+View which does not deal with a resource or a resource list, e.g., landing
+page which shows an overview of a cluster. The page consists of, potentially
+dismissable, blocks that display data.
```
bgrant0607: Like the idea of pointing to docs. 

Would like to chat about how we could make docs a more effective help tool for UI users.

cc @kubernetes/docs

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+![Landing page](mockups/23-03-2016-scale-and-navigation/landing-page.png)
+
+Category view which shows lists of resources. In this mockup it is
```
bgrant0607: I probably wouldn't put pods in the same view as all controllers. 

We should chat about how to organize the pod view: sorting, grouping, filtering, aggregation, etc. By names, labels, controllers, services.

bgrant0607: I don't think status info should only be presented as freeform info in the card. Info like number of available replicas in a Deployment, number of container restarts in a pod, etc. is useful to sort on.

We'll want logs links on pods for debugging use cases. An exec widget would be awesome, too.

Rather than just one labels column, I'd like users to be able to add labels as columns, as with kubectl -L, to facilitate grouping and sorting.

bgrant0607: In general, take a look at what we've been doing with kubectl output. We moved big, unstructured info like images, labels, selectors to optional "wide" output. We also don't display nodes by default for pods.

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+* being explicit about Kubernetes resources displayed in the UI
+* introduction of action bar for performing all actions on resources
+
+## Navigation
+Navigation is realized by left hand side menu. The menu consists of use case
+driven categories that group Kubernetes resources. For example, there may be
+category called "Apps" which deals with Replication Controllers, Pods,
+Deployments, etc., and "Config" which deals with Secrets, Volumes,
+Namespaces, etc. The exact definition of categories will happen in a separate
+design.
```
bgrant0607: We should consider a cluster admin category.

docs/design/navigation-and-scale.md
bgrant0607: Idea: profiles (perhaps customizable/user-definable) to select columns displayed.

For example, a "Debugging" profile might show status info (number of ready pods, number of restarts, etc.), logs, exec, last termination info. A "Rollout" profile might show image and a subset of status info, perhaps highlighting recent changes in some way. A "Resource" profile might show cpu, memory, disk stats.

bgrant0607: We could also think about groups to eliminate the need for columns in some cases. For example, group pods by controller (identified via controllerRef in 1.3 https://github.com/kubernetes/kubernetes/issues/2210#issuecomment-184943728) and/or labels.

bgrant0607:
Also be aware that Templates are coming:
https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/templates.md

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+Details of an individual resource instance are shown on a views which displays
+its properites and all related objects (e.g., Services that target a Pod).
+![Details page template](mockups/23-03-2016-scale-and-navigation/details-page-template.png)
+
+## Concrete pages
+This section shows how templates proposed in the "View templates" section
+can be applied to real examples.
+
+"Apps" category:
+![Apps category](mockups/23-03-2016-scale-and-navigation/apps-category.png)
```
bgrant0607: It's kind of weird that controllers have endpoints.

The unstructured Pods column is problematic IMO

pwittrock:
As of release 1.2, the kubelet exposes disk usage metrics via the new metrics api.  This would be something good to expose as well.

pwittrock:
Would it be of use to have a "Nodes" section and / or have the pod link to which node it is running on.  I am of 2 minds on this.  On the one hand, when things are working well users should be able to abstract away the concept of nodes and just know things run in the cluster.  On the other hand, if a node gets in a bad state the user is going to need to know about it and take action.

bgrant0607:
I'm still looking through this, but overall I like the direction.

pwittrock:
In the Apps overview section, should there be 1 global search bar that applies to all the app types instead of separate search bars for each?

pwittrock:
Does clicking on a replica set link to all of the associated pods?

bgrant0607:
@pwittrock see "Replication Controller details overview". That page includes the pods.

bgrant0607:
@pwittrock Re. search box: One advantage of per section is that the search can automatically narrow to that section.

bgrant0607:
@pwittrock A "Node details" page would be an appropriate cluster-admin view.

bgrant0607:
@bryk Recent changes would be determined how? Watch?

bgrant0607:
cc @mml

bgrant0607:
@janetkuo Right now, it looks like relationships are shown on details pages, such as "Replication Controller details overview". I agree it might be useful to group by service (in the case of definite selector overlap https://github.com/kubernetes/kubernetes/issues/19830) on the Apps page, and to group by controllers in the pods list, which I think should be a separate page, or maybe a toggled view, with the pods interleaved between controllers.

bgrant0607:
Is there a button or selection to view the yaml/json for every resource?

bgrant0607:
@bryk Is there additional information you'd like from the API or in annotations, or label conventions that would be useful?

bgrant0607:
I assume that "ssh" is actually exec?

docs/design/navigation-and-scale.md
bgrant0607: I'd eliminate endpoints, but we should think about where we could provide links to the app, that would hit external IPs/ports of the service, or would use the apiserver proxy.

bgrant0607:
There are some annotations that would be useful to display on details pages, such as "description" and kubernetes.io/change-cause.

For Deployment, we'll want equivalents to the `kubectl rollout` commands, such as history and rollback.

bgrant0607:
Please show termination message in pod details.

bgrant0607:
A graph of event timeseries would be awesome.

bgrant0607:
Pod details overview would be a good place to show the image of each container

bgrant0607:
the search box should accept label selector syntax, or there should be a box next to it for that

bgrant0607:
I like the increased info density, but lightly coloring alternating rows would be helpful to visually break up the display. It would also be nice to select the number of items displayed per section (like gmail inbox) and to be able to collapse sections on the Apps page

bgrant0607:
For some of the views it would be useful to either omit terminated pods, or to put them into a separate section. We do the former in kubectl, unless --show-all is specified. OTOH, for Jobs, users probably want to see the corresponding terminated pods by default.

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+characteristics (like information density) it has been a constant source of
+problems. Users and developers complained about many issues: how do you sort
+cards? how do you change their order? how do you paginate them? how do you
+compare resources against each other? how do you distinct different resources
+in a card stream? how do you do bulk operations?
+
+We have identified that in order to solve these problems the cards can evolve
+into tabularized list items which can show exactly the same information as
+cards. This solution has many advantages, some of which are: pagination,
+ordering, bulk actions or visual comparison. Following mockup outlines
```
bgrant0607: How will bulk actions work? Will each item have a checkbox, similar to gmail?

bgrant0607:
cc @jwforres

bgrant0607:
The circle-style status could be useful for controllers, to show proportion of ready/available replicas or up-to-date replicas (for Deployment):
https://github.com/kubernetes/dashboard/issues/21#issuecomment-151272404

bgrant0607:
Note that if we support sorting by column, it's useful for the sort to be stable, as a way to achieve sorting by secondary and tertiary keys.

bgrant0607:
Deployment is not shown in the mocks. Is that deliberate? I'd put it at the top of the Apps view, above Replica Sets.

docs/design/navigation-and-scale.md
bgrant0607: "Apps" doesn't correspond to anything in the system or docs.

"Controllers" would correspond to system concepts, but might not be intuitive for new users, and is somewhat broader (e.g., HorizontalPodAutoscaler).

Maybe "Workloads"?

bgrant0607:
Is there a "Services" category mock?

We should put Ingress there, also.

HorizontalPodAutoscaler doesn't fit into any of the categories, but should be referenced on the details page of the controller it targets.

As for "Configuration", that term is overloaded, and I'd break it into a few categories.

Namespace is really just a selector. It doesn't really carry any user-facing information at the moment.

ResourceQuota, LimitRange, PodSecurityPolicy, MetadataPolicy could be "Policy" or "Admin". 

Node, PersistentVolume, and cluster resources/info (e.g., componentstatuses, version) could be "Admin" or "Cluster".

PersistentVolumeClaim would be "Volume".

Secret and ConfigMap could be part of "Volume" or "Data".

docs/design/navigation-and-scale.md
bryk: > Cards are suitable for <= 6 IMO

Abstracting from the actual number, I think we agree that cards dont scale. 

>Personally, I think most people running workloads in production won't mutate them with the UI:
http://martinfowler.com/bliki/InfrastructureAsCode.html

Yes, I generally agree. That's why we're focusing on displaying data for now. However, edit functionality in the UI is still very useful for one-off operations that you dot automate (e.g., kill zombie pod, change service IP address, etc).

docs/design/navigation-and-scale.md
bryk: I say total replacement for forseeable future. I agree that it'd be nice to have both. We must be realistic though - we don't have eng capacity to implement and maintain both.

The plus is that we don't propose using raw tables. We propose using tabularized lists, which can display exactly the same information as currently cards do. So all that changes is form factor: cards get wider and get columnized.

docs/design/navigation-and-scale.md
bryk: There will still be the "app creation form" which is basically `kubectl run + expose`. But after you use the form we will show replica sets, not "mystic cards that represent not-well-defined things". In this matter we propose to be very similar to what kubectl does: have something both for novice and expert users.

docs/design/navigation-and-scale.md
bryk: "Apps" category is just an example category to illustrate this design doc. We intentionally didn't propose the categories here, as this requires broader coordination. 

Ideally we'd like to have some initial category definitions in a few weeks and something stable by 1.3. Is there somebody from UX or docs team that is driving this effort? Or should we drive this?

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+
+This section shows concepts used as templates for all views.
+
+View which does not deal with a resource or a resource list, e.g., landing
+page which shows an overview of a cluster. The page consists of, potentially
+dismissable, blocks that display data.
+![Landing page](mockups/23-03-2016-scale-and-navigation/landing-page.png)
+
+Category view which shows lists of resources. In this mockup it is
+(not defined yet) "Apps" category which was selected through the
```
bryk: @bgrant0607 
> I probably wouldn't put pods in the same view as all controllers.
We should chat about how to organize the pod view: sorting, grouping, filtering, aggregation, etc. By names, labels, controllers, services.

Our idea was to have a list of all pods accessible from somewhere. This is because there will always be people using unattached pods (e.g., for experimenting) or people looking for a specific pod.

No matter what menu category pod list lands on, we'd like to make the list very limited (e.g., show 5 pods and "see all" link) and placed at the very bottom to indicate that controllers (replica sets, jobs, etc.) are the thing to use on everyday basis. 

Does this make sense?

docs/design/navigation-and-scale.md
jwforres: checkboxes would be the typical approach, it provides a clear indicator that the row is selectable, and provides a browser-native accessible selector for keyboard+screenreader users

docs/design/navigation-and-scale.md
bryk: > We'll want logs links on pods for debugging use cases. An exec widget would be awesome, too.

Yes, there's a place for that in the actions column and/or action bar.

docs/design/navigation-and-scale.md
jwforres: Would like to stay involved in this particular part of the discussion.  We are looking at reorganizing the OpenShift UI's left nav (some early discussion here https://trello.com/c/dWIKyBWi ).  Would like to keep consistent terminology where we can.

From our users' perspective a service is just a part of their application/microservice, which is why we chose to organize our dashboard with service as the "parent" element and organized the things the service points to underneath of that.

docs/design/navigation-and-scale.md
bryk: I like most of your ideas presented here. Is this something that blocks this design? Or nice-to-have future work?

I'd say this is for future work. We're aiming for Q2 with using this UX language, so we have to be cutting features and complexity aggressively. Can I add your ideas to future work section so that this serves as a base for future development?

docs/design/navigation-and-scale.md
bryk: Yes. I'll add this to the design doc.

docs/design/navigation-and-scale.md
bryk: > I'd eliminate endpoints, but we should think about where we could provide links to the app, that would hit external IPs/ports of the service, or would use the apiserver proxy.
 

Yeah, endpoints are problematic, but, in reality, very useful. Perhaps this needs better title/wording/documentation. Perhaps this can be moved to the unstructured part of the card.

But this is a problem with the specific page, not the proposed new UX language here. I assume non blocker.

bryk: > The unstructured Pods column is problematic IMO

What's the problem? Should it be structured and split into a few columns?

docs/design/navigation-and-scale.md
jwforres: recommendation, keep all help links in one single location / helper utility in the code and reference them based on unique topic names everywhere else, because inevitably doc links will change and its much easier to update a link in one place, see https://github.com/openshift/origin/blob/master/assets/app/scripts/filters/util.js#L168  and  https://github.com/openshift/origin/blob/master/assets/app/scripts/constants.js

bryk:
@janetkuo 

>Will you add filters to tabularized resource lists? For example, filter by status, by label, ...etc. It'd be useful when, say, I want to find all failing pods with label "app=nginx, track=stable", or I want to hide all resources with 0 replicas.

Yes, there will be filtering (as proposed in the mocks). We do not propose exact semantics for it yet. Rough idea is to start with something simple, like text search over data in the table. And then iterate and have some grammar inside the filter. What do you think?

docs/design/navigation-and-scale.md
jwforres: Agree with these points, standalone pods are less important / go at the bottom but you should still see that they are there and be able to get to them.  We took the approach that things that have services pointing to them come first, things that are controllers without services come next, and standalone pods last.

docs/design/navigation-and-scale.md
```diff
@@ -0,0 +1,143 @@
...
+Pod details overview:
+![Pod details overview](mockups/23-03-2016-scale-and-navigation/pod-details-page-overview.png)
+
+Pod details events:
+![Pod details events](mockups/23-03-2016-scale-and-navigation/pod-details-page-events.png)
+
+Pod details logs:
+![Pod details logs](mockups/23-03-2016-scale-and-navigation/pod-details-page-logs.png)
+
+Pod details SSH tab:
```
jwforres: we use term.js for the in-browser terminal, we actually use this fork which is set up for Bower https://github.com/vahe/term.js    it works pretty well, main complaint is the copy/paste behavior is a little weird across different browsers

bryk:
>Some resources have hierarchical relationships. For example, a Deployment manages multiple ReplicaSets, each manages multiple Pods. How do you plan to present that relationship (parent/children) in the detail page of each resource?

>Does it look like this:

>Deployment foo:

>ReplicaSet: foo-123, foo-456, ... (the children)
Pods: foo-123-aaa, foo-456-aaa, ... (the grandchildren?)
Service: foo-service
ReplicaSet foo-123:

>Deployment: foo (the parent)
Pods: foo-123-aaa, foo-123-bbb (the children)
Service: foo-service
Pod foo-123-aaa:

>Deployment: (grandparent?)
ReplicaSet: (parent)
Service: foo-service
Service foo-service:

>Deployment:
ReplicaSet:
Pods:

We propose to show mostly direct relationships and, when it makes sense, indirect. So parents, children and siblings for sure. Grandparents and neighbors sometimes. 

For example, for replica sets show parent deployments, sibling services and child pods. Don't show containers, which is too far away from the context of replica sets. 

What do you think?

bryk:
@janetkuo 

> Do we plan to show that hierarchical relationships in the apps view, too?

No, category views, e.g., "apps/workloads/controllers" have only tabularized resource lists. The lists can, however, display unstructured data that is derived from items' children, e.g., external endpoint that you can use to hit a replica set. 

What do you think?

jwforres:
for label filtering we built a component for that which handles all of the
IN, NOT IN, and EXISTS syntax.
https://github.com/kubernetes-ui/label-selector  and includes some helper
functions for selecting/filtering resources based on the selectors on api
objects

bryk:
@pwittrock 

> Does clicking on a replica set link to all of the associated pods?

This leads to details of the replica set, which, among other information, contains associated pods.

bryk:
@janetkuo @jwforres @bgrant0607 @pwittrock Responded to 50% of the comments. Will address all tomorrow.

docs/design/navigation-and-scale.md
bgrant0607: Ok

docs/design/navigation-and-scale.md
bgrant0607: @bryk We're in active discussion about how we want to describe the system concepts to users.

@jwforres What would you do about jobs, for which there wouldn't (usually) be a service?

@pwittrock @janetkuo @johndmulhausen

docs/design/navigation-and-scale.md
bgrant0607: @bryk We don't need to add features in Q2, but how hard would it be to tune how the data is presented, such as to add or remove columns, or present data in columns rather than freeform text?

docs/design/navigation-and-scale.md
bgrant0607: I like the approach suggested by @jwforres

docs/design/navigation-and-scale.md
bgrant0607: Yes, exactly, pods should be structured and split into a few columns, to facilitate sorting.

docs/design/navigation-and-scale.md
bryk: Such things are not hard, but just require a lot of work. Whenever you do something for the first time in the project, you need to design, discuss and then implement. You know, this takes time and effort.

So, my realistic view is that all we can have for Q2 is: static structured data in static columns + freeform text with unstructured data. Columns can be sortable.

Rearranging, customization, label pivoting, profiles are things I like a lot, but this is for Q3+

bryk:
@pwittrock 
>As of release 1.2, the kubelet exposes disk usage metrics via the new metrics api. This would be something good to expose as well.

You mean this is available for nodes? If yes, I say yes, we should add it to nodes view.

bryk:
@pwittrock 
>Would it be of use to have a "Nodes" section and / or have the pod link to which node it is running on. I am of 2 minds on this. On the one hand, when things are working well users should be able to abstract away the concept of nodes and just know things run in the cluster. On the other hand, if a node gets in a bad state the user is going to need to know about it and take action.

This has been an issue for us for a long time. My best idea so far is to show node that a pod is running on on its details page. Do not show this in a list of pods.Then, on node details page show all pods it is executing, because user is looking at a node anyway.

bryk:
@pwittrock 
> In the Apps overview section, should there be 1 global search bar that applies to all the app types instead of separate search bars for each?

That's a good question. We proposed here a separate search bar for each resource type. The reason was the fact that it is easier to understand what is being filtered if you have a table and directly connected search bar. If there is one global search bar and many tables, this all may get confusing.

However, this is not an easy decision to make. I don't have extremely strong opinion on this.

cc @floreks @maciaszczykm @digitalfishpond @olekzabl @cheld for opinions

bryk:
@bgrant0607 
> Is there a button or selection to view the yaml/json for every resource?

Yes, we discussed to add it to every details page of a resource. The use case is that somebody wants to "export" the resource, do some work on it in CLI or the UI does not handle some case.
I'll update the mocks to reflect this.

bryk:
@bgrant0607 

> @bryk Is there additional information you'd like from the API or in annotations, or label conventions that would be useful?

I think that everything that is standard is useful for the UI. With that the UI can derive information from annotations/labels and preset it to users. Ideally, we'd have a repository/documentation with standard annotations/labels that we can add support for. For example, kuberntes.io/cluster-service=true. Currently this information is just floating around and we cannot be adding support for something that is not formalized.

bryk:
@bgrant0607 

> I assume that "ssh" is actually exec?

Yes

bryk:
@bgrant0607 

> Pod details overview would be a good place to show the image of each container

Yes. There's list of containers, which can have images in a column. Updated the mock.

bryk:
@bgrant0607 

> the search box should accept label selector syntax, or there should be a box next to it for that

Yes. We plan to start with simple text search and then extend it for label and field searching.

bryk:
@bgrant0607 

>I like the increased info density, but lightly coloring alternating rows would be helpful to visually break up the display. It would also be nice to select the number of items displayed per section (like gmail inbox) and to be able to collapse sections on the Apps page

This is just wireframes, so things look big. I hope we'll get dense display. 

Selecting numbers of items displayed and collapsing - I like this. But for Q3+. Will add to future work section.

bryk:
@bgrant0607 
>Deployment is not shown in the mocks. Is that deliberate? I'd put it at the top of the Apps view, above Replica Sets.

No. The example we used just does not cover everything. In actual implementation it'd be there.

bryk:
PTAL. Responded to all comments and updated mocks.

I think we're in pretty good state with this proposal. I see many feature requests that I mostly sadly defer for future, since we have tight Q2/1.3 schedule.

docs/design/navigation-and-scale.md
jwforres: We were just discussing jobs yesterday actually.  Jobs (and things like
jobs) are different and should be visually different.  In OpenShift they
would be less important than Service+Deployment because these are usually
microservices/apps, but Jobs would be more important than standalone pods.

jwforres:
Along those same lines... we exposed a YAML editor for all resources, it
provides a stopgap solution to edit anything until you can provide a rich
GUI editor for that particular resource.  FWIW we picked ACE editor but it
was a toss-up between that and codemirror.  I believe it came down to ACE
having better OOTB support for YAML syntax highlighting

jwforres:
We called our tab "Terminal", and yeah its exec.  One annoying thing is the
API not supporting SIGWINCH signal still, there is an open issue for it.
So the terminal can't be responsive.

docs/design/navigation-and-scale.md
cheld: I like the idea of customizing the columns as I am a bit worried about squeezing too much into too many columns, leaving not enough space for content.

docs/design/navigation-and-scale.md
cheld: Question: what happens if a column contains multiple values (like labels) and the users sorts on this column? Is the result predictable or understandable?

cheld:
CC @fwalker

docs/design/navigation-and-scale.md
bgrant0607: Multiple values make sense for row-oriented formats, not for columns. They should be split into multiple columns

bgrant0607:
@bryk Please file issues for the suggestions you'd like to defer to future work.

bgrant0607:
@bryk Also please post an announcement to kubernetes-dev to point the contributor community at large to this proposal.

bgrant0607:
@bryk Do you plan to update this proposal at all or are we deferring additional details to the designs of individual pages?

docs/design/navigation-and-scale.md
bgrant0607: Thinking out loud:
* I don't think "Apps"  is as applicable to DaemonSets and Jobs, and might be interpreted by some as a whole multi-tier application
* "Controllers" might not be mean anything to non-K8s users and also wouldn't cover controller-less pods. "Collections", the term used in Borg, has similar problems.
* "Containers" might confuse users since none of the subheadings would be about containers per se.

"Workloads" still seems best (relatively). Where we discuss just controllers (and not controller-less pods), I might use "Workload controllers".

A more infrastructure-y approach would be Compute, Networking, Storage.

This could be clarified with tooltips and/or help buttons.

bryk:
I'll update this proposal. Just were waiting for additional comments come it after I posted it on kubernetes-dev. 

More designs of individual pages will come later, but I'll update the designs that are here.

docs/design/navigation-and-scale.md
bryk: Thanks @jwforres for the suggestion!

bryk:
@bgrant0607 

> @bryk Recent changes would be determined how? Watch?

We're hoping to get all data from the apiserver. If apiserver does not support this, we'll most likely drop the feature.

Having a history of changes for a resource was one of top feature requests we've seen. This is an important thing for us.

bryk:
##### Addressed comments to the scale and navigation UX design doc.
* Changed toolbar and left nav: now combined cluster/namespace selector and user info are in the toolbar.
* Added "Future work" section with suggested items that will not happen in 1-2 quarters (mostly due to our staffing).
* Made contents of some propsed resource columns shorter (e.g., "Pods" column is no longer "1 running, 7 desired", but "1 / 3" and format is explained in a tooltip).
* Responded to comments.

PTAL

bryk:
Ping? Is everything resolved or we need some more discussions? 

If all is resolved, then let's merge this.

janetkuo:
It's minor but some of the mockups are still showing "Apps" instead of "Workloads" and you may want to update them too:
- docs/design/mockups/23-03-2016-scale-and-navigation/category-page-template.png
- docs/design/mockups/23-03-2016-scale-and-navigation/landing-page.png

bgrant0607:
@bryk Please notify me directly (with `@`). Otherwise, it's easy for github notifications to get lost in the literally thousands of other notifications per week.

bgrant0607:
There are some other details that could be fixed, also, such as absence of Deployment from the menus, but I'm happy to merge and iterate.

PR #627, opened 2016-04-07
Prototype: list of replication controller cards

bryk:
Just a demo to show to my prototype work. What do you think? Please clone this PR and test yourself. CC @maciaszczykm @floreks @digitalfishpond @olekzabl @cheld 

PS. ignore the ugly actionbar

selection_059

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/627)

### DIFF
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -31,6 +31,7 @@
 }
 
 .kd-chrome-content {
+  background-color: $body;
   padding-top: $toolbar-height-size-base;
 }
 
@@ -58,6 +59,7 @@ a {
 
 .kd-app-content {
   background-color: $body;
+  font-size: $body-font-size-base;
   position: relative;
 }
 
--- a/src/app/frontend/common/components/labels/labels.html
+++ b/src/app/frontend/common/components/labels/labels.html
@@ -14,9 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div flex="95" class="kd-labels" ng-repeat="(key, value) in ::labelsCtrl.labels"
+<div flex="95" class="kd-labels" ng-repeat="(key, value) in labelsCtrl.labels"
      ng-if="labelsCtrl.isVisible($index)">
-  <kd-middle-ellipsis display-string="{{::key}}: {{::value}}"></kd-middle-ellipsis>
+  <kd-middle-ellipsis display-string="{{key}}: {{value}}"></kd-middle-ellipsis>
 </div>
 <div class="kd-labels kd-labels-switch" ng-show="labelsCtrl.isMoreAvailable()"
      ng-click="labelsCtrl.switchLabelsView()">
--- a/src/app/frontend/common/components/middleellipsis/middleellipsis.html
+++ b/src/app/frontend/common/components/middleellipsis/middleellipsis.html
@@ -15,10 +15,10 @@ limitations under the License.
 -->
 
 <div class="kd-middleellipsis-container">
-  <md-tooltip ng-if="::ellipsisCtrl.shouldTruncate()">
-    {{::ellipsisCtrl.displayString}}
+  <md-tooltip ng-if="ellipsisCtrl.shouldTruncate()">
+    {{ellipsisCtrl.displayString}}
   </md-tooltip>
   <span class="kd-middleellipsis"><!-- Collapse whitespace
-    -->{{::(ellipsisCtrl.displayString | middleEllipsis:ellipsisCtrl.maxLength)}}<!-- Collapse whitespace
+    -->{{(ellipsisCtrl.displayString | middleEllipsis:ellipsisCtrl.maxLength)}}<!-- Collapse whitespace
   --></span>
 </div>
--- a/src/app/frontend/common/components/middleellipsis/middleellipsis_directive.js
+++ b/src/app/frontend/common/components/middleellipsis/middleellipsis_directive.js
@@ -53,10 +53,14 @@ export default function middleEllipsisDirective(middleEllipsisFilter) {
       let nonNullContainer = container;
       let nonNullElement = ellipsisElem;
 
-      scope.$watch(() => ctrl.displayString, (displayString) => {
-        ctrl.maxLength = computeTextLength(
-            nonNullContainer, nonNullElement, middleEllipsisFilter, displayString);
-      });
+      scope.$watch(
+          () => {
+
+            let ret = computeTextLength(
+                nonNullContainer, nonNullElement, middleEllipsisFilter, ctrl.displayString);
+            return ret;
+          },
+          (len) => { ctrl.maxLength = len; });
     },
     bindToController: {
       'displayString': '@',
--- a/src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
+++ b/src/app/frontend/replicationcontrollerlist/replicationcontrollercard.html
@@ -15,117 +15,94 @@ limitations under the License.
 -->
 
 <md-card class="kd-replicationcontroller-card">
-  <md-card-content>
-    <div layout="column">
-      <div flex="auto" layout="row" layout-align="space-between center"
-            class="kd-replicationcontroller-card-title-row">
-        <span layout="row" flex="85">
-          <md-icon class="material-icons md-warn kd-replicationcontroller-card-status-icon"
-                   ng-if="::ctrl.hasWarnings()">
-            error
-            <md-tooltip>One or more pods have errors</md-tooltip>
-          </md-icon>
-          <md-icon class="material-icons kd-replicationcontroller-card-status-icon"
-                   ng-if="::ctrl.isPending()">
-            timelapse
-            <md-tooltip>One or more pods are in pending state</md-tooltip>
-          </md-icon>
-          <a ng-href="{{::ctrl.getReplicationControllerDetailHref()}}"
-             class="kd-replicationcontroller-card-name">
-            <span class="md-title kd-replicationcontroller-card-title">
-              <kd-middle-ellipsis display-string="{{::ctrl.replicationController.name}}">
-              </kd-middle-ellipsis>
+  <md-card-content class="kd-replicationcontroller-card-content">
+    <div class="kd-resource-list-container" layout="row">
+      <div style="width: 20px; min-width: 20px; padding-left: 0" flex="none">
+        <md-icon class="material-icons md-warn kd-replicationcontroller-card-status-icon"
+                 ng-if="::ctrl.hasWarnings()">
+          error
+          <md-tooltip>One or more pods have errors</md-tooltip>
+        </md-icon>
+        <md-icon class="material-icons kd-replicationcontroller-card-status-icon"
+                 ng-if="::ctrl.isPending()">
+          timelapse
+          <md-tooltip>One or more pods are in pending state</md-tooltip>
+        </md-icon>
+      </div>
+      <div style="width: 200px; min-width: 150px;" flex="grow">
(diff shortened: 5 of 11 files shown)


cheld:
Very nice. 

We could think about taking some design concepts from inbox: in this case each section (like replica set, deamonset) would get a card and each row in the card is separated by a line. However, in our case we need headers. Slightly different. 

What about the spark lines - do we have enough space to place them? 

What happens if you click on labels in order to sort - is it considered?

The header 'action' is probably not needed.

cheld:
In your screenshot the contents of the columns image and internal endpoint are not really readable. (Of course, one can hover with the mouse and screen can potentially be made wider on a desktop)

cheld:
CC @juliawag

cheld:
CC @juliewag

bryk:
> We could think about taking some design concepts from inbox: in this case each section (like replica set, deamonset) would get a card and each row in the card is separated by a line. However, in our case we need headers. Slightly different.

This is nice suggestion. I like it. Let's keep this in mind while implementing.

bryk:
> What about the spark lines - do we have enough space to place them?

We probably need to get rid of some infor from this view. E.g., container images looks like unnecessary data here.

bryk:
> What happens if you click on labels in order to sort - is it considered?

If sorting makes no sense for a column then it is disabled, I say.

bryk:
> The header 'action' is probably not needed.

Yes.

bryk:
> In your screenshot the contents of the columns image and internal endpoint are not really readable. (Of course, one can hover with the mouse and screen can potentially be made wider on a desktop)

I have ideas about this. We can make middle-ellipsis texts expandable. I.e. click small plus to expand what we had there. Or have copyable tooltip.

Juliewag:
screenclip

Good work! Very good prototype.

It becomes clear that this list view is much easier to visually parse then cards. Cards on dashboards probably only work when the information is less textual and rather pictorial, e.g. graphs and other forms of data visualization.

I have a couple of suggestions that might be helpful. 

## 1. Show relevant information first: 

You already did it, but I want to emphasize that it is important that the controllers with errors are always shown on top. However, after I re-order the table - let's say by title - I cannot get all error-controllers on top again. Users need to order column 'status'.

## 2. Make relevant information easy for humans to parse:
### 2.1. Labels: 
visually group the labels belonging to the same tag (e.g. app) by using colors. Show colored points under 'Label'. when clicking on pink bullet point in column header, user sorts labels by group (and within group alphabetical). 
### 2.2. Status: 


Information displayed under Pods requires too much reading (parsing).
You could sort them into subcolumns as shown in the example. However, maybe it is more relevant to see how many pods there are. If there is one "pending", show a label, but users need to dive into the controller anyways to analyze the problem.

bryk:
@Juliewag Thanks for this feedback. I agree and will implement this into actual product :) I'll share new screenshot with you when ready. Thanks!

bryk:
I've got a new version of the prototype. Still in progress. Take a look and tell me what you think.

selection_061

selection_063

maciaszczykm:
I like it a lot. Few things, that I'd update:

- missing horizontal line between resource info and optional warning message in red
- too small paddings/margins on the right of the page (especially near logs and menu button)

floreks:
This looks awesome. I like the change with pods number. It's simpler and user can get more details by going to details page.

Beside @maciaszczykm comment, I'd limit display of labels to 5? maybe and add option to `show all` as it was discussed some time ago.

maciaszczykm:
> Beside what @maciaszczykm said, I'd limit display of labels to 5? maybe and add option to show all as it was discussed some time ago.

We already have implementation for that :)

floreks:
> We already have implementation for that :)
Yeah, I've forgotten about that. It's not shown on the picture because there are only 4 labels at max.

bryk:
> Yeah, I've forgotten about that. It's not shown on the picture because there are only 4 labels at max.

Yep, we can adjust this treshold here. Or make it auto-magic

digitalfishpond:
> too small paddings/margins on the right of the page (especially near logs and menu button)
@maciaszczykm Do you mean that images, and then the logs and menu buttons aren't spaced out enough? I think the space between logs and menu is ok, the main problem is the small space between images and the logs button which might be reclaimed from the space between pods and age (?). If you increase that, then it would be more clear-cut since logs and menu are invariable and will look neatly aligned anyway. Other than that, I think it looks very neat :)

floreks:
> Yep, we can adjust this treshold here. Or make it auto-magic

Auto-magic based on space that they take (vertical) could be nice. We can have 5/6/7 short labels if they fit in 3 lines but if we have only long labels and each takes whole line then it would be better to use this `show more`.

What do you think?

maciaszczykm:
> @maciaszczykm Do you mean that images, and then the logs and menu buttons aren't spaced out enough? I think the space between logs and menu is ok, the main problem is the small space between images and the logs button which might be reclaimed from the space between pods and age (?). If you increase that, then it would be more clear-cut since logs and menu are invariable and will look neatly aligned anyway. Other than that, I think it looks very neat :)

@digitalfishpond Let's just make it consistent.

digitalfishpond:
@maciaszczykm consistent is better :)

bryk:
>> Yep, we can adjust this treshold here. Or make it auto-magic

> Auto-magic based on space that they take (vertical) could be nice. We can have 5/6/7 short labels if they fit in 3 lines but if we have only long labels and each takes whole line then it would be better to use this show more.
What do you think?

I like this. We definitely cannot be taking 8 lines of text for the labels. Will fix this later in implementation.

bryk:
>> too small paddings/margins on the right of the page (especially near logs and menu button)

> @maciaszczykm Do you mean that images, and then the logs and menu buttons aren't spaced out enough? I think the space between logs and menu is ok, the main problem is the small space between images and the logs button which might be reclaimed from the space between pods and age (?). If you increase that, then it would be more clear-cut since logs and menu are invariable and will look neatly aligned anyway. Other than that, I think it looks very neat :)

Yeah, I have problems with the logs menu. Maybe we should change this into an icon?

bgrant0607:
Are we at the point where it's worth discussing the page details now? Based on comments above, I assume we are.

I'll repeat general feedback on use cases and guidelines: #21

I also made some comments on the overall proposal PR.

In kubectl, we removed (long) configuration information like labels, selectors, containers, images, endpoints from the default view. It's unreadable in this form, anyway, and isn't relevant to most use cases. Additionally, "endpoints" is confusing because there isn't a 1-to-1 correspondence between controllers and services.

Labels are more useful for sorting where configurable as columns (key as header, value in entries). We could do this by default for conventional label keys like app, tier, role, track, etc., if present. Filtering by label selector would be useful, also.

Creation time / age we put at the end in kubectl. Mostly it's useful for sorting: what did I create most recently? It could probably be omitted from the default view.

If we want all the data to fit on one screen (desirable), then we need different views for different use cases. I'd like tabs to flip between them.
* Default view: Status/debugging dashboard. Readiness, failure rate, logs, exec. I agree we should split pod status into separate columns for easier sorting/reading: desired, %/# ready. Maybe details of the most recent failure and/or most recent event on a separate row. I agree sorting by failures/warnings by default is desirable.
* Deployment status: Image/tag info, % ready, age. Replica sets should show Deployment labels/annotations (revision), but these aren't relevant for RCs. When showing images, the registry can be massively summarized. 
* Resource management: CPU, memory requests, limits, usage, recent OOMs
* Launchpad: Link to apiserver proxy for each port, perhaps informed by annotations: https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/service-discovery.md . At minimum, UDP ports are probably not useful. :-)

Speaking of annotations, we could take advantage of annotation conventions, such as description, icon, logging/monitoring URLs, etc.

I agree the actions menu is confusing. Icons SGTM.

Keeping the header fixed is a good idea.

cc @janetkuo @pwittrock

bryk:
> Are we at the point where it's worth discussing the page details now? Based on comments above, I assume we are.

That's correct


> In kubectl, we removed (long) configuration information like labels, selectors, containers, images, endpoints from the default view. It's unreadable in this form, anyway, and isn't relevant to most use cases.

I see. In a browser, however, we have more space and other ways of expression than 80-char rows. E.g., we can do label coloring or compress container images info into an interactive cell. 


> Additionally, "endpoints" is confusing because there isn't a 1-to-1 correspondence between controllers and services.

The endpoints column is by far the most useful one in a list of controllers/services. We've received lots of nice feedback about it, e.g., people said that the info in the "endpoints" column require "like a zillion kubectl commands to get". This could also be in kubectl, why not. 
Of course, there's always some corner cases. However, if we cover typical use cases and explain what exactly the column is, I think that the net gain is positive. 

> Labels are more useful for sorting where configurable as columns (key as header, value in entries). We could do this by default for conventional label keys like app, tier, role, track, etc., if present.

Yes, we will implement this someday. But don't want to be inventing the convention. We'd like to have something ecosystem-wide, so that devs, cli and UIs can base on that. We can drive or participate in discussions, of course. 


> Filtering by label selector would be useful, also.

Yep, will land there. 


> Creation time / age we put at the end in kubectl. Mostly it's useful for sorting: what did I create most recently? It could probably be omitted from the default view.

Nice idea. We will explore this. 


> If we want all the data to fit on one screen (desirable)
The UI is designed to work on all screen sizes. From 24inch to a mobile. 


> ... then we need different views for different use cases. I'd like tabs to flip between them.

Yes, definitely. Our current goal is a set of "generic" views that somewhat fit all cases. Then, to add role-based views. The base framework that we develop now is designed to work well with this, so implementing this is a matter of time and priorities.

bgrant0607:
@bryk Could you please explain the exact use cases for the endpoints? Why do they make sense on RCs rather than on services?

bgrant0607:
I like the idea of using color, in general, but stacking and abbreviating the labels makes them pretty worthless, IMO.

bgrant0607:
Speaking of using color: What makes labels long is typically the prefix (stuff before the `/`). What makes image names long is typically the registry name (again, stuff before the last `/`). What if we put a color-coded "legend" at the top that associated colors with these prefixes and then elided them in the columns? That would make the labels and images more concise and readable. We could leave leading slashes to indicate the prefixes were removed.

bgrant0607:
I'd definitely put endpoints on the services and pods on the RC details page (#694).

bgrant0607:
If users do want service addresses/ports here (rather than cluster-local pod IPs/ports, and in addition to the Services page and RC details page), I'd make the column title clear that that's what it is. Also, "Endpoints" is the term we use for the backend IPs/ports of the service. "Service Ports" would be more accurate/clear.

bryk:
> @bryk Could you please explain the exact use cases for the endpoints? Why do they make sense on RCs rather than on services?

Oh, they make sense on both pages. For services - it is obvious. For RCs, we can either link to "corresponding" services or show directly their endpoint. I like both solutions.

Use case: I want to access the application represented by a Kubernetes object (service, RC, pod, etc). To debug/probe/see it. This is typical for troubleshooting workflows.

bryk:
@bgrant0607 
> I like the idea of using color, in general, but stacking and abbreviating the labels makes them pretty worthless, IMO.

So you say we should remove lables column entirely from the default view? Is that what you did with kubectl? What was the reasoning?

bryk:
@bgrant0607 
> I'd definitely put endpoints on the services and pods on the RC details page (#694).
Yes, this will be done.

bryk:
@bgrant0607 
> If users do want service addresses/ports here (rather than cluster-local pod IPs/ports, and in addition to the Services page and RC details page), I'd make the column title clear that that's what it is. Also, "Endpoints" is the term we use for the backend IPs/ports of the service. "Service Ports" would be more accurate/clear.

I like your proposal. I'll spin this off into a separate issue and continue discussion there. Hopefully leading to better wording.

Issue #636, opened 2016-04-11
Too wide DIV with chechkbox "Expose service externall"


kwodzynski:
##### Environment

```
Dashboard version: gcr.io/google_containers/kubernetes-dashboard-amd64:v1.0.1
Kubernetes version: v1.3.0-alpha.0.604+b494cbc0c17dd7
Operating system: Ubuntu 15.10 4.2.0-34-generic
Node.js version: v0.10.40
Go version: go1.5.1
```

##### Steps to reproduce
I try deploy app by form. Accidentally I click on white field on right side under all of captions.

##### Observed result
I have checked checkbox "Expose service externall". Next click causes unchecking that box. etc.
![zrzut ekranu z 2016-04-11 13-42-03](https://cloud.githubusercontent.com/assets/18006407/14425963/93c63b4c-ffeb-11e5-8060-4cede49bb2bf.png)

##### Expected result
Someone may accidentally select these options. I think the width of this DIV should be reduced. What is your opinion?

bryk:
I think this should be confined to left pane. I.e., the width should be as in all other inputs. Do you agree?

kwodzynski:
Yes, I think it will be fine

digitalfishpond:
I will implement this today

PR #641, opened 2016-04-12
'Expose service externally' checkbox clickable area fix

digitalfishpond:
Expose service externally checkbox: restrict clickable area to label and checkbox #636 

- div added to contain checkbox to enforce desired behaviour, 

- empty kd-user-help section added to make formatting consistent with rest of page ('no padding' which otherwise would require a new css class)


bryk:
LGTM

PR #661, opened 2016-04-19
Refactor action bar and add breadcrumbs framework

floreks:
This has grown quite big. Sorry for that. I can move part that enables actionbar in the UI to different PR if needed.

## Important changes
- Refactored action bar
- Implemented simple breadcrumbs framework

### Actionbar
It is separated into 3 areas. Configurable area is defined per view and needs its own controller and template.

`[breadcrumbs (const width) | flex area | configurable (button) area ]`

### Breadcrumbs
- UI Router based
- Sets label that will be displayed instead of state name
- Sets parent state name without the need to define nested states
- Interpolates label name based on f.e. state params. This only works for the last state in chain for now to keep the implementation fairly simple.
- Limits breadcrumb chain lenght
- Supports middle ellipsis formatting for long breadcrumb labels

Example configuration:
```javascript
'kdBreadcrumbs': {
  'label': '{{$stateParams.replicationController}}',
  'parent': 'replicationcontrollers',
},
```

### Other changes
- Added actionbar to RC list page
- Added actionbar to RC details page
- Supports responsive design

### Update 1:
- Removed sidebar header on RC details page
- Style changes
- Changed buttons to icons on actionbar
- **Tests in progress...**

### Update 2:
 - Removed sidebar completely and enabled details tab
 - Configured icons and added tooltips on actionbar
 - Used angular components instead of directives
 - Added tests

RC details page:
zrzut ekranu z 2016-04-22 13-10-57
zrzut ekranu z 2016-04-22 13-11-07

RC list page:
zrzut ekranu z 2016-04-22 13-10-06

@bryk @maciaszczykm could you please review? :)

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/661)

### DIFF
--- a/src/app/frontend/_variables.scss
+++ b/src/app/frontend/_variables.scss
@@ -26,6 +26,7 @@ $subhead-font-size-base:   rem(1.6) !default;
 $body-font-size-base:      rem(1.4) !default;
 $caption-font-size-base:   rem(1.2) !default;
 $toolbar-height-size-base: rem(6.4) !default;
+$toolbar-height-size-base-sm: rem(4.8) !default;
 
 $primary: #326de6;
 $secondary: #f51200;
--- a/src/app/frontend/chrome/chrome.html
+++ b/src/app/frontend/chrome/chrome.html
@@ -26,9 +26,7 @@ limitations under the License.
   </div>
 </md-toolbar>
 
-<md-toolbar ng-if="ctrl.isActionbarVisible()" ui-view="actionbar"
-    class="kd-toolbar kd-chrome-actionbar md-hue-1">
-</md-toolbar>
+<kd-actionbar ng-if="ctrl.isActionbarVisible()"></kd-actionbar>
 
 <md-content
     ng-class="{'kd-chrome-content': true, 'kd-chrome-has-actionbar': ctrl.isActionbarVisible()}">
--- a/src/app/frontend/chrome/chrome.scss
+++ b/src/app/frontend/chrome/chrome.scss
@@ -26,16 +26,16 @@
   position: fixed !important;
 }
 
-.kd-chrome-actionbar {
-  top: $toolbar-height-size-base;
-}
-
 .kd-chrome-content {
   padding-top: $toolbar-height-size-base;
 }
 
 .kd-chrome-has-actionbar {
   padding-top: 2 * $toolbar-height-size-base;
+
+  @media only screen and (max-width: $layout-breakpoint-sm) {
+    padding-top: $toolbar-height-size-base + $toolbar-height-size-base-sm;
+  }
 }
 
 body {
--- a/src/app/frontend/chrome/chrome_module.js
+++ b/src/app/frontend/chrome/chrome_module.js
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import chromeDirective from './chrome_directive';
+import componentsModule from 'common/components/components_module';
 
 /**
  * Angular module containing navigation chrome for the application.
@@ -23,5 +24,6 @@ export default angular
         [
           'ngMaterial',
           'ui.router',
+          componentsModule.name,
         ])
     .directive('chrome', chromeDirective);
--- a/src/app/frontend/common/components/actionbar/actionbar.html
+++ b/src/app/frontend/common/components/actionbar/actionbar.html
@@ -14,8 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<kd-replication-controller-list-container-deprecated>
-  <kd-replication-controller-card-deprecated ng-repeat="replicationController in ::ctrl.replicationControllers"
-                                  replication-controller="::replicationController">
-  </kd-replication-controller-card-deprecated>
-</kd-replication-controller-list-container-deprecated>
+<md-toolbar class="kd-toolbar kd-actionbar">
+  <div class="md-toolbar-tools" layout="row">
+    <kd-breadcrumbs class="kd-actionbar-breadcrumbs" limit="2" flex></kd-breadcrumbs>
+    <span flex></span>
+    <ng-transclude ui-view="actionbar"></ng-transclude>
+  </div>
+</md-toolbar>
--- a/src/app/frontend/common/components/actionbar/actionbar.scss
+++ b/src/app/frontend/common/components/actionbar/actionbar.scss
@@ -12,25 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/**
- * Controller for the replication controller info directive.
- * @final
- */
-export default class ReplicationControllerInfoController {
-  /**
-   * Constructs replication controller info object.
-   */
-  constructor() {
-    /**
-     * Replication controller details. Initialized from the scope.
-     * @export {!backendApi.ReplicationControllerDetail}
-     */
-    this.details;
-  }
+@import '../../../variables';
 
-  /**
-   * @return {boolean}
-   * @export
-   */
-  areDesiredPodsRunning() { return this.details.podInfo.running === this.details.podInfo.desired; }
+md-toolbar {
+  &.kd-toolbar {
+    &.kd-actionbar {
+      background: $content-background;
+      color: $muted;
+      height: $toolbar-height-size-base;
+      top: $toolbar-height-size-base;
+
+      @media only screen and (min-width: 0) and (max-width: $layout-breakpoint-sm) {
+        height: $toolbar-height-size-base-sm;
+      }
+    }
+  }
+}
+
+.kd-actionbar-breadcrumbs {
+  flex-grow: 2;
+  white-space: nowrap;
 }
--- a/src/app/frontend/common/components/actionbar/actionbar_component.js
+++ b/src/app/frontend/common/components/actionbar/actionbar_component.js
@@ -13,24 +13,11 @@
 // limitations under the License.
 
 /**
- * Controller for the replication controller info directive.
- * @final
+ * Returns actionbar component.
+ *
+ * @return {!angular.Directive}
  */
-export default class ReplicationControllerInfoController {
-  /**
-   * Constructs replication controller info object.
-   */
-  constructor() {
-    /**
-     * Replication controller details. Initialized from the scope.
-     * @export {!backendApi.ReplicationControllerDetail}
-     */
-    this.details;
-  }
-
-  /**
-   * @return {boolean}
-   * @export
-   */
-  areDesiredPodsRunning() { return this.details.podInfo.running === this.details.podInfo.desired; }
-}
+export const actionbarComponent = {
+  templateUrl: 'common/components/actionbar/actionbar.html',
+  transclude: true,
+};
--- a/src/app/frontend/common/components/breadcrumbs/breadcrumb.js
+++ b/src/app/frontend/common/components/breadcrumbs/breadcrumb.js
@@ -13,24 +13,17 @@
 // limitations under the License.
 
 /**
- * Controller for the replication controller info directive.
  * @final
  */
-export default class ReplicationControllerInfoController {
+export default class Breadcrumb {
   /**
-   * Constructs replication controller info object.
+   * Constructs breadcrumb object.
    */
   constructor() {
-    /**
-     * Replication controller details. Initialized from the scope.
-     * @export {!backendApi.ReplicationControllerDetail}
-     */
-    this.details;
-  }
+    /** @export {string} - Label used instead of original state name */
+    this.label;
 
-  /**
-   * @return {boolean}
-   * @export
-   */
-  areDesiredPodsRunning() { return this.details.podInfo.running === this.details.podInfo.desired; }
+    /** @export {string} - Link used to go to the given state */
+    this.stateLink;
+  }
 }
--- a/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html
+++ b/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html
@@ -14,8 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<kd-replication-controller-list-container-deprecated>
-  <kd-replication-controller-card-deprecated ng-repeat="replicationController in ::ctrl.replicationControllers"
-                                  replication-controller="::replicationController">
-  </kd-replication-controller-card-deprecated>
-</kd-replication-controller-list-container-deprecated>
+<span ng-switch="$last" ng-repeat="breadcrumb in ::ctrl.breadcrumbs">
+  <a ng-switch-when="false" href="{{::breadcrumb.stateLink}}">
+    {{::breadcrumb.label}}
+  </a>
+  <span ng-switch-when="true" class="kd-breadcrumbs-ellipsised">
+    <kd-middle-ellipsis display-string="{{::breadcrumb.label}}"></kd-middle-ellipsis>
+  </span>
+  <span ng-switch-when="false">></span>
+</span>
--- a/src/app/frontend/common/components/breadcrumbs/breadcrumbs.scss
+++ b/src/app/frontend/common/components/breadcrumbs/breadcrumbs.scss
@@ -12,25 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/**
- * Controller for the replication controller info directive.
- * @final
- */
-export default class ReplicationControllerInfoController {
-  /**
-   * Constructs replication controller info object.
-   */
-  constructor() {
-    /**
-     * Replication controller details. Initialized from the scope.
-     * @export {!backendApi.ReplicationControllerDetail}
-     */
-    this.details;
-  }
-
-  /**
-   * @return {boolean}
-   * @export
-   */
-  areDesiredPodsRunning() { return this.details.podInfo.running === this.details.podInfo.desired; }
+.kd-breadcrumbs-ellipsised {
+  display: inline-block;
+  width: 100%;
 }
--- a/src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js
+++ b/src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js
@@ -1,0 +1,213 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Breadcrumb from './breadcrumb';
+
+/**
+ * @final
+ */
+export default class BreadcrumbsController {
+  /**
+   * Constructs breadcrumbs controller.
+   *
+   * @param {!ui.router.$state} $state
+   * @param {!angular.$interpolate} $interpolate
+   * @ngInject
+   */
+  constructor($state, $interpolate) {
+    /** @private {!ui.router.$state} */
+    this.state_ = $state;
+
+    /** @private {!angular.$interpolate} */
+    this.interpolate_ = $interpolate;
+
+    /** @export {number} - Initialized from template */
+    this.limit;
+
+    /** @export {!Array<!Breadcrumb>} - Initialized in $onInit method. Used in template */
+    this.breadcrumbs;
+  }
+
+  /**
+   * @export
+   */
+  $onInit() { this.breadcrumbs = this.initBreadcrumbs_(); }
+
+  /**
+   * Initializes breadcrumbs array by traversing states parents until none is found.
+   *
+   * @return {!Array<!Breadcrumb>}
+   * @private
+   */
+  initBreadcrumbs_() {
+    /** @type {!ui.router.$state} */
+    let state = this.state_['$current'];
+    /** @type {!Array<!Breadcrumb>} */
+    let breadcrumbs = [];
+
+    while (state && state['name'] && this.canAddBreadcrumb_(breadcrumbs)) {
+      /** @type {!Breadcrumb} */
+      let breadcrumb = this.getBreadcrumb_(state);
+
+      if (breadcrumb.label) {
+        breadcrumbs.push(breadcrumb);
+      }
+
+      state = this.getParentState_(state);
+    }
+
+    return breadcrumbs.reverse();
+  }
+
+  /**
+   * Returns true if limit is undefined or limit is defined and breadcrumbs count is smaller or
+   * equal to the limit.
+   *
+   * @param {!Array<!Breadcrumb>} breadcrumbs
+   * @return {boolean}
+   * @private
+   */
+  canAddBreadcrumb_(breadcrumbs) {
+    return this.limit === undefined || this.limit > breadcrumbs.length;
+  }
+
+  /**
+   * Returns breadcrumb config object if it is defined on state, undefined otherwise.
+   *
+   * @param {!ui.router.$state} state
+   * @return {Object}
+   * @private
+   */
+  getBreadcrumbConfig_(state) {
+    let conf = state['data'];
+
+    if (conf) {
+      conf = conf['kdBreadcrumbs'];
+    }
+
+    return conf;
+  }
+
+  /**
+   * Returns parent state of the given state based on defined state parent name or if it is not
+   * defined then based on direct parent state.
+   *
+   * @param {!ui.router.$state} state
+   * @return {!ui.router.$state}
+   * @private
+   */
+  getParentState_(state) {
+    let conf = this.getBreadcrumbConfig_(state);
+    let result = state['parent'];
+
+    if (conf && conf.parent) {
+      result = this.state_.get(conf.parent);
+    }
+
+    return result;
+  }
+
+  /**
+   * Creates breadcrumb object based on state object.
+   *
+   * @param {!ui.router.$state} state
+   * @return {!Breadcrumb}
+   * @private
+   */
+  getBreadcrumb_(state) {
+    let breadcrumb = new Breadcrumb();
+
+    breadcrumb.label = this.getDisplayName_(state);
+    breadcrumb.stateLink = this.state_.href(state['name']);
+
+    return breadcrumb;
+  }
+
+  /**
+   * Returns display name for given state.
+   *
+   * If label for the state is defined it is interpolated against given state default view scope.
+   * If interpolation fails then given label string is returned.
+   * If label string is empty then state name is returned.
+   *
+   * @param {!ui.router.$state} state
+   * @returns {string}
+   * @private
+   */
+  getDisplayName_(state) {
+    let conf = this.getBreadcrumbConfig_(state);
+    let areLocalsDefined = !!state['locals'];
+    let interpolationContext = state;
+
+    // When conf is undefined and label is undefined or empty then fallback to state name
+    if (!conf || !conf.label) {
+      return state['name'];
+    }
+
+    if (areLocalsDefined) {
+      // Set context to default view scope
+      interpolationContext = state['locals']['@'];
+    }
+
+    return this.interpolate_(conf.label)(interpolationContext).toString();
+  }
+}
+
+/**
+ * Returns breadcrumbs component. Should be used only withing actionbar component.
+ *
+ * In order to define custom label for the state add: `'kdBreadcrumbs':{'label':'myLabel'}`
+ * to the state config. This label will be used instead of default state name when displaying
+ * breadcrumbs.
+ *
+ * In order to define custom parent for the state add: `'kdBreadcrumbs`:{'parent':'myParentState'}`
+ * to the state config. Parent state will be looked up by this name if it's defined.
+ *
+ * Additionally labels can be interpolated. This applies only to the last state in the
+ * breadcrumb chain, i.e. for given state chain `StateA > StateB > StateC`, only StateC can use
+ * following convention: `'kdBreadcrumbs`:{'label':'{{$stateParams.paramX}}'}`
+ *
+ * Example state config:
+ * $stateProvider.state(stateName, {
+ *   url: '...',
+ *   data: {
+ *      'kdBreadcrumbs': {
+ *        'label': '{{$stateParams.paramX}}',
+ *        'parent': 'parentState',
+ *      },
+ *   },
+ *   views: {
+ *     '': {
+ *       controller: SomeCtrlA,
+ *       controllerAs: 'ctrl',
+ *       templateUrl: '...',
+ *     },
+ *     'actionbar': {
+ *       controller: SomeCtrlB,
+ *       controllerAs: 'ctrl',
+ *       templateUrl: '...',
+ *     },
+ *   },
+ * });
+ *
+ * @return {!angular.Component}
+ */
+export const breadcrumbsComponent = {
+  binding: {
+    'limit': '<',
+  },
+  controller: BreadcrumbsController,
+  controllerAs: 'ctrl',
+  templateUrl: 'common/components/breadcrumbs/breadcrumbs.html',
+};
--- a/src/app/frontend/common/components/components_module.js
+++ b/src/app/frontend/common/components/components_module.js
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {actionbarComponent} from './actionbar/actionbar_component';
+import {breadcrumbsComponent} from './breadcrumbs/breadcrumbs_component';
 import filtersModule from '../filters/filters_module';
 import labelsDirective from './labels/labels_directive';
 import middleEllipsisDirective from './middleellipsis/middleellipsis_directive';
+import resourceCardModule from './resourcecard/resourcecard_module';
 import sparklineDirective from './sparkline/sparkline_directive';
 import warnThresholdDirective from './warnthreshold/warnthreshold_directive';
-import resourceCardModule from './resourcecard/resourcecard_module';
 
 /**
  * Module containing common components for the application.
@@ -27,10 +29,13 @@ export default angular
         'kubernetesDashboard.common.components',
         [
           'ngMaterial',
+          'ui.router',
           filtersModule.name,
           resourceCardModule.name,
         ])
     .directive('kdLabels', labelsDirective)
     .directive('kdMiddleEllipsis', middleEllipsisDirective)
     .directive('kdSparkline', sparklineDirective)
-    .directive('kdWarnThreshold', warnThresholdDirective);
+    .directive('kdWarnThreshold', warnThresholdDirective)
+    .component('kdActionbar', actionbarComponent)
+    .component('kdBreadcrumbs', breadcrumbsComponent);
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.html
@@ -14,47 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<md-content layout="row" layout-fill flex="auto" >
-  <div class="kd-replicationcontrollerdetail-sidebar" layout="column" flex="noshrink"
-       ng-show="ctrl.isSidebarVisible()">
-    <div class="kd-replicationcontrollerdetail-sidebar-header-wrapper kd-replicationcontrollerdetail-sidebar-item"
-        flex="nogrow">
-      <div class="kd-replicationcontrollerdetail-sidebar-header" layout-align="start center"
-           layout="row">
-        <div flex="none">
-          <md-button class="md-icon-button" ui-sref="replicationcontrollersdeprecated">
-            <md-icon md-font-library="material-icons">arrow_back</md-icon>
-          </md-button>
-        </div>
-        <div flex="auto" class="kd-replicationcontrollerdetail-no-overflow">
-          <h1 class="md-title kd-replicationcontrollerdetail-app-name">
-            <kd-middle-ellipsis display-string="{{::ctrl.replicationControllerDetail.name}}">
-            </kd-middle-ellipsis>
-          </h1>
-        </div>
-      </div>
-    </div>
-    <kd-replication-controller-info details="::ctrl.replicationControllerDetail">
-    </kd-replication-controller-info>
-  </div>
-  <div flex="auto" class="kd-replicationcontrollerdetail-no-overflow">
-    <div class="kd-replicationcontrollerdetail-sidebar-header-wrapper kd-replicationcontrollerdetail-sidebar-item"
-         flex="nogrow" ng-show="!ctrl.isSidebarVisible()">
-      <div class="kd-replicationcontrollerdetail-sidebar-header" layout-align="start center"
-           layout="row">
-        <div flex="none">
-          <md-button class="md-icon-button" ui-sref="replicationcontrollersdeprecated">
-            <md-icon md-font-library="material-icons">arrow_back</md-icon>
-          </md-button>
-        </div>
-        <div flex="auto" class="kd-replicationcontrollerdetail-no-overflow">
-          <h1 class="md-title kd-replicationcontrollerdetail-app-name">
-            <kd-middle-ellipsis display-string="{{::ctrl.replicationControllerDetail.name}}">
-            </kd-middle-ellipsis>
-          </h1>
-        </div>
-      </div>
-    </div>
+<md-content layout="row" layout-fill flex="auto">
+    <div flex="auto" class="kd-replicationcontrollerdetail-no-overflow">
     <md-tabs flex="auto" md-border-bottom md-dynamic-height>
       <md-tab label="Pods">
         <md-content>
@@ -302,7 +263,7 @@ limitations under the License.
           </div>
         </md-content>
       </md-tab>
-      <md-tab label="Details" ng-if="!ctrl.isSidebarVisible()">
+      <md-tab label="Details">
         <kd-replication-controller-info details="::ctrl.replicationControllerDetail">
         </kd-replication-controller-info>
       </md-tab>
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail.scss
@@ -14,23 +14,10 @@
 
 @import '../variables';
 
-$replicationcontrollerdetails-sidebar-bg: #fafafa;
-$replicationcontrollerdetails-sidebar-width: 315px;
 
 .kd-replicationcontrollerdetail-app-name {
   color: $foreground-2;
   font-weight: $regular-font-weight;
-}
-
-.kd-replicationcontrollerdetail-sidebar {
-  background-color: $replicationcontrollerdetails-sidebar-bg;
-  max-width: $replicationcontrollerdetails-sidebar-width + 10 * $baseline-grid;
-  min-width: $replicationcontrollerdetails-sidebar-width;
-}
-
-.kd-replicationcontrollerdetail-sidebar-item {
-  padding-left: 2 * $baseline-grid;
-  padding-right: $baseline-grid;
 }
 
 .kd-replicationcontrollerdetail-table,
@@ -52,14 +39,6 @@ $replicationcontrollerdetails-sidebar-width: 315px;
 
 .kd-replicationcontrollerdetail-no-overflow {
   overflow: hidden;
-}
-
-.kd-replicationcontrollerdetail-sidebar-header {
-  height: $table-cell-height;
-}
-
-.kd-replicationcontrollerdetail-sidebar-header-wrapper {
-  border-bottom: 1px solid $border;
 }
 
 .kd-replicationcontrollerdetail-warning-icon {
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_controller.js
@@ -99,13 +99,6 @@ export default class ReplicationControllerDetailController {
   }
 
   /**
-   * Returns true if sidebar is visible, false if it is hidden.
-   * @returns {boolean}
-   * @export
-   */
-  isSidebarVisible() { return this.mdMedia('gt-md'); }
-
-  /**
    * Returns true if event is a warning.
    * @param {!backendApi.Event} event
    * @return {boolean}
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetail_stateconfig.js
@@ -13,7 +13,10 @@
 // limitations under the License.
 
 import ReplicationControllerDetailController from './replicationcontrollerdetail_controller';
+import ReplicationControllerDetailActionBarController from './replicationcontrollerdetailactionbar_controller';
 import {stateName} from './replicationcontrollerdetail_state';
+import {actionbarViewName} from 'chrome/chrome_state';
+import {stateName as replicationControllers} from 'replicationcontrollerlistdeprecated/replicationcontrollerlist_state';
 
 /**
  * Configures states for the service view.
@@ -23,15 +26,30 @@ import {stateName} from './replicationcontrollerdetail_state';
  */
 export default function stateConfig($stateProvider) {
   $stateProvider.state(stateName, {
-    controller: ReplicationControllerDetailController,
-    controllerAs: 'ctrl',
     url: '/replicationcontrollers/:namespace/:replicationController',
-    templateUrl: 'replicationcontrollerdetail/replicationcontrollerdetail.html',
     resolve: {
       'replicationControllerSpecPodsResource': getReplicationControllerSpecPodsResource,
       'replicationControllerDetailResource': getReplicationControllerDetailsResource,
       'replicationControllerDetail': resolveReplicationControllerDetails,
       'replicationControllerEvents': resolveReplicationControllerEvents,
+    },
+    data: {
+      'kdBreadcrumbs': {
+        'label': '{{$stateParams.replicationController}}',
+        'parent': replicationControllers,
+      },
+    },
+    views: {
+      '': {
+        controller: ReplicationControllerDetailController,
+        controllerAs: 'ctrl',
+        templateUrl: 'replicationcontrollerdetail/replicationcontrollerdetail.html',
+      },
+      [actionbarViewName]: {
+        controller: ReplicationControllerDetailActionBarController,
+        controllerAs: 'ctrl',
+        templateUrl: 'replicationcontrollerdetail/replicationcontrollerdetailactionbar.html',
+      },
     },
   });
 }
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html
@@ -1,0 +1,55 @@
+<!--
+Copyright 2015 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div hide show-gt-sm>
+  <md-button class="md-icon-button" ng-click="ctrl.redirectToDeployPage()">
+    <md-icon class="kd-rc-detail-actionbar-icon">add</md-icon>
+    <md-tooltip md-direction="bottom">Deploy a containerized app</md-tooltip>
+  </md-button>
+  <md-button class="md-icon-button" ng-click="ctrl.handleUpdateReplicasDialog()">
+    <md-icon class="kd-rc-detail-actionbar-icon">mode_edit</md-icon>
+    <md-tooltip md-direction="bottom">Edit number of pods</md-tooltip>
+  </md-button>
+  <md-button class="md-icon-button" ng-click="ctrl.handleDeleteReplicationControllerDialog()">
+    <md-icon class="kd-rc-detail-actionbar-icon">delete</md-icon>
+    <md-tooltip md-direction="bottom">Delete replication controller</md-tooltip>
+  </md-button>
+</div>
+
+<div hide-gt-sm>
+  <md-fab-speed-dial md-direction="left" class="md-scale" md-open="ctrl.showFabIcons"
+                     ng-mouseenter="ctrl.showIcons()" ng-mouseleave="ctrl.hideIcons()">
+    <md-fab-trigger>
+      <md-button aria-label="menu" class="md-icon-button">
+        <md-icon class="kd-rc-detail-actionbar-icon">menu</md-icon>
+      </md-button>
+    </md-fab-trigger>
+    <md-fab-actions>
+      <md-button class="md-fab md-raised md-mini" ng-click="ctrl.handleDeleteReplicationControllerDialog()">
+        <md-icon class="kd-rc-detail-actionbar-icon">delete</md-icon>
+        <md-tooltip md-direction="bottom">Delete replication controller</md-tooltip>
+      </md-button>
+      <md-button class="md-fab md-raised md-mini" ng-click="ctrl.handleUpdateReplicasDialog()">
+        <md-icon class="kd-rc-detail-actionbar-icon">mode_edit</md-icon>
+        <md-tooltip md-direction="bottom">Edit number of pods</md-tooltip>
+      </md-button>
+      <md-button class="md-fab md-raised md-mini" ng-click="ctrl.redirectToDeployPage()">
+        <md-icon class="kd-rc-detail-actionbar-icon">add</md-icon>
+        <md-tooltip md-direction="bottom">Deploy a containerized app</md-tooltip>
+      </md-button>
+    </md-fab-actions>
+  </md-fab-speed-dial>
+</div>
--- a/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.scss
+++ b/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.scss
@@ -12,25 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/**
- * Controller for the replication controller info directive.
- * @final
- */
-export default class ReplicationControllerInfoController {
-  /**
-   * Constructs replication controller info object.
-   */
-  constructor() {
-    /**
-     * Replication controller details. Initialized from the scope.
-     * @export {!backendApi.ReplicationControllerDetail}
-     */
-    this.details;
+@import '../variables';
+
+md-toolbar {
+  .md-button {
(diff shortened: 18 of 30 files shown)


bryk:
@maciaszczykm Can you do the initial review?

maciaszczykm:
Okay, so I've got few questions, that I'd like to discuss:

- Why do we hide edit and delete actions under actions button? There is a lot of place in the action bar and I think, that we shouldn't do this. Both mentioned actions are similar to create and in my opinion we should have three buttons here. If we are planning to have more specific actions, then maybe it will be good idea to group them into actions menu/button.




- As @floreks noticed by himself on mobile page there is no create button/action. It needs to be fixed.
- I don't like middle ellipsis filter in top menu, but I know, that it might be necessary in some cases. Can we at least try to avoid it as much as possible? Maybe use smaller font or make single bread crumbs longer? 



- We don't need application name on replication controller details sidebar anymore, but it will be refactored soon anyways (there are a lot more new issues on this page now).



- Styling... I don't like grey font on grey background in the action bar. On the button font color is black, maybe we should switch to black here too? Blue buttons instead of white? Different shade of blue instead of grey as action bar bacground?



I'll look into the code soon.

---

Review status: 0 of 22 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KFmcd3hQzL_kXjG8-o0:1938041201)*

floreks:
> Why do we hide edit and delete actions under actions button?

Looking at other sites it's common practice to either have similar button on the same level. Having 3 button next to each other will just result in showing hamburger menu quicker. Both options are ok for me.

> I don't like middle ellipsis filter in top menu, but I know, that it might be necessary in some cases. Can we at least try to avoid it as much as possible? Maybe use smaller font or make single bread crumbs longer?

Smaller font won't do much. Maybe const labels will be able to fit more often in available space, but still for longer breadcrumbs it will not work as expected. We can't say that some breadcrumb should have more space then the other. They are all equally divided by the available space for whole breadcrumb container. Middle ellipsis can be only applied if container per breadcrumb has width defined. I don't have better solution for that now.

> Styling... I don't like grey font on grey background in the action bar. On the button font color is black, maybe we should switch to black here too? Blue buttons instead of white? Different shade of blue instead of grey as action bar background?

Just a bit lower same style (gray font on dark gray bg) is applied and it looks good. I've tried with black font but it was too visible and distracting in my opinion. Actionbar background is same as on angular material page and as I've checked blue bg color is blending with toolbar, also as the links in breadcrumbs are blue then blue on blue won't work here.

Blue buttons may be also too distracting. They should be visible but not too much I think. Their color should be similar to the background color.

maciaszczykm:
>>Why do we hide edit and delete actions under actions button?

>Looking at other sites it's common practice to either have similar button on the same level. Having 3 button next to each other will just result in showing hamburger menu quicker. Both options are ok for me.

In my opinion all CRUD options should be at the same level and that's why I'd suggest moving delete and edit buttons near the create button. All others action can be grouped into one menu.

>I don't like middle ellipsis filter in top menu, but I know, that it might be necessary in some cases. Can we at least try to avoid it as much as possible? Maybe use smaller font or make single bread crumbs longer?

>>Smaller font won't do much. Maybe const labels will be able to fit more often in available space, but still for longer breadcrumbs it will not work as expected. We can't say that some breadcrumb should have more space then the other. They are all equally divided by the available space for whole breadcrumb container. Middle ellipsis can be only applied if container per breadcrumb has width defined. I don't have better solution for that now.

I agree, we can work on it later.

>>Styling... I don't like grey font on grey background in the action bar. On the button font color is black, maybe we should switch to black here too? Blue buttons instead of white? Different shade of blue instead of grey as action bar background?

>Just a bit lower same style (gray font on dark gray bg) is applied and it looks good. I've tried with black font but it was too visible and distracting in my opinion. Actionbar background is same as on angular material page and as I've checked blue bg color is blending with toolbar, also as the links in breadcrumbs are blue then blue on blue won't work here.

>Blue buttons may be also too distracting. They should be visible but not too much I think. Their color should be similar to the background color.

In the sidebar the contrast between font color and the background is bigger and it makes text easier to read. We can work on it later, it's just my observation.

You mentioned AngularJS webpage and they are actually using black font, because, like I said, it's easier to read.

Moreover, I think we already have too many colors in the two toolbars in the menu. I think we should make it more consistent. I can work on it later :)

floreks:
I've changed buttons to icons and adjusted color of the actionbar to match color of the sidebar. I've also removed sidebar header. What do you think about this apporach?

zrzut ekranu z 2016-04-20 12-58-12
zrzut ekranu z 2016-04-20 12-58-24

maciaszczykm:
@floreks It looks a lot better now. In my opinion we shouldn't display icons in mobile version of menu.

What about making icons blue (same blue as links or top background)?

Now I'll take a look at the code.

floreks:
I agree that icons in mobile version are not needed.

As for the blue icons. I don't think that is a good idea. You've mentioned that colors should be consistent so I think that actionbar components should be kept in shades of gray. Moreover hamburger menu is black and buttons on this menu are  white with black text. They would have to be aligned and in my opinion blue does not look good in there.

Looking at gmail etc. they have light gray actionbar with button (text or icon) but they are slightly darker then the background of the actionbar.

@bryk @cheld we need your opinion on the UI :)

maciaszczykm:
Reviewed 17 of 22 files at r1.
Review status: 17 of 22 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 25 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFnNxs-r3X_ErE7l001:-KFnNxs-r3X_ErE7l002:-429157321) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L25)):*
Is `min-width: 0` needed? Isn't it `0` by default or are you overriding Angular Material?

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 82 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFnOY1_2eWReY2va8vO:-KFnOY1_2eWReY2va8vP:178349153) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L82)):*
`this.limit !== undefined` is not needed here. 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 46 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFnOoJpYFN3newfIM4m:-KFnOoJpYFN3newfIM4n:-1645608664) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L46)):*
Missing empty line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
Review status: 17 of 22 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFn_uxreTJSUjlzB0Q3:467252740) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
>> Styling... I don't like grey font on grey background in the action bar. On the button font color is black, maybe we should switch to black here too? Blue buttons instead of white? Different shade of blue instead of grey as action bar background?

> Just a bit lower same style (gray font on dark gray bg) is applied and it looks good. I've tried with black font but it was too visible and distracting in my opinion. Actionbar background is same as on angular material page and as I've checked blue bg color is blending with toolbar, also as the links in breadcrumbs are blue then blue on blue won't work here.
Blue buttons may be also too distracting. They should be visible but not too much I think. Their color should be similar to the background color.

Can we make the color just white? All cards and places where you do things are white. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFna7EG-AiX-EKQPCP7:-1699073715) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
@maciaszczykm @floreks 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
@maciaszczykm 
> We don't need application name on replication controller details sidebar anymore, but it will be refactored soon anyways (there are a lot more new issues on this page now).

Yeah, I say we should not bother. The RC name should move to the action bar. Which will probably happen soon. 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KFnazLLPQ5fiKUfX02P:-222774847)*

maciaszczykm:
IMO we should consider making toolbar and actionbar blue:



Now it's easier to see, that they have different role than rest of the page. Colors should be aligned, but I like presented configuration.

---

Review status: 17 of 22 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KFnfsej-Cyw4Ts7Gjm4:-930692561)*

maciaszczykm:
Also actionbar could be shinked vertically and we should use icons as @floreks proposed.

---

Review status: 17 of 22 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KFngdKqIMXd-XiFSM8i:-416983311)*

floreks:
IMHO actionbar belongs to the content of the page and not to the toolbar. Colors should be consistent with the look of the page and page content is mostly in white/gray. My initial proposal also considered blue actionbar but looking at angular material and how it composes with rest of the page I think that white/gray actionbar looks a lot better.

I'm not sure about white. It also blend with the content of the page.
zrzut ekranu z 2016-04-20 14-58-36

Just a bit darker white looks better IM
zrzut ekranu z 2016-04-20 14-58-47

I agree that we could shrink it a bit.

maciaszczykm:
@bryk @cheld Which color version do you like more? Blue or white?

@floreks likes white, for me blue is better.

---



vs

zrzut ekranu z 2016-04-20 14-58-47

floreks:
Review status: 17 of 22 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrQAIXF7_yl9fs8qAD:-1635406345) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
IMO white blends with content of the page. I'd make it a bit darker.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 25 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFnNxs-r3X_ErE7l001:-KFrQckS1x_iAsjogwrw:1678284823) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L25)):*
Actually without that it assumes that min-width is equal to layout breakpoint xs and I want it to work for the whole range.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 82 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFnOY1_2eWReY2va8vO:-KFrQoSOJjkFy6f8YaN8:1836248488) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L82)):*
Yep, good catch. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrd0NlquMhfcDjNVIt:759412646) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
I vote for full white. Just checked this and `#fafafa` looks very similar, but different, thus adding noise. White just looks natural and we'll have no problems contrast and coloring fonts. This would be as the rest of application: all content is on white panes (except the toolbar). 
Also, it is better, IMO, to have less color bases, than more. Otherwise we'll finish with 50 shades of grey for backgrounds
@floreks @maciaszczykm @cheld Opinions?



---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
Review status: 15 of 24 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFremrOEWCC_uVS2gn1:362699600) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Hmm.. The middle ellipsis is very aggressive here. Can we ellipsis only when needed? 


---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFreTi4nEm92qSuhoRT:1138587054) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
Why this changed to only icon buttons? 

Can we have icon+text buttons? There's plenty of space now. E.g., `(+Create) (<pencil> Edit) (<trash> Delete)`? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

maciaszczykm:
Review status: 15 of 24 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFrfyqbXeh_0U9ElcjB:168973833) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
In my opinion it looks good. For example Gmail and many other pages use only icons in their action bars. We can add text in tooltips. What do you think about it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFrhTB67dknLfye6h5V:-908149088) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
Do we need array notation here? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 94 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhYFa7GsfA--LV_vX:-KFrhYFa7GsfA--LV_vY:-109043527) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L94)):*
Do we need array notation here? (applies to entire file) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

maciaszczykm:
Review status: 15 of 24 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrghWaNGGqFFYu0R6D:1511117972) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
Still, blue version convinces me the most. 

However, white seems to suit this page a little bit better than grey. And how about buttons? Black seems to create too much contrast.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 25 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFnNxs-r3X_ErE7l001:-KFrhSqOu_LvIMtSfsZ2:24864214) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L25)):*
Okay, understood.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFrjAagBJ3FMTAAga_q:1038074401) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
I see. The problem we have is that every page will have different buttons. Some will have for creation/deletion, some for edit and some for, e.g., rolling updates. It is different case for GMail which has the same buttons for everything. 

What I'm saying is that when we have space, we should add text, to take opportunity to educate users. Then, when space is getting limited, we can fall back to icons only, and then fold into menu with texts+icons. WHht do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 15 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFriZ-5zhzlWCcoVizw:-1108945599) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
It seems that everyone has different opinion. For white color I'd change bottom shadow of the actionbar to delicate line between actionbar and content of the page. Similar to what inbox does for specific mails.



---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFrfsrvc9yUKlZZr4Od:-961436982) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Define 'when needed'.

To make middle ellipsis work we need outer container with defined width. Whole container has now 300px and each breadcrumb container gets equal width. To apply it on demand new algorithm is needed that will calculate space needed by each breadcrumb to be fully displayed and then try to optimally distribute available space between breadcrumbs.

Very naive approach could just try to fit full breadcrumbs starting from the left and shorten the on that do not fit starting from the right.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFrh_0Dm--j7Ph0dWKQ:851336730) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
Unfortunately, yes. Compiler doesn't recognize properties on state.

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFrhwin7jcc7xupAKP1:-259525714) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
I wanted to add tooltips also. Most of the time icons are self explanatory and for some buttons that don't have good icon that will reflect its meaning we can use text.

Icon + Text doesn't look that good. IMO it's better to follow one approach. Either use icons or text.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

maciaszczykm:
Review status: 15 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrlvKIGR62d9ID9iKy:717957993) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
What, I've seen on other pages when there are two toolbars is second bar splits it from rest of the page (different color/border). Having white/grey action bar and white/grey page makes it look too similar IMO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 15 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrsG3IGs0ytkqqGGMm:408100437) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
I'd go with this look. It's similar to what inbox does. Delicate line between actionbar and page content. Same color for both.



For the smaller page when sidebar is gone it looks event better IMO.


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

maciaszczykm:
Review status: 15 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrt1qdHjjQBlezajbP:1324335320) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
This one looks better, I like it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/actionbar/actionbar.scss, line 20 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFn_uxreTJSUjlzB0Q2:-KFrimsrwpGcOEzQ1HSr:1897637675) ([raw file](https://github.com/kubernetes/dashboard/blob/1e2e90c0778e9775179f3d5d4031dd89253b8824/src/app/frontend/common/components/actionbar/actionbar.scss#L20)):*
LGTM. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
Review status: 15 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFrzOT5tMnQi2w6zzjB:-1307618089) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
> Very naive approach could just try to fit full breadcrumbs starting from the left and shorten the on that do not fit starting from the right.

Yeah, I was thinking about this, e.g.,

```html
<action-bar>
  <div class="the-place-where-breadcrubms-land" flex="auto"></div>
  <div class="the-place-for-buttons" flex="auto"></div>
</action-bar>
```

I believe it'll work and and shrink `the-place-where-breadcrubms-land` when needed. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFs-IqecSpAvz0ggP2g:1139981485) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
I see that all this in the externs: https://github.com/google/closure-compiler/blob/master/contrib/externs/angular_ui_router.js

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFs-qjmc1rlqQ0apA75:-731620218) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
Right now we have icons on large screen and text in menus on small screens. There is no correlation between them. If we had text+icon all the time or text all the time or icons all the time, I'd be consistent in all views.

But yeah, I'd like to show this to a UX expert and ask them what they think. Can you talk with @juliewag?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

maciaszczykm:
Hey @juliewag, could you take a look on discussion about styling and tell us which option seems to be the best in your opinion? As a developers we are not experts in this area :)

floreks:
Review status: 15 of 24 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFsU50ZFQs75-c29HU6:967070853) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Only way i know that would work is to calculate width using js and set style on elements or write a directive. Code now looks like this:

```
<breadcrumbs>
  <span>
    <a>ellipsis</a> - this needs const width
    <span>ellipsis</span> - this also needs const width
    <span>separator</span>
  </...>
```

Using flex or `width: 100%` similar to how it's done in table on details page forces elements to grow like in below ss:


Making them shrink to use only as much space as the content needs will result in empty text, because middle ellipsis will try to shorten text(1), then flexbox will shrink(2), then steps 1 and 2 are repated and in result we get empty string.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 18 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFs1z8i9e5gsuYYFfzr:-KFs1z8i9e5gsuYYFfzs:2100118047) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L18)):*
> Yeah, I was thinking about this, e.g.,
```
<action-bar>
  <div class="the-place-where-breadcrubms-land" flex="auto"></div>
  <div class="the-place-for-buttons" flex="auto"></div>
</action-bar>
```
I believe it'll work and and shrink the-place-where-breadcrubms-land when needed.

Now we have something like this:

```
<transclude>
  <span>
    <a>
      <middle-ellipsis>
...
```

`a` element needs defined width for middle ellipsis to even work. Making outer `span` in actionbar flex won't set width on this elements. Only way would be to calculate it in js and set style on elements. In every other place where it's used we have containers with const size (sidebar, cards

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFsaGG16RUPVrKT3icr:1952085011) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
It still says that property is not defined on object. Additionally we have custom `kdBreadcrumbs` object that extends state object. That was workaround that I've found and worked. Do you have any better idea how to solve that?

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFsb5ueh0tj_7ZEOrV_:1086877091) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
To be consistent, on mobile pages we can fallback into FAB Speed Dial: https://material.angularjs.org/latest/demo/fabSpeedDial



Icons would be shown on hover or on click.

What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFsj6-sTo1T48J21oeo:-1014309015) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Try
```html
<breadcrumbs>
  <div layout="row">
    <a flex="nogrow">Replication Controllers</a>
    <span flex="nogrow">separator</span>
    <span flex="nogrow">nginx-10</span>
    <!-- optionally: <span flex="auto"></span> -->
  </div>
</breadcrumbs>
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 15 of 24 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFskk3xLUUpQncwwDaX:73576378) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Neither works when `kd-middle-ellipsis` is used. 

```html
<div layout="row">
  <a flex="nogrow">
    <kd-middle-ellipsis display-string="Replication Controllers"></kd-middle-ellipsis>
  </a>
  <span flex="nogrow">/</span>
  <span flex="nogrow">
    <kd-middle-ellipsis display-string="mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm"></kd-middle-ellipsis>
  </span>
  <!-- optionally: <span flex="auto"></span> -->
</div>
```

In result we get:


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFssJqtW9a5f4zisZ2o:429613489) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Can you maybe just ellipsize the last element? And leave "Replication Controlllers" "Workloads", "Stuff" alone. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFsvms1RLR7KRRXl2A9:-402665355) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
I made this work, see:

```html
<md-toolbar class="kd-toolbar kd-actionbar">
  <div class="md-toolbar-tools" layout="row">
    <kd-breadcrumbs class="kd-actionbar-breadcrumbs" style="white-space:initial;" limit="2" flex="auto"></kd-breadcrumbs>
    <span flex></span>
    <ng-transclude ui-view="actionbar" flex="noshrink"></ng-transclude>
  </div>
</md-toolbar>
```

```html
<div layout="row">
  <div ng-switch="$last" ng-repeat="breadcrumb in ::ctrl.breadcrumbs" flex="auto" style="overflow: hidden;">
    <a flex="none" ng-switch-when="false" href="{{::breadcrumb.stateLink}}">
      {{::breadcrumb.label}}
    </a>
    <kd-middle-ellipsis flex="auto" ng-switch-when="true" style="overflow: hidden;" display-string="{{::breadcrumb.label}}"></kd-middle-ellipsis>
    <span flex="none" ng-switch-when="false">{{::ctrl.separator || '>'}}</span>
  </div>
</div>

```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 15 of 24 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KFsskInfM-wTPNPg5wJ:-647313470) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
I've done it a bit different but result is the same.

Limited breadcrumbs container in actionbar to not exceed 25% of the actionbar space.
```html
<kd-breadcrumbs class="kd-actionbar-breadcrumbs" limit="2" flex="25"></kd-breadcrumbs>
```

Set ellipsised breadcrumb to be able to use 100% of available width:
```html
<span ng-switch="$last" ng-repeat="breadcrumb in ::ctrl.breadcrumbs">
  <a ng-switch-when="false" href="{{::breadcrumb.stateLink}}">
    {{::breadcrumb.label}}
  </a>
  <span ng-switch-when="true" class="kd-breadcrumbs-ellipsised">
    <kd-middle-ellipsis display-string="{{::breadcrumb.label}}"></kd-middle-ellipsis>
  </span>
  <span ng-switch-when="false">{{::ctrl.separator || '>'}}</span>
</span>
```

```css
.kd-breadcrumbs-ellipsised {
  width: 100%;
  display: inline-block;
}
```

It scales nice when browser is resized now.



Still this solution will need adjustment when breadcrumbs framework will be extended.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
I've updated the code. PTAL.

bryk:
Reviewed 2 of 22 files at r1, 1 of 11 files at r4.
Review status: 13 of 26 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/app/frontend/common/components/actionbar/actionbar_directive.js, line 20 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt69sCGmOnaK8SuLUL:-KFt69sDedzSw1O5bR8V:-1827067988) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/actionbar/actionbar_directive.js#L20)):*
Hmm... Can you make this an angular 1.5 component? Like the resource cards are.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 24 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt6Mq6lNBj7ZHbyyn-:-KFt6Mq6lNBj7ZHbyyn0:862970073) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L24)):*
Do we need the ctrl.separator field? Can we just maybe hardcode `>` here? 

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFt6sX3vsB1e57p6HCX:-1414023357) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
This is this.state_.current, not this.state.$current. 

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 38 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt6UQnv_YJAle3ze95:-KFt6UQnv_YJAle3ze96:-25888275) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L38)):*
When you use angular.component, you can move initialziation to $onInit method. It is cleaner and managed by the framework, so that your controller is not littered with bussiness logic. 

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KFt7JiE2U50rfAzhQfx:-1059953262) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
Can you maybe add this to the actionbar view's `data` field? This feels like a thing only connected with the action bar, so why not to move it there? We'll avoid littering global object here and class we don't own. UI router docs say that custom data should go into `data` field. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 13 of 26 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/actionbar/actionbar_directive.js, line 20 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt69sCGmOnaK8SuLUL:-KFt7uwbwY9neIBy3ywn:1314970137) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/actionbar/actionbar_directive.js#L20)):*
Sure. I'll check that.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 24 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt6Mq6lNBj7ZHbyyn-:-KFt81yhnzsdhFu14cuJ:-1466320176) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L24)):*
We don't need it. It's option that framework would have, but in our case it can be hardcoded.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFtCT7E8kvjmZY9uYTc:1718621530) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
This is not the same actually. `state.current` returns pure config object of the state without resolved data, state params, etc.
`state.$current` is an extended state config object that contains `state.current` as `self` object.



---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 38 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt6UQnv_YJAle3ze95:-KFt8Velaqv6FAol_Mss:-252435686) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L38)):*
Great. I have to read about that a bit.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KFt8jmx1BJ_SD1cBrii:1105283627) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
When designing breadcrumbs I was actually thinking about something that can be displayed anywhere and doesn't have to be tied with other components. Like a framework. Also this was the way all other breadcrumb frameworks I've checked were written.

I'll move it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
Review status: 13 of 26 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js, line 54 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFrhTB67dknLfye6h5U:-KFwrZZvsNTSWMxHpbDe:528505022) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs_controller.js#L54)):*
OK, I see. Then use array notation. 

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFws1ERfJS7p1WZIwkt:197780377) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
One more thing. About shadow of the action bar. Currently there is no shadow and the action bar is fixed element, that other fall under. See what material spec says about this: http://www.google.com/design/spec/layout/principles.html#principles-how-paper-works

When some layer is above other, it should be shadowed. Possibly only when scrolling, like in Inbox. @maciaszczykm 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 13 of 26 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KFxG166q8ReRgnLhu5p:-1878817686) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
I've checked that and this change will need some hard logic that will be very tightly coupled with ui router conventions.

Moving to actionbar view data field will need something like that to get config:
```javascript
  getBreadcrumbsConfig(state) {
    let actionbarStateConf = state['views'][`${actionbar}@`];

    if (!actionbarStateConf) {
      actionbarStateConf = state['views'][actionbar];
    }

    return actionbarStateConf['data'] ? actionbarStateConf['data']['kdBreadcrumbs'] : undefined;
  }
```
That is because of how ui router handles naming of [nested views](https://github.com/angular-ui/ui-router/wiki/Multiple-Named-Views). One time we'll have to look under `state.views.actionbar.data.kdBreadcrumbs` and the other time under `state.views.actionbar@.data.kdBreadcrumbs`.

What do you think?

@maciaszczykm @bryk

---

*[src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html, line 19 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFreTi4nEm92qSuhoRS:-KFxFYIvgzQ1PRPUbMsq:213563715) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/replicationcontrollerdetail/replicationcontrollerdetailactionbar.html#L19)):*
Makes sense. I don't want to add too much logic now. Style change on scrolling can be done later if needed. I'll just rollback and use shadow for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

maciaszczykm:
Review status: 13 of 26 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KFxHwy3-pQhsdZqFzij:-870465760) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
This code looks error prone, if there's a way to avoid it I'd consider it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
PTAL

bryk:
*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KFxlpB948V7EFwfn76S:1030387078) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
why is the part needed?
```js
let actionbarStateConf = state['views'][`${actionbar}@`];

    if (!actionbarStateConf) {
      actionbarStateConf = state['views'][actionbar];
    }
```

In chrome.js we have such code, and it works. 
```js
  /**
   * @return {boolean}
   * @export
   */
  isActionbarVisible() {
    return !!this.state_.current && !!this.state_.current.views &&
        !!this.state_.current.views[actionbarViewName] && !this.showLoadingSpinner;
  }
```

The reason I'm worried about adding it to ui.router.$state is the fact that we dont own this class, which means anything can happen between releases. E.g., somebody may be copying state object internally, and omitting our custom fields. The documentation advises to add this info to data field:https://github.com/angular-ui/ui-router/wiki#attach-custom-data-to-state-objects
I expect that something like this should work:
```js
$state.current.views[actionbar].data['kdBreadcrumbs'];
```
If not, then why? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 8 of 28 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KFxnyMpO0UMncQ7aKk4:316099605) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
According to ui-router, when I'm getting `state.$current.views` from the current state I'm on then name of the view defined as `actionbar` will be actually `actionbar@`. Remember that I'm not using `this.state.current` but `this.state.$current` as the starting point.

```
// absolutely targets the 'status' view in root unnamed state.
        // <div ui-view='status'/> within index.html
        "status@" : { }
```

This is different for the parent state that I'm getting with `state.get(...)`.
```
        // Relatively targets the 'detail' view in this state's parent state, 'contacts'.
        // <div ui-view='detail'/> within contacts.html
        "detail" : { }, 
```

This can be skipped by getting initial config from `state.current` while working on `state.$current`, however this logic will not be easily understandable for someone who doesn't know this details about ui router. We could still add this property to data field but on the top level of state config not to actionbar view directly. It would simplify the logic a bit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
Any action item for me? I'm lost in the pool of comments.

floreks:
Sure, thanks! :)

I need answer for discussion about custom breadcrumb object in state config. See https://github.com/kubernetes/dashboard/pull/661#issuecomment-213396453
Also I'd like to know what do you think about current design. I've added screenshots here https://github.com/kubernetes/dashboard/pull/661#issue-149474283.
Last thing is this middle ellipsis formatting https://github.com/kubernetes/dashboard/pull/661#issuecomment-212921666. Is my solution ok?

It's best to download and test it.

bryk:
Review status: 8 of 32 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KGCj7jC-TqbrOLUCMV-:1108513223) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
Looks good. The only issue I have is about the 25%. Why is that? Can we make it "stretch"? So that it uses all available space? Or cap it at N pixels? The reason is that when you resize the window, the name gets middleellipsized for no reason, becasue there is empty space on the right. Looks weird. 

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KGCivkEuEQ0M7FyXk5o:1209793925) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
Ok, I see. Let's do the way you proposed then. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

floreks:
Review status: 8 of 32 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs.html, line 17 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFremrOEWCC_uVS2gn0:-KGDGTZFU_j0nBHIf7gO:-188125020) ([raw file](https://github.com/kubernetes/dashboard/blob/d62a8a22fabd9838d93990d78f5ad4503daa4ee1/src/app/frontend/common/components/breadcrumbs/breadcrumbs.html#L17)):*
`flex=100` or `auto` should stretch it. I wasn't sure if breadcrumb should use so much space.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js, line 34 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KFt7JiE2U50rfAzhQfw:-KGDHBSbCXijbh3PQEE-:-2003718903) ([raw file](https://github.com/kubernetes/dashboard/blob/5cf44f9585dceb413932ea8a4729dac3566cf4c8/src/app/frontend/common/components/breadcrumbs/breadcrumbs_directive.js#L34)):*
Great. I'll adjust the code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
Reviewed 7 of 24 files at r1, 3 of 13 files at r4, 14 of 19 files at r5, 8 of 8 files at r6.
Review status: 30 of 32 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumb.js, line 29 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KGGVVOCTOA3gkir1QLc:-KGGVVOEz1mGZut86bjt:-260864288) ([raw file](https://github.com/kubernetes/dashboard/blob/69e3fddb2d8399888bacb99305ee36c6530044de/src/app/frontend/common/components/breadcrumbs/breadcrumb.js#L29)):*
Can you document what this is? 

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js, line 38 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KGGVjvB1Fv21MPUn50n:-KGGVjvB1Fv21MPUn50o:1064060008) ([raw file](https://github.com/kubernetes/dashboard/blob/69e3fddb2d8399888bacb99305ee36c6530044de/src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js#L38)):*
Document that this is initialized from $onInit. 

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js, line 53 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KGGVpqZJyzlajoujMl5:-KGGVpqZJyzlajoujMl6:2128723576) ([raw file](https://github.com/kubernetes/dashboard/blob/69e3fddb2d8399888bacb99305ee36c6530044de/src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js#L53)):*
Please end private methods with _. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661)*

bryk:
I was going to comment to write tests, but I see that coverage is 100%. Thanks! This is awesome :)

Just a few more unresolved comments and let's merge. 

---

Review status: 30 of 32 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KGGW_WnSpDOAiFN9A6T:-1427955211)*

floreks:
PTAL

---

Review status: 23 of 32 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumb.js, line 29 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KGGVVOCTOA3gkir1QLc:-KGGwO4JE5Hbhv2eLIq4:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/69e3fddb2d8399888bacb99305ee36c6530044de/src/app/frontend/common/components/breadcrumbs/breadcrumb.js#L29)):*
Done.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js, line 38 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KGGVjvB1Fv21MPUn50n:-KGGwQxNrZQWIQO7agMj:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/69e3fddb2d8399888bacb99305ee36c6530044de/src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js#L38)):*
Done.

---

*[src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js, line 53 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-KGGVpqZJyzlajoujMl5:-KGGwRd750Be51XNjrjP:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/69e3fddb2d8399888bacb99305ee36c6530044de/src/app/frontend/common/components/breadcrumbs/breadcrumbs_component.js#L53)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KGGwVDMVtVh41M9Vmoq:676965300)*

floreks:
Yep. I'm working on it. Many conflicts :)

floreks:
huh... done. Should be ok. PTAL

bryk:
:lgtm:

---

Reviewed 15 of 18 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KGHo-0WE9BsWgt5JZ7S:1417088721)*

bryk:
Ok, so let's merge this.

I still have a couple of comments about colors, fonts and buttons. But this is a thing we can resolve later with help from our UX friends. Code looks very nice, so let's merge and iterate. Okay? 

@floreks Please merge.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/661#-:-KGHoC7paYJloYePsDzF:-1547268668)*

PR #663, opened 2016-04-20
implemented validation for expose service

cheld:
Not ready to be merged. I just like to here an opinion.

Basically, the validation works. Error message is shown, however, the elements below move down a few pixels when the error message appears. Not sure how to fix that.

I added a watch to run validation when port-mapping is changed. We have now 2 cyclic watches from port-mapping --> expose service and expose-service --> port-mapping. 

We could think about refactoring and move the expose service into the port mapping directive. In this case we could couple the logic stronger and remove the watches. Not sure if we want this.

@maciaszczykm please have a review

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/663)

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings.html
+++ b/src/app/frontend/deploy/deployfromsettings.html
@@ -114,12 +114,15 @@ limitations under the License.
 </kd-help-section>
 
 <kd-help-section>
-  <div class="md-block">
-    <md-checkbox ng-model="ctrl.isExternal" class="md-primary"
-                 ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }">
-      Expose service externally
-    </md-checkbox>
-  </div>
+  <md-input-container class="md-block">
+      <md-checkbox ng-model="ctrl.isExternal" class="md-primary md-align-top-left" name="exposeService"
+                   ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }"
+                   ng-change="ctrl.validateExposeService()">
+        Expose service externally<br/>
+      <ng-messages for="ctrl.form.exposeService.$error" role="alert" multiple  >
+        <ng-message when="portMappingRequired">Expose service externally cannot be selected without a port mapping.</ng-message>
+      </ng-messages>
+  </md-input-container>
   <kd-user-help>
   </kd-user-help>
 </kd-help-section>
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -173,6 +173,8 @@ export default class DeployFromSettingsController {
 
     /** @private {!md.$dialog} */
     this.mdDialog_ = $mdDialog;
+
+    $scope.$watch('ctrl.portMappings', () => { this.validateExposeService(); }, true);
   }
 
   /**
@@ -368,4 +370,13 @@ export default class DeployFromSettingsController {
    * @export
    */
   switchMoreOptions() { this.detail.showMoreOptions_ = !this.detail.showMoreOptions_; }
+
+  validateExposeService() {
+    let noPortMapping =
+        (this.portMappings !== undefined &&
+         this.portMappings.filter(this.isPortMappingFilled_).length === 0);
+    let isValid = !(this.isExternal && noPortMapping);
+
+    this.form['exposeService'].$setValidity('portMappingRequired', isValid);
+  }
 }


maciaszczykm:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/frontend/deploy/deployfromsettings_controller.js, line 177 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/663#-KFnMLTk6V8VLQNSXrpc:-KFnMLTk6V8VLQNSXrpd:-1975979800) ([raw file](https://github.com/kubernetes/dashboard/blob/2880ca0ca1a1b78fc3be80649b208f3a14bcb64b/src/app/frontend/deploy/deployfromsettings_controller.js#L177)):*
Do we really need `$scope.$watch`? I think there is a possibility to do it from `portmapping_controller.js`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/663)*

cheld:
*[src/app/frontend/deploy/deployfromsettings_controller.js, line 177 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/663#-KFnMLTk6V8VLQNSXrpc:-KFr_bci7SnBUSAevHjS:-748877) ([raw file](https://github.com/kubernetes/dashboard/blob/2880ca0ca1a1b78fc3be80649b208f3a14bcb64b/src/app/frontend/deploy/deployfromsettings_controller.js#L177)):*
Currently, port-mapping and expose-service rely on each other (circular). But they are implemented loosely coupled. So, a watch is needed. From my understanding to remove the watch would require to tightly couple the controller logic of both fields. This would cause a refactoring. Maybe we chat about it...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/663)*

maciaszczykm:
Proposed another solution. PTAL

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/deploy/deployfromsettings_controller.js, line 40 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/663#-KFsC_nH-2fUfoY3Efa7:-KFsC_nH-2fUfoY3Efa8:-964916984) ([raw file](https://github.com/kubernetes/dashboard/blob/2880ca0ca1a1b78fc3be80649b208f3a14bcb64b/src/app/frontend/deploy/deployfromsettings_controller.js#L40)):*
Missing `$scope` injection.

---

*[src/app/frontend/deploy/deployfromsettings_controller.js, line 177 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/663#-KFnMLTk6V8VLQNSXrpc:-KFsAbROsxE0tIN5_KnQ:1950219603) ([raw file](https://github.com/kubernetes/dashboard/blob/2880ca0ca1a1b78fc3be80649b208f3a14bcb64b/src/app/frontend/deploy/deployfromsettings_controller.js#L177)):*
Okay, I took a closer look on it and in my opinion there should be no problem at all to avoid using `$scope.$watch`. In `checkPortMapping()` function from `PortMappingsController` you can do your validation (from `validateExposeService()`) as you have an access to `portMappings` and `isExternal` variables (both are bounded).

To avoid repeating code `validateExposeService()` can be moved to service for example.

Does it make sense to you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/663#-:-KFsCf8LE3EOXxWhUlAP:1765135244)*

bryk:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/deploy/deployfromsettings_controller.js, line 177 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/663#-KFnMLTk6V8VLQNSXrpc:-KFwxIepbSmYlsYfHWXG:1605839730) ([raw file](https://github.com/kubernetes/dashboard/blob/2880ca0ca1a1b78fc3be80649b208f3a14bcb64b/src/app/frontend/deploy/deployfromsettings_controller.js#L177)):*
Just chiming in: you can create a validator directive for this. Should be easier than $watching and stuff. There are examples in our code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/663)*

cheld:
BTW: I forgot to mention. One watch already existed: the watch from port-mapping to expose service. I added the other direction

I discussed with Marcin. The conclusion is to move the expose service checkbox into the directive for the port mapping.

Advantage: we have one controller that can handle the presentation logic for both fields. This is good because they are tightly coupled. I can remove the watches than.

Disadvantage: Port-mapping and checkbox is one UI-component then and will get only one help text together (as it is now anyway). Small layout change might happen.

I will create new PR with refactored version

PR #680, opened 2016-04-26
WIP: Expose external

cheld:
Once again, not intended to be merged. I would like to get opinion. It kind of works, but it is just hacked. Code would need clean up and tests.

I refactored the code, but I am not sure, is it good or for bad?

Positive: moving the expose service checkbox into the directive. 
Negative: removing the watch requires to add direct function calls at any place the model might be changed. There would be still one watch to be removed and the code would get even more complicated, I assume.

Somehow, I feel we are doing something wrong. 

IMHO, the best option is to keep the checkbox in the directive but add a watch again. Or we think about handling the validation differently?

@maciaszczykm  can you please have a look and tell me your opinion?

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/680)

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings.html
+++ b/src/app/frontend/deploy/deployfromsettings.html
@@ -113,19 +113,6 @@ limitations under the License.
   </kd-user-help>
 </kd-help-section>
 
-<kd-help-section>
-  <md-input-container class="md-block">
-      <md-checkbox ng-model="ctrl.isExternal" class="md-primary md-align-top-left" name="exposeService"
-                   ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }"
-                   ng-change="ctrl.validateExposeService()">
-        Expose service externally<br/>
-      <ng-messages for="ctrl.form.exposeService.$error" role="alert" multiple  >
-        <ng-message when="portMappingRequired">Expose service externally cannot be selected without a port mapping.</ng-message>
-      </ng-messages>
-  </md-input-container>
-  <kd-user-help>
-  </kd-user-help>
-</kd-help-section>
 
 <!-- advanced options -->
 <div ng-show="ctrl.isMoreOptionsEnabled()">
--- a/src/app/frontend/deploy/deployfromsettings.html
+++ b/src/app/frontend/deploy/deployfromsettings.html
@@ -114,12 +114,15 @@ limitations under the License.
 </kd-help-section>
 
 <kd-help-section>
-  <div class="md-block">
-    <md-checkbox ng-model="ctrl.isExternal" class="md-primary"
-                 ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }">
-      Expose service externally
-    </md-checkbox>
-  </div>
+  <md-input-container class="md-block">
+      <md-checkbox ng-model="ctrl.isExternal" class="md-primary md-align-top-left" name="exposeService"
+                   ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }"
+                   ng-change="ctrl.validateExposeService()">
+        Expose service externally<br/>
+      <ng-messages for="ctrl.form.exposeService.$error" role="alert" multiple  >
+        <ng-message when="portMappingRequired">Expose service externally cannot be selected without a port mapping.</ng-message>
+      </ng-messages>
+  </md-input-container>
   <kd-user-help>
   </kd-user-help>
 </kd-help-section>
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -173,6 +173,8 @@ export default class DeployFromSettingsController {
 
     /** @private {!md.$dialog} */
     this.mdDialog_ = $mdDialog;
+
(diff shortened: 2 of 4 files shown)


maciaszczykm:
I'm against returning to usage of watch. We can do it like it's done in this pull request or like I suggested in #663.

There are few articles on the web saying, that watches aren't good solution:
- http://www.benlesh.com/2013/10/title.html
- https://www.accelebrate.com/blog/effective-strategies-avoiding-watches-angularjs/
- http://csharperimage.jeremylikness.com/2014/11/the-top-5-mistakes-angularjs-developers_28.html
As you can see main disadvantage is performance downgrade, but it's not the only one.

@bryk @floreks Could you share your opinion on this topic?

bryk:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/deploy/deployfromsettings.html, line 119 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/680#-KGHrp_7oyXU776vAbj2:-KGHrp_7oyXU776vAbj3:2068333911) ([raw file](https://github.com/kubernetes/dashboard/blob/8eaf1909f61a955b49b1af708428664fff26bc30/src/app/frontend/deploy/deployfromsettings.html#L119)):*
Here you can add a custom validator:
```html
<md-checkbox ng-model="ctrl.isExternal" class="md-primary"
                 ng-model-options="{ debounce: { 'default': 500, 'blur': 0 } }"
                 kd-exposed-with-port-mappings="ctrl.portMappings">
```
And have your validator

```js
angular.module('foo').directive('kd-exposed-with-port-mappings', function() {
  return {
    restrict: 'E',
    require: 'ngModel',
    scope: {'kd-exposed-with-port-mappings': '='},
    link: function(scope, elem, attrs, ngModelController) {
       validate(scope.kdExposedWithPortMappings, ngModelController);
    }
  }
});
```

@maciaszczykm @cheld WDYT? 

---

*[src/app/frontend/deploy/portmappings_controller.js, line 116 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/680#-KGHrFVpi9GSR9uVoGC1:-KGHrFVpi9GSR9uVoGC2:1071723521) ([raw file](https://github.com/kubernetes/dashboard/blob/bde4489936d2fa557b2eb0b11712ac832fffae5f/src/app/frontend/deploy/portmappings_controller.js#L116)):*
`$$` properties are angular _private_ things. You should not be using them as they may change at any time. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/680)*

floreks:
How about refactoring it and using components? For example:

Make 3 components: `ServiceConfiguration` `PortMappings` `ExternalService`.
`PortMappings` and `ExternalService` would require `ServiceConfiguration`. There would be only 1 form initialized in `ServiceConfiguration` component ctrl.

```html
<ng-form name="ctrl.serviceConfigForm">
```

Dependencies can be a transclude slots like on resource cards design and both will have access to the form via parent controller `ServiceConfiguration`. Parent ctrl can expose methods needed by `PortMappings` and `ExternalService` for validation.

Child controllers in parent can be initialized this way: https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/resourcecard/resourcecardheadercolumns_component.js#L38

Does that make sense?

bryk:
> How about refactoring it and using components? For example:

Yeah, this also looks nice.

cheld:
@bryk I have not tried your proposal, but I guess it is not do what we want to achieve (as of now). The code snippet will validate only the checkbox when it is changed. It will not validate the checkbox when a port mapping is added or removed - this is the fundamental problem.

What about changing the UI:
 1. we create a header "Service" like for labels
 2. we create a combo "Type" with the values "Internal" & "External" (no validation)
 3. Port mapping below (UDP option is only available when internal is selected)

bryk:
> It will not validate the checkbox when a port mapping is added or removed - this is the fundamental problem.

It can validate on anything - on port mapping change in particular. You just add $watch, e.g., $scope.$watch('portMappings', revalidateFunc);

bryk:
> What about changing the UI:
we create a header "Service" like for labels
we create a combo "Type" with the values "Internal" & "External" (no validation)
Port mapping below (UDP option is only available when internal is selected)

I like this. How about we have a third option for "no service". This would cover the awkward case when you leave port mappings empty and don't know what will happen.

cheld:
@bryk hmm..what you suggest is fundamentally the same I had implemented in 

https://github.com/kubernetes/dashboard/pull/663/files

I also used watch:
https://github.com/kubernetes/dashboard/pull/663/files#diff-9b87a8be20482d83bff42a18e876be7bR177

It is just using $validation instead of ng-change. Or am I wrong?

bryk:
> I also used watch:
https://github.com/kubernetes/dashboard/pull/663/files#diff-9b87a8be20482d83bff42a18e876be7bR177
It is just using $validation instead of ng-change. Or am I wrong?

The $watch is OKay there, but you could also do this with a custom validator. Custom validators are generally considered a better practice, since you separate business from validation and the input-output flow is easy.

cheld:
Ok, thanks for the comments. @floreks so we agreed to redesign the UI, right?

PR #697, opened 2016-05-02
WIP: Redesigned UI for correct validation when service must be created

cheld:
I redesigned the UI as discussed in:

https://github.com/kubernetes/dashboard/pull/680

Code is quite easy ( No watch was used). Test and clean up is missing. I will do it if you like the proposal.

@floreks I am interested if you like the usability and the look and feel. 

First option with header to indicate that both fields belong together. Similar to labels.
expose-externally-redesign

Second option without header (I like better):
expose-external3

(UDP protocol check is not fully wired up)

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/697)

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings.html
+++ b/src/app/frontend/deploy/deployfromsettings.html
@@ -98,11 +98,11 @@ limitations under the License.
 </kd-help-section>
 
 <kd-help-section>
-  <kd-port-mappings port-mappings="ctrl.portMappings" protocols="ctrl.protocols"
-      is-external="ctrl.isExternal">
-  </kd-port-mappings>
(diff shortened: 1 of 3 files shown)


digitalfishpond:
I think that the first option conveys more user-useful information, and, as you say, it's consistent with Labels.

floreks:
I agree with @digitalfishpond. Adding header shows instantly that these fields are somehow related. Now with combobox and additional validation it's more important to show this relation.

I'll checkout the branch and test also.

floreks:
For the usability point of view I think it's fine. First port is always required when service is created and additional port mappings require both ports to be specified.

cheld:
Ok thanks for the comments. I finish it up now

PR #707, opened 2016-05-04
Workloads page with RCs and Replica Sets

bryk:
This is initial version of this view. Will iterate.

Screenshot: 
selection_077

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/707)


bryk:
@digitalfishpond Can you review?

digitalfishpond:
I think it looks really neat :) 

Couple of initial comments:

- Would it be a good idea to have tooltips for the logs and menu icons confirming what they are?

- The logs pop-up drops off the edge of the window when the name is long and it is the first time that has been clicked (subsequent clicks make the whole pop-up appear correctly):

logs_pop_up

- reducing the screen size to small enough for the horizontal scroll-bar to appear causes a box to appear on the far right hand side like this: 

right_side_box

bryk:
> reducing the screen size to small enough for the horizontal scroll-bar to appear causes a box to appear on the far right hand side like this:

I'll do this the following way. The table has horizontal (Y axis) scroll that you can use on mobiles. It is a typical solution to have tables that are still tables (sortable, editable, etc) and work on mobile. 
selection_078

bryk:
> The logs pop-up drops off the edge of the window when the name is long and it is the first time that has been clicked (subsequent clicks make the whole pop-up appear correctly):

Yeah, this is a problem with the logs menu. Can we solve it in another pull request? The menu is likely to be used in all workloads pages, so it'll be easy to fix as long as we use the common component.

bryk:
All addressed. PTAL

digitalfishpond:
> Would it be a good idea to have tooltips for the logs and menu icons confirming what they are?

digitalfishpond:
:lgtm: This looks great as far as I can see :)

---

Reviewed 15 of 15 files at r1, 4 of 5 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/707#-:-KH-nQlHdgzeHgvwDjdG:-1488498164)*

PR #709, opened 2016-05-05
Fix build pipeline where CSS and JS vendor code was not processed

bryk:
Now it is renamed and minified. JS side dropped from 2.5MB to 500KB.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/709)


bryk:
@floreks Can you review?

floreks:
I see we've fixed same issue :) Both solutions are fine for me. 

`gulp-filter` needs more code to get the same result.
`gulp-if` allows to execute only 1 task on filtered files.

I don't have strong opinion which one is better here.

bryk:
This has fewer lines, so I'd go with it. It is a pity that we started working on it at the same time ;)

floreks:
That's fine. We can go with that. `gulp-filter` would be better for more complex build pipeline if we would run multiple tasks on filtered content.

floreks:
>  It is a pity that we started working on it at the same time ;)

Yeah :) On the other side it's good that it's small and took only few minutes to fix, not hours.

PR #710, opened 2016-05-05
Load apiserver config the same way kubectl does it

bryk:
Now it parses kubeconfig files when present

Try running `./dist/amd64/dashboard`. It should get info from your
kubectl file and use it.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/710)

### DIFF
--- a/src/app/backend/client/apiserverclient.go
+++ b/src/app/backend/client/apiserverclient.go
@@ -47,7 +47,7 @@ func CreateApiserverClient(apiserverHost string) (*client.Client, clientcmd.Clie
 	}
 
 	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-		&clientcmd.ClientConfigLoadingRules{}, overrides)
+		clientcmd.NewDefaultClientConfigLoadingRules(), overrides)
 
 	cfg, err := clientConfig.ClientConfig()
 	cfg.QPS = defaultQPS


bryk:
@cheld Can you review?

cheld:
Ah sure. But I am not in office tomorrow.

cheld:
@bryk what is the use case for this change? 

I assume it is development and you work in two environments local and GCE.

You could have also run "kubectl --proxy", but this is even more convenient.

Should we drop the apiServerHost variable and document the new behavior? So, by default we use kubeconfig and if it is not correct you can override with variable.

bryk:
> @bryk what is the use case for this change?

You can mount kubeconfig to a dashboard pod to connect it to any cluster. Currently it is not possible.

bryk:
> Should we drop the apiServerHost variable and document the new behavior? So, by default we use kubeconfig and if it is not correct you can override with variable.

We could do this, but I have nothing against keeping the apiserver-host param. We need to document new behavior, though.

cheld:
So, you want to run Dashboard as a pod inside a cluster, but connect to a API server of a different cluster?

bryk:
> So, you want to run Dashboard as a pod inside a cluster, but connect to an API server of a different cluster?

Not necessarily. Some people have different auth scheme than serviceaccounts. So allowing them to get config from kubeconfig is a door for all cluster setups that we currently don't support.

cheld:
Ok, now I understand :)

LGTM

PR #711, opened 2016-05-05
Revert gulp-filter to version ~3.0.0

### LINKED ISSUE #681: vendor.js is not minified in prod
vendor.js file is not minified on prod. It weights 2.3 MiB now. That's a lot. 

It was being minified in the past, so I expect this is a recent regression in our build pipeline.

floreks:
Closes #681 

It appears that in gulp-filter 4.0.0 there are some issues with resolving paths.

https://github.com/sindresorhus/gulp-filter/issues/62
https://github.com/sindresorhus/gulp-filter/issues/67

There is a PR open that should fix it so we can keep an eye on that and check again once it gets merged.

https://github.com/sindresorhus/gulp-filter/pull/61

For now I've changed version of gulp-filter back to v3 and that fixed filtering issue.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/711)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "gulp-codecov.io": "~1.0.1",
     "gulp-concat": "~2.6.0",
     "gulp-eslint": "~2.0.0",
-    "gulp-filter": "~4.0.0",
+    "gulp-filter": "~3.0.0",
     "gulp-flatten": "~0.2.0",
     "gulp-htmlmin": "~2.0.0",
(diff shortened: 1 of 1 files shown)


bryk:
Oh, the PR https://github.com/kubernetes/dashboard/pull/709 fixes the same problem. Which one gets in?

bryk:
Maybe the gulpIf solution, since it is shorter and is what gulp-useref recommends in their tutorial.

Issue #881, opened 2016-06-11
nil pointer dereference error in deployment details page


ravilr:
#### Issue details
clicking on a deployment resource with spec.strategy.type = Recreate (other than RollingUpdate) leads to a nil pointer dereference:
[restful] recover from panic situation: - runtime error: invalid memory address or nil pointer dereference
    /usr/local/go/src/runtime/asm_amd64.s:472
    /usr/local/go/src/runtime/panic.go:426
    /usr/local/go/src/runtime/panic.go:45
    /usr/local/go/src/runtime/sigpanic_unix.go:24
    /dashboard/.tmp/backend/src/github.com/kubernetes/dashboard/resource/deployment/deploymentdetail.go:150
    /dashboard/.tmp/backend/src/github.com/kubernetes/dashboard/resource/deployment/deploymentdetail.go:123
    /dashboard/.tmp/backend/src/github.com/kubernetes/dashboard/handler/apihandler.go:565
    /dashboard/.tmp/backend/src/github.com/kubernetes/dashboard/handler/apihandler.go:217
    /dashboard/Godeps/_workspace/src/github.com/emicklei/go-restful/container.go:244
    /dashboard/Godeps/_workspace/src/github.com/emicklei/go-restful/filter.go:21
    /dashboard/.tmp/backend/src/github.com/kubernetes/dashboard/handler/apihandler.go:67
    /dashboard/Godeps/_workspace/src/github.com/emicklei/go-restful/filter.go:19
    /dashboard/Godeps/_workspace/src/github.com/emicklei/go-restful/container.go:246
    /dashboard/Godeps/_workspace/src/github.com/emicklei/go-restful/container.go:95
    /usr/local/go/src/net/http/server.go:1618
    /usr/local/go/src/net/http/server.go:1910
    /dashboard/Godeps/_workspace/src/github.com/emicklei/go-restful/container.go:264
    <autogenerated>:32
    /usr/local/go/src/net/http/server.go:1910
    /usr/local/go/src/net/http/server.go:2081
    /usr/local/go/src/net/http/server.go:1472
    /usr/local/go/src/runtime/asm_amd64.s:1998

##### Environment

```
Dashboard version:  gcr.io/google_containers/kubernetes-dashboard-amd64:v1.1.0-beta3
Kubernetes version:  v1.2.4
Operating system: RHEL 7
Node.js version:
Go version: 1.6
```

bryk:
Thanks for reporting this! Will be fixed soon.

PR #886, opened 2016-06-13
Fix a bug where recreate deployments would fail to display

bryk:
Fixes: #881

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/886)


bryk:
@floreks Can you review?

floreks:
Maybe our deployment structure should be similar to what k8s offers?

https://github.com/kubernetes/dashboard/blob/master/Godeps/_workspace/src/k8s.io/kubernetes/pkg/apis/extensions/types.go#L154

It has `DeploymentStrategy` which can be either `Recreate` or `RollingUpdate`. It's more complex than ours and right now we don't support `Recreate` type. What do you think?

bryk:
Yes. AFAIU it is currently very similar to what core deployments have. Recreate deployments do not have additional info. Am I correct?

floreks:
Yes, `Recreate` is just strategy type without additional info. I saw some data that is not taken but it is also not displayed using `kubectl describe` (Paused, RollbackConfig). For now it can stay as it is. I'm not aware how important this additional information are.

floreks:
:lgtm:

<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> Yes, `Recreate` is just strategy type without additional info. I saw some data that is not taken but it is also not displayed using `kubectl describe` (Paused, RollbackConfig). For now it can stay as it is. I'm not aware how important this additional information are.
</details>


---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/886#-:-KK85DrkbVvYyI9J97eb:1417088721)*

PR #892, opened 2016-06-13
Redesign namespace select to make it more usable

bryk:
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/892)


bryk:
@floreks Can you review?

i18n/messages-en.xtb
```diff
@@ -372,4 +372,14 @@
...
 	<translation id="1739547616641575970" key="MSG_PET_SET_LIST_CREATED_AT_TOOLTIP" source="/home/mirchev/go_workspace/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="Tooltip 'Created at [some date]' showing the exact creation time of pet set.">Created at <ph name="CREATION_DATE" /></translation>
 	<translation id="1719241183992408191" key="MSG_PET_SET_LIST_PET_SET_LABEL" source="/home/mirchev/go_workspace/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="Label 'Pet Set' which will appear in the pet set delete dialog opened from a pet set card on the list page.">Pet Set</translation>
-</translationbundle>
+	<translation id="7594899852613234459" key="MSG_NAMESPACE_NOT_SELECTED" source="/usr/local/google/home/bryk/src/github.com/dashboard/.tmp/serve/app-dev.js" desc="Text for dropdown item that indicates that no namespace was selected">all namespaces</translation>
```
marians: I suggest `All namespaces` or `All Namespaces` depending on the general guideline uses (sentence case vs. title case).

floreks:
Code :lgtm:, however I'm still not sure about this `All namespaces` selector which is actually `all but kube-system`. It should be easy to notice that.

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/892#-:-KKCT52tCl9BB4EIqoDv:-684987040)*

marians:
@floreks Agree. The selector behaves weird in that it excludes `kube-system`. Is there useful nomenclature to differentiate between types of namespaces, e.g. "All user namespaces" and "System namespace (kube-system)"?

bryk:
> @floreks Agree. The selector behaves weird in that it excludes kube-system. Is there useful nomenclature to differentiate between types of namespaces, e.g. "All user namespaces" and "System namespace (kube-system)"?

kube-system is a namespace that you shouldn't touch. Unless you are a developer of a system tool. So for typical users, they open Dashboard UI for the first time and they see stuff running in kube-system namespace. That's why we filter it out. Does this make sense?

The other option is to start with `default` namespace and allow users to select a namespace or all namespaces. What do you think about this option?

floreks:
It does make sense not to show user kube-system by default, but most users will understand `All namespaces` as all. `All user namespaces` seems better here as it suggests that i.e. system namespaces will not be shown.

bryk:
> It does make sense not to show user kube-system by default, but most users will understand All namespaces as all. All user namespaces seems better here as it suggests that i.e. system namespaces will not be shown.

Yes, sounds a lot better. How about I change this to "All user namespaces" and add a documentation link somewhere in the dropdown (if possible)?

floreks:
Perfect. Some `?` icon would be good if possible.

bryk:
All right. I'll do this tomorrow or on Monday.

bryk:
So I couldn't easily add the questionmark help button to the select. I'll do this next week anyway as there is still work on the selector (do something on details pages).

PTAL

---

Review status: 4 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[i18n/messages-en.xtb, line 375 \[r1\]](https://reviewable.io:443/reviews/kubernetes/dashboard/892#-KK90jtJpYU6gy8BzQ1x-r1-375:-KKJIHBtHvDg8UCw8P82:-499146009) ([raw file](https://github.com/kubernetes/dashboard/blob/331459079f40e515cbfbe2efa800e92e0d1d8e31/i18n/messages-en.xtb#L375)):*
<details><summary><i>Previously, marians (Marian Steinbach) wrote…</i></summary>
> I suggest `All namespaces` or `All Namespaces` depending on the general guideline used (sentence case vs. title case).
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/892#-:-KKJIMbAxzlsgBTywNwV:-1662475724)*

floreks:
:lgtm:

---

Reviewed 3 of 3 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/892#-:-KKJU2TfU27_XTpSD36d:1417088721)*

Issue #911, opened 2016-06-16
Display all node details


maciaszczykm:
Elements to display:

- [x] basic details
- [x] conditions
- [x] pods
- [x] events
- [x] allocated resources
- [ ] addresses
- [ ] container images

PR #912, opened 2016-06-16
Add more node details, pods and events

maciaszczykm:
Added more node details and connected pods to node detail view. Connected to #911.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/912)


maciaszczykm:
Current state:

image

image

cheld:
Nice

floreks:
Could you also move pods to a list under overview tab? It's done same way on all other pages.

maciaszczykm:
@floreks I started with nodes on overview tab, but if we have them here, there is too much info on it. We are still missing few details on this tab and IMO if we are able to avoid scrollbars adding new tab it's the best solution.

Moreover, nodes are a bit different than workloads that we've already implemented so I wouldn't compare them.

@cheld @bryk Whats your opinion?

floreks:
Scrolling will exist pretty much on every detail page. It doesn't require much, just several pods on f.e. RC detail page. However look and feel of detail page will be different than on other pages. In case of details node is still a `resource` for us and for every resource we're displaying pods on `Overview` tab, together with other related resources.

Events are quite different that's why even on mockups we've prepared separate tab for them.

maciaszczykm:
@floreks Okay, let's move this discussion to the next PR where I'll prepare events.

maciaszczykm:
Added events.

Since there are only 10 pods by default, moved pods to overview tab.

PTAL

maciaszczykm:
Current state:

image

image

floreks:
Reviewed 10 of 10 files at r4, 3 of 3 files at r5, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/912)*

floreks:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/912#-:-KKTbyVKX30KqTF4d9h_:1417088721)*

PR #952, opened 2016-06-24
fix lint issue

cheld:
After a fresh/clean build I get the following lint error:

src/app/frontend/events/eventcardlist.scss
  22:12  error  Property `padding` should be written more concisely as `(4 * $baseline-grid) 0` instead of `(4 * $baseline-grid) 0 (4 * $baseline-grid)`  shorthand-values

I fixed as suggested.

Travis and other have not got into issues because of caching node_modules


cheld:
@floreks can you check the change

floreks:
LGTM. Wait for travis and we can merge.

PR #957, opened 2016-06-27
Remove status check for NodePort services

### LINKED ISSUE #955: Service state incorrectly shown as "Pending" for NodePort type services
#### Issue details
Looking at the code [here](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/servicelist/servicecardlist_component.js#L51-L55). Why does it consider a `LoadBalancer` or `NodePort` type service Pending if `externalEndpoints` is missing? As per the [docs](http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_servicespec), it’s not a “required” value.

##### Environment

Kubernetes cluster running on AWS nodes. Dashboard running as a `ReplicationController` and exposed as a cluster service.

```
Dashboard version: v1.1.0
Kubernetes version: v1.2.4
Operating system: Ubuntu 14.04.3 LTS
```

##### Steps to reproduce

* Create a service of `type=NodePort`
```yaml
apiVersion: v1
kind: Service
metadata:
  name: test
  labels:
    run: test
spec:
  type: NodePort
  ports:
  - port: 80
    protocol: TCP
    nodePort: 31182
  selector:
    run: test
```
* Visit the `#/service` page.

##### Observed result

Service is shown as being in `pending state`.

screen shot 2016-06-26 at 8 57 34 pm
screen shot 2016-06-26 at 8 57 49 pm

##### Expected result

Service should not be shown as pending.

screen shot 2016-06-26 at 9 00 06 pm

bryk:
Fixes #955

### DIFF
--- a/src/app/frontend/servicelist/servicecardlist_component.js
+++ b/src/app/frontend/servicelist/servicecardlist_component.js
@@ -50,8 +50,7 @@ export class ServiceCardListController {
    */
   isPending(service) {
     return service.clusterIP === null ||
-        ((service.type === 'LoadBalancer' || service.type === 'NodePort') &&
-         service.externalEndpoints === null);
+        ((service.type === 'LoadBalancer') && service.externalEndpoints === null);
   }
 
   /**
--- a/src/test/frontend/servicelist/servicecardlist_component_test.js
+++ b/src/test/frontend/servicelist/servicecardlist_component_test.js
@@ -57,16 +57,16 @@ describe('Service list controller', () => {
          clusterIP: '10.67.252.103',
          type: 'LoadBalancer',
          externalEndpoints: null,
-       })).toBeTruthy();
+       })).toBe(true);
      });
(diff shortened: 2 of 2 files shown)


bryk:
@digitalfishpond Can you review?

digitalfishpond:
LGTM (just out of curiosity, is there a particular reason for your decision to use toBe(true) rather than toBeTruthy() which I thought were pretty much equivalent?)

bryk:
>  (just out of curiosity, is there a particular reason for your decision to use toBe(true) rather than toBeTruthy() which I thought were pretty much equivalent?)

When you test a boolean function you should use `toBe(true)` and `toBe(false)`, because `toBeTruthly({})`, `toBeTruthly('  ')`, `toBeFalsy(undefined)`, `toBeFalsy(null)` pass :) This is a potential point for problems.

PR #958, opened 2016-06-27
Update relese guide to include minikube addon update

bryk:


bryk:
@maciaszczykm Can you review this? :)

PR #969, opened 2016-06-28
Prepare hyperkube for 1.3

cheld:
Update below
----------------------------
Not to be merged. FYI only. @maciaszczykm, @floreks, @bryk 

I tried make a couple of changes that will be required for Hyperkube 1.3.  Unfortunately, I was not able to fix travis.

I have currently no idea how I could debug travis easily. Not applying the changes has the following consequences after 1.3:
- The included dashboard of hyperkube will crash during start-up
- Sky-dns will crash during start up

Both are not too important for us, currently. Maybe we fix later

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/969)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,11 @@ addons:
   # Run tests that require a browser on SauceLabs CI provider. Learn more at: https://saucelabs.com
   sauce_connect: true
 
+# Docker is required to set up a simple, single node Kubernetes cluster.
+# Local Docker-based cluster is the simplest way to create kubernetes on the host machine.
+services:
+  - docker
+
 before_script:
   # Prepare environment for the Chrome browser. This is required for PRs from forks where
   # tests do not run on Saucelabs.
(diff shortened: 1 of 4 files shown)


cheld:
I will close after you have taken notice and not continue

bryk:
I like this. If travis is happy, we can merge this, IMO

cheld:
Unfortunately, only my workstation is happy, but not travis.

cheld:
I assume if I setup exactly the same debian version as travis, then I could reproduce....

bryk:
Let's see what travis says...

bryk:
Ok, I see. Travis fails. How about you disable all other gulp tasks and just do `gulp local-up-cluster`. Then dump docker logs to see what is happening with all the containers we run. Then maybe dump what the prober sees (the thing that logs "waiting for...").

What do you think?

cheld:
Ok, I have an idea....

cheld:
@bryk PTAL

cheld:
Ok, hold on I have a better idea....

cheld:
Ok sorry, no better idea...PTAL

bryk:
:) Let me know when ready to review

cheld:
PTAL

bryk:
:lgtm:

---

Reviewed 2 of 3 files at r3, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/969#-:-KLzNzUfow36YvIU8O5a:1417088721)*

PR #990, opened 2016-07-05
Add logs font size toggle

### LINKED ISSUE #879: Add control for log text size
The black-on-white vs white-on-black toggle in the log viewer is very cool. Would it be possible to also add a setting for text size? As it is, it takes too much screen estate on my system. I'd love to see more lines at once. Hard-coding a smaller default might not be the best idea. Bonus points if the preferences are persisted.

maciaszczykm:
Added logs font size toggle.

Closes #879.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/990)

### DIFF
--- a/i18n/messages-en.xtb
+++ b/i18n/messages-en.xtb
@@ -480,4 +480,5 @@
 	<translation id="9200661373319050426" key="MSG_POD_DETAILS_POD_LABEL" source="/home/mirchev/go_workspace/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="Label 'Pod' which appears at the top of the delete dialog, opened from a pod details page.">Pod</translation>
 	<translation id="4032745949688106366" key="MSG_PORT_MAPPINGS_PORT_EMPTY_WARNING" source="/Users/bryk/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="This warning appears when the user specifies an empty port.">Port cannot be empty</translation>
 	<translation id="1976944901672561616" key="MSG_PORT_MAPPINGS_TARGET_PORT_EMPTY_WARNING" source="/Users/bryk/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="This warning appears when the user specifies an empty target port.">Target port cannot be empty</translation>
+	<translation id="712304056244084010" key="MSG_LOGS_FONT_SIZE_LABEL" source="/home/maciaszczykm/workspace/dashboard/.tmp/serve/app-dev.js" desc="Label 'Font size' on the toolbar of the logs page. Ends with colon.">Font size:</translation>
 </translationbundle>
--- a/i18n/messages-ja.xtb
+++ b/i18n/messages-ja.xtb
@@ -669,4 +669,5 @@
 	<translation id="9200661373319050426" key="MSG_POD_DETAILS_POD_LABEL" source="/home/mirchev/go_workspace/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="Label 'Pod' which appears at the top of the delete dialog, opened from a pod details page.">ポッド</translation>
 	<translation id="4032745949688106366" key="MSG_PORT_MAPPINGS_PORT_EMPTY_WARNING" source="/Users/bryk/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="This warning appears when the user specifies an empty port.">Port cannot be empty</translation>
 	<translation id="1976944901672561616" key="MSG_PORT_MAPPINGS_TARGET_PORT_EMPTY_WARNING" source="/Users/bryk/src/github.com/kubernetes/dashboard/.tmp/serve/app-dev.js" desc="This warning appears when the user specifies an empty target port.">Target port cannot be empty</translation>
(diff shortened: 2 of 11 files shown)


bryk:
PTAL

---

Reviewed 5 of 9 files at r2.
Review status: 5 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/logs/logs_controller.js, line 47 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/990#-KM31uoWcj1ChYatwEg4:-KM31uoWcj1ChYatwEg5:549598608) ([raw file](https://github.com/kubernetes/dashboard/blob/d43a219cf1696fe6b918a4e9fd088363562632ec/src/app/frontend/logs/logs_controller.js#L47)):*
> ```JavaScript
>    * @return {Object}
>    */
>   getLogsStyle() { return {'font-size': `${this.logsService_.getFontSize()}px`}; }
> ```

Can you change this to `getLogsClass`? So that we don't have inline styles in the code at all. It is (almost) always better to use CSS  classes than inline styles. With classes you can override things, customize stuff, etc. With inline styles it is extremely hard to change stuff later.  

You could return classes like 'kd-logs-small-size' or 'kd-logs-large-size'

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/990#-:-KM32b-W-VIJC3uHdKhT:676965300)*

maciaszczykm:
Review status: 5 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/logs/logs_controller.js, line 47 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/990#-KM31uoWcj1ChYatwEg4:-KM3BABUUscg0kbqxCBz:-310674799) ([raw file](https://github.com/kubernetes/dashboard/blob/d43a219cf1696fe6b918a4e9fd088363562632ec/src/app/frontend/logs/logs_controller.js#L47)):*
> Can you change this to `getLogsClass`? So that we don't have inline styles in the code at all. It is (almost) always better to use CSS  classes than inline styles. With classes you can override things, customize stuff, etc. With inline styles it is extremely hard to change stuff later.  
> 
> You could return classes like 'kd-logs-small-size' or 'kd-logs-large-size'
</details>
I can do it, but I think, that having more available font sizes is better than keeping `font-size` attribute value in CSS file. Only `font-size` is modified in `ng-style`, so in my opinion it wouldn't be so hard to change stuff later.

Anyways, you decide. If you want to use classes and have 3-5 font sizes then I'll change. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/990)*

bryk:
*[src/app/frontend/logs/logs_controller.js, line 47 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/990#-KM31uoWcj1ChYatwEg4:-KM4JXrYE-2OXaWGJOEy:908376680) ([raw file](https://github.com/kubernetes/dashboard/blob/d43a219cf1696fe6b918a4e9fd088363562632ec/src/app/frontend/logs/logs_controller.js#L47)):*
<details><summary><i>Previously, maciaszczykm (Marcin Maciaszczyk) wrote…</i></summary>
> I can do it, but I think, that having more available font sizes is better than keeping `font-size` attribute value in CSS file. Only `font-size` is modified in `ng-style`, so in my opinion it wouldn't be so hard to change stuff later.
> 
> Anyways, you decide. If you want to use classes and have 3-5 font sizes then I'll change. 
</details>
I think that all we should have is "Compact" and "Normal" mode. If we specify allow to specify too many options the system would just end up being too complicated. 

If a user wants to have 5 types of font size, then, I guess, that we have a different problem: our UI is hard to use. 

So, I'd go with classes. Where, as a benefit, we have all typography in constants. 

Do you agree? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/990)*

bryk:
As I've been testing this I see that the toggle is from "Big font" to "normal font". Can we make it "Normal font" and "small one"? 

---

Reviewed 7 of 7 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/logs/logs.scss, line 37 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/990#-KM8Op2YQjXdEthxt3ya:-KM8Op2YQjXdEthxt3yb:2111402868) ([raw file](https://github.com/kubernetes/dashboard/blob/af8ef9938171a3cb2208af7d71f739aed436ec5f/src/app/frontend/logs/logs.scss#L37)):*
> ```SCSS
> 
> .kd-logs-element {
>   font-size: $headline-font-size-base;
> ```

Here. Can we make this form $body, and the other $caption? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/990#-:-KM8OeMiQUQM1Ou4BlyH:1030228174)*

bryk:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/990)*

bryk:
And btw, please ensure that everything is tested :)

maciaszczykm:
Tested on `serve:prod`. Merging.

bryk:
Hey, I still see `$subhead-font-size-base`, whereas we should have `$caption-...`. Is the diff skewed? if not, can you send a follow up PR to make stuff smaller? :)

maciaszczykm:
@bryk It seems to be okay right now: 

https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/logs/logs.scss#L37

PR #1021, opened 2016-07-13
Fix coverage reporting: no more +6 lines covered

bryk:
... on every pull request.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/1021)


bryk:
Merging manually as this is the only way to verify that it'd work...

Issue #1035, opened 2016-07-15
Support ANSI color codes in dashboard log output


ekuefler:
The dashboard log viewer is great, but it doesn't handle my beautiful colored log output very nicely (it looks ok on the console with `kubectl logs`). Currently it looks like this:

```
2016-07-14T19:55:01.042410815Z [34mI 2016-07-14 19:55:01,024:[0;39m Logging initialized @526ms
2016-07-14T19:55:01.117667521Z [34mI 2016-07-14 19:55:01,114:[0;39m jetty-9.3.z-SNAPSHOT
2016-07-14T19:55:01.144351171Z [34mI 2016-07-14 19:55:01,143:[0;39m Started o.e.j.s.ServletContextHandler@5e4c8041{/,null,AVAILABLE}
2016-07-14T19:55:01.358607587Z [34mI 2016-07-14 19:55:01,358:[0;39m Started ServerConnector@40005471{SSL,[ssl, http/1.1]}{0.0.0.0:443}
2016-07-14T19:55:01.358967764Z [34mI 2016-07-14 19:55:01,358:[0;39m Started @865ms
2016-07-14T19:55:01.360299460Z [34mI 2016-07-14 19:55:01,359:[0;39m jetty-9.3.z-SNAPSHOT
2016-07-14T19:55:01.362919564Z [34mI 2016-07-14 19:55:01,362:[0;39m Started o.e.j.s.ServletContextHandler@2aceadd4{/,null,AVAILABLE}
2016-07-14T19:55:01.364062603Z [34mI 2016-07-14 19:55:01,363:[0;39m Started ServerConnector@24aed80c{HTTP/1.1,[http/1.1]}{0.0.0.0:80}
2016-07-14T19:55:01.364378549Z [34mI 2016-07-14 19:55:01,364:[0;39m Started @870ms
2016-07-14T19:55:29.902895969Z [31mW 2016-07-14 19:55:29,902:[0;39m Requested image via deprecated API
2016-07-14T19:55:29.915895809Z [34mI 2016-07-14 19:55:29,914:[0;39m Requested ImageBlobMetadataId[C5LKlz8-QpshXr6z-OKJ2k] in formats []
2016-07-14T19:55:29.921048965Z [34mI 2016-07-14 19:55:29,920:[0;39m Loading ImageBlobMetadataId[C5LKlz8-QpshXr6z-OKJ2k] from datastore
[...]

PR #1070, opened 2016-07-25
Support ANSI color codes in log viewer (#1035)

ekuefler:
Proof-of-concept, looking for feedback. Some questions/potential issues:

1. Is Dashboard okay with depending on this library? (not one I have experience with, just the first one Google gave me)
2. Is there a better way to load the library than src'ing it from the index page?
3. Should we mirror it somewhere instead of relying on rawgit?
4. Trusting all log output as safe HTML to be injected into the page probably isn't very smart, does Angular provide a better way to handle this?

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/1070)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -10,6 +10,7 @@
     "angular-ui-router": "~0.3.1",
     "angular-resource": "~1.5.7",
     "angular-sanitize": "~1.5.7",
+    "ansi-up": "drudru/ansi_up#~v1.3.0",
     "material-design-icons": "~2.2.3",
     "roboto-fontface": "~0.5.0",
     "ng-jsoneditor": "angular-tools/ng-jsoneditor#~1.0.0",
--- a/src/app/externs/ansiup.js
+++ b/src/app/externs/ansiup.js
@@ -1,0 +1,26 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @fileoverview Externs for https://github.com/drudru/ansi_up
+ *
+ * @externs
+ */
+
+const ansi_up = {};
+
+/**
+ * @param {string} ansi
+ */
+ansi_up.ansi_to_html = function(ansi) {};
--- a/src/app/frontend/index.html
+++ b/src/app/frontend/index.html
@@ -22,7 +22,6 @@ limitations under the License.
     <title>Kubernetes Dashboard</title>
     <link rel="icon" type="image/png" href="assets/images/kubernetes-logo.png"/>
     <meta name="viewport" content="width=device-width">
-    <script src="https://cdn.rawgit.com/drudru/ansi_up/32a3c2deb983579fe6149fba4938ce0f840d2afd/ansi_up.js"></script>
 
     <!-- build:css static/vendor.css -->
     <!-- bower:css -->
--- a/src/app/frontend/index.html
+++ b/src/app/frontend/index.html
@@ -22,6 +22,7 @@ limitations under the License.
     <title>Kubernetes Dashboard</title>
     <link rel="icon" type="image/png" href="assets/images/kubernetes-logo.png"/>
     <meta name="viewport" content="width=device-width">
+    <script src="https://cdn.rawgit.com/drudru/ansi_up/32a3c2deb983579fe6149fba4938ce0f840d2afd/ansi_up.js"></script>
 
     <!-- build:css static/vendor.css -->
     <!-- bower:css -->
--- a/src/app/frontend/logs/logs.html
+++ b/src/app/frontend/logs/logs.html
@@ -17,7 +17,7 @@ limitations under the License.
 <div layout="row" flex="auto" layout-fill>
   <md-content layout-padding flex="auto" layout-fill ng-class=ctrl.getStyleClass()>
     <div ng-repeat="n in ctrl.logsSet track by $index" ng-class=ctrl.getLogsClass()>
-      <pre>{{n}}</pre>
+      <pre ng-bind-html="n"></pre>
     </div>
   </md-content>
 </div>
--- a/src/app/frontend/logs/logs_controller.js
+++ b/src/app/frontend/logs/logs_controller.js
@@ -23,10 +23,11 @@ export class LogsController {
    * @param {!./logs_service.LogsService} logsService
    * @ngInject
    */
-  constructor(podLogs, logsService, $sce) {
+  constructor(podLogs, logsService, $sanitize, $sce) {
     /** @export {!Array<string>} Log set. */
     this.logsSet = podLogs.logs.map(function(line) {
-      return $sce.trustAsHtml(ansi_up.ansi_to_html(line));
+      return $sce.trustAsHtml(ansi_up.ansi_to_html(
+          $sanitize(line).replace(/&#27;/g, "\x1b")));
     });
 
     /** @private {!./logs_service.LogsService} */
--- a/src/app/frontend/logs/logs_controller.js
+++ b/src/app/frontend/logs/logs_controller.js
@@ -25,7 +25,7 @@ export class LogsController {
    */
   constructor(podLogs, logsService, $sce) {
     /** @export {!Array<string>} Log set. */
-    this.logsSet = podLogs.logs.map(line => {
+    this.logsSet = podLogs.logs.map((line) => {
       let escapedLine = this.escapeHtml(line);
       let formattedLine = ansi_up.ansi_to_html(escapedLine);
 
--- a/src/app/frontend/logs/logs_controller.js
+++ b/src/app/frontend/logs/logs_controller.js
@@ -24,12 +24,12 @@ export class LogsController {
    * @ngInject
    */
   constructor(podLogs, logsService, $sce) {
-    var self = this;
+    let self = this;
 
     /** @export {!Array<string>} Log set. */
     this.logsSet = podLogs.logs.map(function(line) {
-      var escapedLine = self.escapeHtml(line);
-      var formattedLine = ansi_up.ansi_to_html(escapedLine);
+      let escapedLine = self.escapeHtml(line);
+      let formattedLine = ansi_up.ansi_to_html(escapedLine);
 
       // We know that trustAsHtml is safe here because escapedLine is escaped
       // to not contain any HTML markup, and formattedLine is the result of
@@ -74,7 +74,7 @@ export class LogsController {
    * @returns {string}
    */
   escapeHtml(html) {
-    var div = document.createElement('div');
+    let div = document.createElement('div');
     div.appendChild(document.createTextNode(html));
     return div.innerHTML;
   }
--- a/src/app/frontend/logs/logs_controller.js
+++ b/src/app/frontend/logs/logs_controller.js
@@ -23,9 +23,11 @@ export class LogsController {
    * @param {!./logs_service.LogsService} logsService
    * @ngInject
    */
-  constructor(podLogs, logsService) {
+  constructor(podLogs, logsService, $sce) {
     /** @export {!Array<string>} Log set. */
-    this.logsSet = podLogs.logs;
+    this.logsSet = podLogs.logs.map(function(line) {
+      return $sce.trustAsHtml(ansi_up.ansi_to_html(line));
+    });
 
     /** @private {!./logs_service.LogsService} */
     this.logsService_ = logsService;
--- a/src/app/frontend/logs/logs_controller.js
+++ b/src/app/frontend/logs/logs_controller.js
@@ -21,20 +21,12 @@ export class LogsController {
   /**
    * @param {!backendApi.Logs} podLogs
    * @param {!./logs_service.LogsService} logsService
+   * @param $sce
    * @ngInject
    */
   constructor(podLogs, logsService, $sce) {
     /** @export {!Array<string>} Log set. */
-    this.logsSet = podLogs.logs.map((line) => {
-      let escapedLine = this.escapeHtml(line);
-      let formattedLine = ansi_up.ansi_to_html(escapedLine);
-
-      // We know that trustAsHtml is safe here because escapedLine is escaped
-      // to not contain any HTML markup, and formattedLine is the result of
-      // passing ecapedLine to ansi_to_html, which is known to only add span
-      // tags.
-      return $sce.trustAsHtml(formattedLine);
-    });
+    this.logsSet = podLogs.logs.map((line) => this.formatLine(line, $sce));
 
     /** @private {!./logs_service.LogsService} */
     this.logsService_ = logsService;
@@ -64,6 +56,20 @@ export class LogsController {
       return `${logsTextColor}-invert`;
     }
     return logsTextColor;
+  }
+
+  /**
+   * Formats the given log line as raw HTML to display to the user.
+   * @returns {string}
+   */
+  formatLine(line, $sce) {
+    let escapedLine = this.escapeHtml(line);
+    let formattedLine = ansi_up.ansi_to_html(escapedLine);
+
+    // We know that trustAsHtml is safe here because escapedLine is escaped to
+    // not contain any HTML markup, and formattedLine is the result of passing
+    // ecapedLine to ansi_to_html, which is known to only add span tags.
+    return $sce.trustAsHtml(formattedLine);
   }
 
   /**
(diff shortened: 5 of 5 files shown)


floreks:
1/2. We shouldn't use externally loaded libraries as it's important that dashboard works 'offline'. Some clusters may be set in internal company networks and they may not have access to these libraries. We should actually use something that can be installed using bower. It will be then automatically injected and available to use. 

I'm not sure if there is such library that will work well with angular but it would be best to avoid using `trustAsHtml`. If there is nothing well suited for us then maybe we could add custom common component for this based on some existing library.

maciaszczykm:
> We shouldn't use externally loaded libraries as it's important that dashboard works 'offline'. Some clusters may be set in internal company networks and they may not have access to these libraries. We should actually use something that can be installed using bower. It will be then automatically injected and available to use.

Yeah, we cannot use externally loaded libraries. 

> I'm not sure if there is such library that will work well with angular but it would be best to avoid using trustAsHtml. If there is nothing well suited for us then maybe we could add custom common component for this based on some existing library.

It's good to avoid it. If we want only colors to be applied, then adding our own HTML parser can solve our problem.

ekuefler:
My knowledge of Bower / Angular / the JS ecosystem is fairly limited, so apologies for basic questions.

It looks like ansi_up defines [bower.json](https://github.com/drudru/ansi_up/blob/master/bower.json), so presumably we could install it through Bower. Are we okay with depending on it in that way or would it be better to avoid external dependencies altogether and reimplement the parser inside the dashboard repo?

Google gives me https://github.com/jorgeecardona/ansi-to-html as an angular-specialized converter, but it doesn't have a bower.json and looks to have no users.

Are there ways to get HTML into an Angular template that don't involve `trustAsHtml`? Assuming we trust the library (it should only generate `span` tags), I think it would theoretically be safe to do something like `trustAsHtml(convertToAnsi(escapeHtml(logLine)))`.

maciaszczykm:
> It looks like ansi_up defines bower.json, so presumably we could install it through Bower. Are we okay with depending on it in that way or would it be better to avoid external dependencies altogether and reimplement the parser inside the dashboard repo?

There is no problem to add new dependency if it would solve the problem, but if it doesn't meet our requirements then we should implement our own parser. I don't have strong opinion on `trustAsHtml`, let's wait for @bryk's opinion.

> Google gives me https://github.com/jorgeecardona/ansi-to-html as an angular-specialized converter, but it doesn't have a bower.json and looks to have no users.

https://github.com/drudru/ansi_up seems to be better choice.

> Are there ways to get HTML into an Angular template that don't involve trustAsHtml? Assuming we trust the library (it should only generate span tags), I think it would theoretically be safe to do something like trustAsHtml(convertToAnsi(escapeHtml(logLine))).

I don't think so. Other option is usage of stylesheets, but it may be too complicated. Perhaps we should stick to `trustAsHtml`...

vincentwoo:
@ekuefler hahaaaaa thissss guy

bryk:
> There is no problem to add new dependency if it would solve the problem, but if it doesn't meet our requirements then we should implement our own parser. I don't have strong opinion on trustAsHtml, let's wait for @bryk's opinion.

Yeah, Marcin is right. If you can find a bower package that solves the problem, then we're good to go. If you need help on how to install the package, let us know. But if this is a matter of small amount of JavaScript to write on our own, we can just do that :) And avoid external dependency at all.

ekuefler:
Okay, updated to install through Bower and do a little bit of sanitizing.

> Other option is usage of stylesheets, but it may be too complicated. Perhaps we should stick to trustAsHtml.

My guess is a pure stylesheet solution isn't possible - the formatting codes are mixed in arbitrarily with the log output, so you need to write spans at some point to decide what gets colored.

Here's a before screenshot:
before

And after:
after

bryk:
Looks nice! I think we now need to really make sure that this has no security holes. Once we have it, let's merge :)

bryk:
Can you add a comment to the code explaining why it is safe and secure? So that others later do not worry about it.

ekuefler:
I realized we probably want to escape rather than sanitize (since we probably want any HTML markup in logs to show up literally), so I switched to using an escaping strategy suggested by a coworker and documented. Probably still good for someone on your side to double-check that it's safe.

Looks like Travis doesn't like the `ansi_up` symbol, how can I tell it that it's safe?

floreks:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/app/frontend/logs/logs_controller.js, line 77 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KNv1EZo8dH8Jn9432z5:-KNv1EZo8dH8Jn9432z6:197296529) ([raw file](https://github.com/kubernetes/dashboard/blob/8102c4f511dfd66420616de4a09abdbb66c14d75/src/app/frontend/logs/logs_controller.js#L77)):*
> ```JavaScript
>    */
>   escapeHtml(html) {
>     var div = document.createElement('div');
> ```

Travis doesn't like this part. `Unexpected var, use let or const instead.` We're using ES6 and `let` is used instead of `var`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070)*

floreks:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/app/frontend/logs/logs_controller.js, line 27 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8wuOTXPtuFDTRrPR1:-KO8wuOTXPtuFDTRrPR2:1728571085) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L27)):*
> ```JavaScript
>    */
>   constructor(podLogs, logsService, $sce) {
>     let self = this;
> ```

You don't have to use `self` workaround. It's not needed when using ES6. 

---

*[src/app/frontend/logs/logs_controller.js, line 30 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8xWzeshR0_vw3-sC7:-KO8xWzeshR0_vw3-sC8:-1505030130) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L30)):*
> ```JavaScript
> 
>     /** @export {!Array<string>} Log set. */
>     this.logsSet = podLogs.logs.map(function(line) {
> ```

Use arrow(fat) functions instead of old style functions.

---

*[src/app/frontend/logs/logs_controller.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8xzHHwKNy9nvhd2fs:-KO8xzHHwKNy9nvhd2ft:-443176974) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L32)):*
> ```JavaScript
>     this.logsSet = podLogs.logs.map(function(line) {
>       let escapedLine = self.escapeHtml(line);
>       let formattedLine = ansi_up.ansi_to_html(escapedLine);
> ```

Closure compiler won't accept this because it's not injected through constructor and it doesn't have JSDoc specified. Basically compiler doesn't know what it is and where it came from.

It should be injected into constructor if it's possible and JSDoc in externs has to be added. You can take a look at other classes to see how it's done or I can try to resolve that if you don't have time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070)*

ekuefler:
Added an externs file, we'll see if Travis likes it. Didn't inject it since I couldn't find any other examples of injecting a third-party lib. Feel free to take this over and finish it off or to point me towards what I should do for it.

floreks:
Nice. Travis passed. Looks nice. I have to test this and I'll have few more remarks I think. Almost ready to merge. :)

maciaszczykm:
@floreks @ekuefler Yep, yesterday Travis flaked due to random error... I've restared it and it seems okay right now :)

floreks:
Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/app/frontend/logs/logs_controller.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8xzHHwKNy9nvhd2fs:-KOPZzf_YNG4vsWOnn1a:b-knzvjh) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L32)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> Closure compiler won't accept this because it's not injected through constructor and it doesn't have JSDoc specified. Basically compiler doesn't know what it is and where it came from.
> 
> It should be injected into constructor if it's possible and JSDoc in externs has to be added. You can take a look at other classes to see how it's done or I can try to resolve that if you don't have time.
</details>
Right now added externs are not working because they are not used by JSDoc. The reason why travis have passed is that compiler has no clue about type of `ansi_up`. I'm not sure how to resolve that, because we can't inject this module through constructor as it is not angular friendly. @bryk any ideas?

---

*[src/app/frontend/logs/logs_controller.js, line 26 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KOPaWcw_3tWx8u2Ihmm:-KOPaWcw_3tWx8u2Ihmn:b33lj6p) ([raw file](https://github.com/kubernetes/dashboard/blob/ecfd9bfed188890484e38bd15ced1fc00b6b28ec/src/app/frontend/logs/logs_controller.js#L26)):*
> ```JavaScript
>    * @ngInject
>    */
>   constructor(podLogs, logsService, $sce) {
> ```

No JSDoc specified for `$sce`

---

*[src/app/frontend/logs/logs_controller.js, line 28 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KOPZjefS3cfScqZxcve:-KOPZjefS3cfScqZxcvf:bkx8a1k) ([raw file](https://github.com/kubernetes/dashboard/blob/ecfd9bfed188890484e38bd15ced1fc00b6b28ec/src/app/frontend/logs/logs_controller.js#L28)):*
> ```JavaScript
>   constructor(podLogs, logsService, $sce) {
>     /** @export {!Array<string>} Log set. */
>     this.logsSet = podLogs.logs.map((line) => {
> ```

I'd extract this logic to new function, i.e.`formatLogs`. It will be easier to test it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070)*

bryk:
Hey @ekuefler this looks awesome! We have one unresolved comment about documentation and we can merge. Do you heed any help on this? 

---

Reviewed 1 of 3 files at r1, 1 of 3 files at r2, 1 of 2 files at r5.
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/app/frontend/logs/logs_controller.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8xzHHwKNy9nvhd2fs:-KOjXqZBsVxC3Wa9fR3w:b2tv7wz) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L32)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> Right now added externs are not working because they are not used by JSDoc. The reason why travis have passed is that compiler has no clue about type of `ansi_up`. I'm not sure how to resolve that, because we can't inject this module through constructor as it is not angular friendly. @bryk any ideas?
</details>
I think this is resolved, right? 

---

*[src/app/frontend/logs/logs_controller.js, line 28 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KOPZjefS3cfScqZxcve:-KOjYdidK9-GRTfz9Yix:b-2worml) ([raw file](https://github.com/kubernetes/dashboard/blob/ecfd9bfed188890484e38bd15ced1fc00b6b28ec/src/app/frontend/logs/logs_controller.js#L28)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> I'd extract this logic to new function, i.e.`formatLogs`. It will be easier to test it.
</details>
I think it is fine :) But I'll let @ekuefler decide. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-:-KOjYSxwIgWFJh15Zorj:b7qsi5l)*

ekuefler:
Updated - should be good unless I need to say something about $sce's type. Also trying reviewable for the first time - this thing's slick.

---

Reviewed 1 of 3 files at r1, 1 of 3 files at r2, 1 of 2 files at r5, 1 of 1 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/app/frontend/logs/logs_controller.js, line 77 \[r3\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KNv1EZo8dH8Jn9432z5:-KOlPZ4UBbHTyEAKOjQv:b-896fix) ([raw file](https://github.com/kubernetes/dashboard/blob/8102c4f511dfd66420616de4a09abdbb66c14d75/src/app/frontend/logs/logs_controller.js#L77)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> Travis doesn't like this part. `Unexpected var, use let or const instead.` We're using ES6 and `let` is used instead of `var`.
</details>
Done.

---

*[src/app/frontend/logs/logs_controller.js, line 27 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8wuOTXPtuFDTRrPR1:-KOlOeArZyv09zYeBJdL:b-896fix) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L27)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> You don't have to use `self` workaround. It's not needed when using ES6. 
</details>
Done.

---

*[src/app/frontend/logs/logs_controller.js, line 30 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8xWzeshR0_vw3-sC7:-KOlPePO9QgJV7KJB2pI:b-896fix) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L30)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> Use arrow(fat) functions instead of old style functions.
</details>
Done.

---

*[src/app/frontend/logs/logs_controller.js, line 32 \[r4\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KO8xzHHwKNy9nvhd2fs:-KOlPQiT6RqtybyC7VfF:b-896fix) ([raw file](https://github.com/kubernetes/dashboard/blob/0ae85e616977a2ec2b6d00cfe6ad28f8ac1a1e3d/src/app/frontend/logs/logs_controller.js#L32)):*
> I think this is resolved, right? 
</details>
Done.

---

*[src/app/frontend/logs/logs_controller.js, line 26 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KOPaWcw_3tWx8u2Ihmm:-KOlOPeo_OrRQbcU4PdS:bxvuvtr) ([raw file](https://github.com/kubernetes/dashboard/blob/ecfd9bfed188890484e38bd15ced1fc00b6b28ec/src/app/frontend/logs/logs_controller.js#L26)):*
<details><summary><i>Previously, floreks (Sebastian Florek) wrote…</i></summary>
> No JSDoc specified for `$sce`
</details>
Added param tag - not sure what the type of this thing is or if it's required

---

*[src/app/frontend/logs/logs_controller.js, line 28 \[r6\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-KOPZjefS3cfScqZxcve:-KOlPReDzAOp4h8iblMt:b-xqj3q) ([raw file](https://github.com/kubernetes/dashboard/blob/ecfd9bfed188890484e38bd15ced1fc00b6b28ec/src/app/frontend/logs/logs_controller.js#L28)):*
> I think it is fine :) But I'll let @ekuefler decide. 
</details>
Seems reasonable to me - I have a couple other formatting ideas, so the logic here might get more complex.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-:-KOlPuOaGmgb4Sf0yESp:b-wl62d7)*

bryk:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070)*

bryk:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1070#-:-KOneBiqXOxNRD2Y5_b4:bnfp4nl)*

bryk:
Travis is down, merging.

PR #1077, opened 2016-07-28
Give logs view sane line spacing

vincentwoo:
This is what logs look like now: 

screenshot 2016-07-27 20 52 59

This is what they look like after this change: 

screenshot 2016-07-27 20 53 22

but honestly having a `<pre>` element for each line of log output seems pretty nonideal

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/1077)

### DIFF
--- a/src/app/frontend/logs/logs.scss
+++ b/src/app/frontend/logs/logs.scss
@@ -35,7 +35,8 @@ $logs-color-white: #fff;
   font-size: $body-font-size-base;
 }
 
-.kd-logs-element, .kd-logs-element-compact {
+.kd-logs-element,
+.kd-logs-element-compact {
   padding: 0 (2 * $baseline-grid);
   word-wrap: break-word;
 
--- a/src/app/frontend/logs/logs.scss
+++ b/src/app/frontend/logs/logs.scss
@@ -27,7 +27,8 @@ $logs-color-white: #fff;
   color: $logs-color-black;
 }
 
-@mixin kd-logs-element-mixin {
+.kd-logs-element {
+  font-size: $body-font-size-base;
   padding: 0 (2 * $baseline-grid);
   word-wrap: break-word;
 
@@ -37,11 +38,7 @@ $logs-color-white: #fff;
 }
 
 .kd-logs-element-compact {
-  @include kd-logs-element-mixin;
-  font-size: $caption-font-size-base;
-}
+  @extend .kd-logs-element;
 
-.kd-logs-element {
-  @include kd-logs-element-mixin;
-  font-size: $body-font-size-base;
+  font-size: $caption-font-size-base;
 }
--- a/src/app/frontend/logs/logs.scss
+++ b/src/app/frontend/logs/logs.scss
@@ -27,20 +27,21 @@ $logs-color-white: #fff;
   color: $logs-color-black;
 }
 
-.kd-logs-element-compact {
-  font-size: $caption-font-size-base;
-}
-
-.kd-logs-element {
-  font-size: $body-font-size-base;
-}
-
-.kd-logs-element,
-.kd-logs-element-compact {
+@mixin kd-logs-element-mixin {
   padding: 0 (2 * $baseline-grid);
   word-wrap: break-word;
 
   pre {
     margin: 0;
   }
+}
+
+.kd-logs-element-compact {
+  font-size: $caption-font-size-base;
+  @include kd-logs-element-mixin;
+}
+
+.kd-logs-element {
+  font-size: $body-font-size-base;
+  @include kd-logs-element-mixin;
 }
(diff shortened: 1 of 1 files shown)


floreks:
I like this change. Use gulp check if you have dev environment configured. It's faster than waiting for travis.

@maciaszczykm what do you think about this change?

---

Reviewed 1 of 1 files at r2.

---

*[src/app/frontend/logs/logs.scss, line 41 \[r2\]](https://reviewable.io:443/reviews/kubernetes/dashboard/1077#-KNk56bbKzX3UHBV0qgG:-KNk56bbKzX3UHBV0qgH:-624340796) ([raw file](https://github.com/kubernetes/dashboard/blob/cf696bc94e59166cd9ef48131af451e0d6c214b8/src/app/frontend/logs/logs.scss#L41)):*
> ```SCSS
>   word-wrap: break-word;
> 
>   pre {
> ```

This would be applied globally and I think it is not expected. I'd use custom css class instead and follow our naming convention `kd-logs-line`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubernetes/dashboard/1077#-:-KNk5mDVxMXC3WXrCj09:-412250310)*

vincentwoo:
I'm making the change through the github editor.

The change is not global, expand the context above the change site in your code viewer.

floreks:
Ok, I haven't seen that. Still there is a slight issue with that as `kd-logs-element` is added dynamically. `kd-logs-element-compact` would need same css style in order to make it work when font is changed.

I'd go with custom class as it would work in both cases with just 1 change.

vincentwoo:
Adding a class for a nonsemantic style seems crazy, I've just restructured your scss instead.

floreks:
It's best not to split class styling into multiple parts. We're using scss, so if you think introducing new class is too much then use power of scss.

```
@mixin kd-logs-element-mixin($font-size) {
  font-size: $font-size;
  padding: 0 (2 * $baseline-grid);
  word-wrap: break-word;

  pre {
    margin: 0;
  }
}
```

```
.kd-logs-element-compact {
  @include kd-logs-element-mixin($caption-font-size-base);
}

.kd-logs-element {
  @include kd-logs-element-mixin($body-font-size-base);
}
```

vincentwoo:
I personally think the coupling involve in your suggestion is overkill, and that a split class definition is much clearer. If you wish to make that change for your project, I'll let you go for it.

floreks:
From my experience split class definition may create some issues if they get separated. Some code may land between the definitions or some definition lands in another file. Dev is not aware of that and tries to add i.e. padding to first one, but it gets overridden by the second definition. It takes time to find the cause because it's not that obvious.

That's why I believe introducing only single point of failure is a better choice, despite if it's code or css style, but I'll wait for @maciaszczykm and @bryk opinion.

maciaszczykm:
IMO we shouldn't split definition of same CSS class into two if it's not neccessary. It may look clear now, but when logs view will evolve and we will add more lines to this file, we may forget about it while refactoring etc. Copying content of `kd-logs-element` to `kd-logs-element-compact` is also a solution, not the best, because it's against DRY rule, but at least we can avoid issues later on.

vincentwoo:
at your insistence i have implemented floreks proposal minus the font-size-as-argument bit because that level of indirection physically pains me to see

maciaszczykm:
@vincentwoo We can merge it after CI will be green:

https://travis-ci.org/kubernetes/dashboard/builds/147967819

```
41:3  error  Mixins should come before declarations  mixins-before-declarations
46:3  error  Mixins should come before declarations  mixins-before-declarations
```

floreks:
I've only proposed mixin because it gives ability to pass parameter. If you've removed it then it's best to use extend/inheritance in order to share base style as described in [sass guide](http://sass-lang.com/guide).

My proposal with extend:
```scss
.kd-logs-element {
  font-size: $body-font-size-base;
  padding: 0 (2 * $baseline-grid);
  word-wrap: break-word;

  pre {
    margin: 0;
  }
}

.kd-logs-element-compact {
  @extend .kd-logs-element;

  font-size: $caption-font-size-base;
}
```

@maciaszczykm what do you think?

vincentwoo:
seems very reasonable

vincentwoo:
Your build system threw some kind of weird error and it's getting late. Please take this PR and do what you want with it, I don't really care. Merge/rebase away.

The kube dashboard isn't very good. I believe in you guys. You can fix it. Believe in yourselves. Godspeed.

bryk:
@floreks @maciaszczykm Are we doing something with this? Maybe merge and then quickly fix maser?

maciaszczykm:
@bryk Yes, it's good idea. This makes logs look better, so let's not abandon it. We can do it tomorrow.

bryk:
Awesome, thanks.

floreks:
@bryk would you rather disable [this](https://github.com/kubernetes/dashboard/blob/master/.sass-lint.yml#L25) in order to allow extending of existing classes or use [placeholder class](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#placeholder_selectors_)?

bryk:
Ok, I'll merge this and fix travis soon.

PR #1163, opened 2016-08-30
Changed serve.js to use InClusterHeapsterClient by default.

PiotrDabkowski:
Most of the people use in cluster heapster.

What do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/1163)

### DIFF
--- a/build/serve.js
+++ b/build/serve.js
@@ -38,7 +38,6 @@ export const browserSyncInstance = browserSync.create();
 const backendDevArgs = [
   `--apiserver-host=${conf.backend.apiServerHost}`,
   `--port=${conf.backend.devServerPort}`,
-  `--heapster-host=${conf.backend.heapsterServerHost}`,
 ];
 
 /**
@@ -49,7 +48,6 @@ const backendDevArgs = [
 const backendArgs = [
   `--apiserver-host=${conf.backend.apiServerHost}`,
   `--port=${conf.frontend.serverPort}`,
-  `--heapster-host=${conf.backend.heapsterServerHost}`,
 ];
 
 /**


bryk:
@maciaszczykm @floreks @digitalfishpond @cheld What do you think? Does this break your workflow?

cheld:
Heapster is not part of Hyperkube, yet (because heapster is not cross-compiled, @zreigz is just looking into this topic). The settings are aligned with build/hyperkube.sh. I am wondering why this change is not breaking the build. I assume there are no tests to check heapster integration. 

I assume you want to change, because you are developing against GKE. Isn't `kubectl proxy --port=8080` are more convenient way?

cheld:
@batikanu

floreks:
I think most devs install heapster as part of their test cluster anyway. This change is also good for me because I'm deploying all-in-one heapster yaml and always have to change config in order to use in cluster heapster rather than the one outside the cluster.

LGTM

cheld:
Again, this change will remove heapster from TravisCI. I don't think this is intended. 

Not sure what is requested/required. Maybe a way to supersede settings without producing a git change set? 

In general, IMHO we have the following requirements:
1. automatic setup of a test environment for travis
2. automatic setup of a personal test environment identical to travis (for debugging)
3. one reference setup for manual tests

floreks:
I understand that it would remove heapster from Travis. We don't have any tests for that unfortunately.

We could solve that by removing standalone Heapster and deploying in cluster one using kubectl on Travis. This way devs with their own clusters won't have to alter configuration and heapster will still be available on Travis.

Personally I've used `hyperkube.sh` only few times so this change would be nice.

@cheld @maciaszczykm what do you think?

bryk:
@floreks I agree with what @floreks states here. Default to in-cluster Heapster and use it. We had out-of-cluster Heapster installation because of legacy installation reasons that are not a problem anymore. 

@rf232 Do you have any opinions?

rf232:
I don't have much of a feeling for what would be the best for the project, I don't know the workflow of enough people yet. I do know that I have a similar change for myself as well to aid in my own dev work. So perhaps at least making this settable from the commandline or making a serve:inClusterHeap target available would be prudent.

rf232:
```
git diff
diff --git a/build/conf.js b/build/conf.js
index c10a4a1..384837b 100644
--- a/build/conf.js
+++ b/build/conf.js
@@ -21,6 +21,7 @@ import path from 'path';
  * Load the i18n and l10n configuration. Used when dashboard is built in production.
  */
 let localization = require('../i18n/locale_conf.json');
+let util = require('gulp-util');
 
 /**
  * Base path for all other paths.
@@ -96,7 +97,7 @@ export default {
      * Address for the Heapster API server. If blank, the dashboard
      * will attempt to connect to Heapster via a service proxy.
      */
-    heapsterServerHost: 'http://localhost:8082',
+    heapsterServerHost: util.env.heapsterServerHost !== undefined ? util.env.heapsterServerHost : 'http://localhost:8082',
   },
 
   /**
```

allows me to run `gulp serve --heapsterServerHost ''` to use the in cluster running heapster

floreks:
What do you think about flipping the logic here? It feels more natural for me if by default we use in-cluster heapster and when this parameter is actually provided then we can fall back to external one.

bryk:
I think this should be closed now as @rf232 reimplemented identical logic.

Issue #1249, opened 2016-09-16
Heapster metrics can go below 0


bryk:
Sometimes Heapster metrics go below 0. We should protect from this. 

See https://raw.githubusercontent.com/jskswamy/dashboard/b9b34f9a1a1bdc0178e00f8e035cbe129ed66db4/docs/dashboard-ui.png

maciaszczykm:
+ plot sizes should be equal.

floreks:
Fixed. Closing.

PR #1250, opened 2016-09-16
Make it impossible to metrics go under 0

bryk:
Re #1249


bryk:
@maciaszczykm Can you review?

PR #1254, opened 2016-09-19
Fix gutter between graph cards

bryk:
Now it is correctly 20px. Previously it was 2x this.

### DIFF
--- a/src/app/frontend/common/components/graph/graphcard.scss
+++ b/src/app/frontend/common/components/graph/graphcard.scss
@@ -16,4 +16,14 @@
 
 kd-graph-card {
   flex: 1;
+
+  &>kd-content-card {
+    margin-bottom: 0;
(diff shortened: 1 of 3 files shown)


bryk:
@floreks Can you review this?

src/app/frontend/common/components/graph/graphcard.scss
```diff
@@ -16,4 +16,21 @@
...
+  }
+
+  &:last-child {
+    &>kd-content-card {
+      margin-left: 0;
+    }
+  }
+}
+
+.kd-right-graph {
```
floreks: This class should no longer exist so this part is not needed :)

bryk: Yep, not needed. Removed :)

PR #1295, opened 2016-09-29
Hide basicAuth and opaque secret data

### LINKED ISSUE #1291: Dashboard v1.4.0 incorrectly renders opaque secrets
#### Issue details

the new dashboard shows any secret including opaque ones I would have expected this to have the same behaviour as kubectl which will only list the file names of an opaque secret not the contents
##### Environment

```
Dashboard version: v1.4.0
Kubernetes version: v1.4.0
Operating system: coreos
Node.js version: n/a
Go version: n/a
```
##### Steps to reproduce

deploy the currently manifest for v1.4.0 dashboard and add an opaque secret to your kubernetes cluster then try to view it in the ui
##### Observed result

I can see the full details of any secret via the web ui dashboard even opaque ones
##### Expected result

I would expect to see a listing on the secrets but no contents similar to what kubectl shows.

batikanu:
This fixes #1291

### DIFF
--- a/i18n/messages-en.xtb
+++ b/i18n/messages-en.xtb
@@ -1152,7 +1152,8 @@
   <translation id="4826002371920064937" key="MSG_RESOURCE_DETAIL_INFO_CARD_NAME_LABEL" desc="Label 'Name' for a resource.">Name</translation>
   <translation id="6116578277969213046" key="MSG_RESOURCE_DETAIL_INFO_CARD_START_TIME_LABEL" desc="Label 'Creation time' for a resource.">Creation time</translation>
   <translation id="1320482404001625698" key="MSG_SECRETDETAIL_ACTIONBAR_0" desc="Label \'Secret\' which appears at the top of the delete dialog, opened from a secret details page.">Secret</translation>
-  <translation id="7988357395949564443" key="MSG_SECRETDETAIL_DETAIL_0" desc="Config map info details section name.">Data</translation>
+  <translation id="1561742373508692276" key="MSG_SECRETDETAIL_DETAIL_0" desc="Secrets info details section data.">Data </translation>
+  <translation id="1940368443923988450" key="MSG_SECRETDETAIL_DETAIL_1" desc="Secrets info details section bytes.">bytes </translation>
   <translation id="4201489447726655383" key="MSG_SECRETDETAIL_INFO_0" desc="Config map info details section name.">Details</translation>
   <translation id="2838644310432685833" key="MSG_SECRETLIST_CARDLIST_0" desc="Label \'Name\' which appears as a column label in the table of secrets (secret list view).">Name</translation>
   <translation id="3059701171980133521" key="MSG_SECRETLIST_CARDLIST_1" desc="Label \'Namespace\' which appears as a column label in the table of secrets (secret list view).">Namespace</translation>
--- a/src/app/backend/resource/secret/secretdetail.go
+++ b/src/app/backend/resource/secret/secretdetail.go
@@ -36,6 +36,9 @@ type SecretDetail struct {
 
 	// Used to facilitate programmatic handling of secret data.
 	Type api.SecretType `json:"type"`
+
+	// Determines if the secret is hidden type in case of opaque and basicAuth types.
+	IsHiddenType bool `json:"isHiddenType"`
 }
 
 // GetSecretDetail returns returns detailed information about a secret
@@ -51,11 +54,18 @@ func GetSecretDetail(client *client.Client, namespace, name string) (*SecretDeta
 	return getSecretDetail(rawSecret), nil
 }
 
+// For opaque and basicAuth secret types the credentials like username, password and etc. should be hidden.
+func isHiddenType(secretType api.SecretType) bool {
+	return secretType == api.SecretTypeBasicAuth || secretType == api.SecretTypeOpaque
+}
+
+
 func getSecretDetail(rawSecret *api.Secret) *SecretDetail {
 	return &SecretDetail{
 		ObjectMeta: common.NewObjectMeta(rawSecret.ObjectMeta),
 		TypeMeta:   common.NewTypeMeta(common.ResourceKindSecret),
 		Data:       rawSecret.Data,
+		IsHiddenType: isHiddenType(rawSecret.Type),
 		Type:       rawSecret.Type,
 	}
 }
--- a/src/app/frontend/common/components/togglehiddentext/togglehiddentext.html
+++ b/src/app/frontend/common/components/togglehiddentext/togglehiddentext.html
@@ -1,0 +1,20 @@
+<div layout="row" flex="auto" class="kd-toggle-hidden-text-row">
+    <input readonly flex="nogrow" ng-if="isActive" type="text"
+           class="kd-toggle-hidden-text-input" value="{{::$ctrl.text}}">
+    <span ng-if="!isActive">
+        <input ng-if="::$ctrl.placeholder.length" readonly flex="nogrow"
+               type="text" class="kd-toggle-hidden-text-input"
+               value="{{::$ctrl.placeholder}}">
+        <input ng-if="::!$ctrl.placeholder.length" type="password" readonly
+               flex="nogrow" class="kd-toggle-hidden-text-input"
+               value="{{::$ctrl.text}}">
+    </span>
+    <div>
+        <md-icon ng-show="!isActive" ng-click="isActive = !isActive"
+                 md-font-set="material-icons"
+                 class="kd-toggle-eye-icon kd-toggle-eye-icon-off "> remove_red_eye </md-icon>
+        <md-icon ng-show="isActive" ng-click="isActive = !isActive"
+                 md-font-set="material-icons"
+                 class="kd-toggle-eye-icon kd-toggle-eye-icon-on"> remove_red_eye </md-icon>
+    </div>
+</div>
--- a/src/app/frontend/secretdetail/detail.html
+++ b/src/app/frontend/secretdetail/detail.html
@@ -15,14 +15,17 @@ limitations under the License.
 -->
 
 <div layout="column">
-  <kd-secret-info secret="::$ctrl.secretDetail"></kd-secret-info>
-  <kd-info-card>
-    <kd-info-card-section name="[[Data|Config map info details section name.]]">
-      <kd-info-card-entry ng-repeat="(key, value) in ::$ctrl.secretDetail.data"
-          title="{{::key}}">
-        <!-- TODO(#1059): Don't show plain text value. -->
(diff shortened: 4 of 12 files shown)


src/app/backend/resource/secret/secretdetail.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
 	"k8s.io/kubernetes/pkg/api"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
+	//"fmt"
```
maciaszczykm: Not needed.

batikanu: Thanks. PTAL

src/app/backend/resource/secret/secretdetail.go
```diff
@@ -51,11 +52,24 @@ func GetSecretDetail(client *client.Client, namespace, name string) (*SecretDeta
 	return getSecretDetail(rawSecret), nil
 }
 
+func hideSecretData(rawSecret *api.Secret) map[string][]byte {
+	// For opaque and basicAuth secret types the credentials like username, password and etc. should be hidden.
+	if rawSecret.Type == api.SecretTypeOpaque || rawSecret.Type == api.SecretTypeBasicAuth {
+		data := make(map[string][]byte)
+		for key, value := range rawSecret.Data {
+			data[key] = []byte(fmt.Sprintf("%d bytes", len(value)))
```
bryk: Hmm... This is not translatable. Can you send the data and render it on the client as the `FOO bytes`? 

And can you make sure to enable the option of rendering this fully? I.e., `foo-secret: 1234 bytes (show value)`. This is because we cannot entirely hide the value of the secret. Hiding it by default is nice, but entirely, not that much. 

@maciaszczykm @digitalfishpond What do you think?

maciaszczykm: I agree, the best option seems to be adding switch like `(show value)` and hiding value by default.

ashmere: @bryk @maciaszczykm  personally unless it is a startup flag for the dashboard I don't think there should be an option on a production kubernetes dashboard to see opaque secret, as this can expose a lot of things the general kubernetes admin/user thought was secure to anyone with access to the UI.

bryk: But such thing would be only an illusion of security. CLI and API returns opaque secrets with full info to all users. So if a user can see the secret, they can see its opaque content.

batikanu:
How about this?

hiddentext

bryk:
@batikanu Oh, yeah, this looks cool! 

I'm wondering whether this should be links or toggles. We have links "show more" +then "show less" when there are more than X labels on a resource. It feels more natural to me, because here it is unclear what happens when you toggle the toggle. Will it "show more" or is currently "showing more"? 
@romlein @rf232  Can you comment on this?

maciaszczykm:
+1 for links, I'd like to be consistent and besides that, such a big toggle looks strange near small text.

batikanu:
@bryk @maciaszczykm  thanks for the feedbacks. Actually, the intend was to implement a new component for such fields and also password fields which shows by default "****" and displays the hidden value if you click on eye eye
  or switch 
switch.
 It was also stated by @bryk in the code as todo. I've just realized that such a component must include also label to have a better layout. It requires a bit more effort therefore I wanted to discuss it before implementing. 
I can change it to simple links or make it reusable component. What do you think?

maciaszczykm:
@batikanu New reusable component sounds very nice :)

rf232:
I personally think the usecases between more or less labels is quite different from showing/hiding secrets data. And for that something like the eye or switch actually feels quite natural. I just now looked at my password storage and it indeed uses such a eye icon.

minor nitpick: is your length in bytes correct for (non ascii) unicode strings?

batikanu:
> minor nitpick: is your length in bytes correct for (non ascii) unicode strings?
I already receive the data as byte from the backend. Therefore I just use its length.

bryk:
+1 on eye icon

romlein:
If I'm understanding the use case correctly, I would use the eye pattern as shown in this interactive [InVision prototype](https://invis.io/6M91K6GQG#/198955708_Hidden) which references this Material spec for [icon toggles](https://material.google.com/components/buttons.html#buttons-toggle-buttons).
The problem with this solution is that the icon has to be a smaller than is ideal to fit in between the current spacing of rows, but I think this is a good start.
Toggles are typically for turning something _on_ or _off_.

bryk:
All right, I see general agreement on the eye-icon approach. @batikanu Can you progress with this option? :)

batikanu:
@bryk  I will proceed with eye-icon approach. @romlein why don't you prefer [eye-off icon ]( https://materialdesignicons.com/icon/eye-off ) when it is focused?

src/app/backend/resource/secret/secretdetail.go
```diff
@@ -36,6 +36,8 @@ type SecretDetail struct {
 
 	// Used to facilitate programmatic handling of secret data.
 	Type api.SecretType `json:"type"`
+
```
bryk: Please add small docs info here. About what this is.

src/app/frontend/secretdetail/detail.html
```diff
@@ -15,14 +15,27 @@
...
-      <kd-info-card-entry ng-repeat="(key, value) in ::$ctrl.secretDetail.data"
-          title="{{::key}}">
-        <!-- TODO(#1059): Don't show plain text value. -->
-        <pre class="kd-pre-block">{{::$ctrl.formatDataValue(value)}}</pre>
-      </kd-info-card-entry>
-    </kd-info-card-section>
-  </kd-info-card>
+    <kd-secret-info secret="::$ctrl.secretDetail"></kd-secret-info>
+    <kd-info-card>
+        <kd-info-card-section name="{{::$ctrl.i18n.MSG_SECRET_INFO_DATA_SECTION}}">
```
bryk: Please use the new i18n framework here :) It is much easier for you to write.

src/app/frontend/secretdetail/detail.html
```diff
@@ -15,14 +15,27 @@
...
+            <kd-info-card-entry ng-repeat="(key, value) in ::$ctrl.secretDetail.data"
+                                title="{{::key}}">
+                <!-- TODO(#1059): Don't show plain text value. -->
```
bryk: Is this TODO done?

src/app/frontend/secretdetail/togglehiddentext.scss
```diff
@@ -0,0 +1,34 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.kd-toggle-hidden-text-input {
+  border: 0 none;
+  font-size: 14px;
```
bryk: Can you import the variables.scss file here and use variables from there?

romlein:
@batikanu the eye on / off is potentially confusing for users, as there isn't good consistency around whether the icon shown should suggest [what will happen if clicked or be a label or what's currently shown](http://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan/1319#1319). Also, this is more closely aligned with the [Material spec for this UI element](https://material.google.com/components/buttons.html#buttons-toggle-buttons).

This is somewhat contingent on the use case though: how frequently will users be toggling visibility on and off, and what will the default be? If most users will choose to show all fields, I could see the more subtle eye-off icon as being less distracting and therefore more appropriate.

src/app/backend/resource/secret/secretdetail.go
batikanu: done.

src/app/frontend/secretdetail/detail.html
batikanu: done.

src/app/frontend/secretdetail/togglehiddentext.scss
batikanu: I've imported and used from them as much as I can use.

batikanu:
Now it looks like this. PTAL
secret

src/app/frontend/common/components/togglehiddentext/togglehiddentext.html
```diff
@@ -0,0 +1,52 @@
...
+
+    <!--<md-button  ng-show="!isActive" ng-click="isActive = !isAcitve" class="kd-toggle-eye-button-off md-icon-button md-mini" aria-label="Settings">-->
+        <!--<md-icon class="kd-toggle-eye-icon-off" md-font-set="material-icons"> remove_red_eye </md-icon>-->
+    <!--</md-button>-->
+
+
+    <!--<md-button ng-show="isActive" ng-click="isActive = isAcitve" class="kd-toggle-eye-button-on md-icon-button md-mini" aria-label="Settings">-->
+        <!--<md-icon class="kd-toggle-eye-icon-on" md-font-set="material-icons"> remove_red_eye </md-icon>-->
+    <!--</md-button>-->
+    <!--<i ng-show="isActive" ng-click="isActive = !isActive" class="material-icons kd-toggle-eye-icon-on">remove_red_eye</i>-->
```
rf232: I'm not a big fan of commented out code, can you clean this up?

i18n/messages-en.xtb
```diff
@@ -813,16 +813,25 @@
   <translation id="5847167541694942228" key="MSG_PETSETDETAIL_PETSETINFO_9" desc="The message says that that many pods are pending (pet set details page).">{{ $ctrl.petSet.podInfo.pending}} pending</translation>
   <translation id="7332879554461510589" key="MSG_PETSETLIST_PETSETCARDLIST_0" desc="Column header \'Kind\' of a Pet Set List">Kind</translation>
   <translation id="2816069801101820166" key="MSG_PETSETLIST_PETSETCARDLIST_1" desc="Pet set list header: name.">Name</translation>
+  <translation id="3262303717958393654" key="MSG_PETSETLIST_PETSETCARDLIST_1" desc="Pet set list header: name.">Name</translation>
```
rf232: Eek, lot of added duplications in this file. Not a problem with your code, but we might need to take a look at why this happens. Might be a problem in my i18n fix commit from PR #1386

src/app/frontend/common/components/togglehiddentext/togglehiddentext.html
batikanu: done.

i18n/messages-en.xtb
batikanu: Strange. I cleaned duplication translations of this pull request.

batikanu:
PTAL

PR #1307, opened 2016-10-05
Display common object meta for all resources

bryk:
1. Now all resources display name, namespace, labels and annotations (if
the property is set)
2. Commonized and removed a lot of code.


bryk:
@rf232 @digitalfishpond Can you review this?

bryk:
Thanks

Issue #1121, opened 2016-08-12
Pod state not correct


cheld:
#### Issue details
Pod state is not correct in UI

##### Environment

```
Dashboard version: master (12. Aug)
Kubernetes version: v1.4.0-alpha.2
```

##### Steps to reproduce
```
$ kubectl run test --image=debian -- echo hello
$ kubectl get pods
NAME                    READY     STATUS             RESTARTS   AGE
test-2172490464-elqbi   0/1       CrashLoopBackOff   18         1h
```
##### Observed result
UI claims pod is in "Running" state

##### Expected result
"Crashed"

bryk:
Interesting. Is CrashLoopBackOff status something new in 1.4 ?

cheld:
Hmmm...I would say I have seen at least the "CrashLoopBack" part before. Not sure if I have seen the suffix "Off", though

bryk:
Closing as stale.

cheld:
Reproducible in v1.4.0-alpha.3

maciaszczykm:
I guess we have to follow `kubectl` logic on backend side to fix it.

cheld:
An alternative (not so nice) would be to show something unspecific e.g. 'unkown', 'incorrect'  as state in the UI. Than we do not have to copy the logic which seems a bit fragile

bryk:
Ok, thanks for reopening!

batikanu:
I made some research about it. I think we can reflect the logic of the [resource printer](https://github.com/kubernetes/kubernetes/blob/64f5d9c3f4b3997b59638448c76e58a56328b984/pkg/kubectl/resource_printer.go#L677) in kubectl. I can take it if you approve the solution.

cheld:
I did not dig into the code. Maybe it is possible to extract a closed function (without side effects) that could be used by kubectl and Dashboard?

floreks:
I think main assumption of this issue is wrong. It assumes that dashboard relies on simplified data description and we are actually trying to reflect `kubectl describe` not `kubectl get`.

Let's compare:
`kubectl describe pod test-496409221-psuql` returns
```
NAME                   READY     STATUS             RESTARTS   AGE
test-496409221-psuql   0/1       CrashLoopBackOff   6          9m
```

`kubectl describe pod test-496409221-psuql` returns
```
Name:		test-496409221-psuql
Namespace:	default
Node:		floreks-celsius-w530-power/192.168.30.230
[...]

PR #1308, opened 2016-10-06
Show more detailed pod status on pod list page

IanLewis:
This PR is meant to address #1121. Changed status on the pod list page to show the same status as kubectl.

Future work needed:
- Logic for whether to show the pod as successful or not needs to be updated to incorporate `status.containerStatuses` and/or `status.conditions`.
- Messages on the pod list page need to be updated to reflect new success logic.
- More info on container status needs to be shown on the Pod detail page.
- Make use of future printer API work? (https://github.com/kubernetes/kubernetes/issues/29472)


src/app/backend/resource/pod/podcommon.go
```diff
@@ -30,14 +32,77 @@ func getRestartCount(pod api.Pod) int32 {
...
+				if container.State.Terminated.Signal != 0 {
+					reason = fmt.Sprintf("Signal:%d", container.State.Terminated.Signal)
+				} else {
+					reason = fmt.Sprintf("ExitCode:%d", container.State.Terminated.ExitCode)
+				}
+			}
+		}
+	}
+	if pod.DeletionTimestamp != nil {
+		reason = "Terminating"
```
bryk: Do you have thoughts on internationalizing this? Entire UI is translated, but messages from the server aren't. Any solutions to this?

src/app/backend/resource/pod/podcommon.go
```diff
@@ -30,14 +32,77 @@ func getRestartCount(pod api.Pod) int32 {
 	return restartCount
 }
 
+// getDisplayPods gets the status of the pod for display in a compact view like a list view. Logic is copied from kubectl ResourcePrinter implementation.
+func getDisplayStatus(pod api.Pod) string {
```
bryk: Can we have tests for this fuction? Looks like an easy candidate for a few unit tests.

src/app/backend/resource/pod/podcommon.go
IanLewis: Yah. This is a difficult one. Kubectl displays internal error names/strings rather than "messages" in the translatable sense. I'm not sure it's really translatable unless we break with kubectl and show something like "Error: %(reason)s" or something like that.

Perhaps that's a better approach?

bryk: Yes. I think that's best solution. With the new i18n framework you could do this in the HTML via:
`[[Error: {{$ctrl.errorReason}}|A description]]`

IanLewis:
Updated to have better messages that are translatable. Tests will be forthcoming.

bryk:
All right. Looks good. Please add tests now :)

IanLewis:
The linters... They hate me...

IanLewis:
@bryk Added tests and tamed the linters/compilers. PTAL.

PR #1311, opened 2016-10-06
Fix i18n bug for translations with non ascii chars

bryk:
Previously translations like [[Message|ą 'fff']] wouldn't get
translated, because non ascii and control chars were escaped and the
system couldn't find it the pureHtmlContent var. I've fixed this by
working on js-escaped HTML and then evaulating it at the end.


bryk:
@rf232 @olekzabl Can you review this?

PR #1322, opened 2016-10-11
Adding note about Heapster dependency

puja108:
Not sure if we might also want to add that it is sufficient to run it in standalone mode, so people don't go deploying InfluxDB and Grafana without actually needing them.


cheld:
lgtm. 

BTW: the fundamental problem is that kubeadm does not deploy heapster and dashboard by default (or easily)

CC @floreks, @kenan435

PR #1344, opened 2016-10-17
Start with showing more info about creator of a pod

rf232:
Show more info on the poddetail page about it's controller.

This PR is mostly a proof of concept with just implementation for pods belonging to a Job. please take a look at it and tell me if i should continue on the chosen way.

pod_creator


rf232:
PTAL @bryk @digitalfishpond :)

digitalfishpond:
Travis seems to be reporting a problem...

rf232:
Should be fixed with the new push (A)

-- 
Rob

src/app/backend/resource/pod/poddetail.go
```diff
@@ -112,11 +143,56 @@ func GetPodDetail(client k8sClient.Interface, heapsterClient client.HeapsterClie
...
+		return nil, err
+	}
+
+	events := <-channels.EventList.List
+	if err := <-channels.EventList.Error; err != nil {
+		return nil, err
+	}
+	reference := serializedReference.Reference
+	kind := reference.Kind
+	switch kind {
```
maciaszczykm: Hmm, I think we need more generic solution for this issue or at least split implementation to separate functions. I'll take a closer look on it today.

maciaszczykm: Hmm, I think we need more generic solution for this issue or at least split implementation to separate functions. I'll take a closer look on it today.

maciaszczykm:
It's very valuable feature :)

At the moment just tested to reply on your question regarding chosen way. Considering following view:

image

There is no information what kind of resource it is. In my opinion it's necessary in all cases.

Moreover, I'm wondering if it's possible to display it in other format, because currently it's displayed in table with only one entry inside. What about making it card similar to details card (available on top of every resource details view)?

I'll dig into code later on.

src/app/backend/resource/pod/poddetail.go
```diff
@@ -15,17 +15,23 @@
...
 
 	"github.com/kubernetes/dashboard/src/app/backend/client"
 	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
 	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/job/joblist"
 	"github.com/kubernetes/dashboard/src/app/backend/resource/metric"
 )
 
+const (
+	CreatedByAnnotation = "kubernetes.io/created-by"
```
maciaszczykm: @bryk I think it's good idea to update Go dependencies, because string like this are already added into core code: 

``` go
// CreatedByAnnotation represents the key used to store the spec(json)
// used to create the resource.
CreatedByAnnotation = "kubernetes.io/created-by"
```

It's likely, that we'll find more useful code.

src/app/backend/resource/pod/poddetail.go
```diff
@@ -45,13 +51,26 @@ type PodDetail struct {
...
 
 	// Metrics collected for this resource
 	Metrics []metric.Metric `json:"metrics"`
 }
 
+// Creator is a view of the creator of a given pod, in List for for ease of use
+// in the frontend logic.
+//
+// Has 'oneof' semantics on the non-Kind fields decided by which Kind is there
+type Creator struct {
```
maciaszczykm: s/Creator/Controller? Mainly because of `kubectl` parity.

src/app/backend/resource/pod/poddetail.go
maciaszczykm: I'm trying to find a solution to avoid switch here, but it seems there aren't any. What about splitting code to separate functions?

src/app/backend/resource/pod/poddetail.go
bryk: Maybe we spawn an issue and add a todo here to update the deps, so that Rob is unblocked?

src/app/backend/resource/pod/poddetail.go
```diff
@@ -45,13 +51,26 @@ type PodDetail struct {
...
+	Kind string `json:"kind"`
+
+	JobList *joblist.JobList `json:"joblist,omitempty"`
```
bryk: Should this be just a `*Job`? I think so. Is there anything that prevents us from doing so?

src/app/backend/resource/pod/poddetail.go
bryk: This looks fine to me. We could move entire switch to a separate function and that's all what I see.

maciaszczykm:
@bryk I agree, let's move dependencies update to another issue. Moving switch to another function is also good.

src/app/backend/resource/pod/poddetail.go
rf232: Ack, I'll move it to it's own function.

src/app/backend/resource/pod/poddetail.go
rf232: It was at least a lot easier to get the list view to work when sending a list with one element, will take a look if i can get around that. Probably also depends a bit on what @maciaszczykm thinks about display as 'list' vs display as resource card?

rf232:
Still to do:
Show what the kind is of the creator/controller (so either add column, or switch to some detailcard)

decide where the boundary between creator and controller exactly lies

more reviews?

src/app/frontend/poddetail/creatorinfo.html
```diff
@@ -0,0 +1,27 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<kd-content-card>
+  <kd-title>{{$ctrl.i18n.MSG_CREATOR_DETAILS_SUBTITLE}}</kd-title>
```
bryk: Can you use the "new" i18n framework here?

src/app/backend/resource/pod/poddetail.go
```diff
@@ -45,13 +53,26 @@ type PodDetail struct {
...
 	// Metrics collected for this resource
 	Metrics []metric.Metric `json:"metrics"`
 }
 
+// Creator is a view of the creator of a given pod, in List for for ease of use
+// in the frontend logic.
+//
+// Has 'oneof' semantics on the non-Kind fields decided by which Kind is there
+type Controller struct {
+	Kind string `json:"kind"`
```
bryk: Can you document this?

src/app/backend/resource/pod/poddetail.go
bryk: Ok, if this makes implementation easier, then good. 

@maciaszczykm Can you chime in here re showing this as a list row vs full details? I'm for showing this as a list row because all we want to do is to link to the creator, not show all details, since the user is looking at a pod here, not the creator.

src/app/frontend/poddetail/creatorinfo_component.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @final
+ */
+export default class CreatorInfoController {
+  /**
+   * @ngInject
+   */
+  constructor() {
+    /** @export {!Object} Initialized from a binding. */
```
bryk: Can you add this to backendApi.js and add `@type` info?

src/app/frontend/poddetail/creatorinfo_component.js
```diff
@@ -0,0 +1,50 @@
...
+    this.creator;
+
+    /** @export */
+    this.listResource = 'api/v1/fake';
```
bryk: Don't need this. You can omit the parameter in the list directive.

src/app/frontend/poddetail/creatorinfo.html
rf232: ack

src/app/frontend/poddetail/creatorinfo_component.js
rf232: ack

src/app/frontend/poddetail/creatorinfo_component.js
rf232: ack

src/app/backend/resource/pod/poddetail.go
rf232: or @romlein? a singleton list as a rich info link, or should we go for a resource card like look?

src/app/backend/resource/pod/poddetail.go
```diff
@@ -19,25 +19,57 @@ import (
...
+	k8sClient.ConfigMapsNamespacer
+	Apps() k8sClient.AppsInterface //Added this one
+	Autoscaling() k8sClient.AutoscalingInterface
+	Batch() k8sClient.BatchInterface
+	Extensions() k8sClient.ExtensionsInterface
+	Rbac() k8sClient.RbacInterface
+	Discovery() discovery.DiscoveryInterface
+	Certificates() k8sClient.CertificatesInterface
+}
+// End TODO:
```
rf232: We really need to update the vendor/k8s.io code now to get PetSets all around in the api. Embedding this interface here makes me trick the go type system for a bit, but this won't help for the fakeClient used by testing

rf232: cc @bryk

rf232:
not finished for review yet. Did the refactor part, and the other 'creator/controller' types, but not yet the added column in the tables to show the type of the controller.

src/app/backend/resource/pod/poddetail.go
bryk: Do this in a separate PR and then rebase? I can guide you how it is done.

rf232: PR #1378 exists now and does the update vendor packages part

rf232:
Will merge master back in when the vendor packages are updated and will write tests then. Think I'm feature complete though

rf232:
@bryk: Can you take a look at the i18n changes in my last commit, I think it may be a bit problematic, because I added translatable tokens in the template and now the key 'follow_numbers' are shifted for some and thus the japanese translation will get lost again. I think we should open an issue for that, do you agree?

bryk:
> @bryk: Can you take a look at the i18n changes in my last commit, I think it may be a bit problematic, because I added translatable tokens in the template and now the key 'follow_numbers' are shifted for some and thus the japanese translation will get lost again. I think we should open an issue for that, do you agree?

Yes, that's a huge problem. Can you open an issue for this?

rf232:
This should be ready for merge now after travis is happy

Issue #1309, opened 2016-10-06
FAQ about dashboard setup


IanLewis:
We get a lot of issues related to the dashboard not working in new clusters which almost invariably turn out to be improperly set up Kubernetes clusters rather than issues with the dashboard. We should have a FAQ or suggestions when installing the dashboard that users can reference for guidance about where to create issues.

bryk:
My idea was to start our installation guide with references to tools/providers that we know that work. And only in the advanced section explain how to run the UI on a hand-crafted cluster. 

E.g., 
```To access the UI open http://master/ui on your cluster provider (e.g., GKE, stackpoint, Foo) or use minikube to create a new cluster.

(advanced)
To install the UI on your claster do $Foo and $Bar. 
```

IanLewis:
Ok. I think we should at least do what you suggested. I think it might also be good to call out network related issues as most likely being cluster related rather than dashboard related. Maybe a "Troubleshooting" section?

cheld:
I will start with a trouble shooting section regarding service account and kubeconfig file. We can add other topics later.

maciaszczykm:
It is already available in project's wikis.

PR #1348, opened 2016-10-20
Link from crash message to the troubleshooting guide

bryk:
Re #1309

### DIFF
--- a/docs/user-guide/troubleshooting.md
+++ b/docs/user-guide/troubleshooting.md
@@ -1,3 +1,11 @@
+<!--
+-----------------NOTICE------------------------
+This file is referenced in code as
+https://github.com/kubernetes/dashboard/blob/master/docs/user-guide/troubleshooting.md
+Do not move it without providing redirects.
+-----------------------------------------------
+-->
+
 # Troubleshooting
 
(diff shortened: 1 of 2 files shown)


bryk:
@rf232 @cheld @IanLewis Please take a look.

docs/user-guide/troubleshooting.md
```diff
@@ -1,3 +1,11 @@
+<!--
+-----------------NOTICE------------------------
+This file is referenced in code as
+https://github.com/kubernetes/dashboard/blob/master/docs/user-guide/troubleshooting.md
+Do not move it without providing redirects.
```
IanLewis: Are redirects possible on github?

rf232: I think the project wiki is intended for these usecases, but not really sure how to set that up nicely

IanLewis:
Thinking about it we could leave a file with a link to the new file if it's
ever moved. Not ideal but probably better than moving to the wiki.

rf232:
Agreed, switching docu to wiki is not something we should do just for this usecase, and a non automatic HTTP 301 file sounds like a better solution :)

PR #1351, opened 2016-10-20
Use the incluster heapster by default

rf232:
also update travis config to still use a seperate heapster instance

compare PR #1163

### DIFF
--- a/build/conf.js
+++ b/build/conf.js
@@ -21,7 +21,7 @@ import path from 'path';
  * Load the i18n and l10n configuration. Used when dashboard is built in production.
  */
 let localization = require('../i18n/locale_conf.json');
-let util = require('gulp-util');
+let util = import('gulp-util');
 
 /**
  * Base path for all other paths.
--- a/build/conf.js
+++ b/build/conf.js
@@ -15,8 +15,8 @@
 /**
  * @fileoverview Common configuration constants used in other build/test files.
  */
-import path from 'path';
 import gulpUtil from 'gulp-util';
+import path from 'path';
 
 /**
  * Load the i18n and l10n configuration. Used when dashboard is built in production.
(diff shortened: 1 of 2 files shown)


rf232:
PTAL @floreks

build/conf.js
```diff
@@ -21,6 +21,7 @@ import path from 'path';
  * Load the i18n and l10n configuration. Used when dashboard is built in production.
  */
 let localization = require('../i18n/locale_conf.json');
+let util = require('gulp-util');
```
floreks: Import can be used instead here. Require can be used when we need to 'import' some custom file outside of `node_modules`.

rf232: ack

floreks: Sorry if I haven't clearly explained it. Look few lines up at line 18. `import path from 'path'`. I've wanted to move it there to common imports and use this syntax :)

floreks:
LGTM.

@bryk @cheld any comments on that?

floreks:
`gulp format-javascript` should fix travis.

bryk:
Looks cool, thanks!

batikanu:
@rf232 gulp local-up-cluster runs heapster in standalone mode and gulp serve goes to in-cluster heapster by default. Is it per purpose that developer has to pass heapster api as parameter when using gulp serve?

rf232:
No it is not the purpose of this changelist to make developer workflow more tedious. Using the incluster one by default was done because we thought/think most developers use a dev cluster with heapster in it. I know for a fact that @bryk and me have that. And @floreks asked for that default on my preview of this change in #1163. 

If we know a good way to detect that we have done this `gulp local-up-cluster` to switch the default I'd be happy with that though. (Or change gulp local-up-cluster to run heapster in the cluster :))

batikanu:
Yes currently [the getting-started guide](https://github.com/kubernetes/dashboard/blob/master/docs/devel/getting-started.md) is misleading after this change. If the developer follows the guide, he will wait forever to finish gulp local-up-cluster because of hidden parameter which has to be passed for heapster. Shall I create an issue for that?

rf232:
Yes please

-- 
Rob

PR #1359, opened 2016-10-24
Previously some cards had 20px gutter and some 40px (due to

bryk:
margin-collapse CSS property). Some views had page-bottom margin, some
not.

This PR fixes the issue by introducing page-filled configuration
framerowk. Pages that need to fill app content can configure it (logs
view). Other pages will have any height and bottom margin of 4 baseline
grids.


bryk:
@batikanu @rf232 Can you review this?

Issue #1361, opened 2016-10-24
XSS risk on ?namespace= get parameters


rf232:
when giving a get paramer of `?namespace=%3Cimg%20src%3Dx%20onerror%3Dalert(document.domain)%3E` to the dashboard we see a broken image and get a popup. We should fix this and check if more such XSS vulns are there

rf232:
okay, the namespaceParam is the only place we touch the $location.search or queryParams at all, the rest of our router code is all guarded agains xss by angular

rf232:
closed by PR #1362

PR #1362, opened 2016-10-24
Sanitize our access to $location.search().namespace

rf232:
quick way to make sure we only show sanitized namespaces, perhaps research if this is intended behaviour of this select component and if we should do more.

#1361


rf232:
hmm, quick manual testing shows that this works only half the times, (in a nice oscillating sequence)

bryk:
> hmm, quick manual testing shows that this works only half the times, (in a nice oscillating sequence)

You say that the content is not sanitized sometimes?

rf232:
yes, or at least later than needed.

bryk:
Let me test this...

rf232:
PTAL

PR #1363, opened 2016-10-24
Add pre-commit hooks

floreks:
Often travis builds fail due to file formatting. This is a little help to automatically check that before committing our changes. Just as a start I've added file formatting checks but we could extend that and i.e. execute tests.

zrzut ekranu z 2016-10-24 15-06-29


bryk:
Merging.

PR #1367, opened 2016-10-25
Strings update with new i18n and text conventions pt.1

digitalfishpond:
Part 1 of updating strings for the new i18n framework with correct text conventions (up to and including 'deploymentlist' package).

### DIFF
--- a/src/app/frontend/common/components/labels/labels.html
+++ b/src/app/frontend/common/components/labels/labels.html
@@ -23,7 +23,8 @@ limitations under the License.
   <div class="kd-chips kd-chips-switch" ng-show="labelsCtrl.isMoreAvailable()"
        ng-click="labelsCtrl.switchLabelsView()">
     {{labelsCtrl.isShowingAll() ?
-      labelsCtrl.i18n.MSG_LABELS_SHOW_LESS_TOOLTIP : labelsCtrl.i18n.MSG_LABELS_SHOW_ALL_TOOLTIP}}
+      '[[show less labels|Tooltip on the "show less" button for the labels of any kubernetes resource. Usually appears on the resource details pages.]]' :
+      '[[show all labels|Tooltip on the "show all" button for the labels of any kubernetes resource. Usually appears on the resource details pages.]]' }}
   </div>
 </div>
 <div ng-hide="::labelsCtrl.labels">-</div>
--- a/src/app/frontend/common/resource/deleteresource.html
+++ b/src/app/frontend/common/resource/deleteresource.html
@@ -14,18 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<md-dialog aria-label="{{::$ctrl.i18n.MSG_DELETE_RESOURCE_DIALOG_TITLE}}" layout="column">
+<md-dialog aria-label="[[Delete a {{::$ctrl.resourceKindName}}|Title for a delete resource dialog.]]" layout="column">
   <md-dialog-content layout-padding>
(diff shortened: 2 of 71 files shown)


src/app/frontend/common/components/labels/labels.html
```diff
@@ -23,7 +23,9 @@
   <div class="kd-chips kd-chips-switch" ng-show="labelsCtrl.isMoreAvailable()"
        ng-click="labelsCtrl.switchLabelsView()">
     {{labelsCtrl.isShowingAll() ?
-      labelsCtrl.i18n.MSG_LABELS_SHOW_LESS_TOOLTIP : labelsCtrl.i18n.MSG_LABELS_SHOW_ALL_TOOLTIP}}
+      '[[show less labels|Tooltip on the "show less" button for the labels of any kubernetes resource. ' +
```
bryk: Hmm... I think that `'` are wrong here. You should not add the `' +` part, I think.

src/app/frontend/common/resource/deleteresource.html
```diff
@@ -14,18 +14,18 @@
...
 -->
 
-<md-dialog aria-label="{{::$ctrl.i18n.MSG_DELETE_RESOURCE_DIALOG_TITLE}}" layout="column">
+<md-dialog aria-label="[[Delete a {$ctrl.resourceKindName}|Title for a delete resource dialog.]]" layout="column">
   <md-dialog-content layout-padding>
-    <h4 class="md-title">{{::$ctrl.i18n.MSG_DELETE_RESOURCE_DIALOG_TITLE}}</h4>
+    <h4 class="md-title">[[Delete a {$ctrl.resourceKindName}|Title for a delete resource dialog.]]</h4>
     <div>
-      <kd-i18n message="{{::$ctrl.i18n.MSG_DELETE_RESOURCE_CONFIRM_QUESTION}}"></kd-i18n>
+      [[Are you sure you want to delete {$ctrl._resourceKindName} <i>{$ctrl._objectMeta.name}</i> in namespace <i>{$ctrl._objectMeta.namespace}</i>?|Confirmation question, appears before any dashboard resource will be deleted.]]
```
bryk: I love this now! :)

src/app/frontend/common/resource/deleteresource_controller.js
```diff
@@ -40,7 +40,10 @@ export class DeleteResourceController {
     this.resource_ = $resource;
 
     /** @export */
-    this.i18n = i18n(resourceKindName, objectMeta);
+    this._resourceKindName = resourceKindName;
```
bryk: Why prefixing with `_`? We've never done this before?

digitalfishpond:
I can do some transferring of the translations (maybe as a separate PR at the end of everything?), but I will definitely need @Yamagami-Takuma to check the result!

digitalfishpond:
PTAL

src/app/frontend/deploy/createnamespace_controller.js
```diff
@@ -67,10 +67,8 @@ export default class NamespaceDialogController {
     /** @export {!angular.FormController} */
     this.namespaceForm;
 
-    /**
-     * @export
-     */
-    this.i18n = i18n;
+    /** @export */
+    this.maxLength = 64;
```
rf232: previous message said 63, while indeed the spec says it is 64, why change it now hidden in this huge refactor CL?

digitalfishpond: Agreed, particularly since the variable already exists a few lines up. I have removed this variable altogether in a new commit.

Yamagami-Takuma:
@digitalfishpond  
About ja.xtb, I will update translations :)

PR #1369, opened 2016-10-25
Fix to i18n framework (calling variables)

digitalfishpond:
Make it possible to call values of variables from the controller within html i18n.


build/i18n.js
```diff
@@ -100,7 +100,7 @@ export function processI18nMessages(file, minifiedHtml) {
         let exec = regexpClone(I18N_REGEX).exec(match);
         // Default to no description when it is not provided.
         let desc = (exec[2] || '(no description provided)').trim();
-        return {text: exec[1], desc: desc, original: match};
+        return {text: exec[1].replace('{$', '{ $'), desc: desc, original: match};
```
bryk: Can you add an explanation comment why this is needed?

bryk:
LGTM, please merge when travis is happy.

PR #1370, opened 2016-10-25
Remove unused source attribute from translation files

floreks:
This will save us some space as `source` attribute is not needed.

I'm not sure if `sortxtb.xslt` is the right place to do this kind of clean up, but it's only 1 line.


build/sortxtb.xslt
```diff
@@ -1,10 +1,9 @@
 <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
-  <xsl:output method="xml" indent="yes" omit-xml-declaration="no" encoding="UTF-8" />
+  <xsl:output method="xml" indent="yes" omit-xml-declaration="no" encoding="UTF-8"/>
   <xsl:strip-space elements="*"/>
   <xsl:template match="translationbundle">
-      <!-- XSLT doesn't support adding doctype without PUBLIC or SYSTEM annotation, so add it to the doc raw -->
-      <xsl:text disable-output-escaping="yes">&lt;!DOCTYPE translationbundle&gt;
-</xsl:text>
+    <!-- XSLT doesn't support adding doctype without PUBLIC or SYSTEM annotation, so add it to the doc raw -->
+    <xsl:text disable-output-escaping="yes">&lt;!DOCTYPE translationbundle&gt;</xsl:text>
```
rf232: I added this newline to keep the formatting the same as in the old case, but am fine with removing it

floreks:
> I guess you checked that the xtb generator still recognizes the file for its reuse logic?
Yep, I've regenerated them and checked on `serve:prod`. Everything works as intended. Files are correctly picked up.

PR #1371, opened 2016-10-25
Custom handle last-applied-configuration annotations

bryk:
This is a custom handler for the
kubectl.kubernetes.io/last-applied-configuration annotation that you
often see on your objects. This is v1 and the popup window requires some
follow-up polish.

selection_139

selection_140

### DIFF
--- a/src/app/frontend/common/components/annotations/annotations_component.js
+++ b/src/app/frontend/common/components/annotations/annotations_component.js
@@ -15,18 +15,14 @@
 import LabelsController from '../labels/labels_controller';
 
 /**
- * Returns directive definition for annotations.
- *
- * @return {!angular.Directive}
+ * @return {!angular.Component}
  */
-export default function annotationsDirective() {
-  return {
-    controller: LabelsController,
-    controllerAs: 'annotationsCtrl',
-    templateUrl: 'common/components/annotations/annotations.html',
-    scope: {},
-    bindToController: {
-      'labels': '=',
-    },
-  };
-}
+export const annotationsComponent = {
+  controller: LabelsController,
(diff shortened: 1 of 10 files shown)


src/app/frontend/common/components/annotations/annotations_component.js
```diff
@@ -15,18 +15,14 @@
 import LabelsController from '../labels/labels_controller';
 
 /**
- * Returns directive definition for annotations.
- *
  * @return {!angular.Directive}
```
rf232: is this still a directive now?

src/app/frontend/common/components/annotations/lastappliedconfiguration_component.js
```diff
@@ -0,0 +1,57 @@
...
+    this.mdDialog_ = $mdDialog;
+  }
+
+  /** @export */
+  openDetails() {
+    // TODO make this dialog prettier
+    let dialog = this.mdDialog_.alert()
+                     .title(i18n.MSG_CONFIG_DIALOG_TITLE)
+                     .textContent(`kubectl.kubernetes.io/last-applied-configuration: ${this.value}`)
+                     .ok(i18n.MSG_CONFIG_DIALOG_OK_ACTION);
```
rf232: 'OK' what, for me OK feels like something is going to happen, unless the dialog text is very clearly purely informational. I think we need to make sure that people don't think the OK button will roll us back to that config.

rf232: Sounds like a clear solution

rf232:
> How about "close"? 
Sounds like a clear solution

bryk:
PTAL

PR #1375, opened 2016-10-26
docs(frontend): Fixed some i18n constants/docstrings typos

nicr9:
Found a few of these when I was working on helm integration and I thought it'd be nice to fix em separately.


i18n/messages-ja.xtb
```diff
@@ -759,12 +759,12 @@
...
+  <translation id="7172190985346183509" key="MSG_RC_DETAIL_PODS_FAILED_LABEL" desc="The message says how that many pods have failed (replication controller details page)."><ph name="PODS_COUNT"/> の失敗</translation>
   <translation id="8107580983200519386" key="MSG_RC_DETAIL_PODS_LABEL" desc="Label 'Pods' for the pods in a replication controller on its details page.">ポッド</translation>
-  <translation id="6107745197793205489" key="MSG_RC_DETAIL_PODS_PENDING_LABEL" desc="The message says that that many pods are pending (replication controller details page)."><ph name="PODS_COUNT"/> 待機中</translation>
-  <translation id="5308765422355681699" key="MSG_RC_DETAIL_PODS_RUNNING_LABEL" desc="The message says that that many pods are running (replication controller details page)."><ph name="PODS_COUNT"/> 実行中</translation>
+  <translation id="6107745197793205489" key="MSG_RC_DETAIL_PODS_PENDING_LABEL" desc="The message says how that many pods are pending (replication controller details page)."><ph name="PODS_COUNT"/> 待機中</translation>
```
floreks: How about removing second 'that'? :)

nicr9: Oh damn, I thought I got em all!

bryk:
Hmm... Don't we need to fix this in the HTML and JS source files? Not in the .xtb file that would be overridden.

nicr9:
As far as I can tell the vast majority of these changes (the "that that" -> "how")

floreks:
Oh... yes, you should actually update all htmls and just run gulp task to regenerate translations. I've missed that somehow.

As I've briefly checked our html files do not have these changes applied.

rf232:
I see you added commit #f19f770, and that you did a lot of edits to templates there. Is this the 'full' list of changes you also did to the xtb files, or should we expect another commit?

As soon as you are ready please say something in the line of PTAL :)

nicr9:
Hey @floreks @rf232 , sorry about the delay!

I ran `gulp generate-xtbs` (I think that's the right one?) and it didn't have any complaints or make any changes to the xtbs that were there so I think we're good to go!

PTAL when you're ready

PR #1377, opened 2016-10-27
Build and push HEAD image on every build on master 

bryk:
Also rename canary to head for images that we build. This name conveys
better what we actually do.

This bases on @floreks work that he did on this topic. With a couple of alterations from me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,12 +20,14 @@
 # According to the documentation: http://docs.travis-ci.com/user/docker/
 # sudo is required to enable docker in Travis CI.
 sudo: true
+
 # Cache downloaded Node.JS modules & Bower frontend dependencies for faster builds.
 cache:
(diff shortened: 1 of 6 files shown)


floreks:
@googlebot I'm ok with these commits being contributed to Google.

.travis.yml
```diff
@@ -20,12 +20,14 @@
...
 # sudo is required to enable docker in Travis CI.
 sudo: true
+
 # Cache downloaded Node.JS modules & Bower frontend dependencies for faster builds.
 cache:
   directories:
     - node_modules
     - bower_components
     - .tools
+    - "$HOME/google-cloud-sdk/"
```
floreks: Not needed if we are not pushing to gcr also.

bryk:
Done. PTAL

Issue #1346, opened 2016-10-19
Incorrect values for memory usage being displayed 


andrewmclagan:
##### Environment

Default GKE cluster installation.

```
Dashboard version: 0.1.4
Kubernetes version: 1.4.3 
Operating system: Default GKE cluster installation.
Node.js version: Default GKE cluster installation.
Go version: Default GKE cluster installation.
```
##### Steps to reproduce

View `Workloads >> Pods` then sum the total memory usage and compare that to the graphs. On our system there is a large disparity. 
##### Expected result

Memory usage for pods should sum to the total in the graphs?
##### Comments

See attached screen grab 
![screen shot 2016-10-19 at 3 52 39 pm](https://cloud.githubusercontent.com/assets/1073259/19506087/1a4e7c56-9614-11e6-9dcb-0cf9c22ac5df.png)

bryk:
Is this something that always happens always, or just sometimes? 

Can you paste the output of `http://<master-IP>/ui/api/v1/pod/default`? (you can obfuscate everything that is not about metrics).

andrewmclagan:
Not sure if im reading the endpoint correctly, although that redirects to:

`/api/v1/proxy/namespaces/kube-system/services/kubernetes-dashboard/#/workload?namespace=default`

Being the kubernetes bash not an API endpoint. Is that correct? If so i can dump that info

bryk:
The full endpoint is: `/api/v1/proxy/namespaces/kube-system/services/kubernetes-dashboard/api/v1/pod/default`. Use it for the dump.

andrewmclagan:
```json
{
  "listMeta": {
   "totalItems": 11
  },
  "pods": [
   {
    "objectMeta": {
     "name": "es-client-2600856264-0bubb",
     "namespace": "default",
     "labels": {
      "name": "elasticsearch",
      "pod-template-hash": "2600856264",
      "role": "client"
     },
     "annotations": {
      "kubernetes.io/created-by": "{\"kind\":\"SerializedReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"ReplicaSet\",\"namespace\":\"default\",\"name\":\"es-client-2600856264\",\"uid\":\"27094f41-9a7b-11e6-8a15-42010af000c6\",\"apiVersion\":\"extensions\",\"resourceVersion\":\"633398\"}}\n",
      "kubernetes.io/limit-ranger": "LimitRanger plugin set: cpu request for container es-client"
     },
[...]

PR #1381, opened 2016-10-28
Make pod metrics sum to aggregate resouce usage graphs

bryk:
The issue was with different metrics downloaded in the list and in the graphs. The other problem was that "current usage" was off by 15 mins.

Re #1346


bryk:
@rf232 If you happen to do a release, try to include this PR there :)

Issue #1384, opened 2016-10-31
i18n framework relabels translations


rf232:
#### Issue details
When adding new translations to a template and later in the file translations already exist the keys will be renumbered, and thus invalidate existing translations

##### Steps to reproduce
- Add a new translatable key to a template file and
- regenerate i18n/messages 

##### Observed result
new translation key is inserted in the middle of the list and all later ones are shifted up

##### Expected result
new translation key gets numbered after existing translations.

cheld:
CC @taimir, @digitalfishpond

taimir:
@rf232 just to confirm the problem, you are saying that if in a template file we already have 2 translated messages:

```
MSG_EXAMPLE_1 = "foo" ----> messages-ja.xtb has translation for  MSG_EXAMPLE_1
MSG_EXAMPLE_2 = "bar" ----> messages-ja.xtb has translation for  MSG_EXAMPLE_2
```

and now I were to introduce a message in the beginning of the file, those existing two would get renamed (the index will shift by one):

```
MSG_EXAMPLE_1 = "foobar" ----> new message
MSG_EXAMPLE_2 = "foo"
MSG_EXAMPLE_3 = "bar"
```
and this would wrongly lead to "foobar" receiving the old translation of "foo", "foo" receiving the old translation of "bar" and "bar" having no translation whatsoever (even though it did not change).

That should be about it, right? And then the solution is to just append the new message at the end of the list (to avoid the shifts).

rf232:
the XTB generator is a bit smarten than that, so the result is that all three have no translations (it 'throws away' old translation if the base english text changes.)

If you look at my PR that is basically what I do, but in the old case we didn't know anything about the previous iteration of the messages file and thus looked at the template file entirely fresh. So the code has some more logic to it than perhaps you first might expect

taimir:
Ah ok, that was the second possibility. So in the current state nothing is translated wrongly, but we are losing a ton of valid translations each time some new text is introduced in the beginning of a template.

rf232:
Indeed

PR #1386, opened 2016-11-02
Reuse existing translation keys to prevent renumbering

rf232:
Load in the existing english translations and see if we can match existing translations to tokens we find. If we find a match we will reuse the 'old' key instead of just renumbering

Fixes: #1384


rf232:
PTAL @bryk, @digitalfishpond

PR #593, opened 2016-03-30
Remove sauce access keys from travis config

bryk:
With that tests on sauce will not run on devs' forks, which will
hopefully make them less flaky.

I've added the key to travis config variable and made publicly
available.

Partially addresses: #494

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/dashboard/593)

PR #1388, opened 2016-11-03
Fix travis break

rf232:
first commit fixes the break (hopefully, don't have IE11 around to test it :))

second commit enables the saucelabs to also run on pull requests, so we can see if this actually fixes the break

### DIFF
--- a/src/test/frontend/poddetail/creatorinfo_directive_test.js
+++ b/src/test/frontend/poddetail/creatorinfo_directive_test.js
@@ -50,7 +50,7 @@ describe('CreatorInfo directive', () => {
     ['PetSet', 'pet-set'],
   ]);
 
-  for (let [kind, directive] of kind2directive) {
+  kind2directive.forEach((directive, kind) => {
     it(`should render a ${kind} as a kd-${directive}-card-list`, () => {
       // given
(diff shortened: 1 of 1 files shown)


rf232:
@olekzabl, @digitalfishpond, @maciaszczykm 

TIL: we run our tests on IE11, but not on Pull Requests, only on commits in master. Apparently my previous CL broke that. here a PR that fixes the tests

the second commit in the PR is a quick hack to get the IE11 tests to run on this PR, I'm not sure if we can keep this change or not. I don't know the details of why we don't use this on all PRs but that might be a different discussion, so I'm more than happy to revert that one before merging.

maciaszczykm:
@rf232 In the past we had a lot of problems with flaky tests from sauce labs and we had to temporarily disable them on development banches. @bryk in pull request #593 did the change, that might get you more insight.

We can test if the issue still occurs.

Also +1 for first commit if Travis will pass.

rf232:
grmbl, now unittests work even in IE11, but somehow the protractor tests (integration tests) fail to start the browser now.

rf232:
run succeeded with ie11, removed second commit from pull request, ready for merging I think

Issue #1223, opened 2016-09-13
Messages consistency


maciaszczykm:
There still is some inconsistency across Dashboard's messages. Take a look on following screenshot:

![](https://cloud.githubusercontent.com/assets/1159999/18435351/d9e01d98-78f2-11e6-9d69-f149abdf7f9c.png)

Message title shouldn't end with dot, but following message should. Like in Inbox. Now it's inconsistent.

Thing to consider is naming of Kubernetes' resources. Are we using capital letters everywhere? Do we consider resources like event, endpoint etc. as Kubernetes'? We should decide it here and then fix all messages.

bryk:
@maciaszczykm Can you document guidelines on how we should write messages? 

I can give some input, based on my experience with other UI projects. E.g., when to use sentences, when to end with dot, etc. 

cc @Lukenickerson @romlein

maciaszczykm:
@bryk Sure, let's continue discussion and I'll create summary of it.

maciaszczykm:
I'll create short document with guidelines for the repository and after that maybe also I will be able to refactor existing messages.

maciaszczykm:
The next step is to review all messages across the app and fix all issues. We can also apply new localization framework everywhere.

bryk:
@digitalfishpond Can you go over this? To apply i18n framework and go over messages?

digitalfishpond:
Yep, it's what I'm currently doing :)

Issue #1025, opened 2016-07-14
Kubernetes Dashboard shows "???" over the places


mumoshu:
#### Issue details

Kubernetes Dashboard shows "???" over the places.

(I'm replicating the [according issue in the minikube repo](https://github.com/kubernetes/minikube/issues/326) here because I'm not sure where is this issue coming from or have no idea how to drill down to the root cause)

##### Environment

minikube v0.5.0

```
Dashboard version: v1.1.0
Kubernetes version: v1.3.0
Operating system: boot2docker running on minikubeVM(VirtualBox)
```

##### Steps to reproduce

With minikube v0.5.0 installed:

```
$ minikube start
$ minikube dashboard
```

##### Observed result

As the title says, all the texts other than kubernetes objects' names are shown as '?????'

![image](https://cloud.githubusercontent.com/assets/22009/16828112/e6527918-49c7-11e6-80f5-b05e0d8bebe0.png)

##### Expected result

There are meaningful texts.

bryk:
Hah, so this is funny. From you github profile I assume that you are from Japan. Our UI is translated into Japanese and when it detects that your default language is JA, then it switches to Japanese strings.

You can quickly fix this by changing your preferred browser lang to English. 

The question is why you have `??` instead of Japan characters. Do you maybe miss some fonts? Do you usually see Japanese characters on other websites?

bryk:
cc @taimir

mumoshu:
@bryk Thanks for taking your time.
Yes, I'm from Japan and switching preferred lang on my browser to English did fix the issue, like this:

![image](https://cloud.githubusercontent.com/assets/22009/16831043/b83bfbf6-49dd-11e6-88bd-1b13558e3edf.png)

I guess it isn't a fonts related issue because:

* I usually see Japanese characters without any problem on other websites.
* looking into HTML sources also shows many `??`s:

![image](https://cloud.githubusercontent.com/assets/22009/16831137/35dd3e44-49de-11e6-9bf2-0c9e9e83e72a.png)

taimir:
OK, I'm on it. Let's see what causes this, `serve:prod` definitely displays the japanese text correctly so I suspect it might be a packaging issue.

bryk:
@taimir Can you run this via minikube and see what's happened

taimir:
Mhm, I'll try it out.

taimir:
@bryk
Hey, sorry for the slight delay - I wanted to double check everything.

[...]

PR #1391, opened 2016-11-04
s/show less foos/show fewer foos/g

anguslees:
("labels" and "attributes" are countable nouns)

### DIFF
--- a/i18n/messages-en.xtb
+++ b/i18n/messages-en.xtb
@@ -61,9 +61,11 @@
   <translation id="5675030862734752753" key="MSG_COMMON_COMPONENTS_ACTIONBAR_ACTIONBAR_1" desc="Label for global action bar create button.">Create</translation>
   <translation id="1745702155571319443" key="MSG_COMMON_COMPONENTS_ANNOTATIONS_ANNOTATIONS_0" desc="Friendly name of the kubernetest.io/created-by annotation">Created by:</translation>
   <translation id="820028802315082613" key="MSG_COMMON_COMPONENTS_ANNOTATIONS_ANNOTATIONS_1" desc="Text on the button to show less annotations">show less annotations</translation>
+  <translation id="1417978160910275828" key="MSG_COMMON_COMPONENTS_ANNOTATIONS_ANNOTATIONS_1" desc="Text on the button to show fewer annotations">show fewer annotations</translation>
   <translation id="6884163988875905686" key="MSG_COMMON_COMPONENTS_ANNOTATIONS_ANNOTATIONS_2" desc="Text on the button to show all the annotations">show all annotations</translation>
   <translation id="6935188165410090417" key="MSG_COMMON_COMPONENTS_ANNOTATIONS_LASTAPPLIEDCONFIGURATION_0" desc="Label for the kubectl.kubernetes.io/last-applied-configuration annotation">last applied configuration</translation>
   <translation id="2631421290135072028" key="MSG_COMMON_COMPONENTS_LABELS_LABELS_0" desc="Tooltip on the &quot;show less&quot; button for the labels of any kubernetes resource. Usually appears on the resource details pages.">show less labels</translation>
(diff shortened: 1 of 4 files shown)


digitalfishpond:
@anguslees the strings need to be changed in the code. The xtb files are generated automatically from the html and JS files. I am currently updating all the strings in the application as part of issue #1223 and I have already planned to include this change in my next PR.

maciaszczykm:
Yeah, my mistake, that's true @digitalfishpond. Totally forgot about auto-generation of XTB entries.

digitalfishpond:
Tell you what, I'll remove my change in my current work, and you can update your pool request in the following way: revert the change to messages-en.xtb, and change the strings directly in frontend/common/components/annotations.html and frontend/common/components/labels.html, then run gulp build and you should see the messages-en.xtb and messages-ja.xtb files changed automatically.

anguslees:
(ack, will do.  thanks!)

digitalfishpond:
Hi @anguslees, any progress on this? Can I help you further in any way?

anguslees:
When I run `gulp build` (actually running it in the container that gets produced by `./build/run-gulp-in-docker.sh`), `messages-en.xtb` contains the expected change but `messages-ja.xtb` has *all* the japanese characters replaced with question marks.  Presumably it's some utf8 encoding issue.  Is there some trick to regenerating this file correctly?

digitalfishpond:
@anguslees , unfortunately, we don't have any automatic translation generation. Our esteemed colleagues in Japan provide the translations manually. So you can just push your changes as they are and we can update the translation subsequently.

digitalfishpond:
Little clarification: do you mean all the japanese characters from _every_ translation in the entire file? in which case don't push, and I'll try to find a solution!

digitalfishpond:
Ok, if it is all the Japanese characters in the entire file, the first step would be to check the encoding of your IDE which should be UTF-8. If you are using Webstorm (IntelliJ), you can find that under File>Settings>Editor>File Encodings. let me know how you get on.

anguslees:
Yes, every japanese character in the entire file was `???`.  I never edited `messages-ja.xtb`, this was just me looking at what output came from `gulp build`, and comparing before/after.

digitalfishpond:
Ah. So it's probably a good idea to fix this so that you don't have similar problems in the future if you want to add or change a string. Ok, my first thought was that maybe the IDE you are using was not using UTF8 encoding, but I just tried building using US-ASCII and then again with ISO-8859-1, and the translations remained intact. 
@taimir @bryk - Any possible ideas why @anguslees ' build is generating a messages-ja.xtb file with every single Japanese character replaced with a question mark?

taimir:
> When I run gulp build (actually running it in the container that gets produced by ./build/run-gulp-in-docker.sh), messages-en.xtb contains the expected change but messages-ja.xtb has all the japanese characters replaced with question marks. 

@bryk If I remember correctly there was a similar issue a couple of months back, the problem was a user was seeing `???` instead of japanese in the production version of dashboard. It turned out that the build server where the dashboard releases were put together was missing localization. I assume this is the same issue, probably something must be installed in the docker container where `gulp` is run. I'll try to find the issue I'm referring to.

taimir:
There it is: https://github.com/kubernetes/dashboard/issues/1025. In this case, I suspect the problem is that the gulp container has no locale support.

bryk:
@taimir @digitalfishpond I remember this happening on one of my machines. I dont know why. Maybe some packages weren't installed. 

@digitalfishpond You are maintainer of this repo. You can "Add more commits by pushing to the patch-1 branch on anguslees/dashboard.". This means that you can fetch this commit, rebuilt locally, push to anguslees/dashboard and merge. Can you do this?

taimir:
@anguslees @digitalfishpond This is just a wild guess, but can one of you guys try adding the following:

```
apt-get install -y locales
locale-gen ja_JP.UTF-8
```

to the Dockerfile for the build container: [this one](https://github.com/taimir/dashboard/blob/master/build/Dockerfile). And then just run `build/run-gulp-in-docker.sh` as usual. Just make sure the build container image is rebuilt.

Maybe that could help...

PR #1397, opened 2016-11-07
Part 3 of updating strings for the new i18n framework

### LINKED ISSUE #1223: Messages consistency

digitalfishpond:
Part 3 of updating strings for the new i18n framework with correct text conventions.
This concludes the update and should fix #1223


rf232:
Can you rebase on current master? currently conflicts in PetSet and ReplicationController.

Otherwise LGTM

digitalfishpond:
PTAL

PR #1399, opened 2016-11-07
Adapted documentation and yaml to fit to kubeadm

cheld:
- Improved the setup description, especially for kubeadm
- Added toleration to yaml, so Dashboard can also be deployed on master. The reason for this is that users execute the installation instructions from top to bottom and expect that all Kubernetes core components (this includes Dashboard) are running.

More information on taints and tolerations:
https://github.com/kubernetes/kubernetes/blob/master/docs/design/taint-toleration-dedicated.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,23 +9,28 @@ itself.
 
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
-## Usage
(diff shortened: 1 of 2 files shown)


cheld:
https://github.com/kubernetes/kubernetes/issues/35536

PR #1406, opened 2016-11-09
Minor message fixes

maciaszczykm:


i18n/messages-en.xtb
```diff
@@ -813,16 +814,25 @@
   <translation id="5847167541694942228" key="MSG_PETSETDETAIL_PETSETINFO_9" desc="The message says that that many pods are pending (pet set details page).">{{ $ctrl.petSet.podInfo.pending}} pending</translation>
   <translation id="7332879554461510589" key="MSG_PETSETLIST_PETSETCARDLIST_0" desc="Column header \'Kind\' of a Pet Set List">Kind</translation>
   <translation id="2816069801101820166" key="MSG_PETSETLIST_PETSETCARDLIST_1" desc="Pet set list header: name.">Name</translation>
+  <translation id="3262303717958393654" key="MSG_PETSETLIST_PETSETCARDLIST_1" desc="Pet set list header: name.">Name</translation>
```
rf232: apart from the id, these two lines are identical aren't they. Somehow about every PR we get has these kind of changes to the message files, and I don't understand why that is happening.

Verified a bit and if I'm on a clean version of master and running gulp generate-xtbs gets the same problem. Inspecting the xtbgenerator code shows me that these IDs should be stable. so don't know how this happened. but I appear to get the same IDs generated on my machine, so lets just submit these versions and hope?

PR #1408, opened 2016-11-10
Update kubernetes deps, refactor and rename PetSet to StatefulSet

### LINKED ISSUE #1395: Rename PetSet to StatefulSet
@bryk In k8s 1.5 we're renaming PetSet to StatefulSet. We should do the rename in dashboard too (both frontend and backend). 

@kow3ns @foxish @erictune

floreks:
Closes #1395

## Changes

 - Updated kubernetes dependencies to newest version
 - Refactored code to use newest clientset/interface
 - Renamed PetSet to StatefulSet
 - Disabled some backend tests due to unsupported code (they have to be rewritten)

zrzut ekranu z 2016-11-10 15-57-23

For now [internalclient](https://github.com/kubernetes/kubernetes/tree/master/pkg/client/clientset_generated/internalclientset) is used due to major differences between types when `client-go` is used. I have different branch with `client-go` integration but there are still many conflicts and it requires much more changes than using `internalclientset`. We can wait for `client-go` to use same types as main repo and then do the migration.

zrzut ekranu z 2016-11-10 14-26-42

@bryk @rf232 PTAL


rf232:
Code LGTM
Shame about the tests, I would need to do some manual validation if most stuff still works. While I'd agree on merging this, I would really not want to go to a release for 1.5 without having them back.
Also I want to have a second pair of eyes doing the review + approving this.

floreks:
Yes, this needs more tests. I've tested most functionalities but it would be great if more people test this change. 

As for the tests, I'll check new way of mocking client and start rewriting them on new branch forked from this one. This PR is already quite big :)

rf232:
@floreks, agreed on this PR being big enough already, so I would indeed do the test fixes on a branch. I'm afraid I won't have time to test before Monday though.

rf232:
Testing on a GKE - Alpha features enabled clusters gives me:
 * Workloads page stays empty, I think the 'shouldShowZeroState' doesn't gracefully degradate.
 * StatefullSet lists page gives a 404, shouldn't it give an empty list, and thus show zerostate
both of these are minor issues, people should upgrade to core 1.5 if they want to use dashboard 1.5, still questions to think about.

Now I'll try to set up a local cluster with an alpha version of kubernetes to run your new fancy code with the new fancy features of kubernetes

floreks:
I'm not sure how GKE configures kubernetes, but I've checked it on cluster version: `1.3.7`, `1.4.7` and `HEAD`. Both workloads and stateful set list pages work fine. I can see zero state page on stateful set list page and Workloads page displays all objects correctly.

There should be no problem running it on older clusters as long as they are configured correctly.

rf232:
I tried to delete a Job and got:
```
Internal Server Error

api.DeleteOptions is not suitable for converting to "batch/v1"
```

rf232:
> Testing on a GKE - Alpha features enabled clusters gives me:

> * Workloads page stays empty, I think the 'shouldShowZeroState' doesn't gracefully degradate.
> * StatefullSet lists page gives a 404, shouldn't it give an empty list, and thus show zerostate  
both of these are minor issues, people should upgrade to core 1.5 if they want to use dashboard 1.5, still questions to think about.

Ah, these where my own problem, running gulp serve:prod fixes them

floreks:
> I tried to delete a Job and got:

>>Internal Server Error

>>api.DeleteOptions is not suitable for converting to "batch/v1"

Good catch. I'll investigate that.

rf232:
One more issue, could you update build/hyperkube.sh to use a newer K8S_VERSION with the new Stateful Sets in it?

```diff
diff --git a/build/hyperkube.sh b/build/hyperkube.sh
index 5ab4c7d..890cb86 100755
--- a/build/hyperkube.sh
+++ b/build/hyperkube.sh
@@ -17,7 +17,7 @@
 # Learn more at https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/docker.md
 
 # Version of kubernetes to use.
-K8S_VERSION="v1.4.0-alpha.3"
+K8S_VERSION="v1.5.0-beta.0"
 # Version heapster to use.
 HEAPSTER_VERSION="v1.1.0"
 # Port of the heapster to serve on.
```

floreks:
> One more issue, could you update build/hyperkube.sh to use a newer K8S_VERSION with the new Stateful Sets in it?

I'll update it.

Regarding Job deletion issue. I've found a fix for that, but it requires changes in the core. I'm not sure if this is intentional or not so I'll create an issue in the core repo.

rf232:
@floreks, since the Job deletion is not a new regression could you get the hyperkube change in this and rebase/merge with head. I would like to get this change in asap :)

floreks:
Sure, I'll do that soon. I have also almost finished fixing tests.

floreks:
Done. Hyperkube version is not actually the newest one but `1.5.0-beta.0` should be more stable than `1.6.0-alpha.0` and dependencies version matches `1.5.0`.

Issue #1389, opened 2016-11-03
Show Horizontal Pod Autoscaler resource


bryk:
We should show HPA resources in the UI as described here: http://kubernetes.io/docs/user-guide/horizontal-pod-autoscaling/

My opinion now is that we should show this as a property of a RS/Deployment, rather than a separate menu item.

rf232:
hmm, technically a RS/RC/Deployment can have more than 1 HPA linked to it, so I think we do need to do something more like a normal resource instead of just as a property

PR #1410, opened 2016-11-10
1389 display hpa

rf232:
#1389 

Known issues:
* Yaml view + edit doesn't work
* Delete doesn't work
* No menu entry for the list of HPA's yet

### Resource with no HPA 
zero-state-hpa
### Resource with 2 HPAs
2-hpa-rc
### Resource with 1 HPA
1-hpa-deployment
### Listview of HPA's when not on resource page
hpa-list
### Detail page of single HPA
hpa-detail


rf232:
Also while developing I had some issues with `gulp serve` not finding my new routes and I had to use `gulp serve:prod` to get it to work. I don't know what I did wrong there

bryk:
I had a look on this. Looks like a solid chunk of work with most cases covered.

1. I'm not sure, though, how we should represent relationships between objects. I.e., whether target here should be a link, or a "rich row":
selection_142


2. How many HPAs can object have? 0? 1? more?

3. Editing does not work

4. Should this have a separate menu item? Where?

rf232:
Re: your questions

> 1\. I'm not sure, though, how we should represent relationships between objects. I.e., whether target here should be a link, or a "rich row":

Went with a display similar to the serializedreference for the controller/creator here to get a quick working version. We might do something similar here as we did for the creator, but with the same 'issue' that it always is a singleton list.

> 2\. How many HPAs can object have? 0? 1? more?

Yes, there is basically no limit to how many HPA's can target a single resource. What the real meaning is if they contradict eachother I don't know.

> 3\. Editing does not work

I know, that is one of the reasons this is still WIP :)

> 4\. Should this have a separate menu item?

Yes, I think

> Where?

I don't know yet, it doesn't like it belongs in any of the current sections

bryk:
>> 4. Should this have a separate menu item?

> Yes, I think

I dont know. I'm not entirely sure of it... There are pros and cons of putting this in a separate menu item and making this listable. Let's discuss this when we met.

rf232:
PTAL @bryk, @maciaszczykm 

I added testcases now.

Known to not work is still the deletion. Is basically the same issue as https://github.com/kubernetes/kubernetes/issues/36806 but for another api group

src/app/frontend/deploymentdetail/deploymentdetail.html
```diff
@@ -65,5 +65,20 @@
   </kd-content>
 </kd-content-card>
 
+<kd-content-card>
+  <kd-title>[[Horizontal Pod Autoscalers|Title for the 'Horizontal Pod Autoscalers' card on the Replication Controller detail page]]
```
bryk: selection_143

I think that now this should be removed.

rf232: Ack, somehow this got changed between me writing the code and this review getting there. Also somehow that change got merged very quickly ;)

rf232:
I vote for keeping the kind of the target in there in someway, especially since there are different kinds of resource that can be targeted here. I'm not sure if the current way of rendering it is the best though

rf232:
Guessing my change will break when #1437 will break, and I will resolve the issues in this PR (at least if @floreks can make it happen to get that one in)

floreks:
> Known to not work is still the deletion. Is basically the same issue as kubernetes/kubernetes#36806 but for another api group

I'll fix autoscaling group also in my PR, so both Jobs and HPAs deletion will be soon fixed in the core.

src/app/backend/resource/common/types.go
```diff
@@ -96,24 +96,25 @@ type ResourceKind string
...
-	ResourceKindLimitRange            = "limitrange"
-	ResourceKindStatefulSet           = "statefulset"
-	ResourceKindNamespace             = "namespace"
-	ResourceKindNode                  = "node"
-	ResourceKindSecret                = "secret"
-	ResourceKindConfigMap             = "configmap"
-	ResourceKindPersistentVolume      = "persistentvolume"
-	ResourceKindPersistentVolumeClaim = "persistentvolumeclaim"
-	ResourceKindResourceQuota         = "resourcequota"
+	ResourceKindConfigMap               = "configmap"
```
bryk: +1 :)

src/app/backend/resource/horizontalpodautoscaler/horizontalpodautoscalercommon.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package horizontalpodautoscaler
+
+import ()
```
bryk: Hmm...? Remove?

src/app/backend/resource/horizontalpodautoscaler/horizontalpodautoscalercommon.go
```diff
@@ -0,0 +1,22 @@
...
+
+type ScaleTargetRef struct {
```
bryk: Add a comment that this corresponds to a API object type?

src/app/backend/resource/horizontalpodautoscaler/horizontalpodautoscalerdetail/horizontalpodautoscalerdetail.go
```diff
@@ -0,0 +1,80 @@
...
+	"log"
+
+	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/horizontalpodautoscaler"
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	"k8s.io/kubernetes/pkg/apis/autoscaling"
+	client "k8s.io/kubernetes/pkg/client/clientset_generated/internalclientset"
+)
+
+// HorizontalPodAutoscalerDetail provides the presentation layer view of Kubernetes Horizontal Pod Autoscaler resource.
```
bryk: Add a note that this is very similar to the api object and users should look for docs there?

src/app/frontend/horizontalpodautoscalerlist/horizontalpodautoscalercard.html
```diff
@@ -0,0 +1,78 @@
...
+      </div>
+    </kd-resource-card-column>
+    <kd-resource-card-column ng-if="$ctrl.showScaleTarget">
+      <a ng-href="{{::$ctrl.getScaleTargetHref()}}">
+        {{::$ctrl.horizontalPodAutoscaler.scaleTargetRef.kind}}: {{::$ctrl.horizontalPodAutoscaler.scaleTargetRef.name}}
+      </a>
+    </kd-resource-card-column>
+    <kd-resource-card-column>
+      <span ng-show="::$ctrl.horizontalPodAutoscaler.targetCPUUtilizationPercentage">
+      {{::$ctrl.horizontalPodAutoscaler.targetCPUUtilizationPercentage}}%
```
bryk: Should the percent be localized here? Are there languages that percent comes first? Ask olek maybe :)

rf232: cc: @olekzabl, should it?

olekzabl: Good catch :) I knew nothing and would probably not notice, but:
https://en.wikipedia.org/wiki/Percent_sign#Correct_style

src/app/backend/resource/horizontalpodautoscaler/horizontalpodautoscalerdetail/horizontalpodautoscalerdetail.go
rf232: ack

src/app/backend/resource/horizontalpodautoscaler/horizontalpodautoscalercommon.go
rf232: ack

src/app/backend/resource/horizontalpodautoscaler/horizontalpodautoscalercommon.go
rf232: ack

bryk:
Let me know when ready for approval.

rf232:
PTAL @bryk

src/app/externs/backendapi.js
```diff
@@ -1147,6 +1147,51 @@ backendApi.PersistentVolumeClaimList;
 
 /**
  * @typedef {{
+ *   objectMeta: !backendApi.ObjectMeta,
+ *   typeMeta: !backendApi.TypeMeta,
+ *   scaleTargetRef: !backendApi.ScaleTargetRef,
+ *   minReplicas: !number,
```
bryk: Why is some `!number` and some `number` here? They are actually the same types. Primitve types are non-nullable by default. Nice reading at: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler

Same for strings below.

bryk: Nullable number is `?number`

bryk:
LGTM, please merge when travis is happy :)

PR #1418, opened 2016-11-14
fixes broken graphs

### LINKED ISSUE #1416: i18n updates break graphs
Current implementation of the i18n framework breaks the graphs throughout the application

kd-graph-card attribute graph-title needs to be escaped with an additional single speech mark character

digitalfishpond:
Added single quotes on graph-title value for kd-graph-card tags throughout application
fixes #1416


digitalfishpond:
@bryk , makes sense. PTAL

PR #1419, opened 2016-11-14
Surface conditions in list views

### LINKED ISSUE #1353: Surface pod condition information in pod lists
The issue https://github.com/kubernetes/dashboard/issues/1317 started a discussion that we should surface pod condition info in pod lists. Or, at leas, pod lists on service pages. 

I.e., this section

```
Conditions:
  Type      Status
  Initialized   True 
  Ready     False 
  PodScheduled  True 
Volumes:
  configuration-volume:
    Type:   ConfigMap (a volume populated by a ConfigMap)
.....
```

I propose that we get rid of Cluster IP column and use this one. Cluster IP for a pod is not worth showing on all lists. This column is useless.

maciaszczykm:
Updated backend to return conditions of pods and nodes. Added `kd-conditions` directive to display conditions in list views.

I'll send commit with updated pods and nodes list views shortly. Please post your comments about current solution:

image

image

Closes #1353.

I'll also add conditions table to pod details view after it'll be merged (I did it for nodes some time ago).


bryk:
Hmmm.... So you're saying that when a condition is not met it is gray?

maciaszczykm:
@bryk Yes, I've added tooltip to the column:

image

I think this can save us space and help users to differ conditions by their status. Displaying conditions in separate columns takes too much space. Not displaying conditions with `False` status seems to be confusing for me...

What do you think about it?

maciaszczykm:
One more note...

I've started to implement it using red and green colors, but it's a bit problematic, because sometimes it's good situation when condition is not met (node and `OutOfSpace` condition case). So that's why I tried normal and muted color.

bryk:
@rf232 Can you help here with UX over this? I'm wondering how to display such diverse set of information in one row.

bryk:
@maciaszczykm Idea: we could list conditions manually and say which one is 'good' and which one is 'bad'. For a pod we could say 'ready' when all three are met. For nodes, we could show ready when ready is met and others are failing. Just brainstorming. WDYT?

maciaszczykm:
@bryk I was thinking on making list of "good" and "bad" conditions:
- when a "good" condition is met it's displayed using green color, when not we use red to warn user,
- when a "bad" condition is met it's diplayed using red color, when not we use green, because everything is okay.

Your idea is good, but I've some doubts:
- we have to update on every API change,
- more important - in node cases, it'll difficult to make conditions hierarchy - I think we have to display all of them (for pods it may be okay at the moment).

maciaszczykm:
Ready for review.

rf232:
> I've started to implement it using red and green colors, but it's a bit problematic, because sometimes it's good situation when condition is not met (case with node and `OutOfSpace` condition). So that's why I tried normal and muted color.

I 100% agree with this. Even if we do classifications on 'good' and 'bad' conditions it can be quite confusing, since then you have the question what does it mean if the `OutOfSpace` condition is green. 

But on the other hand I think the most important thing is that someone looking at the page has a quick understanding of the states of these conditions and how they affect the status, and then the Green - Red dichotomy might be nice.

So all in all I'm most definitely not sure what the best way to display this is (and this is assuming that we can easily classify all condititions on being 'good' or 'bad')

-- 
I personally think the design of these conditions is bad on the backend. IMHO they should have chosen them in such a way that they all were positive or all negative, not the current mishmash.

rf232:
codewise: LGTM

bryk:
Idea: how about we pull all this information into the status column? I.e., process it and assign status based on what we think is important. E.g., when a pod is "initialized, scheduled and not ready" we mark it as "waiting: not ready" or "waiting: not initialized". Similarly for nodes, we could have "status: out of disk" and mark it with red or "not yet ready" and mark it as gray or "has memory pressure" and mark it as yellow. 

@rf232 @maciaszczykm What do you think?

maciaszczykm:
@bryk It's good idea. We could display more warning icons if something is not okay. In a tooltip we may display all conditions and their statuses.

Main concern is number of conditions, that we have to handle now and in the future.

> Similarly for nodes, we could have "status: out of disk" and mark it with red or "not yet ready" and mark it as gray or "has memory pressure" and mark it as yellow.

We need to determine what to do when both conditions are true. It seems to be obvious, that we should mark it with red, but we have to implement logic for it. Shouldn't be that hard tho.

bryk:
@rf232 What do you think about such proposal?

rf232:
So let me get this straight. 
We want to build some sort of graph of the dependencies of `condition`s which build up the `status` of a resource. and with this hope to be able to summarize the conditions of a resource in 1 or 2 words?

Isn't the status already doing that in a way?

So not to be a debbie downer, but what information would this summarized version of the condition add to our list view?

bryk:
> So let me get this straight. 
We want to build some sort of graph of the dependencies of conditions which build up the status of a resource. and with this hope to be able to summarize the conditions of a resource in 1 or 2 words?

Yes

> Isn't the status already doing that in a way?

Kind of. Not for nodes. And for pods not entirely.

bryk:
How about we add conditions to the details page of nodes and pods? Then implement status handling that covers all conditions we already know about? 

This would, IMO, solve the original problem we had reported by the customer, yet the list of objects would not be cluttered with info not needed most times.

What do you think?

rf232:
as discussed during bug triage, that sounds like indeed the best way to go ahead, and the changes on the detail pages are already in (thanks, and awesome @maciaszczykm )

maciaszczykm:
Closing as we agreed for a different solution.

PR #1424, opened 2016-11-15
Fix graph titles for memory and CPU timelines

bryk:
Graph title on workloads page was invalid. Fixed it.

Then removed "history" suffix from all titles, since I think it is
redundant and piece of information that adds no value.


bryk:
@rf232 Can you review this?

PR #1428, opened 2016-11-16
Implement better UX around pagination and resource lists

bryk:
This PR is a big one and contains a lot of changes, I'm sorry for this.

* Moved pagination to top of resource card lists.
* Embeded headers in resource card lists. With that the header can be
  sticky in a next PR.
* Resource lists do not link to list pages on aggregate pages. It is
  doble only via menu.
* Reordered lists on workloads page to match menu.
* Changed "Creator" to "Created by" as that was the feedback from small
  UX study we did recently.
* Removed rows-per-page selector entirely. The reason for this is that
  as long as this is not persisted across URL changes, it is little
  useful. We can reintroduce it later.

Future work:
* Make list headers sticky.


bryk:
@rf232 @floreks Can you review this?

floreks:
Merged so quick :) I was just about to review it. I'll check it anyway though.

bryk:
Yep, please add your comments :) I'll do a follow up PR with tests and cleanup soon. So comment on all now.

PR #1434, opened 2016-11-17
Fix typo in deployment detail template

bryk:
It was causing errors on the console and broken new replica sets list


bryk:
@rf232 Can you review this?

PR #1437, opened 2016-11-17
Fix backend tests

floreks:
Rewritten tests disabled during dependency update.


rf232:
Code looks fine, is it intentional though that the TestGetNodeDetail test from the nodedetail tests is not reenabled again?

floreks:
>Code looks fine, is it intentional though that the TestGetNodeDetail test from the nodedetail tests is not reenabled again?

I was going through with the list from my previous PR. This test was not skipped and was not listed on my list and I've missed it :) Nice catch!

>I really love the simplifications with helper functions you made for creating stuff :)

I've had to go through core code to fix them and noticed similar approach. I've used that to simplify some tests a bit. It's hard to follow them sometimes :)

floreks:
Done. PTAL.

PR #1452, opened 2016-11-18
Redesign secret detail page

bryk:
This is my second iteration to the details page upgrade effort started by @batikanu.

* Made all secrets hidden by default. This is the spirit of secrets anyway. 
* I propose moving eye icon to the left of the list of secrets
* Changed card titles of cards there to be same as in resource lists
* Changed eye-off eye-on icons to reflect material design guidelines (one icon is atfitificially created using SVG)


bryk:
@rf232 @batikanu Can you review this?

src/app/frontend/secretdetail/detail.html
```diff
@@ -15,17 +15,34 @@
...
+  <kd-secret-info secret="::$ctrl.secretDetail"></kd-secret-info>
+  <kd-info-card>
+    <kd-info-card-header>[[Data | Secrets info details section data.]]</kd-info-card-header>
+    <kd-info-card-section>
+      <div layout="row" class="kd-info-card-entry kd-secret-detail-row" ng-repeat="(key, value) in ::$ctrl.secretDetail.data">
+        <div flex="nogrow" style="position: relative">
+          <md-button ng-click="isActive = !isActive" class="md-icon-button">
+            <!-- SVG drawing to simulate crossed-eye icon.
+                 This should be removed if possible. -->
+            <svg ng-if="isActive" width="40px" height="40px" viewBox="0 0 40 40"
```
batikanu: there is [eye-off icon](https://materialdesignicons.com/icon/eye-off) which you can use instead of svg directly.

bryk: Sadly, we do not use this library :( We use https://material.io/icons/, which is limited in scope. There is no eye-off icon there. 

We could try evaluating and including other material design libraries, but this requires much effort of integration, validation and license eveluation. Can we do this in a separate task? I can spawn an issue.

src/app/backend/resource/secret/secretdetail.go
```diff
@@ -36,9 +36,6 @@ type SecretDetail struct {
 
 	// Used to facilitate programmatic handling of secret data.
 	Type api.SecretType `json:"type"`
-
-	// Determines if the secret is hidden type in case of opaque and basicAuth types.
-	IsHiddenType bool `json:"isHiddenType"`
```
rf232: We now just hide them all? When I looked at the default token I now can't see the 'namespace' anymore in a useful way

src/app/frontend/secretdetail/detail.html
bryk: Created https://github.com/kubernetes/dashboard/issues/1460

src/app/backend/resource/secret/secretdetail.go
bryk: Yes, there's change in behavior here, yes. I propose hiding all secrets by default, and only showing some of them we pick by hand (possibly none). This is to follow security "principle" similar to what you have in, e.g., firewalls, i.e., you disallow all traffic and allow only selected ports and protocols.

Does this make sense?

src/app/frontend/common/components/togglehiddentext/togglehiddentext.scss
```diff
@@ -1,51 +0,0 @@
...
-
-@import '../../../variables';
-
-$eye-icon-size: $body-font-size-base * 1.8;
-$eye-icon-color-on: rgba(0, 0, 0, .84);
-$eye-icon-color-off:  rgba(0, 0, 0, .54);
-$eye-icon-background-color-on: rgba(224, 224, 224, 1);
-
-.kd-toggle-hidden-text-input {
-  border: 0 none;
```
bryk: Yes, makes total sense. Done.

bryk:
@rf232 @batikanu PTAL

src/app/backend/resource/secret/secretdetail.go
bryk: kubectl describe does obfuscate namespace, so we do here. I'd say to obfuscate everything for now and when feature requests come to show individual types/fields, we'd show them. But hiding everything by default is safer.

Issue #1345, opened 2016-10-18
Exec into Pod


IanLewis:
I am planning on working on "exec into Pod" functionality. I'm planning on gathering requirements via this issue and creating a proof of concept.
- [x]  Gather requirements
- [x]  Create POC
- [x]  Create UX design
- [ ]  Implementation

My general thought is that I will search for and use an existing library on the client side to display the terminal and take input. I'll use something like websockets or whatever transport the client side library supports for terminal events. The server side will use the [exec endpoint](http://kubernetes.io/kubernetes/third_party/swagger-ui/#!/api%2Fv1/connectPostNamespacedPodExec) from the Kubernetes API for connecting to the actual pod.

Related issues: #21 #550

IanLewis:
I may be getting ahead of myself, but a couple of thoughts re: requirements:

1. `kubectl exec` supports the `-t` option to enable tty. Do we need to have the ability to turn tty on/off? I can think of cases where it isn't needed but are there cases where the user would need to turn these off? My first thought is that we can leave this option out and just enable tty all the time for simplicity.
1. `kubectl exec` supports the `-i` option to enable stdin. Is there a need for some kind of stdin option? i.e. some kind of text box where the user can put in text that can be sent on stdin? My first thought is no, but I want to make sure.
1. `kubectl attach` is very similar to `kubectl exec` and my feeling is that the machinery to make `exec` and `attach` work would be pretty much the same. They could probably be developed at the same time or attach could be added quickly after exec into pod.

IanLewis:
Here are a few browser based terminal libraries I'm looking at:

- http://xtermjs.org/
- http://terminal.js.org/
- http://terminal.jcubic.pl/
- http://cb.vu/

xterm.js seems like the best based on a cursory look. It's a fairly active project and is used as part of Microsoft Visual Studio Code for their integrated terminal.

IanLewis:
In terms of requirements it looks like debugging is the main use case for exec into pod.

This might affect the UI and how users discover the feature, but for now it doesn't seem like it's going to affect base functionality from what I described above.

colemickens:
No `hterm` (https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm)? This is what Google's Secure Shell extension for Chrome uses, IIRC.

IanLewis:
@colemickens Cool. I didn't know about that one.

bryk:
[...]

PR #1455, opened 2016-11-21
Exec into pod prototype

IanLewis:
A simple prototype for the exec into pod feature (#1345). Currently it's only an echo server connected to an xterm.js frontend over websockets. Currently, doesn't work in development mode because browserSync doesn't support websockets.

Steps to test:
1. Run `gulp serve:prod`
1. View a pod detail page.
1. Click on the "Exec" link under a container card

Current issues
- Need to figure out how to create a remote command client since it's separate from the normal API client.
- Browsersync doesn't support websockets. Need to figure out a workaround.

I will update this PR until I can exec a command into a pod using the API.


src/app/backend/handler/apihandler.go
```diff
@@ -996,6 +1007,55 @@ func (apiHandler *APIHandler) handleGetPods(
...
+	// config := &restclient.Config{}
+	// exec, err := remotecommand.NewExecutor(config, "POST", req.URL())
+	// if err != nil {
+	// 	log.Println(err)
+	// 	return
+	// }
+
+	// err = exec.Stream(remotecommand.StreamOptions{
+	// 	SupportedProtocols: remotecommandserver.SupportedStreamingProtocols,
+	// 	Stdin:              struct{
```
IanLewis: I expect I will have to create custom Readers and Writers for the remote command client here. The Readers and Writers will read and write data to the websocket.

src/app/backend/handler/apihandler.go
```diff
@@ -996,6 +1007,55 @@ func (apiHandler *APIHandler) handleGetPods(
...
+	// For example about how kubexec works see: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/exec.go
+	// req := apiHandler.client.Post().
+	// 	Resource("pods").
+	// 	Name(pod.Name).
+	// 	Namespace("default").
+	// 	SubResource("exec").
+	// 	Param("container", containerName)
+
+	// config := &restclient.Config{}
+	// exec, err := remotecommand.NewExecutor(config, "POST", req.URL())
```
IanLewis: I need to create a new executor which is like an API client. I need to pass it a restclient.Config. Not sure the best way to get the config from the existing API client to set up this Config object.

build/serve.js
```diff
@@ -20,7 +20,8 @@ import browserSyncSpa from 'browser-sync-spa';
 import child from 'child_process';
 import gulp from 'gulp';
 import path from 'path';
-import proxyMiddleware from 'proxy-middleware';
+// import proxyMiddleware from 'proxy-middleware';
+import proxyMiddleware from 'http-proxy-middleware';
```
IanLewis: Changed to using http-proxy-middleware because it supports websockets but BrowserSync itself doesn't seem to work with websockets.

src/app/backend/handler/apihandler.go
```diff
@@ -252,6 +257,12 @@ func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.Hea
 			To(apiHandler.handleGetPodContainers).
 			Writes(pod.PodDetail{}))
 	apiV1Ws.Route(
+		apiV1Ws.GET("/pod/{namespace}/{pod}/exec").
+			To(apiHandler.handleExecIntoPod))
```
IanLewis: I plan to make this endpoint default to using the first container in the pod.

src/app/externs/xterm.js
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @constructor
+ */
+var Terminal = function() {};
```
IanLewis: Added externs for xterm.js since we have to use serve:prod and the closure compiler will complain without them.

src/app/frontend/podexec/afterrender_directive.js
```diff
@@ -0,0 +1,36 @@
...
+ * @return {!angular.Directive}
+ * @ngInject
+ */
+export default function afterRenderDirective($timeout) {
+  var def = {
+    restrict: 'A',
+    terminal: true,
+    transclude: false,
+    /** @param {!AfterRender.attrs} attrs */
+    link: function (scope, element, attrs) {
```
IanLewis: This probably doesn't have to be generic. Would appreciate some pointers as to how to clean this up a bit.

src/app/frontend/podexec/afterrender_directive.js
```diff
@@ -0,0 +1,36 @@
...
+      // We call $timeout with a zero timeout so that the function code
+      // is placed in the browser execution queue. Since the render is
+      // also already in the queue this ensures that the code is run after
+      // the template is rendered.
+      $timeout(scope.$eval(attrs.afterRender), 0);  //Calling a scoped method
```
IanLewis: Could I somehow bind the Terminal to the element object rather than using $timeout() to render the Terminal after it's been attached to the DOM?

build/serve.js
bryk: Sounds good.

src/app/backend/handler/apihandler.go
bryk: Sounds good.

src/app/backend/handler/apihandler.go
bryk: Yes, sounds like a doable plan.

src/app/externs/xterm.js
bryk: Yes, that's what we need. We can upstream this eventually.

src/app/frontend/podexec/afterrender_directive.js
bryk: As discussed offline, this is my suggestion that should work.

1. Rename this from generic 'after-render' to 'terminal-foo' or something.
2. Attach terminal to the `element` variable you have here, e.g., by taking the lines from podexec_controller.js:
```js
   // TODO: Is there a better way to get the element to attach the terminal to?
    this.term.open(this.document_.getElementById('kd-podexec-terminal'));
```

To:

```js
   this.term.open(element);
```

bryk:
@rf232 Can you help reviewing this?

src/app/frontend/podexec/afterrender_directive.js
IanLewis: Thanks. I'll clean this up at some point.

bryk:
Let me know when there is another version to fetch and test locally :)

IanLewis:
@byrk I've pushed a version that's working end-to-end. Currently it doesn't have a way to specify the command; it just uses "bash" and the command for testing purposes. I hope to resolve how to specify the command as part of UX discussions.

For now the terminal works (doesn't support color yet). I hope this can suffice as a prototype.

bryk:
Cool, let me test this.

bryk:
Tested. Works really cool! 

Felt natural to type bash commands there and all worked flawlessly. 

Should we start the UX discussion about this now?

smarterclayton:
@jwforres if there were any lessons learned from the OpenShift implementation of this or places we can share code would be good.

IanLewis:
@smarterclayton @smarterclayton Could you point us to the relevant code in OpenShift? I'm also not that familiar with it. Is it pretty easy to get it up and running on something like GKE and try out the feature?

I did look at screenshots and descriptions of the feature a bit and noticed that it altered the container endpoint (or maybe ran the command in a new container in the pod?) and we explicitly decided not to do that and just use the exec API as is.

It would be good to know what UX ideas came out of the feature in OpenShift though.

jwforres:
@IanLewis OpenShift and Cockpit both use https://github.com/kubernetes-ui/container-terminal

It's built on term.js which is unfortunately now deprecated for support.  We want to migrate it to xterm.js instead but it has slightly different rendering and support options so we have to do some work before we can migrate.

A few things we've found:

1) People want an option to expand to fullscreen, without creating a new connection. So open in a new tab doesn't work for this.  On a related note, now that resize signals can pass through the websocket down into the container, we automatically resize the terminal window to the space available in the browser, we've only had positive feedback for this feature. https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/pod.js#L115

2) People expect things like the "clear" command to just work. To make that happen we decided to set the TERM env var as part of the exec call that runs /bin/sh.   This is similar to what kubectl does already. https://github.com/openshift/origin-web-console/blob/master/app/views/browse/pod.html#L201

3) Multi-container pods we came up with a design for switching between the containers while keeping the connections open in the background.  This lets people run commands across multiple containers in the pod without having to have multiple browser tabs open.

IanLewis:
> @IanLewis OpenShift and Cockpit both use https://github.com/kubernetes-ui/container-terminal

@jwforres Thanks Jessica! Thanks for the pointer. I had not heard of container-terminal before! I suppose it relies on directly accessing the Kubernetes API?

Really great points below too. This is super helpful.

> 1. People want an option to expand to fullscreen, without creating a new connection. So open in a new tab doesn't work for this. On a related note, now that resize signals can pass through the websocket down into the container, we automatically resize the terminal window to the space available in the browser, we've only had positive feedback for this feature. 

Yah. We had some discussion about popping the terminal out into a new window. I had thought I would do something like create a connection token which could be used to re-establish a connection but holding the connection in the backend is risky since you need to be able to time it out if it's not closed properly.

You're talking about real fullscreen right? Like Youtube's fullscreen feature. If opening in a new tab/window worked would users be ok with that in your opinion?

I planned on supporting resize at some point but I'm quite frankly ignorant as to how terminal sizing works.

> People expect things like the "clear" command to just work. To make that happen we decided to set the TERM env var as part of the exec call that runs /bin/sh. This is similar to what kubectl does already. 

Makes sense. I'll look for the relevant part of kubectl as well.

> Multi-container pods we came up with a design for switching between the containers while keeping the connections open in the background. This lets people run commands across multiple containers in the pod without having to have multiple browser tabs open.

Interesting. I'm curious what sort of use cases users have for something like this. What are they actually doing?

jwforres:
On Fri, Dec 2, 2016 at 7:49 PM, Ian Lewis <notifications@github.com> wrote:

> @IanLewis <https://github.com/IanLewis> OpenShift and Cockpit both use
> https://github.com/kubernetes-ui/container-terminal
>
> @jwforres <https://github.com/jwforres> Thanks Jessica! Thanks for the
> pointer. I had not heard of container-terminal before! I suppose it relies
> on directly accessing the Kubernetes API?
>
> Really great points below too. This is super helpful.
>
>
>    1. People want an option to expand to fullscreen, without creating a
>    new connection. So open in a new tab doesn't work for this. On a related
>    note, now that resize signals can pass through the websocket down into the
>    container, we automatically resize the terminal window to the space
>    available in the browser, we've only had positive feedback for this feature.
>
> Yah. We had some discussion about popping the terminal out into a new
> window. I had thought I would do something like create a connection token
> which could be used to re-establish a connection but holding the connection
> in the backend is risky since you need to be able to time it out if it's
> not closed properly.
>
> You're talking about real fullscreen right? Like Youtube's fullscreen
> feature. If opening in a new tab/window worked would users be ok with that
> in your opinion?
>
I was actually just referring to taking up the whole browser viewport and
dropping things that take up space like left navigation, not a whole
monitor fullscreen.  I haven't heard any requests for the latter.


> I planned on supporting resize at some point but I'm quite frankly
> ignorant as to how terminal sizing works.
>

I've you take a look in the container-terminal code you'll see a
sizeTerminal method that shows what data needs to be sent through the
websocket
https://github.com/kubernetes-ui/container-terminal/blob/master/container-terminal.js#L124


> People expect things like the "clear" command to just work. To make that
> happen we decided to set the TERM env var as part of the exec call that
> runs /bin/sh. This is similar to what kubectl does already.
>
> Makes sense. I'll look for the relevant part of kubectl as well.
>
> Multi-container pods we came up with a design for switching between the
> containers while keeping the connections open in the background. This lets
> people run commands across multiple containers in the pod without having to
> have multiple browser tabs open.
>
> Interesting. I'm curious what sort of use cases users have for something
> like this. What are they actually doing?
>
We didn't hear explicit requests for this, we just wanted to support
multi-container pods well.  Originally we showed terminals for every
container in the pod and didn't establish the connection to one until you
clicked into it, but this prevented us from doing nicer things when there
is only one container, like taking up the available space in the browser
automatically, and auto-connecting and auto-focusing the cursor.   You
don't necessarily need to maintain the sessions when you switch between
containers, but I don't recommend our original approach of showing
terminals for all of them all at once, it was always a temporary solution
for us.

rf232:
Also when running on an exposed dashboard and something goes wrong the whole dashboard crashes, perhaps something to look into :)
```
[2016-12-14T12:44:04Z] Incoming HTTP/1.1 GET /api/v1/pod/kube-system/kube-dns-v20-l5k58/exec/healthz?command=bash request from X.X.X.X:47560
Unable to create stream: error executing remote command: Error executing command in container: Error executing in Docker Container: -1
[2016-12-14T12:44:04Z] Outcoming response to X.X.X.X:47560 with 200 status code
http: response.Write on hijacked connection
E1214 12:44:04.360473       1 v2.go:105] read tcp X.X.X.X:9090->X.X.X.X:47560: use of closed network connection
E1214 12:44:04.360473       1 v2.go:105] read tcp X.X.X.X:9090->X.X.X.X:47560: use of closed network connection
log: exiting because of error: log: cannot create log: open /tmp/dashboard.dash-demo-2-mdzr6.unknownuser.log.ERROR.20161214-124404.1: no such file or directory
```

src/app/backend/handler/apihandler.go
```diff
@@ -996,6 +1008,107 @@ func (apiHandler *APIHandler) handleGetPods(
...
+	for i := range m {
+		p[i] = m[i]
+	}
+	return len(m), nil
+}
+
+// Write is called by the remotecommand Executor stream and writes the given
+// data to the client websocket.
+func (t WebSocketIO) Write(p []byte) (int, error) {
+	// TODO: Should binary messages be used? xterm.js seems to send text based messages.
```
jwforres: we send/receive binary in container-terminal, we convert in the browser https://github.com/kubernetes-ui/container-terminal/blob/master/container-terminal.js#L31

src/app/frontend/podexec/podexec_controller.js
```diff
@@ -0,0 +1,85 @@
...
+    // TODO: Is there a better way to get the element to attach the terminal to?
+    this.term.open(this.document_.getElementById('kd-podexec-terminal'));
+
+    let namespace = this.stateParams_.objectNamespace;
+    let podId = this.stateParams_.objectName;
+    let container = this.stateParams_.container || '';
+
+    let url = `ws://${location.host}/api/v1/pod/${namespace}/${podId}/exec/${container}`.replace(/\/+$/, "");
+    // TODO: Get a command from the user. Needs UX work.
+    url += "?command=bash";
```
jwforres: we opted to use sh not bash since it is slightly more likely to exist in the container

IanLewis: I don't think my idea was to default to a particular command and have the user initiate command. I just didn't go far enough into the UX to actually implement it and I just chose bash because it was handy to test with.

IanLewis:
@rf232 Yes. The error handling is not what it could be.

IanLewis:
Possibly relevant link
https://gist.github.com/TamalSaha/4e6a5c1144e2daf370398520e20fdef7

floreks:
Closing in favour of #1939 .

Issue #1296, opened 2016-09-29
Introduce auto-refresh mechanism


arun-gupta:
Moving from https://github.com/kubernetes/kubernetes/issues/33755

Using Kubernetes 1.4, `master-ip/ui` shows the dashboard nicely. 

For example, https://52.41.10.66/api/v1/proxy/namespaces/kube-system/services/kubernetes-dashboard/#/replicationcontroller?namespace=default shows all the Replication Controllers.

New RCs were created and the page had to be manually refreshed to show the newly created RCs.

An auto-refresh capability in the UI would be great.

cheld:
Also, deleted items would disappear :)

bryk:
We had initial thinking on this topic, but when you dig deeper, some issues arise. For example, what if your system is highly dynamic and pods come and go every second? This would cause complete mess.


Would a button like "there are new RCs, refresh" or "turn on autorefresh" help?

cheld:
@batikanu told me there should be an auto-refresh after every own action. Maybe it is missing at some places... Of course, still missing is refresh after own CLI-actions, actions from other users and refresh of graphs.

Some monitoring dashboards offer a auto-refresh button that reloads the model every couple second or so. @bryk could give an example of the problems that it would cause?

bryk:
> @batikanu told me there should be an auto-refresh after every own action

Yes, that's true. But this is only a single refresh.

>@bryk could give an example of the problems that it would cause?

E.g., you're doing a rolling update and your pods appear and disappear from the pod list every second. You want to click a pod, but it disappears from under your pointer (this may be actually the feature users look for, actually).

quinnj:
I also think auto-refresh would be super useful. There could perhaps be some kind of setting to limit how often it was able to refresh (in cases where pods are coming/going sub-second).

guenhter:
Kibana has a nice approach doing the auto-update. 
![GitHub Logo](https://www.elastic.co/guide/en/kibana/6.0/images/autorefresh-intervals.png)

andrewrk:
Auto-refresh is a hack. What we actually want is #2541 - live updates pushed to the browser when the information changes.

ruippeixotog:
While push updates would surely be the ideal solution, I believe that it would need a lot of work on the backend in order to support it. On the other hand, an auto-refresh feature could be easily implemented on the browser side and it would already help a lot.

[...]

PR #1459, opened 2016-11-21
[WIP] auto-refresh implementation

batikanu:
- This is a sample auto-refresh implementation for dashboard resources which was requested at #1296
- Currently it is implemented only for replication controller list and node list but can be used for every other resources and views.
- It is not complete yet. 
- It requires improvement and discussion before further development if it is needed at all. 

Auto-refresh feature is added to desired list or detail page as a directive with parameters of data resource delay and etc. It can be also used as a service but it has some pros and cons. 

It is based on polling action in particular intervals for retrieving data from back-end. 

Auto-refresh service automatically stops when user changes the states to avoid unnecessary polling unrelated to the selected view. It also breaks if any pop up menu is open to avoid disappearing menus involuntarily on update and it continues polling after the pop up menu closes. 

Multiple polling is possible for one view also with different delays. 

The graph component wasn't suitable for auto-refresh, therefore I made some small changes there to enable it.

### DIFF
--- a/src/app/frontend/common/autorefresh/autorefresh_controller.js
+++ b/src/app/frontend/common/autorefresh/autorefresh_controller.js
@@ -1,0 +1,133 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Controller for the autorefresh directive.
+ * @final
+ */
+export class AutoRefreshController {
+  /**
+   * @param {!./../../chrome/chrome_state.StateParams} $stateParams
+   * @param {!./../pagination/pagination_service.PaginationService} kdPaginationService
+   * @param {!./poll_service.PollService} kdPoll
+   * @ngInject */
+  constructor($stateParams, kdPaginationService, kdPoll) {
+    /**
+     * Initialized from the scope.
+     * @export {!angular.$resource}
+     */
+    this.source;
+
+    /**
+     * Initialized from the scope.
+     * @export {!Object}
+     */
+    this.target;
+
+    /**
+     * Initialized from the scope.
+     * @export {!string}
+     */
+    this.type;
+
+    /**
+     * Initialized from the scope
+     * @export {number}
+     */
+    this.delay;
+
+    /**
+     * Initialized from the scope
+     * @export {!boolean}
+     */
+    this.namespace;
+
+    /**
+     * @type {!./poll_service.PollService}
+     * @private
+     */
+    this.pollService_ = kdPoll;
+
+    /**
+     * @type {!./../../chrome/chrome_state.StateParams}
+     * @private
+     */
+    this.stateParams_ = $stateParams;
+
+    /**
+     * @type {!./../pagination/pagination_service.PaginationService}
+     * @private
+     */
+    this.kdPaginationService_ = kdPaginationService;
+
+    this.startAutoRefresh_();
+  }
+
+  /**
+   * Starts auto-refreshing according to settings
+   * @private
+   */
+  startAutoRefresh_() {
+    if (!this.type) {
+      return this.startRefresh(this.source, this.target, null);
+    }
+    if (this.type.toLocaleLowerCase() === TYPE_LIST) {
+      return this.startRefresh(this.source, this.target, this.createPaginationQueryWithNamespace());
+    }
+    if (this.type.toLocaleLowerCase() === TYPE_LIST_WITH_NO_NAMESPACE) {
+      return this.startRefresh(this.source, this.target, this.createPaginationQuery());
+    }
+    return this.startRefresh(this.source, this.target, null);
+  }
+
+  /**
+   * Starts auto refresh and returns notification promise for each update.
+   * @param resource {!angular.Resource}
+   * @param target {!Object}
+   * @param params {angular.Resource.ParamsOrCallback=}
+   * @return {!angular.$q.Promise}
+   * @export
+   */
+  startRefresh(resource, target, params) {
+    let promise = this.pollService_.poll(resource, params, this.delay);
+    promise.then(null, null, (newObj) => {
+      // Update target object with the new object.
+      Object.assign(target, newObj);
+    });
+    return promise;
+  }
+
+  /**
+   * Creates pagination default resource query with namespace parameter.
+   * @return {!backendApi.PaginationQuery}
+   * @export
+   */
+  createPaginationQueryWithNamespace() {
+    return this.kdPaginationService_.getDefaultResourceQuery(this.stateParams_.namespace);
+  }
+
+  /**
+   * Creates pagination default resource query.
+   * @return {!backendApi.PaginationQuery}
+   * @export
+   */
+  createPaginationQuery() {
+    return this.kdPaginationService_.getDefaultResourceQuery();
+  }
+}
+
+/** @type {number} */
+export const DEFAULT_DELAY = 30000;
+export const TYPE_LIST_WITH_NO_NAMESPACE = 'list-without-namespace';
+export const TYPE_LIST = 'list';
--- a/src/app/frontend/common/autorefresh/autorefresh_directive.js
+++ b/src/app/frontend/common/autorefresh/autorefresh_directive.js
@@ -1,0 +1,66 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {AutoRefreshController} from './autorefresh_controller';
+/**
+ * Provides upload file mechanism and updates the model with file parameter for the selected file.
+ * @param {!./poll_service.PollService} kdPoll
+ * @return {!angular.Directive}
+ * @ngInject
+ */
+export default function autoRefreshDirective(kdPoll) {
+  return {
+    scope: {},
+    bindToController: {
+      // Resource object to query data.
+      'source': '=',
+      // The the model object which is updated in each poll.
+      'target': '=',
+      // Type of the page like detail list, list-without-namespace, detail page and etc.
+      'type': '@?',
+      // Polling timeout in milliseconds.
+      'delay': '@?',
+    },
+    controller: AutoRefreshController,
+    controllerAs: 'ctrl',
+    templateUrl: '',
+    link: function(scope) {
+      /**
+       * @type {./poll_service.PollService}
+       * @private
+       */
+      let kdPoll_ = kdPoll;
+
+      // Cancel all polls in case of state change to avoid accumulated polls.
+      scope.$on('$stateChangeStart', () => {
+        kdPoll_.removeAll();
+      });
+
+      // Stop all polls when any menu is open to avoid disappearing them while auto-refresh.
+      /** @type {function()} */
+      let menuOpenEvent = scope.$root.$on('$mdMenuOpen', () => {
+        kdPoll_.stopAll();
+      });
+
+      scope.$on('$destroy', menuOpenEvent);
+
+      // Continue polling if any stopped after any menu is open.
+      /** @type {function()} */
+      let menuCloseEvent = scope.$root.$on('$mdMenuClose', () => {
+        kdPoll_.startAll();
+      });
+      scope.$on('$destroy', menuCloseEvent);
+    },
+  };
+}
--- a/src/app/frontend/common/autorefresh/poll_service.js
+++ b/src/app/frontend/common/autorefresh/poll_service.js
@@ -1,0 +1,207 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Service class for polling.
+ * @final
+ */
+export class PollService {
+  /**
+   * @param $interval {!angular.$interval}
+   * @param $q {!angular.$q}
+   * @ngInject
+   */
+  constructor($interval, $q) {
+    /**
+     * @type {!angular.$interval}
+     * @private
+     */
+    this.intervalService_ = $interval;
+
+    /**
+     * @type {!angular.$q}
+     * @private
+     */
+    this.q_ = $q;
+
+    /**
+     * @type {!Array<PollOptions>}
+     * @private
+     */
+    this.optionsList_;
+  }
+
+  /**
+   * Starts polling and returns polling promise.
+   * @param resource {!angular.$resource}
+   * @param params {angular.Resource.ParamsOrCallback=}
+   * @param delay {number}
+   * @export
+   * @returns {!angular.$q.Promise}
+   */
+  poll(resource, params, delay) {
+    this.optionsList_ = this.optionsList_ || new Array();
+    let options = this.getOptionsByResource_(resource);
+    if (!options) {
+      options = new PollOptions(resource, params, delay);
+      this.optionsList_.push(options);
+    }
+    // Stop if there is any previously running interval
+    this.cancelInterval_(options);
+    return this.createInterval_(options);
+  }
+
+  /**
+   * Stops all running polls.
+   * @export
+   */
+  stopAll() {
+    this.optionsList_.map((options) => {
+      this.cancelInterval_(options);
+    });
+  }
+
+  /**
+   * Stops all running polls and removes all registered options.
+   * @export
+   */
+  removeAll() {
+    this.stopAll();
+    this.optionsList_ = new Array();
+  }
+
+  /**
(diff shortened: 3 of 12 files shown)


bryk:
Wow, this is cool! Checked the behavior and it works amazingly well.

Let's start a discussion regarding polling versus watching. You proposed polling here, which is nice and easy to implement. But can we do watching for changes (with throttling perhaps)?

This would be much faster and responsive system, as the delay between a change and population into the UI would be instantaneous. Also, the reason I push for watching is that I've seen and worked on UIs that do polling, and it never worked well. It was either spamming backend with requests or having huuge lag in the UI. Watching is harder to implement at first, but user benefit is great.

For watching we could have one-endpoint-per-page to listen to. E.g., RC detail and list pages would have a websocket endpoit `/api/v1/replicationcontroller/default/watch` that UI can attach to. Whenever backend thinks something has changed, it'd send a new version of the resource there. Backend can use watch endpoints from the apiserver or send new values in intervals (for graphs). 

What do you think? cc @rf232 @cheld

rf232:
>
> Let's start a discussion regarding polling versus watching. You proposed
> polling here, which is nice and easy to implement. But can we do watching
> for changes (with throttling perhaps)?
>
I agree with the preference for watching, although it can indeed be quite a
PITA to implement at first.

> For watching we could have one-endpoint-per-page to listen to. E.g., RC
> detail and list pages would have a websocket endpoit
> /api/v1/replicationcontroller/default/watch that UI can attach to.
> Whenever backend thinks something has changed, it'd send a new version of
> the resource there. Backend can use watch endpoints from the apiserver or
> send new values in intervals (for graphs).
>
 The graphs I have to think about, but indeed for the more 'vanilla'
resources getting these updates pushed from the backend sounds like the way
to go.



On a more project management kind of view. I rather get this feature in
after we did the 1.5 release because it is quite a big change and might be
a bit tricky to get right in one go.

cheld:
It seems that kubernetes API server supports watching...I didn't know that. So we should try to use this functionality.

@bryk if I understood your proposal correctly, you suggest to watch for changes, but only send a hint to the UI to indicate if it is worthwhile to refresh. Seems a good approach to me. 

Of course we do not have to watch for changes for the graphs - we know they update periodically. But it would be good to synchronize the updates of UI with heapster backend. Not sure if this is possible.

BTW: a couple of UX problems arise as indicated by Batikan. e.g.
- the user selects item (pop up menu open) and item disappears because the item is deleted or 'paginated' to other page
- the user opens a modal dialog and the page in the background disappears, because item is deleted
- the user has one resource (e.g. deployment) selected and the resource is deleted.
I assume any solution that is simple and somehow reasonable is good enough...

bryk:
> BTW: a couple of UX problems arise as indicated by Batikan. e.g.

> the user selects item (pop up menu open) and item disappears because the item is deleted or 'paginated' to other page

> the user opens a modal dialog and the page in the background disappears, because item is deleted

I think we can stop autorefresh when a modal or menu is opened. Should prevent from things disappearing at the time you modify them.

> the user has one resource (e.g. deployment) selected and the resource is deleted.
I assume any solution that is simple and somehow reasonable is good enough...

Yes, agreed :) I guess when on details page of an object, we'll stop updating it when it is deleted.

batikanu:
@bryk Thank you for the feedback :) The websocket proposal is more promising for me too. I will check how can I implement it in the backend. 
@cheld     

> the user selects item (pop up menu open) and item disappears because the item is deleted or 'paginated' to other page

> the user opens a modal dialog and the page in the background disappears, because item is deleted

in the current solution, polling just waits if any menu is open and continues after the menu is closed. So these problems shouldn't occur.

marcacohen:
Hello! I'm new to this project (thanks to @IanLewis for pointing me here). The lack of auto-refresh was one thing I noticed right away so was happy to learn about this PR. Using this approach, I've experimented with adding auto-refresh to podlist and job list (commits on forked repo [here](https://github.com/marcacohen/dashboard/commit/cb500c9404197b141d572ea7117856db7f3095f7)) and it seems to work quite well. I see above the comments about observing vs. polling and agree that would be ideal. Is anyone actively working on this? (adding support for auto-refresh via backend notifications?) If not, I'd be happy to give it some time but didn't want to dup effort if someone else is already working on it.

batikanu:
Hi @marcacohen welcome. Actually I was working on this. I implemented basic websocket implementation by upgrading existing webservice and using watch for a sample resource. Frontend  implementation is missing. Unfortunately, I have another tasks right now to complete therefore I can only work on it in spare time. You can do it much faster I think. 

**Things to consider:**

- Websockets don't work with browsersync therefore you can only work with production as far as I know. There was a discussion about dropping browsersync but it requires additional effort to make it work because we use it to serve frontend in dev modus. It is also relevant to https://github.com/kubernetes/dashboard/pull/1455. I think it should be an additional pull request and has to be solved anyway in my opinion. 

- Current graphs are not refresh-able. I fixed it in this pool request. 

- Opening menus, popups should be taken into attention before making any update in the page to avoid magical disappearing menus or item in the list etc. I implemented a workaround in this pull request to avoid such scenarios. 

- The watches may trigger too many updates from backend. There might be some filter or throttling implementation needed. 

I can support you if you need anything.

marcacohen:
Hi @batikanu, thanks for the guidance! In reviewing your thoughts (esp. the concerns about testing and throttling), I have the sense that this may turn into a relatively complex undertaking, requiring subtle changes to both FE & BE. I'm wondering if it might be advisable to move ahead with your polling approach for now, to get something, even if suboptimal, working soon. 

The way you inhibit polling from all but the active view is relatively efficient, works with some simple client-only changes, and can be easily reused/extended to other pages. The main downside seems to be suboptimal responsiveness but slightly delayed refresh still beats none/manual refresh. 

So, I'm thinking let's finish this PR and get some improvements now (I'm happy to run with it if you're too busy) and then take on the more systemic notification approach as a follow on. Wdyt?

(cc @IanLewis)

batikanu:
Hi @marcacohen thanks for the feedback. Sure I can update the pull request to make it able to merge. 
What do others think? @bryk @rf232

bryk:
> Hi @marcacohen thanks for the feedback. Sure I can update the pull request to make it able to merge.
What do others think? @bryk @rf232

That would be awesome! :)

jwforres:
websockets are great and we use them almost exclusively in the openshift console, but a few warnings:

1) IE has a limitation that you can only have 6 websockets open for the same domain. kube today only gives you one websocket per resource type, so if you need to keep an eye on more than 6 resource types, you end up still having to poll some of them.  We opted to only fall back to polling some resources and only in IE. We make decisions on what to watch vs poll based on frequency of updates.  There is an open kube issue to support a multi-resource watch.

2) To watch a kube resource it is generally recommended to do a list, get the resourceVersion from that, and then open a watch from that resourceVersion.  OpenShift console has a lot of code in place already for handling list then watch, and handling of watch retries.  You will need retries because in a lot of environments there is a limited time a connection is allowed to remain open with no data transfer before it will get closed.  If you are interested this is our DataService https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/data.js

Another advantage to doing list then watch, if websockets fail to establish for whatever reason, you still have the initial load of data. (3) below is an example where this has been helpful.  But also environments where people have proxies or LBs and haven't set them up to work correctly with websockets.  Helps to diagnose things to see the List call work and then the Watch fail.

3) In Safari websockets + self-signed certs is problematic. If you just accept the cert using the cert exception dialog that pops up, it does not extend that acceptance to websocket connections (for some strange reason).  You have to have the cert accepted at the system level.

IanLewis:
I agree that we will want to move to some kind of websocket interface in the future rather than polling. OpenShift's console does pretty well here. That said, I think we should get the current implementation in first given limited bandwidth.

> 1) IE has a limitation that you can only have 6 websockets open for the same domain. kube today only gives you one websocket per resource type, so if you need to keep an eye on more than 6 resource types, you end up still having to poll some of them. We opted to only fall back to polling some resources and only in IE. We make decisions on what to watch vs poll based on frequency of updates. There is an open kube issue to support a multi-resource watch.

Dashboard has a server side component and clients don't directly hit the Kubernetes API so we could relatively easily handle the individual watch API websockets on the server side and only need to have 1 websocket going to the client. We could solve the IE issue that way.

> 3) In Safari websockets + self-signed certs is problematic. If you just accept the cert using the cert exception dialog that pops up, it does not extend that acceptance to websocket connections (for some strange reason).  You have to have the cert accepted at the system level.

This sounds gnarly. I guess that, as a practical matter, Safari won't work with websockets at all without a proper cert then? No terminal etc. Is there anything that can be done to mitigate it?

jwforres:
On Wed, Jan 25, 2017 at 9:56 PM, Ian Lewis <notifications@github.com> wrote:

> I agree that we will want to move to some kind of websocket interface in
> the future rather than polling. OpenShift's console does pretty well here.
> That said, I think we should get the current implementation in first given
> limited bandwidth.
>
>
>    1. IE has a limitation that you can only have 6 websockets open for
>    the same domain. kube today only gives you one websocket per resource type,
>    so if you need to keep an eye on more than 6 resource types, you end up
>    still having to poll some of them. We opted to only fall back to polling
>    some resources and only in IE. We make decisions on what to watch vs poll
>    based on frequency of updates. There is an open kube issue to support a
>    multi-resource watch.
>
> Dashboard has a server side component and clients don't directly hit the
> Kubernetes API so we could relatively easily handle the individual watch
> API websockets on the server side and only need to have 1 websocket going
> to the client. We could solve the IE issue that way.
>
Forgot to mention it also affects Edge.


>
>    1. In Safari websockets + self-signed certs is problematic. If you
>    just accept the cert using the cert exception dialog that pops up, it does
>    not extend that acceptance to websocket connections (for some strange
>    reason). You have to have the cert accepted at the system level.
>
> This sounds gnarly. I guess that, as a practical matter, Safari won't work
> with websockets at all without a proper cert then? No terminal etc. Is
> there anything that can be done to mitigate it?
>
IIRC if you hit 'Continue' on the dialog that pops up then that it is
effectively a one time allowance, and that situation is where the problem
occurs. Instead if you hit Advanced or whatever it says then there is an
"Always Allow" option you can choose, by picking that it will let the
websockets work.  It is definitely frustrating and even more so on iOS.  We
generally recommend Chrome or Firefox instead if you are working in an env
where you have self-signed certs.

batikanu:
@marcacohen @bryk 
It is rebased now.
I've activated auto-refresh polling for replicationcontrollerlist, replicasetlist, nodelist and podlist. It can be done for other list pages easily.
For detail pages it can be a bit tricky because of reused podlists but I think it is doable with some changes.
Don't forget to remove '::' from metrics parameter assignments in templates like here (metrics="::$ctrl.replicationControllerList.cumulativeMetrics"). Otherwise graphic components can't get model update. 
The tests are missing currently. 
Let me know if any change is needed.

src/app/frontend/replicasetlist/replicasetlist.html
```diff
@@ -34,3 +34,9 @@
   </kd-content>
 </kd-content-card>
 <kd-zero-state ng-if="$ctrl.shouldShowZeroState()"></kd-zero-state>
+
+<kd-auto-refresh source="::$ctrl.replicaSetListResource"
```
olekzabl: Another problematic scenario (here, I've no idea why it behaves so):

1. Open a new Dashboard window; go to the pod list.
2. Change the namespace to "kube-system".
3. After 10 seconds, the list of pods will change back to that for namespace "default".
   (However, the namespace selector will still show "kube-system").

src/app/frontend/common/autorefresh/autorefresh_directive.js
```diff
@@ -0,0 +1,81 @@
...
+       * @private
+       */
+      let kdPoll_ = kdPoll;
+
+      // Cancel all polls in case of state change to avoid accumulated polls.
+      /** @type {function()} */
+      let stateChangeEvent = scope.$root.$on('$stateChangeStart', () => {
+        kdPoll_.stopAll();
+      });
+      scope.$on('$destroy', stateChangeEvent);
```
olekzabl: I don't understand what's happening here.
Why are we hitting as high as scope.$root, and doing this as late as on $destroy?
I'm not claiming it's wrong, but could you explain?

batikanu: I can't catch $mdMenuOpen event in scope. I need to remove it on scope.$destroy to avoid triggering in unregistered pages as well. 
For $stateChangeStart I will fix it as sope.$on and $destroy event won't be needed for it because it is removed with the scope already. 

Any other suggestion for mdMenuOpen ?

src/app/frontend/common/autorefresh/autorefresh_directive.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Provides upload file mechanism and updates the model with file parameter for the selected file.
+ * @param {!./poll_service.PollService} kdPoll
+ * @return {!angular.Directive}
+ * @ngInject
+ */
+export default function autoRefreshDirective(kdPoll) {
+  return {
+    scope: {},
+    bindToController: {
```
batikanu: Ok I write short desc.

src/app/frontend/common/autorefresh/autorefresh_directive.js
```diff
@@ -0,0 +1,81 @@
...
+        let delay = attrs.delay ? attrs.delay : DEFAULT_DELAY;
+        let startDelayForRestart = () => {
+          timeout = setTimeout(() => {
+            kdPoll_.startAll();
+            isRestartInProgress = false;
+          }, delay);
+          isRestartInProgress = true;
+        };
+        if (isRestartInProgress && !timeout) {
+          clearTimeout(timeout);
```
batikanu: it is failure thanks. 
It should be 

let timeout;
      // Continue polling if any stopped after any menu is open.
      /** @type {function()} */
      let menuCloseEvent = scope.$root.$on('$mdMenuClose', () => {
      ...
        if (isRestartInProgress && timeout) {
          clearTimeout(timeout);
          startDelayForRestart();
          return;
        }
        startDelayForRestart();
      });

src/app/frontend/common/autorefresh/poll_service.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /**
+   * Starts polling and returns polling promise.
+   * @param resource {!angular.$resource}
+   * @param params {angular.Resource.ParamsOrCallback=}
+   * @param delay {number}
+   * @returns {!angular.$q.Promise}
+   */
+  poll(resource, params, delay) {
```
batikanu: Thank you for spotting the bug! It wasn't working as intended. I've created new method for pollService to remove all optionsList on state change. It was designed to keep multiple running polls on the same state. In this case it was holding the options from previous state which had to be removed from the list.

src/app/frontend/common/autorefresh/poll_service.js
```diff
@@ -0,0 +1,208 @@
...
+    options.interval = this.intervalService(() => {
+      let promise = options.resource_.get(options.params).$promise;
+      promise.then(
+          (obj) => {
+            options.deferred.notify(obj);
+          },
+          (err) => {
+            options.deferred.reject(err);
+            // Resource can't retrieve any record. It is not necessary to continue polling anymore.
+            this.stop();
```
batikanu: Correct. 
In this case the failing options is not necessarily current one. I will modify it to stop and remove the failing one instead of current one.

src/app/frontend/replicasetlist/replicasetlist.html
batikanu: I can't reproduce it in my version. Maybe related to optionList bug which I've solved in my copy.

src/app/frontend/common/autorefresh/autorefresh_controller.js
```diff
@@ -0,0 +1,133 @@
...
+   * @param {!./../../chrome/chrome_state.StateParams} $stateParams
+   * @param {!./../pagination/pagination_service.PaginationService} kdPaginationService
+   * @param {!./poll_service.PollService} kdPoll
+   * @ngInject */
+  constructor($stateParams, kdPaginationService, kdPoll) {
+    /**
+     * Initialized from the scope.
+     * @private {!angular.$resource}
+     */
+    this.source;
```
batikanu: Sure

batikanu:
@olekzabl Thank you for the testing. I made quick fixes. PTAL. 
I think I can't deliver tests very soon bcs I work on that only in free time. What about creating a different pull request for the tests?

src/app/frontend/replicasetlist/replicasetlist.html
olekzabl: Indeed, now I can't reproduce it as well.

src/app/frontend/common/autorefresh/autorefresh_directive.js
```diff
@@ -0,0 +1,86 @@
...
+    link: function(scope, element, attrs) {
+      /**
+       * @type {./poll_service.PollService}
+       * @private
+       */
+      let kdPoll_ = kdPoll;
+
+      // Cancel all polls in case of state change to avoid accumulated polls.
+      scope.$on('$stateChangeStart', () => {
+        kdPoll_.removeAll();
```
olekzabl: So this is the place where you fixed the failure scenario, right?
Indeed, now it works for me :)
Still, I have a feeling that the code would be a little safer if you just passed the "options" thing from poll() to start() and stop() as a parameter, rather than introducing "currentOptions" as a de-facto global variable. (I'm fine about "optionsList"). Or am I missing something? Anyway, I won't insist on that.

src/app/frontend/common/autorefresh/autorefresh_directive.js
```diff
@@ -0,0 +1,86 @@
...
+        /** @type {number} */
+        let delay = attrs.delay ? attrs.delay : DEFAULT_DELAY;
+        let startDelayForRestart = () => {
+          timeout = setTimeout(() => {
+            kdPoll_.startAll();
+            isRestartInProgress = false;
+          }, delay);
+          isRestartInProgress = true;
+        };
+        if (isRestartInProgress && timeout) {
```
batikanu: Previously If the user opens and closes menu several times in a row, then it was polling faster than delay parameter. I implemented this to avoid that situation but I realized that later I modified the logic in poll service to avoid this case. I can't reproduce it anymore. So it can be omitted :)

src/app/frontend/common/autorefresh/autorefresh_directive.js
batikanu: Honestly I can't remember exactly the scenario. It is long time ago but they were intended to be public methods can be called outside without knowing about options. However, I see that they are not used publicly anywhere. I will remove them completely. They can be implemented later in a suitable way according to the need.

batikanu: Yes it was the solution. It also solved namespaces issue I think.

batikanu:
@olekzabl thanks for the improvements. PTAL

batikanu:
@olekzabl It is easy to fix it but I've found bigger issue with pagination lists. 
If user is in page 2 and all the pods in page 2 are deleted before polling then user gets "There is nothing to display here" page. Other way around, if user has less than 10 pods which means single page and if new pods are created which require 2 pages, pagination arrows are not displayed for navigation. Currently I am thinking about how to update resourcecardlistpagination on each poll. Any ideas are more than welcome :)

maciaszczykm:
It will be done in a different way. Closing it as stale. We can still check it when needed.

Issue #1441, opened 2016-11-17
UI for managing RBAC-related resources


mlamina:
Following #964, I would like to draft a feature that would allow users to manage [Roles, RolesBindings, ClusterRoles, and ClusterRoleBindings](http://kubernetes.io/docs/admin/authorization/) in the dashboard.

**Issue Goal**
Develop use cases, UI workflows and wireframes for managing RBAC resources.

**Mockups**
There are clickable mockups in [InVision](https://invis.io/AF9E3F55W#/206606225_RBAC__List_Of_Roles).

**Initial Thoughts on Workflows**
- Roles and ClusterRoles are almost identical in how they are used/created, so the UI could display them as one single concept that has a flag (cluster yes/no)
- The UI could be centered around Roles, meaning that users would be presented with a list of existing roles which contains buttons for CRUD operations
- Existing roles could then have a _bind_ action, allowing the user to grant a particular permission to a user or set of users
- In general, the UI could be a simplified, more user-friendly way to manage a cluster's access control by abstracting the raw concepts of Roles, RolesBindings, ClusterRoles, and ClusterRoleBindings 

CC @rf232 @bryk

rf232:
cc @romlein

rf232:
Great! Also I love that this explicitly is about managing the RBAC resources and *not* about the dashboard using and handling RBAC related like issue #964 describes. Please keep it that way!

mlamina:
I agree. Deep integration of AuthN and AuthZ into the dashboard should be a separate topic.

mlamina:
Two quick ideas for a simple UI...

"Landing page" showing a list of existing roles. If namespace is set to "*", the role is a ClusterRole:

![list of roles](https://cloud.githubusercontent.com/assets/474314/20413893/4490f2b4-ace4-11e6-888f-169c6eeb736c.png)

------------------
Clicking on one of the list items would lead to a detail view for a role:

![role details](https://cloud.githubusercontent.com/assets/474314/20413912/645fbdc8-ace4-11e6-9b90-bafbbd464e3b.png)

------------------

There were a couple of things that popped into my mind while creating the wireframes:

- Where in the main menu should we put this section of the dashboard? Since it's namespace-independent, I would put it right next to Nodes and Persistent Volumes under the Admin section?
- The only difference between Roles and ClusterRoles is that ClusterRoles apply to any namespace. In the UI, this could be represented by either choosing a namespace for a role or selecting "*".
[...]

Issue #1423, opened 2016-11-15
Usage metrics


ianlewis:
The purpose of this issue is to discuss the feasabity of collecting some kind of usage metrics for the dashboard UI.

Brought up originally here: https://github.com/kubernetes/dashboard/issues/1354#issuecomment-255637633

bryk:
I'd really love to have this!

I remember having these discussion in the team and the consensus was that this is a hard technical problem in the environment we have. I.e., it is hard to collect metrics directly from the browser if users are running this in a non-Internet env. The same for collecting from the backend: if users run this in an isolated cluster, metrics would not get transmitted.

If anybody has ideas here, let us know. CC @rf232

danielromlein:
Has anyone have any ideas around telemetry for Dashboard? I'd love to be able to pull in a UX researcher to both validate the decisions we've been making and shape what we choose to do in the future.
/ @rf232 @bryk @maciaszczykm @floreks

IanLewis:
To answer @bryk's question I think this can be opt-in via a command line flag or something to the app in the Deployment and in that case we can assume they have access to the internet.

cheld:
This might help: https://github.com/kubernetes-incubator/spartakus.

floreks:
We don't really plan to automatically store and analyze information about how people use Dashboard. I don't think this will change unless we get many more contributors to be able to properly support such feature.

/close

Issue #1248, opened 2016-09-16
Display all info for a Pod


digitalfishpond:
##### Steps to reproduce

Navigate to the details page of any pod:
##### Observed result

![screenshot from 2016-09-16 11-32-14](https://cloud.githubusercontent.com/assets/13361549/18581534/5a19f766-7c01-11e6-9d3f-c440ebc703b7.png)
##### Expected result

Details should include all info from a pod, such as:

```
# Please edit the object below. Lines beginning with a '#' will be ignored,`
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.io/created-by: |
      {"kind":"SerializedReference","apiVersion":"v1","reference":{"kind":"ReplicationController","namespace":"default","name":"integration-test-valid-rc-saddssdadsasdadsadsasdasdasaddsasdas-sss","uid":"f926c650-1842-11e6-8952-42010af000fb","apiVersion":"v1","resourceVersion":"18735332"}}
    kubernetes.io/limit-ranger: 'LimitRanger plugin set: cpu request for container
      php-redis'
  creationTimestamp: 2016-09-05T13:01:10Z
  generateName: integration-test-valid-rc-saddssdadsasdadsadsasdasdasaddsasdas-sss-
  labels:
    app: guestbook
    tier: frontend2
  name: integration-test-valid-rc-saddssdadsasdadsadsasdasdasaddsapj3nx
  namespace: default
  resourceVersion: "18735345"
  selfLink: /api/v1/namespaces/default/pods/integration-test-valid-rc-saddssdadsasdadsadsasdasdasaddsapj3nx
  uid: d13174b3-7368-11e6-8d2f-42010af000fb
spec:
  containers:
  - image: gcr.io/google_samples/gb-frontend:v4
    imagePullPolicy: IfNotPresent
    name: php-redis
    ports:
    - containerPort: 80
      protocol: TCP
    resources:
      requests:
        cpu: 100m
    terminationMessagePath: /dev/termination-log
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: default-token-3kz4k
      readOnly: true
  dnsPolicy: ClusterFirst
  nodeName: gke-bryk-cluster-v1-2-default-pool-b632f6bc-2nvp
  restartPolicy: Always
  securityContext: {}
  serviceAccount: default
[...]

PR #1463, opened 2016-11-22
Design doc for access control management UI

mlamina:
WIP, trying to formalize the ideas from #1441.


bryk:
This is a good start @mlamina! I see that this is a template only as of now, which is fine. Should we marge it or wait for you to fill the contents? Both ways work for me.


In the meantime, try to fix the issue with the CLA. Click "details" link on the red line above :)

bryk:
Btw, join SIG-UI meeting tomorrow to talk about this, if you can. 

Details here: https://github.com/kubernetes/community/tree/master/sig-ui

mlamina:
I signed the CLA!
I would suggest to keep the doc in this PR right now to have an active discussion?

mlamina:
Aright, I worked on the design document to make it a little more complete. Would be great to get some feedback from @danielromlein :)

cheld:
CC @floreks, @kenan435

danielromlein:
Questions / thoughts after looking through the [design doc](https://github.com/mlamina/dashboard/blob/33439145126fca1ab03584d358219c7e05206d16/docs/design/access-control.md):
- I think it'd be great to flesh out the use cases more. _Why_ as a cluster admin do I want to create roles and assign them to users? 
- What sort of ranges do we expect to see for Roles and Bindings? If they're going to potentially grow large, might it make sense to use tabs for them, or to paginate the table?
- Do we want some view where a user can see both what a role’s permissions are _and_ to whom that roll is assigned? Currently only link between them.
- I'm struggling to understand Bindings... Would there ever be a reason to have multiple bindings for the same role in the same namespace?
- @bryk Would this require further engineering work to get implemented, or is Dashboard set up to handle RBAC today? I'm just curious.

@normalnorman your input around this would be super valuable, as you have more experience with PRDs than me 😊

bryk:
Sorry @mlamina if I'm stealing your answers here :) Feel free to correct me. 

> What sort of ranges do we expect to see for Roles and Bindings? If they're going to potentially grow large, might it make sense to use tabs for them, or to paginate the table?

Dozens of roles and bindings is what I'd expect for a large org. I think pagination or a way of handling large number of roles is a must here. 

>@bryk Would this require further engineering work to get implemented, or is Dashboard set up to handle RBAC today? I'm just curious.

We can handle RBAC admin as it is here easily. If a tricky role is then assigned to the UI pod, it'll break. In Q1 it should be handling all roles.

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+## Abstraction / Simplification
+
+The only difference between Roles and ClusterRoles is that Roles apply to a specific namespace 
+and ClusterRoles apply to all namespaces. The same is true for RoleBindings and ClusterRoleBindings. 
+The UI should help users to understand these concepts and user flows should be designed to enforce
+the correct usage. 
+
+NOTE: The design proposed in this document will combine certain concepts:
```
bryk: Interesting proposal. I think I like it. Simplifies the problem from user's perspective. We could say in some non-prominent place that * means that this is cluster role object, but overall this looks nice.

WDYT? @rf232 @normalnorman

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+* Edit a role's rules
+* Unbind a subject from a role
+
+# Design
+
+
+
+## Overview
+
+The overview can be reached by clicking the "Access Control" item in the admin section of the
```
bryk: Wondering about name here. Should this be called "Access Control"? Or maybe "RBAC"? "Roles and Bindings"? Maybe "IAM"? Or just "Roles" as bindings are auxiliary objects? 

Listing a couple of ideas to explore the space.

@rf232 @normalnorman what do you think?

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+left-hand menu. It shows existing Roles/Bindings, some important details for every item and allows
+the user to create a new Role or Binding right away.
+
+![Overview](mockups/21-11-2016-access-control/overview.png)
```
bryk: Mock LGTM

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Creating a role
+
+Clicking the "+" under the existing Roles will open a dialog. After entering the name and
```
bryk: This is a tricky part. We've never had such proposal for creating objects. All we had is this +create button, which is mildly useful:
image

I like the proposal here to move creation to the context of where it is used. We can apply this pattern project wide if it turns out to be good.


Other option would be to add "+create" button to the list toolbar, as this is our long term place to put things related to a list displayed below.  

@normalnorman @danielromlein What do you think re this?

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+picking a namespace ("*" will create a ClusterRole), clicking "Create" will create the Role
```
bryk: I'm wondering whether the `*` is too magic. From my talks with users I remember that they do not like magic.

How about changing `*` to `all namespaces` or `all namespaces (will create a ClusterRole)` or add `info` bubble somewhere there to explain this. All I want here is to be explicit to what we do.

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+
+Clicking the "+" under the existing Bindings will open a dialog. After the selecting the
+target role and the namespace ("*" will create a ClusterRoleBinding), clicking "Create"
+will create the Binding and redirect the user to the Binding's detail page.
+
+![Overview](mockups/21-11-2016-access-control/create-binding.png)
+
+## Edit a role
+
+The only property of a Role that cannot be changed is its namespace, because it determines
```
bryk: How do I get into the state of editing a role? What do I click?

docs/design/access-control.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Bind subjects to a role
+
+Clicking the "+" under the existing Bindings will open a dialog. After the selecting the
+target role and the namespace ("*" will create a ClusterRoleBinding), clicking "Create"
+will create the Binding and redirect the user to the Binding's detail page.
+
+![Overview](mockups/21-11-2016-access-control/create-binding.png)
+
+## Edit a role
```
bryk: Do you propose a page that would show details of a role? E.g., the pages that we currently have for other objects.

docs/design/access-control.md
mlamina: Yes, I think `all namespaces` would be more explicit and having an info bubble for further clarification is a good idea!

docs/design/access-control.md
mlamina: My idea was to have clickable list items in the overview. Clicking an item leads to the edit page.

docs/design/access-control.md
bryk: I would kind of expect a page to show single binding or single role. Just to be parallel with all other views in the UI. But I see your point that such view is not required.

I guess this is call to action for out friendly UX designers who should have opinions on this. 

@danielromlein @normalnorman Can you suggest something here?

docs/design/access-control.md
bryk: Typically what we do in other views is that when we click on a resource, we show its details page. Then there user can edit the thing. Does this sound sane?

danielromlein: To me it makes sense to have a detail page with an edit option there. RE my earlier question
>Do we want some view where a user can see both what a Role’s permissions are _and_ to whom that Role is assigned?

The detail page would give us the opportunity to display information like, for example, what Bindings a given Role applies to (if that's important to users...?)

mlamina: Good idea! Would anyone mind to make some changes in the mockups? I joined a new project at work and am quite busy at the moment, but I'll do my best to keep the ball rolling!

danielromlein: @mlamina I'll try to make changes! Right now my priority is the Pod Details page #1248, but this is next on my list.

mlamina:
@bryk @danielromlein @normalnorman I updated the mockups and the proposal. There is now a separate detail page for roles and instead of the "*", we use "all namespaces". I thought about adding a separate detail page for Roles as well, but there wouldn't be any difference between the detail page and the edit page, because all the user would be able to do is add or remove subjects. Ideas? Comments?

docs/design/access-control.md
mlamina: IMHO "Access Control" is the most precise and unambiguous term?

danielromlein:
One Role can have multiple Bindings, correct? I'm not seeing how that would work in the current [wireframe](https://github.com/mlamina/dashboard/blob/7fe272e1bef40d7cd1f3b3df146a289d24d9a428/docs/design/access-control.md#view-a-role). Is there a wireframe accommodating that?

I think we'd agreed putting the _Create_ buttons up [at the top of each card](https://invis.io/969TJXX4B#/208427844_Wireframe) was a better solution and more in-line with the Material spec than below.

👍  on the change to "All Namespaces."

mlamina:
@danielromlein yes, a Role can have multiple Bindings. The "Role detail view" you referenced even shows it: The "details" section shows the Role's properties while the "Bindings" section below that lists the Bindings associated with that Role.

You're absolutely right, the Create buttons should be at the top of each card, I'll make that adjustment. Great mockups by the way! :)

bryk:
Lets merge! @mlamina let us know when you need any help coding this.

mlamina:
Thanks @bryk ! Should I just create another pull request and get on with it or are there any procedures / best practices that you guys would like me to follow?

bryk:
Just create a pull request :)

danielromlein:
Sorry I missed this @mlamina! Yeah go ahead and create a PR this LGTM.

mlamina:
https://github.com/kubernetes/dashboard/pull/1576

mlamina:
OK so I started implementing and found that the design is missing an important detail: I assumed that a role simply consists of a set of verbs and a set of resources. However, A role can have multiple rules and each of these rules contains one set of resources and one set of verbs. This clearly not only impacts the design, but also the implementation. Now I see two possible ways to handle this:

1. Adjust the design to account for the 1:n relationship between roles and rules
2. Restrict the implementation to a 1:1 relationship

Option (1) is obviously the easy way out and might need to be fixed in the future. However, option (2) would probably have a heavy impact on the design document. Would love to hear your opinions!

bryk:
I'd suggest adjusting the design. We should create UIs that handle most cases. If we handle only 1:1 relationships this would be unusable for users who have 1:n relationships.

How much does this affect the implementation? Can you comment?

mlamina:
Yes, I agree! The backend part of the implementation should not be affected too much, it's just a small change in data types. However the fronted would look very different, since every role can now have a list of "rules" instead of just some verbs and resources. I'm not a designer, so if somebody with more expertise in this could take a look at it that would be cool! :)

danielromlein:
Interactive [wireframes](https://invis.io/2GA8MURUF) incorporating the concept of 1:n Rules per Role. 
@mlamina PTAL
I still don't have a good understanding of API groups.
>look at them simply as a list of strings

☝️ I don't know what you mean by this...?

mlamina:
OK maybe check out [this link](https://kubernetes.io/docs/api/) under "API Groups"? They are a concept to split API endpoints into groups, which makes them easier to upgrade and manage.
In the context of Roles and their rules, this is how you can think of the semantics of a rule:

> Allow all subjects bound to this rule to apply [list of verbs] to [list of resources] under [list of API groups]

[list of X] is simply a list of strings, so for example we could have something like:

apiGroups: ["", "rbac.authorization.k8s.io", "some.other.group"]
resources: ["pods"]
verbs: ["get", "watch", "list"]

danielromlein:
Ok @mlamina that makes sense – thanks for the explanation!
WDYT about the [wireframes](https://invis.io/7Z9NNP4E8)?
cc @normalnorman thoughts on this IA pattern?
/ @bryk

mlamina:
Oh right, forgot my feedback on the wireframes :)
- I love the split pane design! 
- When creating a role, it might be useful to have a "done" or "create" button to finish the action
- The popup for creating a rule is missing a field for API groups
- What does the "+ Create" button in the upper right corner do?
- In the popup for creating a RoleBinding:
  - the first dropdown is probably supposed to say "Role" for selecting a Role, not "Rule"?
  - the "Role Type" can be inferred from the selected Role, so we could simplify the view by removing this field
  - Is  the idea to only have one subject to begin with or to be able to add multiple subjects right away? Just wondering because there is no "+ Subject" button

danielromlein:
@mlamina thanks! 👍 
 > When creating a role, it might be useful to have a "done" or "create" button to finish the action

There's Create / Cancel, but your missing them might indicate that they should be closer in proximity to inputs 😊 
 > The popup for creating a rule is missing a field for API groups

Should this be a multi-select pattern?
 > What does the "+ Create" button in the upper right corner do?

That's our global "Create" that's present across all pages that brings a user to the _Deploy a Containerized App_ page. We haven't tested that very methodically though to assess if people get confused with page-specific actions.

> the first dropdown is probably supposed to say "Role" for selecting a Role, not "Rule"?

Correct. Changed.
> the "Role Type" can be inferred from the selected Role, so we could simplify the view by removing this field

Great. Changed.
> Is the idea to only have one subject to begin with or to be able to add multiple subjects right away? Just wondering because there is no "+ Subject" button

Yes. My thought was once a user had entered values for both inputs, another set of inputs would appear below. Think that's clear enough?

I should have recommended this, but for future feedback it's easier to track in InVision itself (toggle on "Comments" in the lower RH corner, then click anywhere to leave a comment)

Where I'd like to go next with this is to mock it up in Sketch / InVision so we have a clickable prototype that we could test with Kubernetes users.

mlamina:
> Should this be a multi-select pattern?

I'm not exactly sure what the multi-select pattern looks like, but the end result needs to be a list of (possibly arbitrary) strings!

> That's our global "Create" that's present across all pages that brings a user to the Deploy a Containerized App page

For me personally it was a bit confusing to see that button in the _Access Control_ section. I've always wondered: Are there usage statistics / metrics being collected in the dashboard? Do we know which parts of the UI are used and in which way?

> My thought was once a user had entered values for both inputs, another set of inputs would appear below

Cool! I like it :)

> I should have recommended this, but for future feedback it's easier to track in InVision itself

Good idea, will do that next time! Looking forward to seeing the clickable prototype!

danielromlein:
>Are there usage statistics / metrics being collected in the dashboard?

There are not, but I agree that this would be tremendously valuable to UX. We opened #1423 to explore the subject of telemetry but it hasn't gone anywhere. If you're willing to drive that forward I'd encourage you to go for it!

danielromlein:
Interactive [InVision prototype](https://invis.io/9HAB8JPMW) I'd like to get people's feedback on. Click around and let me know what you think of the flow. Are any use cases not accounted for?

Please comment using the built-in [InVision functionality](https://support.invisionapp.com/hc/en-us/articles/209192426-How-do-I-comment-on-a-prototype-) so we can keep track of what comments apply to which areas of the UI.

cc @mlamina @bryk @rf232

bryk:
I looked at InVision mocks. They look very solid! I left a few comments.

PR #1465, opened 2016-11-22
Bump version to 1.5.0-beta1 in prep for release

bryk:


rf232:
should package.json also be updated?

bryk:
PTAL

PR #1467, opened 2016-11-22
Display errors on pod list page

floreks:
- Added events retrieval on pod list page and applied same logic for error recognition as on other list pages.
- Extracted `podcard_component` from `podcardlist_component`

zrzut ekranu z 2016-11-22 14-45-51

### DIFF
--- a/src/app/backend/resource/node/nodedetail.go
+++ b/src/app/backend/resource/node/nodedetail.go
@@ -218,7 +218,7 @@ func GetNodePods(client k8sClient.Interface, heapsterClient client.HeapsterClien
 		return nil, err
 	}
 
-	podList := pod.CreatePodList(pods.Items, dsQuery, heapsterClient)
+	podList := pod.CreatePodList(pods.Items, []api.Event{}, dsQuery, heapsterClient)
 	return &podList, nil
 }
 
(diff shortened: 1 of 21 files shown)


bryk:
Checked how it works. This is totally awesome. I'm wondering why we didn't have it before...

bryk:
Btw, CPU usage column is broken. See NPEs in the console. Typo?

floreks:
> Checked how it works. This is totally awesome. I'm wondering why we didn't have it before...

I don't know either :) We have it on other list pages and logic is pretty much the same. We can also implement that on detail pages of resources that display pod list.

> Btw, CPU usage column is broken. See NPEs in the console. Typo?

Yeah, there is a typo. Missing `$` before `ctrl`. Will fix that tomorrow morning.

floreks:
Fixed typo. Rebased the code.

bryk:
> I don't know either :) We have it on other list pages and logic is pretty much the same. We can also implement that on detail pages of resources that display pod list.


I had a comment here that for some reason I did not post it... My comment was why we're sending empty lists of events in details pages? How hard would it be to get events there and send them out? AFAIK, the events are already downloaded, so we need to pass them through. Should this be done in this PR or another?

src/app/backend/resource/node/nodedetail.go
```diff
@@ -218,7 +218,7 @@ func GetNodePods(client k8sClient.Interface, heapsterClient client.HeapsterClien
 		return nil, err
 	}
 
-	podList := pod.CreatePodList(pods.Items, dsQuery, heapsterClient)
+	podList := pod.CreatePodList(pods.Items, []api.Event{}, dsQuery, heapsterClient)
```
bryk: Look at line 136&138 - we already have events to pass there :)

bryk: All righty, let's merge this then :)

Issue #1447, opened 2016-11-18
Verify that the UI works on safari, firefox and IE


bryk:
A process to verify this for 1.5. @rf232 can you add this to the project?

rf232:
Related to #794, close that one if the UI works on IE.

IanLewis:
What can users expect for browser support again? What versions are supported?

bryk:
Latest versions of evergreen browsers + latest IE

bryk:
Re what is support: UI works there in sane way

bryk:
Once #1481 is merged, we can close this. I did a roundtrip of all browsers and everything seems to work.

PR #1481, opened 2016-11-24
Fix navigation menu on IE

bryk:
* nav elements were collapsing into each other
* scrollbar was overlaping with the toolbar

Re #1447

Tested on other browsers. All ok there.


bryk:
@digitalfishpond Can you review this?

PR #1482, opened 2016-11-24
Fix an issue when showing lists of identical containers

### LINKED ISSUE #1475: PetSet view does not show multi-container pods with same image ("Duplicates in a repeater are not allowed")
#### Issue details

See screenshot: My PetSet pods are not show, an error is logged in the console.

##### Environment

Minikube 0.12.2 environment, petset with a two-container-pod template, both containers use the same image (but run different processes).

```
Dashboard version: 1.4.2 (as included in minikube 0.12.2: gcr.io/google_containers/kubernetes-dashboard-amd64:v1.4.2)
Kubernetes version: client 1.4.4, server 1.4.3
Operating system: Fedora Linux 24
Node.js version: 7.1.0
Go version: (N/A)
```

##### Steps to reproduce

1. Define a petset like this:
~~~~
# Internal "Headless" service for the petset
kind: Service
apiVersion: v1
metadata:
  name: service
  labels:
    service: service
spec:
  ports:
  - port: 8080
    name: http
  clusterIP: None
  selector:
    service: service
---
kind: PetSet
apiVersion: apps/v1alpha1
metadata:
  name: petset
spec:
  serviceName: service
  replicas: 3
  template:
    metadata:
      labels:
        service: service
      annotations:
        pod.alpha.kubernetes.io/initialized: "true"
    spec:
      containers:
        - name: container1
          image: busybox
          command:
          - /bin/sh
          - -c
          - "while true; do sleep 5; done"
        - name: container2
          image: busybox
          command:
          - /bin/sh
          - -c
          - "while true; do sleep 5; done"
  volumeClaimTemplates:
  - metadata:
      name: data
      annotations:
        volume.alpha.kubernetes.io/storage-class: "slow"
    spec:
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 4Gi

~~~~

(Simplified down from my mongo example)

##### Observed result

See screenshot: no pods visible.

##### Expected result

Visible pods :)

This does work when there is only one container in the pod.

screen-kube_dashboard-petset-dup-repeater

bryk:
The UI would crash when showing lists of identical containers in a couple of places.

Fixes #1475

### DIFF
--- a/src/deploy/kubernetes-dashboard.yaml
+++ b/src/deploy/kubernetes-dashboard.yaml
@@ -19,17 +19,19 @@
 kind: Deployment
 apiVersion: extensions/v1beta1
 metadata:
-  name: dashboard-same-img
-  namespace: default
+  labels:
+    app: kubernetes-dashboard
+  name: kubernetes-dashboard
+  namespace: kube-system
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app: dashboard-same-img
+      app: kubernetes-dashboard
   template:
     metadata:
       labels:
-        app: dashboard-same-img
+        app: kubernetes-dashboard
       # Comment the following annotaion if Dashboard must not be deployed on master
       annotations:
         scheduler.alpha.kubernetes.io/tolerations: |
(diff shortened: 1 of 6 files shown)


src/deploy/kubernetes-dashboard.yaml
```diff
@@ -19,19 +19,17 @@
 kind: Deployment
```
maciaszczykm: Isn't it committed by mistake? :)

bryk: Ooops, my mistake. Removed. PTAL :)

bryk:
Updated, PTAL

PR #1484, opened 2016-11-24
Fix pod status sections in the RC and RS details pages

bryk:
We had a lot of typos in templates.


bryk:
@digitalfishpond Can you review this? :)

PR #1490, opened 2016-11-25
Decode b64 secret values in the UI

bryk:
They were not decoded by mistake. Once you decode them, the output is the same as in kubectl.


bryk:
@rf232 Can you review this?

PR #1491, opened 2016-11-25
Fix pod list cores column: we had a typo in template

bryk:


bryk:
@olekzabl Can you review this?

PR #1495, opened 2016-11-28
Add a info message that the memory usage graph includes caches

### LINKED ISSUE #1432: Add documentation tooltip to metric graphs
Frequently I hear users complaining that the metric values we display are incorrect. But in fact, they didn't know what the value was. This is obviously our fault, because we do not explain the problem. 

How about we add, e.g., a documentation bubble explaining what the metrics there are?

Example problem: https://github.com/kubernetes/heapster/issues/1356

rf232:
closes #1432 

memusage-notooltip

memusage-tooltip

### DIFF
--- a/src/app/frontend/poddetail/poddetail.html
+++ b/src/app/frontend/poddetail/poddetail.html
@@ -20,7 +20,7 @@ limitations under the License.
                  selected-metric-names="'cpu/usage_rate'">
   </kd-graph-card>
   <kd-graph-card graph-title="[[Memory usage|Title for graph card displaying memory metric of one pod.]]"
-                 graph-info="[[The memory usage includes the caches in the containers running on this pod.|Help message detailing what is included in the memory usage]]"
+                 graph-info="[[The memory usage includes the caches in the pods running on this pod.|Help message detailing what is included in the memory usage]]"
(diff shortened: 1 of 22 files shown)


rf232:
PTAL @bryk

src/app/frontend/poddetail/poddetail.html
```diff
@@ -20,6 +20,7 @@
                  selected-metric-names="'cpu/usage_rate'">
   </kd-graph-card>
   <kd-graph-card graph-title="[[Memory usage|Title for graph card displaying memory metric of one pod.]]"
+                 graph-info="[[The memory usage includes the caches in the pods running on this pod.|Help message detailing what is included in the memory usage]]"
```
bryk: Fix?

src/app/frontend/podlist/podlist.html
```diff
@@ -20,6 +20,7 @@
                  selected-metric-names="'cpu/usage_rate'">
   </kd-graph-card>
   <kd-graph-card graph-title="[[Memory usage|Title for graph card displaying memory metric of pods.]]"
+                 graph-info="[[The memory usage includes the caches in the pods running in these pods.|Help message detailing what is included in the memory usage]]"
```
bryk: Same here.

PR #1498, opened 2016-11-28
Fix time tooltip on podlist

rf232:
Time display used wrong reference
Also the time format was the wrong format. To prevent that happening in the future decorated the default format to use a different default format and timezone.


rf232:
PTAL @bryk

Issue #1497, opened 2016-11-28
Chrome wont load the dashboard


bpasson:
#### Issue details
Chrome will not start the dashboard due to strict mime type checking. The appConfig.json
has mime type text/plain and chrome wont execute it with that mime type. 

##### Environment
```
Chrome version: 54.0.2840.98 (64-bit)
Dashboard version: 1.5.0-beta1
Kubernetes version: 1.4.6
Operating system: Ubuntu 16.04
Node.js version: n.a. 
Go version: n.a.
```

##### Steps to reproduce
To reproduce the issue use the following steps:
* Deploy the dashboard using kubectl create -f https://rawgit.com/kubernetes/dashboard/master/src/deploy/kubernetes-dashboard.yaml
* Open the dashboard in Chrome 

##### Observed result
Chrome shows a grey page. 

##### Expected result
I would expect the dashboard to show.

PR #1499, opened 2016-11-29
Set content type on app config handler

bryk:
Fixes: #1497


bryk:
@digitalfishpond Can you review this?

PR #1500, opened 2016-11-29
version bump for 1.5.0 release

rf232:


rf232:
Getting version number up, we should then also push image and send PR to minikube and core to use the new images

actual release is questionable, do we want to do this together with core or just do it now?

bryk:
> actual release is questionable, do we want to do this together with core or just do it now?

Do this now. We need to push images to make all our dependencies work. We can theoretically update tags later, but we've been doing this in the past and people didn't like it (nor me - updating existing tags is doubtful...)

PR #1501, opened 2016-11-30
added prometheus metrics for api calls

cheld:
I added prometheus instrumentation to dashboard backend. I think most Kubernetes components do so, except Dashboard.

Raw metrics are available at /metrics

I tried keep the format similar to k8s-apiserver, but I had two problems:
1. I could not distinguish between Get and List easily. So everything is GET or PUT
2. I could not describe the resource properly in all cases. 

Of course, both aspects could be achieved by cluttering the dashboard-apihandler a bit. Maybe there is another clever option.

So, overall the metrics are not as precise as for kubernetes apiserver. Unless someone has clever idea, I would suggest to start with current solution and improve later on.

Some code changes to consider:
1. I added a dependency to k8s.io/kubernetes/pkg/util/net. The utility method is not of much use. I can copy or remove if we try to get rid of k8s dependency.
2. I wanted to add a simple test for the metrics collection. Basically check that no error is thrown. And verify that metric reached prometheus. However, I could not read from prometheus. Suggestion welcome...

BTW: after this is merged we should instrument more code. E.g. I would suggest to _consider_ adding metrics at places we make debug output.

### DIFF
--- a/src/app/backend/handler/apihandler.go
+++ b/src/app/backend/handler/apihandler.go
@@ -60,6 +60,7 @@ import (
 	clientK8s "k8s.io/kubernetes/pkg/client/clientset_generated/internalclientset"
 	"k8s.io/kubernetes/pkg/client/unversioned/clientcmd"
 	"k8s.io/kubernetes/pkg/runtime"
+	utilnet "k8s.io/kubernetes/pkg/util/net"
 )
 
 const (
@@ -77,6 +78,29 @@ type APIHandler struct {
 	heapsterClient client.HeapsterClient
 	clientConfig   clientcmd.ClientConfig
 	verber         common.ResourceVerber
+}
+
+func wsMetrics(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
+	startTime := time.Now()
+	verb := req.Request.Method
+	resource := mapUrlToResource(req.SelectedRoutePath())
+	client := utilnet.GetHTTPClient(req.Request)
+	chain.ProcessFilter(req, resp)
+	code := resp.StatusCode()
+	contentType := resp.Header().Get("Content-Type")
+	if resource != nil {
+		Monitor(verb, *resource, client, contentType, code, startTime)
+	}
+}
+
+// Extract the resource from the path
+// Third part of URL (/api/v1/<resource>) and ignore potential subresources
+func mapUrlToResource(url string) *string {
+	parts := strings.Split(url, "/")
+	if len(parts) < 3 {
+		return nil
+	}
+	return &parts[3]
 }
 
 // Web-service filter function used for request and response logging.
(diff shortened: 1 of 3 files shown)


cheld:
@floreks can you have a look?

bryk:
I need to understand this: so you propose to expose prometheus instrumentation data points?

floreks:
As I see this is similar to what kubernetes does. It will expose data that we can use for instrumentation, i.e. request times.

Overall LGTM. It just needs `gofmt`. It would be also good to have some tests if it's possible.

bryk:
> Raw metrics are available at /metrics

What other addons do here? Can this be configurable? Can we maybe allow for multiple implementations, e.g., prometheus is on "/metrics/prometheus" and stackdriver on "/metrics/stackdriver"?

cheld:
@bryk 

Currently, the endpoint serves latency metrics for API calls and some default go metrics for memory and threads etc. It is similar to other k8s components like etcd, apiserver, kubelet and probably more.

kubernetes only exports to prometheus format. In my understanding prometheus is the defacto standard. I am not sure if stackdrive can scrape metrics. I did a quick research, but I could not find anything. So I don't think multiple formats such as /metrics/stackdriver makes sense as no alternatives are available, anyway. In OpenStack the plan is to provide an agent that scrape metrics in prometheus format and forwards it to OpenStack. Stackdriver could do the same

The metrics data can be used in various ways:
1. it can be used by prometheus (or other tools) to create nice monitoring dashboards
2. it can be used for system testing. E.g. this is the official kubernetes performance dashboard http://146.148.52.109/#/builds . The data is processed using prometheus metrics. It is also used by kubernetes integration tests in the build scripts to detect slow api calls.
3. Users can inspect a running system. The data is potentially also available in the logs, however you need a tool like ELK to aggregate and digest the logs into numbers

src/app/backend/handler/apihandler.go
```diff
@@ -79,13 +80,25 @@ type APIHandler struct {
 	verber         common.ResourceVerber
 }
 
+func wsMetrics(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
+	startTime := time.Now()
+	verb := req.Request.Method
+	resource := strings.Split(req.SelectedRoutePath(), "/")[3]
```
cheld: agree

cheld:
BTW: If the performance dashboard is very slow :-) trust me, it is not broken, just slow

bryk:
Ok, this looks cool to me then. @rf232 Can you also take a look?

src/app/backend/handler/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+
+func ResetMetrics() {
+	requestCounter.Reset()
+	requestLatencies.Reset()
+	requestLatenciesSummary.Reset()
+}
+
+
+
+// Small optimization over Itoa
```
rf232: Is this optimization really needed, 

if so, the go compiler should be fixed :)

cheld: Yea you are write. Actually I have seen this snippet many times in go code, that's why I have blindly copied the idiom.

cheld:
PTAL

rf232:
can you rebase/merge with master, then travis will pick up my change to get protractor running again

Issue #1349, opened 2016-10-20
Some waiting pods have incorrect icons


bryk:
I recently checked my dev cluster and saw this: 
![selection_135](https://cloud.githubusercontent.com/assets/1159999/19553074/ee74d1b2-96b2-11e6-8211-f9fe439fe1f5.png)

Everything is green or gray, but it should not be. Pods prefixed with `ewa-app-` should have red icon. The `maxmaxmax...` one should be potentially red and `mysql` perhaps too. 

cc @IanLewis @rf232

floreks:
This is similar to #1121. Origin of this issue is the same, because pod is considered running/waiting by the kubernetes itself and containers are failing. Right now icon reflects only pod state not pod with containers and this should be changed.

Additionally on pods page we don't handle events and our error recognition is based on them.

IanLewis:
The icon is still looking at the pod state currently and it's "Running" for
those with pods in a crash loop.

I explicitly left changing this behavior out of my last change as it needed
some discussion. We need to define what situations should show red icons vs
green. I think we may need a yellow too for initializing pods, pods with
containers waiting for an image pull etc.

I'm willing to take this on as it's relatively simple change. The hard part
is deciding on design

IanLewis:
I think we can close #1121 as addressed by my earlier changes and use this issue for discussion of the icon.

bryk:
If I was to propose the algorithm, I'd say:
1. Only pods with status equal to "Running" have green icon.
2. "Waiting" status may be red or grey. I'd go through all waiting reasons and make categorize them using common sense. If a reason ends with "BackOff", I'd make it red. 

What do you think?

rf232:
Sounds like a nice algorithm

IanLewis:
For reference, here are the definitions of some errors we need to care about:

https://github.com/kubernetes/kubernetes/blob/4f32cc073b68850979e648996df0b09cd11a2263/pkg/kubelet/images/types.go
https://github.com/kubernetes/kubernetes/blob/8fd414537b5143ab039cb910590237cabf4af783/pkg/kubelet/container/sync_result.go

IanLewis:
Should we use things like the pod conditions (Initialized, Ready) for this. I'm wary of using the status string as we display it here to run the algorithm.

bryk:
> Should we use things like the pod conditions (Initialized, Ready) for this. I'm wary of using the status string as we display it here to run the algorithm.

[...]

PR #1502, opened 2016-11-30
Check pod conditions when displaying state icon

IanLewis:
Shows an error icon in pod lists when the pod's readiness is false. This has the benefit of not relying on events. It should also show an error icon when the pod is failing readiness probes.

I'll clean this up and add tests if the concept looks ok. It's meant as a solution for #1349

### DIFF
--- a/src/app/frontend/podlist/podcard_component.js
+++ b/src/app/frontend/podlist/podcard_component.js
@@ -82,7 +82,24 @@ export class PodCardController {
    * @export
    */
   isFailed() {
-    return this.pod.podStatus.podPhase === 'Failed' || this.hasWarnings();
+    if (this.pod.podStatus.podPhase === 'Failed' || this.hasWarnings()) {
+      return true;
+    }
+
+    // Check the pod's readiness state.
+    let ready = false;
+    let initialized = false;
+    for (let i = this.pod.podStatus.conditions.length - 1; i >= 0; i--) {
+      let state = this.pod.podStatus.conditions[i];
+      if (state.type == "Initialized") {
+        initialized = state.status === "True";
+      }
+      if (state.type == "Ready") {
+        ready = state.status === "True";
+      }
+    }
+
+    return initialized && !ready;
   }
 
   /**
--- a/src/app/frontend/podlist/podcard_component.js
+++ b/src/app/frontend/podlist/podcard_component.js
@@ -66,7 +66,7 @@ export class PodCardController {
    */
   isPending() {
     // podPhase should be Pending if init containers are running but we are being extra thorough.
-    return this.pod.podStatus.status === "pending";
+    return this.pod.podStatus.status === 'pending';
   }
 
   /**
@@ -74,7 +74,7 @@ export class PodCardController {
    * @export
    */
   isSuccess() {
-    return this.pod.podStatus.status === "success";
+    return this.pod.podStatus.status === 'success';
   }
 
   /**
@@ -83,7 +83,7 @@ export class PodCardController {
    * @export
    */
   isFailed() {
-    return this.pod.podStatus.status === "failed";
+    return this.pod.podStatus.status === 'failed';
   }
 
   /**
--- a/src/app/frontend/podlist/podcard_component.js
+++ b/src/app/frontend/podlist/podcard_component.js
@@ -65,7 +65,8 @@ export class PodCardController {
    * @export
    */
   isPending() {
-    return !this.hasWarnings() && this.pod.podStatus.podPhase === 'Pending';
+    // podPhase should be Pending if init containers are running but we are being extra thorough.
+    return (this.pod.podStatus.podPhase === 'Pending' || !this.isInitialized()) && !this.isFailed();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/app/frontend/podlist/podcard_component.js
```diff
@@ -82,7 +82,24 @@ export class PodCardController {
    * @export
    */
   isFailed() {
-    return this.pod.podStatus.podPhase === 'Failed' || this.hasWarnings();
+    if (this.pod.podStatus.podPhase === 'Failed' || this.hasWarnings()) {
```
floreks: I'm thinking about this check and I think it should be changed, because it will be true in case there are some warnings and pod is in running state. Pod state should be more important than events.

I think it should be: `if(podPhase === 'Failed' || (hasWarnings() && podPhase !== 'Running'))`

What do you think?

bryk: Agreed.

src/app/frontend/podlist/podcard_component.js
```diff
@@ -82,7 +82,24 @@ export class PodCardController {
...
+      return true;
+    }
+
+    // Check the pod's readiness state.
+    let ready = false;
+    let initialized = false;
+    for (let i = this.pod.podStatus.conditions.length - 1; i >= 0; i--) {
+      let state = this.pod.podStatus.conditions[i];
+      if (state.type == "Initialized") {
+        initialized = state.status === "True";
```
bryk: Should we maybe alter the status text here? Or do we do this already?

src/app/frontend/podlist/podcard_component.js
IanLewis: Yep. I pointed that out on #1349 but didn't implement it here yet.

src/app/frontend/podlist/podcard_component.js
IanLewis: What do you mean by alter the status text? You mean make this a proper boolean in the JSON returned by the AJAX api?

IanLewis:
Since we seem to be liking this idea I'll update this PR with proper logic and tests.

src/app/frontend/podlist/podcard_component.js
bryk: I meant that this should be also included in the `getDisplayStatus` method. I.e., when we say that status is red, the status text column should say: "Running: not initialized yet" or similar.

src/app/frontend/podlist/podcard_component.js
IanLewis: What about Succeeded state? should we show an error if the pod is Succeeded but has warnings somehow?

IanLewis:
This is a bit more complicated than I thought since there are issues to take into account like containers that terminate with success or failure statuses, init containers and their statuses. I plan to rewrite this logic a bit and update this PR.

IanLewis:
@floreks @bryk I updated the PR logic. Still no tests but I'd like you to take a look at the logic and test it to make sure it's sound.

I didn't update the getDisplayStatus() because I think it works well as is. Let me know if there's a specific status that you want me to support or change the status of.

Here are some sample manifests I used for testing: https://gist.github.com/IanLewis/d55c133751ca1d2db638c53af798e515

The pod statuses look something like this:

selection_007

src/app/frontend/podlist/podcard_component.js
```diff
@@ -82,7 +83,59 @@ export class PodCardController {
    * @export
    */
   isFailed() {
-    return this.pod.podStatus.podPhase === 'Failed' || this.hasWarnings();
+    if (this.pod.podStatus.podPhase === 'Failed') {
```
bryk: For some reason now none of my pods are "Green". Everything is pending or failed.

src/app/frontend/podlist/podcard_component.js
```diff
@@ -82,7 +83,59 @@ export class PodCardController {
...
+    }
+    return ready;
+  }
+
+  /**
+   * Checks if pod status is initalized. i.e. pod init containers have run. 
+   * @return {boolean}
+   * @export
+   */
+  isInitialized() {
```
bryk: I think this should return true when there are no init containers. WDYT?

bryk: When I change default to true here, pods have reasonable statuses now.

src/app/frontend/podlist/podcard_component.js
```diff
@@ -82,7 +83,59 @@ export class PodCardController {
...
+    return this.isInitialized() && !this.isReady();
+  }
+
+  /**
+   * Checks if pod status is ready. i.e. it's running and passess readiness checks.
+   * @return {boolean}
+   * @export
+   */
+  isReady() {
+    let ready = false;
```
bryk: Should this default to true? I.e., are we sure that pod condition of type "Ready" will always be there?

bryk:
High level comment: should this be done on the backend. The more I think about it the more I feel this is backend's responsibility. The reason is that other components should be able to reuse the logic here. For example, a Deployment should perhaps turn red when one of its pods is red. What do you think?

IanLewis:
I don't totally disagree but I don't think we should go overboard with
reusing the pod status. I think for a sufficiently large set of pods, some
being in a failure state is perfectly normal. Also makes it a bit expensive
to compute. I don't think failure states for a Deployment/ReplicaSet/etc
isn't trivial.

I'd be willing to move it to the backend if/when we could find a
sufficiently good use case.

bryk:
> I don't totally disagree but I don't think we should go overboard with
reusing the pod status. I think for a sufficiently large set of pods, some
being in a failure state is perfectly normal. Also makes it a bit expensive
to compute. I don't think failure states for a Deployment/ReplicaSet/etc
isn't trivial.
I'd be willing to move it to the backend if/when we could find a
sufficiently good use case.

All right, I think I'm convinced. I'll try to merge this now.

bryk:
There's still one pending comment. Can you resolve it?

bryk:
@IanLewis Ping?

IanLewis:
@bryk Sorry. I actually came around and did as you suggested and moved it to the server side and wrote some tests. Will have an update today.

IanLewis:
Finally checks have passed. @bryk PTAL

bryk:
I like this. Thanks! :)

PR #1510, opened 2016-12-05
Explicitly select a JDK version in travis

rf232:
Mostly here for checking why builds are failing on travis now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@
 # According to the documentation: http://docs.travis-ci.com/user/docker/
 # sudo is required to enable docker in Travis CI.
 sudo: true
-group: edge
 
 # Cache downloaded Node.JS modules & Bower frontend dependencies for faster builds.
 cache:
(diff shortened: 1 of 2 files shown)


rf232:
4 days ago the travis-ci/travis-build repo pushed a change to the way they generate the build script and while I'm not sure how this change exactly broke our code (as in why did it work before at all) but now explicitly select a JDK version to use.

rf232:
ptal @bryk

rf232:
Hmm, there appears to be more going on, I keep on getting:
```
/home/travis/build/kubernetes/dashboard/node_modules/selenium-webdriver/error.js:26
  constructor(opt_error) {
                         ^
WebDriverError: Connection refused (Connection refused)
```
now, but that appears to be a different problem from the jdk change (or does protratactor use Java, then it might be that we need another JDK than the openjdk actually

Also it might even be that our change is not needed if i look at: https://github.com/travis-ci/travis-ci/issues/6928

bryk:
LGTM again

PR #1530, opened 2016-12-21
Fix typo of Japanese localization

tk3fftk:
"なめ" maybe mean typing alphabet "name" as  Japanese.
It has no meaning so I fix same as other words. (but we laugh when I found it.)


bryk:
@Yamagami-Takuma Can you take a look at this? :)

Yamagami-Takuma:
@tk3fftk  
Exactly. This is my typo. (why I tried to type "name" here...) 
Thanks for fix this!

LGTM

rf232:
@tk3fftk, it might sound a bit over done for a typo fox, but for all pull requests we need the author to sign the CLA. If you could do that we can merge the PR

tk3fftk:
Thank you for review!
I signed the CLA. (Does it have any delay?)

PR #1533, opened 2016-12-21
Csrf protection

rf232:
Fix our Post requests to not be open to CSRF attacks

### DIFF
--- a/src/app/backend/handler/apihandler.go
+++ b/src/app/backend/handler/apihandler.go
@@ -15,6 +15,7 @@
 package handler
 
 import (
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"log"
@@ -60,6 +61,7 @@ import (
 	"github.com/kubernetes/dashboard/src/app/backend/validation"
 	"golang.org/x/net/xsrftoken"
 	clientK8s "k8s.io/kubernetes/pkg/client/clientset_generated/internalclientset"
+	"k8s.io/kubernetes/pkg/client/restclient"
 	"k8s.io/kubernetes/pkg/client/unversioned/clientcmd"
 	"k8s.io/kubernetes/pkg/runtime"
 	utilnet "k8s.io/kubernetes/pkg/util/net"
@@ -80,6 +82,7 @@ type APIHandler struct {
 	heapsterClient client.HeapsterClient
 	clientConfig   clientcmd.ClientConfig
 	verber         common.ResourceVerber
+	csrfKey        string
 }
 
 type CsrfToken struct {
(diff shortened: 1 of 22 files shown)


src/app/backend/handler/apihandler.go
```diff
@@ -93,6 +102,42 @@ func wsMetrics(req *restful.Request, resp *restful.Response, chain *restful.Filt
...
+	return func(req *restful.Request, resp *restful.Response, chain *restful.FilterChain) {
+		resource := mapUrlToResource(req.SelectedRoutePath())
+		if resource == nil || (shouldDoCsrfValidation(req) &&
+			!xsrftoken.Valid(req.HeaderParameter("X-CSRF-TOKEN"),
+				csrfKey,
+				"none",
+				*resource)) {
+
+			err := errors.New("CSRF validation failed")
+			log.Print(err)
```
bryk: Should we print the error here? Wouldn't it be printed by our error handlers anyway?

bryk: If not, can we add more context to the log line? E.g., "CSRF validation failed"

src/app/backend/handler/apihandler.go
```diff
@@ -93,6 +102,42 @@ func wsMetrics(req *restful.Request, resp *restful.Response, chain *restful.Filt
...
+// should either not edit anything or be already safe to CSRF attacks (PUT
+// and DELETE)
+//
+// Validation handlers are also idempotent functions, and not actual data
+// modification opperations
+func shouldDoCsrfValidation(req *restful.Request) bool {
+	if req.Request.Method != "POST" {
+		return false
+	}
+	if strings.HasPrefix(req.SelectedRoutePath(), "/api/v1/appdeployment/validate/") {
```
bryk: Can you comment why this route has been excluded?

src/app/backend/handler/apihandler.go
rf232: We don't go through our error handler, I don't throw an error, I only create an error and follow the same pattern as handleInternalError. I could refactor that function to accept an http StatusCode also, but didn't think it was really worth the effort
https://github.com/kubernetes/dashboard/blob/master/src/app/backend/handler/apihandler.go#L1389

src/app/backend/handler/apihandler.go
rf232: Look at the comment of the function, second paragraph :)

src/app/backend/handler/apihandler.go
bryk: Ok, now I see that my comment was wrong. Discard it.

src/app/backend/handler/apihandler.go
bryk: Yeah, I see, but should this be function documentation? Or implementation documentation inside the function? I'd  put it here because it stays inside the context.

PR #1539, opened 2016-12-23
Allow overriding the bearer token via an authorization header

hanikesn:
This patch allows injecting the bearer token for every request via the `Authorization` header. This can be used together with a reverse proxy in front of the dashboard (e.g. [lua-resty-openidc](https://github.com/pingidentity/lua-resty-openidc)) to handle the actual authentication flow.

This is a work-in-progress. And the following still needs to be done:
- DeployAppFromFile expects a ClientConfig object instead of an actual ApiClient
- Cache API Clients per unqiue bearer token
- The dashboard still needs access to the api server on start to fetch the server version

Implementing an authentication flow inside the dashboard would be out of scope for this PR.

I would be happy to hear your comments, especially because I saw that handling IAM is already on the roadmap.


src/app/backend/handler/apihandler.go
```diff
@@ -527,13 +526,39 @@ func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.Hea
...
 	namespace := parseNamespacePathParameter(request)
 	dataSelect := parseDataSelectPathParameter(request)
 	dataSelect.MetricQuery = dataselect.StandardMetrics
-	result, err := statefulsetlist.GetStatefulSetList(apiHandler.client, namespace, dataSelect, &apiHandler.heapsterClient)
+
+	apiclient, err := getApiCient(apiHandler.clientConfig, request)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
```
rf232: hmm, shame that this requires so much code duplication. Is it possible to abstract this away in a wrapping function?

src/app/backend/handler/apihandler.go
```diff
@@ -136,7 +135,7 @@ func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.Hea
 	verber := common.NewResourceVerber(client.Core().RESTClient(),
```
rf232: I think the RESTClients here won't forward the bearer token now. I think that this is a serious security issue if you want to use this Bearer token feature in some enforced way

bryk:
I also like the idea of this PR :)

src/app/backend/handler/apihandler.go
```diff
@@ -816,13 +850,14 @@ func (apiHandler *APIHandler) handleDeploy(request *restful.Request, response *r
...
 // Handles deploy from file API call.
-func (apiHandler *APIHandler) handleDeployFromFile(request *restful.Request, response *restful.Response) {
+func (apiHandler *APIHandler) handleDeployFromFile(apiClient *clientK8s.Clientset, request *restful.Request, response *restful.Response) {
 	deploymentSpec := new(deployment.AppDeploymentFromFileSpec)
 	if err := request.ReadEntity(deploymentSpec); err != nil {
 		handleInternalError(response, err)
 		return
 	}
 
+	// TODO inject bearer token
```
hanikesn: I'm not sure how to handle injecting the token into the clientConfig directly. Any ideas?

bryk: Hmm... Can you explain the problem? Cannot you pass the apiClient as you did in other places in the code?

hanikesn: deployment.DeployAppFromFile takes apiHandler.clientConfig directly. From my perspective I have 2 options:
- Modify deployment.DeployAppFromFile to take an apiClientSet (I guess there was a good reason to use clientConfig directly)
- Figure out a way to inject the token into the clientConfig directly.

gurvindersingh:
I look forward to have this merged and use it with our kubernetes. We are using kubernetes with JWT token support for RBAC and having this patch merged, we can run dashboard behind this proxy (https://github.com/UNINETT/goidc-proxy) to pass JWT token and we will have fully functioning multi user dashboard/k8s setup.

src/app/backend/handler/apihandler.go
bryk: > Modify deployment.DeployAppFromFile to take an apiClientSet (I guess there was a good reason to use clientConfig directly)

I think this is good enough. Go this way if it feels easier. We used apihandler.clientConfig there because this was best at the time of writing.

bryk:
@gurvindersingh Could you describe how did you set up everything and add this to documentation? I'd love to have such reference guide.

gurvindersingh:
@bryk I will write the whole setup and will send PR, once this patch is merged and I can test the whole setup in practice. In our current setup, we are running K8s with `OIDC` authentication. This populates user and group information from `JWT` token. So the plan is to run the proxy mentioned above in front of dashboard which will forward user's `JWT` which has all the info about him/her and then we have RBAC policies enabled in K8s which will restrict what list of namespaces, dashboard gets and corresponding objects. 

I will know if this works as expected after testing, as I am wondering how dashboard will react to some endpoints being blocked with 403 or so. Also as the author mentioned need to provide a basic service-account for dashboard itself to get version and other information which it needs for itself not on user's behalf.

bryk:
> I will know if this works as expected after testing, as I am wondering how dashboard will react to some endpoints being blocked with 403 or so. Also as the author mentioned need to provide a basic service-account for dashboard itself to get version and other information which it needs for itself not on user's behalf.

AFAIK all read only paths should work fine. Dashboard may have some problems with write paths, though. But let's verify this once we have the PR in.

src/app/backend/handler/apihandler.go
```diff
@@ -173,13 +171,38 @@ func FormatResponseLog(resp *restful.Response, req *restful.Request) string {
...
+func (apiHandler *APIHandler) injectApiClient(fn func(*clientK8s.Clientset, *restful.Request, *restful.Response)) restful.RouteFunction {
+	return func(request *restful.Request, response *restful.Response) {
+		cfg, err := apiHandler.clientConfig.ClientConfig()
+		if err != nil {
+			handleInternalError(response, err)
+			return
+		}
+
+		token := request.HeaderParameter("Authorization")
+		if strings.HasPrefix(token, "Bearer ") {
```
ericchiang: >  Unless otherwise noted, all the protocol parameter names and values
>   are case sensitive.

https://tools.ietf.org/html/rfc6750#section-1.1

src/app/backend/handler/apihandler.go
```diff
@@ -173,13 +171,38 @@ func FormatResponseLog(resp *restful.Response, req *restful.Request) string {
 		req.Request.RemoteAddr, resp.StatusCode())
 }
 
-// CreateHTTPAPIHandler creates a new HTTP handler that handles all requests to the API of the backend.
-func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.HeapsterClient,
-	clientConfig clientcmd.ClientConfig) (http.Handler, error) {
+func (apiHandler *APIHandler) injectApiClient(fn func(*clientK8s.Clientset, *restful.Request, *restful.Response)) restful.RouteFunction {
+	return func(request *restful.Request, response *restful.Response) {
+		cfg, err := apiHandler.clientConfig.ClientConfig()
```
ericchiang: Are you allowed to modify the results of `ClientConfig()`? I don't see anything in the godoc of the interface. Maybe play it safe and make a shallow clone to ensure modifying the `BearerToken` wont impact other callers? Or add a test to ensure it doesn't.

src/app/backend/handler/apihandler.go
hanikesn: I exepected https://golang.org/pkg/strings/#HasPrefix to be case sensitive so I'm not sure what you're trying to tell me.

bryk:
@hanikesn Do you need any help with merging or working on this?

hanikesn:
@bryk I'd be happy to get some help or hand this over. Unfortunately I was sick last week and probably will also not have the time in the future to get it done completely.

src/app/backend/handler/apihandler.go
```diff
@@ -220,22 +243,22 @@ func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.Hea
 
 	apiV1Ws.Route(
 		apiV1Ws.POST("/appdeployment").
-			To(apiHandler.handleDeploy).
+			To(apiHandler.injectApiClient(apiHandler.handleDeploy)).
```
liggitt: This style of invocation concerns me... it means the request starts out as the shared in-cluster-config user unless overridden by a call to injectApiClient. That means if a new API call is added, and the call to injectApiClient is forgotten, an unintentional escalation can occur.

If the goal is to make the dashboard safe to expose to multiple users without risking permission escalations, the override should go in the other direction... by default, make a request with no credentials, add in user-provided credentials if present, and only use the in-cluster-config if a particular use of apiHandler explicitly chooses to use the dashboard's credentials.

src/app/backend/handler/apihandler.go
ericchiang: Woops, sorry. Thought that "Bearer" was generally case insensitive. That's what the API server does.[0] Posted to the RFC without reading it correctly.

[0] https://github.com/kubernetes/apiserver/blob/58aa310d2f43105ec9a78bed26d97b3b4cdeb524/pkg/authentication/request/bearertoken/bearertoken.go#L44

src/app/backend/handler/apihandler.go
liggitt: run the cfg through `anonymousConfig := restclient.AnonymousClientConfig(cfg)` to strip existing credentials from the config, before adding in the user's credentials

src/app/backend/handler/apihandler.go
ericchiang: Agreed maybe modify the APIHandler client to take a method?

```
type APIHandler struct {
    client func(r *restful.Request) (*clientK8s.Clientset, error)
}
```

bryk: Agreed. This should be "forward and require Bearer token for all requests, ignore Bearer for hand-crafted list of request".

Josh-West-Bose:
Greatly looking forward to this!  Then we can give our users a standard URL for dashboard access, instead of only `kubectl proxy`!

src/app/backend/handler/apihandler.go
hanikesn: The shared in cluster user must have only access to the kubernetes api version in order for the dashboard to start. I think we should try to get rid of it entirely if it's run behind a reverse proxy.

With the current logic you can also just not set the bearer token and you will use the shared user. Maybe it makes sense to add a cli flag to disable this behaviour.

>Agreed maybe modify the APIHandler client to take a method?

Sounds reasonable.

src/app/backend/handler/apihandler.go
hanikesn: So I looked deeper into this, but TBH there's a whole jungle of apiClients in the code and it's quite hard to navigate. In the end I sttill think it would be best to modify clientConfig directly, but couldn't find a good way, yet.

hanikesn:
So I finally got around to actually test this on our staging environment. Biggest issue was cookie sizes larger than 4k which lead to:

- https://github.com/pingidentity/lua-resty-openidc/issues/33
- https://github.com/bungle/lua-resty-session/issues/35

Also I had to set `proxy_buffer_size 8k;` on our Nginx Ingress Loadbalancer.

Here's a gist with a working setup: https://gist.github.com/hanikesn/336ae235c1045c9f2b3dcca017359ad4

src/app/backend/handler/apihandler.go
hanikesn: Taking a closer look at the code I specifically added this wrapper function so it can handle all the error handling and we get less code duplication. The other approach would look like:

```
func (apiHandler *APIHandler) getApiClient(request *restful.Request) (*clientK8s.Clientset, error) {
	cfg, err := apiHandler.clientConfig.ClientConfig()
	if err != nil {
		return nil, err
	}

	token := request.HeaderParameter("Authorization")
	if strings.HasPrefix(token, "Bearer ") {
		cfg.BearerToken = strings.TrimPrefix(token, "Bearer ")
	}

	apiclient, err := clientK8s.NewForConfig(cfg)
	if err != nil {
		return nil, err
	}

	return apiclient, nil
}
```


```
	apiClient, err := apiHandler.getApiClient(request)
	
	if err != nil {
		handleInternalError(response, err)
		return
	}
```

If you're ok with duplicating the last part about for every endpoint I can make that change.

hanikesn:
I now moved the whole client creation into apihandler.go and removed all static clients from ApiHandler. `handleDeployFromFile` now also uses the credentials from the header.

There's now both `func (apiHandler *APIHandler) getApiClient(request *restful.Request) (*clientK8s.Clientset, clientcmd.ClientConfig, error)` and `func (apiHandler *APIHandler) injectApiClient(fn func(*clientK8s.Clientset, *restful.Request, *restful.Response)) restful.RouteFunction`. Not sure what's the better approach. But I like the idea of having error handling centralized.

src/app/backend/handler/apihandler.go
```diff
@@ -173,13 +177,53 @@ func FormatResponseLog(resp *restful.Response, req *restful.Request) string {
...
+	authorizationHeader := request.HeaderParameter("Authorization")
+	token := ""
+	if strings.HasPrefix(authorizationHeader, "Bearer ") {
+		token = strings.TrimPrefix(authorizationHeader, "Bearer ")
+	}
+
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ ExplicitPath: apiHandler.clientConfig.KubeConfigFile},
+		&clientcmd.ConfigOverrides{
+			AuthInfo:clientcmdapi.AuthInfo{Token: token},
```
hanikesn: Turns out that overriding the token this way only works outside the cluster when e.g. specifying a kubeconfig file on the command line. But when run inside the cluster the token that's read from disk takes precedence over the one specified here.

bryk:
@maciaszczykm @floreks Can you help here? Do you have ideas how to fix problems mentioned by @hanikesn?

src/app/backend/handler/apihandler.go
rf232: Is it an idea to use the Filter chaining and storing the client in the Request.Attributes? the comment on the attributes member of the Request struct says it is for request scoped variables, and it seems like the client should be. Error handling can be done in the filter too (see the CSRF filter)
[https://github.com/emicklei/go-restful/blob/master/request.go#L18](https://github.com/emicklei/go-restful/blob/master/request.go#L18)

hanikesn:
Any updates from your side @floreks @maciaszczykm ?

maciaszczykm:
@hanikesn We're preparing to the release right now and were focused on other tasks, sorry for not responding here! I'll take a look on it and @floreks will hopefully join me. It could take a while though.

Could you rebase and verify if it works with TPR implementation which uses yet another REST client?

src/app/backend/handler/apihandler.go
maciaszczykm: Is it the only problem left in current implementation, that you're aware of?

maciaszczykm: Quoting from https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/overrides.go#L29:

``` go
// ConfigOverrides holds values that should override whatever information is pulled from the actual Config object.  You can't
// simply use an actual Config object, because Configs hold maps, but overrides are restricted to "at most one"
type ConfigOverrides struct {
	AuthInfo clientcmdapi.AuthInfo
	// ClusterDefaults are applied before the configured cluster info is loaded.
	ClusterDefaults clientcmdapi.Cluster
	ClusterInfo     clientcmdapi.Cluster
	Context         clientcmdapi.Context
	CurrentContext  string
	Timeout         string
}
```

I'm not sure how to understand "at most one" in this context.

src/app/backend/handler/apihandler.go
```diff
@@ -173,13 +177,60 @@ func FormatResponseLog(resp *restful.Response, req *restful.Request) string {
...
+			ClusterInfo: clientcmdapi.Cluster{Server: apiHandler.clientConfig.ApiserverHost}})
+
+	cfg, err := clientConfig.ClientConfig()
+
+	// TODO Workaround, because ClientConfig doesn't override the token from the tokenfile with the one specified
+	// when run inside a cluster
+	if token != "" {
+		cfg.BearerToken = token
```
joshwget: We should move the error check up before this so there isn't a panic when `cfg == nil`. I just hit this attempting to rebase on master.

joshwget:
@hanikesn If you'd like I can pick up where you left off and continue work on this PR. We're definitely pretty interested in this feature.

hanikesn:
@joshwget Feel free to take over. Unfortunately I couldn't spend the necessary time on this PR.

src/app/backend/handler/apihandler.go
hanikesn: Fixed.

hanikesn:
Ok, I had some time and resolved the conflicts with current master. Can we work on a strategy on how to merge this? I could split up the changes into a bigger refactoring change. And then one where we handle the auth injection?

hanikesn:
I followed @rf232 advice and use the request filter chain to inject the apihandler/apiconfig now. Codewise I'd say it doesn't get much cleaner than this.

maciaszczykm:
@hanikesn Okay, thank you. We will take a look on it tommorow and return to you with feedback.

joshwget:
Thanks for beating me to this @hanikesn! The filter chain implementation looks really clean.

Going back to an earlier discussion, I really like the idea of a flag to disable all access except through an authorization header bearer token. Would this be a part of the remaining refactoring work? I guess I'm not actually sure what other refactoring is currently missing from this PR.

src/app/backend/handler/apihandler.go
```diff
@@ -179,13 +223,15 @@ func formatResponseLog(response *restful.Response, request *restful.Request) str
...
-
-	verber := common.NewResourceVerber(client.CoreV1().RESTClient(),
+func createVerberFromApiClient(client *clientK8s.Clientset) common.ResourceVerber {
+	return common.NewResourceVerber(client.CoreV1().RESTClient(),
 		client.ExtensionsV1beta1().RESTClient(), client.AppsV1beta1().RESTClient(),
 		client.BatchV1().RESTClient(), client.AutoscalingV1().RESTClient(), client.StorageV1beta1().RESTClient())
+}
+
+// CreateHTTPAPIHandler creates a new HTTP handler that handles all requests to the API of the backend.
+func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.HeapsterClient,
```
maciaszczykm: `client` seems to be not used here.

src/app/backend/handler/apihandler.go
```diff
@@ -204,7 +250,7 @@ func CreateHTTPAPIHandler(client *clientK8s.Clientset, heapsterClient client.Hea
 		csrfKey = string(bytes)
 	}
 
-	apiHandler := APIHandler{client, heapsterClient, clientConfig, verber, csrfKey}
+	apiHandler := APIHandler{heapsterClient, clientConfig, csrfKey, }
```
maciaszczykm: Redundant comma.

src/app/backend/client/apiserverclient.go
```diff
@@ -56,10 +56,10 @@ func buildConfigFromFlags(masterUrl, kubeconfigPath string) (*rest.Config, error
 //
 // apiserverHost param is in the format of protocol://address:port/pathPrefix, e.g.http://localhost:8001.
 // kubeConfig location of kubeconfig file
-func CreateApiserverClient(apiserverHost string, kubeConfig string) (*kubernetes.Clientset, *rest.Config, error) {
+func CreateApiserverClient(apiserverHost string, kubeConfig string) (*kubernetes.Clientset, error) {
```
maciaszczykm: Please also run `gulp format`.

src/app/backend/handler/apihandler.go
```diff
@@ -1442,7 +1494,7 @@ func (apiHandler *APIHandler) handleGetPersistentVolumeList(request *restful.Req
 func (apiHandler *APIHandler) handleGetThirdPartyResource(request *restful.Request,
 	response *restful.Response) {
 	dataSelect := parseDataSelectPathParameter(request)
-	result, err := thirdpartyresource.GetThirdPartyResourceList(apiHandler.client, dataSelect)
+	result, err := thirdpartyresource.GetThirdPartyResourceList(request.Attribute("apiclient").(*clientK8s.Clientset), dataSelect)
```
maciaszczykm: Perhaps you could move `request.Attribute("apiclient").(*clientK8s.Clientset)` to a function?

maciaszczykm:
@hanikesn Tommorow we are going to run more tests and provide you more details :-)

maciaszczykm:
@bryk @rf232 Can you also check it on GCE?

src/app/backend/handler/apihandler.go
hanikesn: Isn't Go gonna panic on its own when trying to cast `nil` to `*clientK8s.Clientset`?

floreks: It will but it's better to do cast with error check so we can adjust the error message. This way it will be more clear for devs and also we will have single method where error can occur.

maciaszczykm:
At the moment we are discovering some strange issues with Dashboard after building Docker images, we need to fix it before.

floreks:
We have tested it and everything seems to be working fine.

I have created SA that acts as a user and when granted `cluster-admin` role then everything worked. When changed role binding to `basic-user` my access was revoked.

zrzut ekranu z 2017-05-11 11-39-09

It is a major step forward. We will be able to save some user settings that i.e. could be downloaded from config map tied to the user. Also admin now will be able to grant/revoke access rights to users accessing dashboard.

To extend this we can think about also checking `X-Remote-User` header that can be used together with some authenticating proxy and IDP.

Thanks for this contribution!

Issue #1532, opened 2016-12-21
Issue deleting PV from dashboard


Dmitry1987:
#### Issue details
Tried to delete 3 PV through dashboard, got error (on screenshot).
kubectl delete was ok.

##### Environment
AWS, K8s installed with kubeadm on ubuntu 16 instances,
Dashboard was launched with "kubectl create -f https://rawgit.com/kubernetes/dashboard/master/src/deploy/kubernetes-dashboard.yaml"   3 days ago
```
Dashboard version: 1.5.0
Kubernetes version: 1.5.1
Operating system: ubuntu 16
Node.js version: -
Go version: -
```

##### Steps to reproduce
Create StatefulSet or something else with Volume Claim, AWS will dynamically provision the volumes.
Then try to delete from dashboard. 

##### Observed result
![image](https://cloud.githubusercontent.com/assets/9369080/21389689/95be6826-c78b-11e6-9287-8f8bf043975a.png)

##### Expected result
Delete volume resources.

##### Comments
Seems like a small bug. The error message needs to state "volumes not found" if underlying AWS\other volumes can't be reached, or something similar.  I'm not sure I had those volumes at all (maybe deleted manually some time ago) but it shouldn't matter:  if underlying cloud volume exist, delete it, if not, show information about it, and delete the K8s resource anyway.

Hope it helps,
thanks a lot to all devs for the good job, the dashboard is awesome anyway =)

Dmitry1987:
Some more info about versions, if will help to locate the bug:

kubectl version
Client Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"2016-12-14T00:57:05Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"2016-12-14T00:52:01Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}

rf232:
hmm, my first guess this was an issue in the backend, but `kubectl delete pv <pv>` works, so we do something wrong here.

rf232:
[...]

PR #1545, opened 2016-12-28
Let Verber work with resources without namespace

rf232:
Fixes: #1532


Dmitry1987:
@rf232  thanks for fixing this!  
btw how do I know in which version a particular fix will be released?  if it was merged to master, then we'll see it in nearest minor release?

P.S.  I liked the "Namespacedness" from "ShouldRespectNamespacedness" lol  :D

rf232:
Yesterday we released v1.5.1 with this fix in it.

-- 
Rob

PR #1556, opened 2017-01-06
fix variable name mismatch

hougangliu:
fix variable name mismatch to avoid misunderstanding.


hougangliu:
I signed it!

hougangliu:
I had signed CLA

PR #1566, opened 2017-01-13
there should be not related pods for service without selector

hougangliu:
for a service without selector, there should be not any pod.
and but in current logic, get pods by labels.SelectorFromSet(nil), all pods are selected.


rf232:
Codepart LGTM, but is it possible to add a test were we verify we still do a list request if we do have a label selector?

hougangliu:
Sure, I had added a test with list request.

PR #1571, opened 2017-01-16
Return http status 200 instead of 201 for getting resource

hougangliu:
when get resource from backend successfully, it should return 200 http status code instead of 201


rf232:
Thanks, I saw the issue earlier, but was too lazy, So thanks for stepping up while I was slacking :)

PR #1576, opened 2017-01-17
WIP: RBAC resource management

mlamina:
Implementation of [this design document](https://github.com/kubernetes/dashboard/blob/master/docs/design/access-control.md).


bryk:
Went through the code and it generally looks good :) 

Should we review it before the discussion about 1:n relationship is resolved?

mlamina:
This is just "empty UI" so far, I've already started working on the backend part. While doing so I stumbled upon the 1:n relationship problem. So I suggest we solve this before continuing the work here?

bryk:
There are two compile errors while trying to `gulp serve:prod`.

bryk:
Not sure if this is intended at the moment, but I got 500 error when trying to access the page: 
screen shot 2017-01-23 at 8 35 53 am

bryk:
> While doing so I stumbled upon the 1:n relationship problem. So I suggest we solve this before continuing the work here?

Can you talk with @danielromlein about this? :) He knows stuff.

mlamina:
Thanks for your comments @bryk! This PR is still WIP. I'm trying to keep the state of the application as clean as possible with every commit though :)

@danielromlein Your input would be greatly appreciated!

bryk:
@mlamina Should we merge this PR and then follow up with another one? Or wait with merging until it is done?

mlamina:
Right now it is not in a state that I would consider usable. I think we should sort out the design first, then I'll continue implementing and then we merge!

danielromlein:
Sorry for the delay in getting back to you guys on this! 
@bryk where does this fall in the UX design priorities? 
@mlamina I'm struggling to understand the 1:n issue for Roles > Rules that you described; are you available to jump on a call at some point so we could discuss and wireframe some options?

mlamina:
@danielromlein absolutely! I'll message you on Slack

danielromlein:
@mlamina I'm struggling to understand how Rules fit in to the scheme of things. Looking at the docs page for [authorization](https://kubernetes.io/docs/admin/authorization/), it seems like their use of `rules` (plural) simply describes the various parameters that constitute one Role. What would an API example of multiple Rule ("groups") applied to a single Role look like?

Also: would a Rule have a name and exist similarly to a Role in that it could be applied to multiple Roles? Or would the creation of Rules be limited to a small-enough number of variables that it'd just make sense to have its creation be from those constituent variables on a case-by-case basis? Related: would we need a list view of Rules like we have for Roles and Bindings?

@bryk question for you: Would it make sense to have a Binding's subject(s) selection incorporated as part of the creation process of a Binding? Currently the "New Role Binding" dialogue mockup just has _Name_, _Target Role_, and _Namespace_ – a user then has to go to the detail view for the Binding to add subjects one by one.

mlamina:
@danielromlein You're absolutely right! Rules are no objects by themselves. Rules don't have names and are not reusable like Roles! They live and die with their Role, being no more than a list of parameters. So I think there's no need to put them on the same conceptual level with Roles and RoleBindings. I just read through the documentation again - now that we're doing it properly, we might also consider adding the __ apiGroup__ property next to __verbs__ and __resources__!

Hope that answers your questions? :)

danielromlein:
@mlamina OK! This is good. Follow up questions:
- How do API Groups work? Would there be one per Rule?
- Rather than creating a Binding one _add_ at a time, would we want a way to select multiple subjects to a binding all at once? i.e. a multi-select pattern?
- How many Resource types could there be? That influences what UI pattern would be best.

/ @bryk

mlamina:
- You can find more information about API Groups [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-group.md). I couldn't find any API endpoint which would allow us to get a list of enabled API groups, but I think for now it is fine if we look at them simply as a list of strings, speaking from the perspective of a Role. Next to _verbs_ and _resources_, every rule in a Role has a list of API groups
- Sure, why not! Every binding can have multiple subjects, so some sort of multi-select pattern would make sense. However, AFAIK there is no list of available subjects available (yet), which means the user will need to specify every subjects type and name by hand.
- Good point! The list is actually quite long, longer than I initially thought. Check [this](https://kubernetes.io/docs/resources-reference/v1.5/) out

maciaszczykm:
What's state of this PR?

mlamina:
@maciaszczykm The work on this got delayed, because the design needed to be updated. See #1620 for details! The update is almost done and I will continue to work on this

mlamina:
@bryk @danielromlein Do you know anyone who could help/assist me with the implementation? The design contains tabs, an extended action bar and a new type of detail view. My experience with Angular is quite limited, so I'm afraid to build something that doesn't comply with your already established best practices regarding navigation, states, nested views, etc.

maciaszczykm:
@floreks let's help to finish frontend part here.

mlamina:
@floreks @maciaszczykm thank you for offering your help! Let me know if you have any questions or if there's anything I can do to contribute

maciaszczykm:
Will send commit soon. We'll merge initial version and then iterate on it.

joshwget:
Was this PR a complete implementation of the linked design document? I only see the ability to list roles and cluster roles on master. Nothing regarding role bindings or the ability to add/edit roles.

floreks:
This PR added only partial support to view ClusterRoles and Roles without details and edit option. Creation of resources is a different topic and can be tracked from #1600.

Issue #1504, opened 2016-11-30
ThirdPartyResources Page


IanLewis:
We should allow viewing/creating/editing of ThirdPartyResources. There is a bit of UX to think about since the ThirdPartyResource types themselves can be viewed as well as the custom objects for each resource.

- How to display each ThirdPartyResource?
- How to we display the objects of each ThirdPartyResource type

IanLewis:
This is something I'd like to look at doing; perhaps after I've finished the exec into pod feature.

bryk:
@IanLewis Yes, this is an important feature. Definitely on track to work on after pod exec. And I agree that there are many UX challenges in this problem...

foxish:
We have some thoughts on this as we start using TPRs alongside custom controllers as a way to represent them.
@kubernetes/sig-apps-misc

erictune:
My take on this: 
- Users think about their cluster as holding a number of "apps".
- Useful: an "app summary" view in dashboard that lists all their apps, regardless of what Kinds make up that app.
- Not as useful: Having one view which is just Deployments, another which is just StatefulSet, and another which is just TPR, etc.
- With Helm Charts, and similar packaging systems, users often won't even be aware what Kind implements their app.
- An "app" is a collection of resources, but you don't want to see most of them most of the time.  There is a clear "parent" object which should be shown
  in the "app summary" view, and child objects should be hidden in a summary view, but show when you drill down on a pariticular app. 
- Examples of "apps":
  - nginx app, parent object is `Deployment`, children are `ReplicaSet` and `Pods`.
  - mysql app, parent object is a `StatefulSet`, children are `Pods` and `PVCs`.
  - a spark job called, parent object is a TPR of type `SparkJob`, children are a `ReplicaSets` and `Pods` 
  - there are too many possibilities to hard code.

IanLewis:
@erictune Thanks for the input. Is this something that really has to do with ThirdPartyResources? Maybe it deserves it's own issue?

erictune:
I comment on this issue because I am questioning the premise of this issue.  Should third party resources have their own separate page, as the title of this issue suggests.  Or should they be integrated into various pages that alongside "first party resources"?

IanLewis:
Ok. It sounded like a much broader idea that even things like Deployments, Services, Pods etc. should be shown differently which is why I thought different issue. And that issue could easily invalidate this issue.

erictune:
Sorry if I came off a little strong there.  

[...]

Issue #1457, opened 2016-11-21
Migrate to client-go


IanLewis:
It may take a while but since the dashboard is essentially just a client application using the Kubernetes API, it would be good to migrate to the official Go client at some point.

Is this even feasible?

bryk:
Yes, this should be doable. When we started the UI project there was no client-go project yet. That wa sthe reason.

floreks:
We've discussed that already in #1395. Unfortunately at this point `client-go` can not be used by dashboard. We are using some utility functions available only in the core for now. Additionally there is a problem with object types that requires very much work to resolve. I have started to work on that and even have a branch but we have to wait until these issues are resolved.

bryk:
> We are using some utility functions available only in the core for now.

Do you remember what functions are those?

floreks:
`k8s.io/kubernetes/pkg/controller/deployment/util`, `k8s.io/kubernetes/pkg/kubectl/cmd/util` and probably `k8s.io/kubernetes/pkg/kubectl/resource`.

There might be something else though.

IanLewis:
A cursory look also leads me to believe that it currently doesn't support exec/attach API which is required for #1345

PR #1586, opened 2017-01-24
Added ThirdPartyResource design doc and mockups

IanLewis:
This PR is a design document meant to address issue #1504.


floreks:
Is anyone working on implementation of third party resources support in dashboard?

bryk:
@IanLewis are you? @mijajasinski is going to help with reviewing

IanLewis:
@floreks @bryk I haven't really started on the implementation if that's what you mean. I have a *very* rough mock on my local machine but that's it. I was starting to get wary of starting implementation since, while I know it's an important feature people are already using, TPR is still beta so the API may change and there is talk of adding [Aggregated API Servers (AAS)](https://github.com/kubernetes/dashboard/issues/1504#issuecomment-276543001) and I was thinking it may be best to hold off a bit and see how that plays out. WDYT?

I know we have added beta features like Deployments before, but TPR seems a bit more in the air to me. Maybe adding it but indicating it's still in beta might be best?

IanLewis:
I also wanted to make sure we were cool good to merge this design doc PR before starting on the implementation in earnest.

floreks:
@IanLewis I have worked a bit with API for TPR and first we need to migrate to client-go. Current k8s dependencies are a bit outdated and contain some old code for TPR. Without update we are able only to get basic info about TPR templates not objects created by them.

I have already started migration to newest client-go but it will take some time. Regarding AAS I'm not sure if they will help us with TPR. As stated it will be different way of extending API. More flexible and powerful. TPRs are easy to create and use but have some constraints.

docs/design/thirdpartyresources.md
```diff
@@ -0,0 +1,44 @@
...
+
+Third Party Resources will be listed in the sidebar of the UI in a new section
+alongside first party objects. It will function similarly to existing menu
+items except that the items in the menu will be dynamically populated from the
+server.
+
+![](mockups/24-01-2017-thirdpartyresources/workloads.png)
+
+Clicking on a menu item brings up the Third Party Resource's detail page where
+the details of the ThirdPartyResource itself is displayed and objects created
```
mijajasinski: I think it should be "the details of the ThirdPartyResource itself are displayed"

IanLewis:
@floreks 

> I have worked a bit with API for TPR and first we need to migrate to client-go. Current k8s dependencies are a bit outdated and contain some old code for TPR. Without update we are able only to get basic info about TPR templates not objects created by them.

Hmm. I've done some work using client-go as well, but I've not used the dashboard dependencies. I'll take your word that client-go is needed.

> I have already started migration to newest client-go but it will take some time. Regarding AAS I'm not sure if they will help us with TPR. As stated it will be different way of extending API. More flexible and powerful. TPRs are easy to create and use but have some constraints.

Yah, my understanding is that all of this is still quite a bit in flux and it's hard to see how it will land.

> Backend for this will be quite complex as it requires dynamic client creation for every TPR APIGroup. Similar to how it's done in kubectl.

Yah. It won't be super bad because the dynamic client will only be needed on a TPR detail page. TPRs themselves can be listed with the normal client so, if we go with the current design, the backend for the sidebar can just use the regular client.

IanLewis:
For reference the issue for migrating to client-go is #1457

IanLewis:
@floreks 

> Current k8s dependencies are a bit outdated and contain some old code for TPR. Without update we are able only to get basic info about TPR templates not objects created by them.

Can we just update the current dependencies for now to get the dynamic client? I'm afraid client-go will affect the implementation of other PRs like the terminal (#1345)

docs/design/thirdpartyresources.md
IanLewis: Yah.

floreks:
@IanLewis 

> Can we just update the current dependencies for now to get the dynamic client? I'm afraid client-go will affect the implementation of other PRs like the terminal (#1345)

I will check. Our deps are few months old and much has changed since then. It should be easier to update only k8s deps for sure.

floreks:
@IanLewis @bryk Small update.

It appears that kubernetes is already using client-go as dependency and when using govendor to fetch latest deps we are also getting client-go and apimachinery. There is however [small issue on client-go](https://github.com/kubernetes/client-go/issues/93) side and it is not possible to build it right now. I wouldn't try to manually fix dependencies but rather wait until it gets fixed and we can safely use govendor to update and build them correctly.

Let's hope it will take few days not weeks to fix it :)

IanLewis:
If you back out client-go to one prev revision it builds. I suspect they
updated client-go but not the apimachinery repo yet.

Issue #1508, opened 2016-12-02
Sample manifests for development/debugging


IanLewis:
I want to create some sample manifests for debugging purposes that allow for testing different pod states in the cluster (Crashloopbackoff, Failed, Pending, Initializing, etc.). It can be difficult sometimes to recreate these states in the cluster.

cheld:
I had the same idea a while ago and as created some resources with max values for layouting testing 

https://github.com/kubernetes/dashboard/tree/master/docs/test/examples/poison-lab

However, it was not maintained or extended afterwards. So, we should clean up here too.

IanLewis:
@cheld Ok. I'm not sure that I know what exactly the example there is for. Do you think you could create a PR to clean it up? I'm planning on creating yaml files with comments at the top explaining exactly what each file is for.

Should I put my files in the docs/test/examples/<something> directory also?

cheld:
@IanLewis 

The first example is 
https://github.com/kubernetes/dashboard/blob/master/docs/test/examples/demo/guestbook-all-in-one.yaml

it is the guestbook example which was shown in kubecon. It has a few small adoptions to run on a disconnected notebook/dev environment and resource quotas so you can create scheduling errors when scaling. The general idea was to have a yaml that fills the dashboard with some meaningful resources. I assume we can remove.

Again, the other example was intended to test layout. It would have to been extended to all resource types to be useful. I tend to remove this one as well.

So, I would suggest you create a new folder and we remove the complete doc/test folder as it is outdated anyway.

IanLewis:
@cheld Ok. I'm aiming to create manifests that test specific situations rather than doing a lot of things at once. These are the ones I have so far: https://gist.github.com/IanLewis/d55c133751ca1d2db638c53af798e515

PTAL

cheld:
Yea, like it

bryk:
I also like this. @IanLewis Propose a place or directory layout where to put such files and put them. Maybe add comments to the files to document what they are supposed to do (if not trivial).

IanLewis:
After thinking about it I think that docs is the right place for this. I propose we create a new docs/testing folder with these manifests and a simple README.md for the time being to expand on it later.

We should also delete the docs/test folder as @cheld suggested.

PR #1588, opened 2017-01-24
Removed docs/test directory.

IanLewis:
Deleted the docs/test directory per #1508


cheld:
What about the example folder. Do you want to keep it?

PR #1620, opened 2017-02-07
Access control design update

mlamina:
Since the doc was created there have been some conceptual changes and new amazing mockups thanks to @danielromlein . This PR syncs these changes with the document.


danielromlein:
>The only property of a Role that cannot be changed is its namespace, because it determines if the Role is a ClusterRole or not.

I don't understand why a user will be prevented from changing a Role's namespace...?

>Clicking "Create" will create the Role and redirect the user to the Role's detail page.

"Redirect" is misleading here. How about "Clicking "Create" will create the Role and select that Role from the list, displaying that Role's details in the panel on the right.

>Shortly introduce the concept of roles and bindings to the user

This is minor, but I think "shortly" should be "briefly." 🤓

mlamina:
> I don't understand why a user will be prevented from changing a Role's namespace...?

Maybe more precise: The namespace of a Role might be changed from one namespace to another, but not between "all namespaces" and something else. Reason: A Role with the namespace "all namespaces" is a ClusterRole by definition. Changing its namespace would mean that the ClusterRole would have to be transformed into a regular Role.

I agree with your other comments!

seh:
Per [my comments in the "sig-ui" Slack channel](https://kubernetes.slack.com/archives/sig-ui/p1486147397000187?thread_ts=1486144422.000175&cid=C09R1JVU6), I find the treatment of _ClusterRoleBinding_ and _RoleBinding_ resources confusing.

Saying that a binding applies to all namespaces _implies_ that the binding is a _ClusterRoleBinding_, but that's too subtle; an administrator who then falls back to _kubectl_ commands needs to know that he's dealing with a _ClusterRoleBinding_, and not a _RoleBinding_. I would prefer to see the binding type mentioned explicitly. These are different resource kinds.

mlamina:
I agree that being explicit is always preferable and that there should be no ambiguity. @danielromlein maybe we should display a "type" property with every object? I still think having one consolidated view for creating Roles/ClusterRoles or RoleBindings/ClusterRoleBindings is a good idea. It keeps things simple

bryk:
@mlamina Can you work with @danielromlein to resolve @seh's comments? The comments are valid.

danielromlein:
Just adding a column for "Binding Type" to the _Bindings_ table seems like the most straightforward solution – what do you think @mlamina?

mlamina:
Yes, we should also have a "Role Type" column for Roles!

danielromlein:
I updated the [InVision prototype](https://invis.io/7Z9NNP4E8) to include "Role Type" and "Binding Type" columns.
Also, _Bindings_ is now _Role Bindings_ throughout the UI for consistency's sake (@normalnorman's recommendation). 
PTAL @mlamina
I thought @normalnorman had a good point too about being able to view users:
> Is there somewhere in Dashboard where an Admin can see all users, or is the [Create Role Binding](https://invis.io/2SAGARYFV#/217422807_New_Role_Binding) screen the only place where that list is currently surfaced? [currently in an [autocomplete](https://material.angularjs.org/latest/demo/autocomplete)]  I'm wondering if the flow should be available in the opposite direction: Admin goes to a list of users, picks an individual user, assigns Role(s).  This might remove the need for an object type of "Role Binding".  A user would just have one or more Roles assigned to them. 

@mlamina @bryk @rf232 WDYT?

bryk:
@danielromlein 
> I updated the InVision prototype to include "Role Type" and "Binding Type" columns.

The link is no longer valid. Can you update?

danielromlein:
@bryk hmmm, it's working for me...? May have just been a temporary InVision bug? Give this a try: [prototype](https://invis.io/7Z9NNP4E8)

bryk:
@danielromlein For some reason I cannot comment in Invision, so asking here.

1. The right side panel does not have padding/border on the right and bottom. Is this intentional?

danielromlein:
@bryk yes it's intentional: there's no padding so it functions as a panel sliding in from the right of the viewport.
Based on some feedback from designers, I'm assessing whether one of [these two options](https://invis.io/2DAKGTLWS) would more clearly establish the master / detail relationship.

danielromlein:
@bryk @mlamina WDYT about a view for Subjects? Perhaps a table under another tab? Is there a use case for wanting to view all available subjects and take CRUD actions on them?

mlamina:
@danielromlein AFAIK the Kubernetes API does not provide a way to query all available subjects. If a subject actually exists or not cannot be verified as of know, so I don't think such a view would be possible at the moment.

mlamina:
@bryk @danielromlein So what do we think - are we good to go? If the InVision mockups are OK with everybody, I'd put the screenshots into the document and update the texts accordingly so we can go ahead and merge!

danielromlein:
@mlamina things look good to me! I think leaving out subjects for the MVP makes sense.
@bryk WDYT?

maciaszczykm:
What's state of this PR? Can we merge it or is there anything to do?

danielromlein:
@maciaszczykm the screenshots still need to be updated with the lastest [InVision mockups](https://invis.io/7Z9NNP4E8). @mlamina is this something you will have time to do?

danielromlein:
Also, @mlamina RE the following point, want to add something to the design doc mentioning that technical limitation?
> AFAIK the Kubernetes API does not provide a way to query all available subjects. If a subject actually exists or not cannot be verified as of know, so I don't think such a view would be possible at the moment.

mlamina:
Hey folks, sorry for the delay! I will update the screenshots and add the technical limitation to the doc as soon as I can, then we can merge :)

maciaszczykm:
@mlamina Thank you, really appreciate it!

mlamina:
@danielromlein @maciaszczykm We're good to go from my side

maciaszczykm:
@danielromlein Can you take a look before we merge it?

danielromlein:
Made a few capitalization and grammar changes and committed to your fork @mlamina (hope that was the right way to go about it??)

mlamina:
Way to go @danielromlein ! :)

danielromlein:
LGTM

maciaszczykm:
@mlamina @danielromlein Thank you, I'm merging it :)

PR #1638, opened 2017-02-14
Client-go migration

### LINKED ISSUE #1457: Migrate to client-go

floreks:
This is a big change and extensive tests are needed. It will take some time to review and I can finish what's left in the meantime.

## What's done

- Client-go - updated to current master, commit hash: [86a2be1b447d7abddae88de0a5f642935992b803](https://github.com/kubernetes/client-go/tree/86a2be1b447d7abddae88de0a5f642935992b803)
- 90% of the code migrated to use client-go instead of kubernetes deps
- Fixed most backend tests
- **Cleaned up kubernetes dependencies. Left only the one needed for deploy from file (kubectl utils) and updated, commit hash: 871fc426b3576bfe9a92b3ca1d4bfa83ea71edd5**

## TODO

- [x] ~Get rid of kubernetes dependencies~
- [x] Fix 10 tests that are skipped
- [x] Fix deployment detail and not use deployment utils from kubernetes
- [x] Fix deployment deploy not to use kubectl utils
- [x] Enable endpoint for getting deployments old replica sets

~Dashboard is running and travis should be green, however 1 endpoint is disabled and 10 tests are skipped.~

~In order to get rid of kubernetes as dependency and fully migrate to client-go we'd need to rewrite some code. Rewriting deployment details and getting replica sets should not be hard but getting rid of kubectl as dependency might be harder. If we want to do that then I will create issues for that.~

~I have not updated kubernetes dependencies because current master and 1.6.0-alpha are broken. Same dependency error as in [client-go](https://github.com/kubernetes/client-go/issues/93).~

Closes #1457


rf232:
Awesome, current changes look good (apart from a run of gofmt on the test code.)

In a way I'm happy that you haven't fixed the Deployment utils things yet, since I fear those are the more interesting parts to review, so being able to view them somewhat in isolation is a good thing (hint, don't squash it in the current changes :))

floreks:
@rf232 @IanLewis @maciaszczykm 

Most of the work is done. Travis is green. Only thing left is to fix some tests but I need to look a bit deeper into some of them because of the migration and this may take some time. Could you test it and maybe merge? Faster we merge, easier it will be for others to rebase their changes.

rf232:
> Most of the work is done. Travis is green. Only thing left is to fix some tests but I need to look a bit deeper into some of them because of the migration and this may take some time. Could you test it and maybe merge? Faster we merge, easier it will be for others to rebase their changes.

Agreed, reviewing now

> I can create issue to fix skipped tests and work on that separately.

SGTM

src/app/backend/resource/deployment/util/deploymentutil.go
```diff
@@ -0,0 +1,100 @@
+package util
```
floreks: Ye, It would be great if they could move utils also. Keeping deployment utils deps also downloads significant amount of files so I guess it's better for now to just copy these few functions and keep them in sync.

floreks:
> To get the diamond zombie slayer achievement please run:
`govendor list +u | cut -c5- | xargs govendor remove`

Good tip. Achievement unlocked :)

src/app/backend/resource/deployment/util/deploymentutil.go
floreks: Just tried to reuse them and unfortunately the are using different extensions package version. We have to leave it as is for now.

PR #1640, opened 2017-02-14
Add custom page title for each view

### LINKED ISSUE #1609: <title>Kubernetes Dashboard</title> should change based on location
When you've got a bunch of tabs open, it's impossible to tell them apart. It should be something like `Daemon Sets - Kubernetes Dashboard`. The most important part of the title should be at the left-hand side of it, since the right-hand side gets elided first.

Bonus: When you're working with multiple clusters (instances of `kubectl proxy`), display some sort of "cluster name"... I'm not sure if a cluster has a concept of a name though.

maciaszczykm:
Fixes #1609:

image

image

image


maciaszczykm:
Still WIP.

PR #1642, opened 2017-02-15
Third party resource backend

maciaszczykm:
To do:

- [x] list endpoint
- [x] detail endpoint
- [x] tests

Question marks:

- ~~Should third party resources be namespaced?~~

Currently handling third party resources as templates, in the next pull request I'll add endpoints for getting third party objects.


maciaszczykm:
Regarding namespacing third party resources, from [documentation](https://kubernetes.io/docs/user-guide/thirdpartyresources/):

> When you create a new ThirdPartyResource, the Kubernetes API Server reacts by creating a new, namespaced RESTful resource path. For now, non-namespaced objects are not supported. As with existing built-in objects, deleting a namespace deletes all custom objects in that namespace. **ThirdPartyResources themselves are non-namespaced and are available to all namespaces.**

maciaszczykm:
@kubernetes/dashboard-maintainers One more remark, what about changing other package file names to make them shorter like it's done here?  `common.go`, `details.go` etc.

PR #1654, opened 2017-02-17
cap the replica set limit at 10

jsloyer:
capping the cleanup policy at a reasonable limit so a bunch of replica sets are left around


rf232:
Thanks, looks like a good default indeed

PR #1663, opened 2017-02-21
WIP: patch version number and git commit into dashboard

### LINKED ISSUE #1507: Add debug information to the frontend
Dashboard UI should have the option to show debug info layer in the browser. It'd say its version, which SHA commit it was built at and such. This info can then be passed when sending out feedback reports from within the UI. This should help a lot with debugging.

cheld:
Adding debug information to UI. Fixes #1507

Small patch - still I need some help :)

Question 1: In which step should I patch the source files?

In my understanding I have to options:
1: build/script.js
2: build/build.js somewhere at localize task

Both options are possible. I just wonder which version is cleaner and more obvious

- In script.js I could limit the replace scope to a single source file. In build.js the source files are concatenated  already to multiple app.js files. So the patch must run over the full files. 
- the localize task seems simpler to refactor to make the patching more explicit in the code

Question 2: should I reuse existing configuration service or create new one?

Existing services:
1: index_config.js
2. appconfig_service.js

or new one

Any preference?

BTW: I also simplified the error submit dialog a bit. The intention was to motivate the user to fill the fields. Especially, I removed the preset title to avoid users to be able to press submit without typing :)


cheld:
@floreks please have a look

build/script.js
```diff
@@ -176,6 +177,8 @@ function createCompileTask(translation) {
                     conf.paths.bowerComponents, 'google-closure-library/closure/goog/base.js'),
               ])
               .pipe(gulpClosureCompiler(closureCompilerConfig))
+              .pipe(gulpReplace('BUILD_DASHBOARD_VERSION', conf.deploy.version.release))
```
floreks: I'd export `BUILD_DASHBOARD_VERSION` constant to conf file and comment it properly. It should be clear what's it for.

Additionally I'd prefer to limit replace only to a single js file where those strings should be replaced and do it before the concatenation and minification. There is always a chance that someone will use such a string in some other place and it should not be replaced there.

src/app/frontend/common/appconfig/appconfig_service.js
```diff
@@ -44,4 +44,12 @@ export class AppConfigService {
       return null;
     }
   }
+
+  getDashboardVersion() {
```
floreks: I think app config should be a good place to store these values. We also could extract it to a new service but I don't have any preferences here.

@maciaszczykm what do you think?

src/app/frontend/error/internalerror_controller.js
```diff
@@ -18,14 +18,18 @@
 export class InternalErrorController {
   /**
    * @param {!./internalerror_state.StateParams} $stateParams
+   * @param {!Object} kdAppConfigService
```
floreks: It should be possible to provide correct type for this service. Object is too generic.

src/app/frontend/error/internalerror_controller.js
```diff
@@ -18,14 +18,18 @@
...
    */
-  constructor($stateParams) {
+  constructor($stateParams, kdAppConfigService) {
     /** @export {!angular.$http.Response} */
     this.error = $stateParams.error;
 
     /** @export */
     this.i18n = i18n;
+
+    this.dashboardVersion = kdAppConfigService.getDashboardVersion();
```
floreks: private variables should end with `_`.

build/script.js
maciaszczykm: Agreed here.

src/app/frontend/common/appconfig/appconfig_service.js
maciaszczykm: It seems to be good place in terms of similar capabilities of all those functions, but you need to double-check if replace won't harm existing functionality. It should not, but in the future we might need more service methods and I would not mix them with your methods (that will be modified by gulp).

So, I'd move it to another service and comment it properly just in case, but it's not required at the moment.

src/app/frontend/error/internalerror_controller.js
```diff
@@ -66,16 +70,13 @@ export class InternalErrorController {
...
-    let body = `#### Issue details\n\n##### Environment\n<!-- Describe how do you run Kubernetes ` +
-        `and Dashboard.\n      Versions of Node.js, Go etc. are needed only from developers. To ` +
-        `get them use console:\n      $ node --version\n      $ go version\n-->\n\n\`\`\`\n` +
-        `Dashboard version:\nKubernetes version:\nOperating system:\nNode.js version:\nGo version` +
-        `:\n\`\`\`\n##### Steps to reproduce\n<!-- Describe all steps needed to reproduce the ` +
-        `issue. It is a good place to use numbered list. -->\n\n##### Observed result\nDashboard ` +
-        `reported ${this.getErrorStatus()}:\n\`\`\`\n${this.getErrorData()}\n\`\`\`\n\n##### ` +
-        `Expected result\n<!-- Describe expected result as precisely as possible. -->\n\n##### ` +
-        `Comments\n<!-- If you have any comments or more details, put them here. -->`;
+    let title = ``;
```
maciaszczykm: Is it intentional to leave it blank?

src/app/frontend/error/internalerror_controller.js
```diff
@@ -66,16 +70,13 @@ export class InternalErrorController {
...
+    let body = `##### Steps to reproduce\n<!-- Describe all steps needed to reproduce the ` +
```
maciaszczykm: It doesn't match Dashboard's issue template, but it may be better, because most of the users do not use Dashboard directly from their workspace but in container.

You may also add missing headers and spaces while using `<!-- Comment here. -->` comments to make it prettier.

src/app/frontend/error/internalerror_controller.js
maciaszczykm: Okay, just finished reading. Nevermind.

src/app/frontend/error/internalerror_controller.js
maciaszczykm: https://github.com/kubernetes/dashboard/pull/1663#discussion_r102226358

build/script.js
```diff
@@ -102,6 +104,34 @@ gulp.task('scripts-watch', createScriptsStream(false));
...
+                // Partials generated by other tasks, e.g., Angular templates.
+                path.join(conf.paths.partials, '**/*.js'),
+                // Include base.js to enable some compiler functions, e.g., @export annotation
+                // handling and getMsg() translations.
+                path.join(
+                    conf.paths.bowerComponents, 'google-closure-library/closure/goog/base.js'),
+              ])
+              .pipe(closureCompiler(translation))
+              .pipe(gulpIf('**/app.js',gulpReplace('BUILD_DASHBOARD_VERSION', conf.deploy.version.release)))
+              .pipe(gulpIf('**/app.js',gulpReplace('BUILD_GIT_COMMIT', process.env.TRAVIS_COMMIT)))
```
cheld: it seems that gulpIf works on destination files - not source files. I tried a lot (including reordering) but something like `**/appconfig_service.js` just did not work. So it seems I have to go through the hole file for replace. I could make the string more unique to make accidental replacement unlikely. 

I could use gulp-replace-task (https://www.npmjs.com/package/gulp-replace-task) instead of gulp-replace to refactor the to replace lines in a single function. 

Not sure. Maybe I should place the replace code into build.js WDYT?

floreks: Did you try pipe before closureCompiler task? gulp.src first takes all the files in frontend src directory + some partial files and pipes them to closure compiler. Closure compiler bundles them to app.js so that's probably why appconfig is not found.

You can always just write some debug function as a pipe and check what files are available in there. There are also gulp-debug, gulp-print modules that can be useful for debugging. Gulp task replace looks promising. You could use that once everything else is working.

maciaszczykm: Changing `app.js` (whole aplication source) is very risky and should be avoided if it's possible. We should search for other solutions, doing it before build seems to be better idea, but not the only one. Moreover, if it will be done before minimizing to `app.js` it cannot change sources in `/src` directory.

maciaszczykm: >to replace lines in a single function

+1

cheld: I guess the problem is that closure compiler does not take the result from gulp replace, but loads the files from disk and discarding my changes

build/script.js
```diff
@@ -102,6 +105,34 @@ gulp.task('scripts-watch', createScriptsStream(false));
...
+                // Partials generated by other tasks, e.g., Angular templates.
+                path.join(conf.paths.partials, '**/*.js'),
+                // Include base.js to enable some compiler functions, e.g., @export annotation
+                // handling and getMsg() translations.
+                path.join(
+                    conf.paths.bowerComponents, 'google-closure-library/closure/goog/base.js'),
+              ])
+              .pipe(gulpIf('src/app/frontend/common/appconfig/appconfig_service.js',gulpReplace('BUILD_DASHBOARD_VERSION', conf.deploy.version.release)))
+              .pipe(gulpIf('src/app/frontend/common/appconfig/appconfig_service.js',gulpReplace('BUILD_GIT_COMMIT', process.env.TRAVIS_COMMIT)))
+              .pipe(closureCompiler(translation))
```
cheld: I have copied the path from gulp-print, but it is not working. 

I guess the problem is that closure compiler does not take the result from gulp replace, but loads the files from disk and discards my changes. 

Maybe because it is not a real gulp task but just a wrapper around a java program?

cheld:
PTAL

build/script.js
```diff
@@ -19,8 +19,11 @@ import async from 'async';
 import gulp from 'gulp';
 import gulpClosureCompiler from 'gulp-closure-compiler';
 import gulpHtmlmin from 'gulp-htmlmin';
+import gulpIf from 'gulp-if';
 import gulpModify from 'gulp-modify';
+import gulpPrint from 'gulp-print';
```
maciaszczykm: It seems like `gulp-print` is not saved to `package.json`:

```
$ gulp serve:prod
[15:25:37] Requiring external module babel-core/register
module.js:472
    throw err;
    ^

Error: Cannot find module 'gulp-print'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/maciaszczykm/Workspace/go/src/github.com/kubernetes/dashboard/build/script.js:24:1)
    at Module._compile (module.js:571:32)
    at loader (/Users/maciaszczykm/Workspace/go/src/github.com/kubernetes/dashboard/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/maciaszczykm/Workspace/go/src/github.com/kubernetes/dashboard/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/maciaszczykm/Workspace/go/src/github.com/kubernetes/dashboard/gulpfile.babel.js:29:1)
    at Module._compile (module.js:571:32)
    at loader (/Users/maciaszczykm/Workspace/go/src/github.com/kubernetes/dashboard/node_modules/babel-register/lib/node.js:144:5)
```

cheld: ah, thats debug left over...

cheld: fixed

maciaszczykm:
Tested on `gulp serve:prod` and it works (well, at least Dashboard version, commit should work then building on Travis):

screenshot at lut 27 15-37-33

I'd change just one thing, `<!-- comment -->` doesn't work inside code markups. What about changing it?

Anyways, LGTM from my side.

maciaszczykm:
BTW, what about @floreks changes from #1675?

He made it possible to do all changes outside `app.js` which is super important. If we merge this we should adjust gulp taks to execute before closure compilation.

cheld:
I guess @floreks should merge first

maciaszczykm:
@cheld You can use @floreks' code now. 

Please remember to rebase, because you have some conflits.

cheld:
Merged. PTAL

floreks:
Overall LGTM. Fix travis and we can merge.

src/app/frontend/error/internalerror_controller.js
```diff
@@ -35,6 +37,10 @@ export class InternalErrorController {
      * Hide side menu while entering internal error page.
      */
     this.kdNavService_.setVisibility(false);
+
+    /** build information*/
+    this.dashboardVersion_ = kdAppConfigService.getDashboardVersion();
```
floreks: Missing private annotation and type.

src/app/frontend/error/internalerror_controller.js
```diff
@@ -35,6 +37,10 @@ export class InternalErrorController {
...
+    this.gitCommit_ = kdAppConfigService.getGitCommit();
```
floreks: Same here

cheld:
PTAL

floreks:
Travis still fails.

PR #1664, opened 2017-02-22
[WIP] Pod detail tabs

IanLewis:
This PR is the first part of updating the Pod Detail UI to have an "Overview", "Logs", and "Terminal" tab per the [Troubleshooting design doc](https://github.com/kubernetes/dashboard/blob/master/docs/design/troubleshooting.md)

This PR will move the logs to a Tab interface along on the Pod details page with the pod overview. Terminal will be added in a later PR.


bryk:
Cool, thanks for doing this!

bryk:
@floreks Can you also review this?

floreks:
I see that it's still a raw version. Ping me once it's ready for review.

IanLewis:
Yes. It's still very raw. Trying to figure out how I will make the tabs work. i.e. whether I wanted to use stateconfig/router to switch between tabs, or use the tabs mechanism itself.

AFAICT, It seems I'll probably have to use tabs because if using state/router it would unload the current tab when switching to a new tab. But the md-tabs loads all controllers for content in each tab on initialization, but I don't want that for logs, terminal. So I guess I will have to lazy load logs content, and catch the event when the users switches to the new tab.

Also not sure how or if I should support a direct URL to pod logs like there is currently. I don't currently see it as a must but without using state/router, it seems like it would be hard. Maybe I can have a parameter that causes it to automatically switch to the logs tab? Not sure. Is this even necessary?

danielromlein:
@IanLewis is this on an environment somewhere that I could check it out?

maciaszczykm:
Closing as stale. Currently, action bar buttons are used instead of tabs.

PR #1667, opened 2017-02-22
"Send feedback" button v2

### LINKED ISSUE #1506: "Send feedback" button v2
Copying discussion from already merged PR: https://github.com/kubernetes/dashboard/pull/1461#issuecomment-263959659

@danielromlein said:
> Sorry for not chiming in earlier! I see this is merged, but if you feel like refining further, I think it would be better to 1) Have the card centered horizontally, 2) Have it be closer to the app bar (48dp), and hide the side nav. If there's a strong reason for the side nav to be shown, I would center the card in the remaining screen area. As per the [mockup](https://invis.io/8V9FX5Q9R#/207252275_500_Error).

maciaszczykm:
Closes #1506.

### DIFF
--- a/src/app/frontend/chrome/nav/nav_component.js
+++ b/src/app/frontend/chrome/nav/nav_component.js
@@ -82,6 +82,13 @@ export class NavController {
   toggle() {
     this.isVisible = !this.isVisible;
   }
+
+  /**
+   * Sets visibility of the navigation component.
+   */
+  set(isVisible) {
+    this.isVisible = isVisible;
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/app/frontend/chrome/nav/nav_component.js
```diff
@@ -82,6 +82,13 @@ export class NavController {
   toggle() {
     this.isVisible = !this.isVisible;
   }
+
+  /**
+   * Sets visibility of the navigation component.
+   */
+  set(isVisible) {
```
floreks: s/set/setVisible|setVisiblity

PR #1673, opened 2017-02-24
fix prompt

foxyriver:
I think when the labelForm.key is valid, you need cancel the prompt.


mijajasinski:
Hi!
I'm not sure I got it right, but it seems your change is actually preventing the ng-messages from being hidden when $invalid gets set to false. Note that the error message still disappears when there are no errors, since none of the subsequent ng-when is satisfied.

Let me know if I'm missing something.

foxyriver:
@mijajasinski I used dashboard release 1.5, and found the problem which is the error message will not disappear when there are no errors. Now I debug the latest dashboard code, this problem is gone. but I still didn't find where it fixed.
I think I am wrong, it didn't need to delete $invalid. And I will still seek how it fixed to resolve this problem.

mijajasinski:
OK, I'll close this PR for now, then.

PR #1676, opened 2017-02-27
List third party resource objects backend

maciaszczykm:
Open for a discussion here, because it's still raw version. What would you change? Any suggestions are appreciated.

Sample answer from http://localhost:9091/api/v1/thirdpartyresource/cron-tab.stable.example.com:
``` json

{
  "objectMeta": {
   "name": "cron-tab.stable.example.com",
   "creationTimestamp": "2017-02-20T09:05:08Z"
  },
  "typeMeta": {
   "kind": "thirdpartyresource"
  },
  "description": "A specification of a Pod to run on a cron style schedule",
  "versions": [
   {
    "name": "v1"
   }
  ],
  "objects": {
   "kind": "CronTabList",
   "apiVersion": "stable.example.com/v1",
   "metadata": {
    "selfLink": "/apis/stable.example.com/v1/crontabs",
    "resourceVersion": "232892"
   },
   "items": [
    {
     "kind": "CronTab",
     "apiVersion": "stable.example.com/v1",
     "metadata": {
      "name": "my-new-cron-object",
      "namespace": "default",
      "selfLink": "/apis/stable.example.com/v1/namespaces/default/crontabs/my-new-cron-object",
      "uid": "56f42f16-f9b3-11e6-a67e-acbc32b15303",
      "resourceVersion": "122472",
      "creationTimestamp": "2017-02-23T10:32:13Z"
     }
    }
   ]
  }
 }
```

To check it on your own use examples from https://kubernetes.io/docs/user-guide/thirdpartyresources/.

### DIFF
--- a/src/app/backend/resource/thirdpartyresource/detail.go
+++ b/src/app/backend/resource/thirdpartyresource/detail.go
@@ -45,17 +45,22 @@ func GetThirdPartyResourceDetail(client k8sClient.Interface, config clientcmd.Cl
 		return nil, err
 	}
 
-	return getThirdPartyResourceDetail(config, thirdPartyResource), nil
+	return getThirdPartyResourceDetail(config, thirdPartyResource)
 }
 
-func getThirdPartyResourceDetail(config clientcmd.ClientConfig, thirdPartyResource *extensions.ThirdPartyResource) *ThirdPartyResourceDetail {
+func getThirdPartyResourceDetail(config clientcmd.ClientConfig, thirdPartyResource *extensions.ThirdPartyResource) (*ThirdPartyResourceDetail, error) {
+	objects, err := getThirdPartyResourceObjects(config, thirdPartyResource)
+	if err != nil {
+		return nil, err
+	}
+
 	return &ThirdPartyResourceDetail{
 		ObjectMeta:  common.NewObjectMeta(thirdPartyResource.ObjectMeta),
 		TypeMeta:    common.NewTypeMeta(common.ResourceKindThirdPartyResource),
 		Description: thirdPartyResource.Description,
 		Versions:    thirdPartyResource.Versions,
-		Objects:     getThirdPartyResourceObjects(config, thirdPartyResource),
-	}
+		Objects:     objects,
+	}, err
 }
 
 // getThirdPartyResourceGroupVersion returns first group version of third party resource. It's also known as
--- a/src/app/backend/resource/thirdpartyresource/detail.go
+++ b/src/app/backend/resource/thirdpartyresource/detail.go
@@ -45,22 +45,22 @@ func GetThirdPartyResourceDetail(client k8sClient.Interface, config clientcmd.Cl
 		return nil, err
 	}
 
-	return getThirdPartyResourceDetail(config, thirdPartyResource)
-}
-
-func getThirdPartyResourceDetail(config clientcmd.ClientConfig, thirdPartyResource *extensions.ThirdPartyResource) (*ThirdPartyResourceDetail, error) {
 	objects, err := getThirdPartyResourceObjects(config, thirdPartyResource)
 	if err != nil {
 		return nil, err
 	}
 
+	return getThirdPartyResourceDetail(thirdPartyResource, objects), nil
+}
+
+func getThirdPartyResourceDetail(thirdPartyResource *extensions.ThirdPartyResource, objects ThirdPartyResourceObjectList) *ThirdPartyResourceDetail {
 	return &ThirdPartyResourceDetail{
 		ObjectMeta:  common.NewObjectMeta(thirdPartyResource.ObjectMeta),
 		TypeMeta:    common.NewTypeMeta(common.ResourceKindThirdPartyResource),
 		Description: thirdPartyResource.Description,
 		Versions:    thirdPartyResource.Versions,
 		Objects:     objects,
-	}, err
+	}
 }
 
 // getThirdPartyResourceGroupVersion returns first group version of third party resource. It's also known as
--- a/src/app/backend/resource/thirdpartyresource/detail.go
+++ b/src/app/backend/resource/thirdpartyresource/detail.go
@@ -16,23 +16,27 @@ package thirdpartyresource
 
 import (
 	"log"
+	"strings"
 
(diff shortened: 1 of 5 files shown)


maciaszczykm:
https://github.com/kubernetes/features/issues/95#issue-180289753 list of TPR missing functionalities.

src/app/backend/resource/thirdpartyresource/detail.go
```diff
@@ -41,14 +45,50 @@ func GetThirdPartyResourceDetail(client k8sClient.Interface, name string) (*Thir
...
+		group = thirdPartyResource.ObjectMeta.Name
+	}
+
+	return schema.GroupVersion{
+		Group:   group,
+		Version: version,
+	}
+}
+
+// TODO(maciaszczykm): Is there any Kubernetes API to use instead of this function?
```
maciaszczykm: Didn't found anything useful yet.

src/app/backend/resource/thirdpartyresource/object.go
```diff
@@ -0,0 +1,62 @@
...
+func (list *ThirdPartyResourceObjectList) GetListMeta() metav1.List {
+	return &list.Metadata
+}
+
+func getThirdPartyResourceObjects(config clientcmd.ClientConfig, thirdPartyResource *extensions.ThirdPartyResource) ThirdPartyResourceObjectList {
+	var list ThirdPartyResourceObjectList
+
+	// TODO(maciaszczykm): Is there a way to alter existing rest client to use different group version?
+	// It would allow to create REST client only once, not for each call.
+	// Creating cache for clients also seems to be risky (could take a lot of memory).
```
maciaszczykm: Didn't found anything useful yet. Might consider implementing client cache... WDYT?

src/app/backend/resource/thirdpartyresource/detail.go
```diff
@@ -41,14 +45,50 @@ func GetThirdPartyResourceDetail(client k8sClient.Interface, name string) (*Thir
...
 	return &ThirdPartyResourceDetail{
 		ObjectMeta:  common.NewObjectMeta(thirdPartyResource.ObjectMeta),
 		TypeMeta:    common.NewTypeMeta(common.ResourceKindThirdPartyResource),
 		Description: thirdPartyResource.Description,
 		Versions:    thirdPartyResource.Versions,
+		Objects:     getThirdPartyResourceObjects(config, thirdPartyResource),
 	}
 }
+
+// TODO(maciaszczykm): It should always take preffered version. Is there any Kubernetes API to use instead of this function?
```
maciaszczykm: Didn't found anything useful yet. Do you know any useful API?

src/app/backend/resource/thirdpartyresource/object.go
```diff
@@ -0,0 +1,72 @@
...
+
+	restConfig, err := newClientConfig(config, getThirdPartyResourceGroupVersion(thirdPartyResource))
+	if err != nil {
+		log.Println(err)
+		return list
+	}
+
+	// TODO(maciaszczykm): Is there a way to alter existing rest client to use different group version?
+	// It would allow to create REST client only once, not for each call.
+	// Creating cache for clients also seems to be risky (could take a lot of memory).
```
maciaszczykm: Didn't found anything useful yet. Might consider implementing client cache... WDYT?

src/app/backend/resource/thirdpartyresource/detail.go
bryk: What is the preferred version? Can you link to any docs?

src/app/backend/resource/thirdpartyresource/detail.go
bryk: AFAIR, no. But this issue has some interesting links you can dig into: https://github.com/kubernetes/kubernetes/issues/18622

src/app/backend/resource/thirdpartyresource/object.go
bryk: I'd go for easiest solution for now. AFAIU clients are just sets of a few properties. There are no long-living connections or expensive threads.

maciaszczykm:
@bryk `preferredVersion` seems to be prefereed version of API to be used in storage, quoting from https://github.com/kubernetes/kubernetes/blob/81d788dd6e0748e5d53a62c16d933c5f7a0718af/federation/apis/swagger-spec/federation.json#L66:

```
preferredVersion is the version preferred by the API server, which probably is the storage version.
```

TPR objects are only accessible through that version, it doesn't work with other ones... In most of the cases, that I've checked it was the first version from resource and actually there is old PR at core repository, that should do it:

https://github.com/kubernetes/kubernetes/pull/24006/files

Code was moved, but it seems that my assumption is still correct:

https://github.com/kubernetes/kubernetes/blob/320f7ce6f348a85a9a27e128e0eddee023e7478c/staging/src/k8s.io/client-go/discovery/discovery_client.go#L115

However, it seems that PR to support multiple versions is already there:

https://github.com/kubernetes/kubernetes/pull/36977/files

maciaszczykm:
Plus one more comment from core: https://github.com/kubernetes/kubernetes/pull/36977/files#r96520383

It seems, that we need to work with what we've got.

floreks:
Agree. I've also checked code and existing TPRs and it seems that currently there is not better way to handle them. We'll be in sync with the changes in core and for now this implementation LGTM.

maciaszczykm:
PTAL

src/app/backend/resource/thirdpartyresource/object.go
```diff
@@ -0,0 +1,69 @@
...
+	return &list.Metadata
+}
+
+func getThirdPartyResourceObjects(config clientcmd.ClientConfig, thirdPartyResource *extensions.ThirdPartyResource) ThirdPartyResourceObjectList {
+	var list ThirdPartyResourceObjectList
+
+	restConfig, err := newClientConfig(config, getThirdPartyResourceGroupVersion(thirdPartyResource))
+	if err != nil {
+		log.Println(err)
+		return list
```
floreks: Shouldn't we return an error here and let it be propagated to frontend?

maciaszczykm: Agreed.

maciaszczykm:
PTAL

PR #1685, opened 2017-03-01
Parse URLs in annotations

### LINKED ISSUE #1680: Feature request: parse URLs in annotations
Not sure how common this is but we are storing URLs to various external resources (jenkins build, git repo, documentation++),  as annotations for deployments, services and all other objects in Kubernetes and it would be nice if they where parsed and clickable from the UI without having to copy/paste them.

maciaszczykm:
Fixes #1680:

image

### DIFF
--- a/src/app/frontend/common/components/labels/labels_controller.js
+++ b/src/app/frontend/common/components/labels/labels_controller.js
@@ -19,10 +19,18 @@
 const alwaysVisibleLabelsNumber = 5;
 
 /**
- * Regular expression matching URLs.
+ * Regular expression for URL validation created by @dperini.
+ * https://gist.github.com/dperini/729294
+ *
  * @type {RegExp}
  */
-const urlRegexp = new RegExp('(https?|ftp)://(-\.)?([^\s/?\.#-]+\.?)+(/[^\s]*)?', 'i');
+const urlRegexp = new RegExp(
+    '^(?:(?:https?|ftp)://)(?:\\S+(?::\\S*)?@)?(?:(?!(?:10|127)(?:\\.\\d{1,3}){3})(?!(?:169\\.254|192\\.168)' +
+        '(?:\\.\\d{1,3}){2})(?!172\\.(?:1[6-9]|2\\d|3[0-1])(?:\\.\\d{1,3}){2})(?:[1-9]\\d?|1\\d\\d|2[01]\\d|' +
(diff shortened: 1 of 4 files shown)


src/app/frontend/common/components/labels/labels_controller.js
```diff
@@ -19,6 +19,12 @@
 const alwaysVisibleLabelsNumber = 5;
 
 /**
+ * Regular expression matching URLs.
+ * @type {RegExp}
+ */
+const urlRegexp = new RegExp('(https?|ftp)://(-\.)?([^\s/?\.#-]+\.?)+(/[^\s]*)?', 'i');
```
rf232: should we anchor the regexp? since we don't try to extract the actual url, and just make a href of the whole annotation

maciaszczykm: Hmm, with regexp it's possible to determine if annotation value is URL or not. Without it we would make all annotations hrefs.

maciaszczykm: Plus the fact, that at the moment only value is clickable link, not whole chip with annotation.

rf232: no i mean to change the regexp to '^(https?|ftp)://(-\.)?([^\s/?\.#-]+\.?)+(/[^\s]*)?$' because otherwise an annotation like "mixed-text-link: Foo http://k8s.io" would also match

maciaszczykm: Ahh, okay. You're totally right.

maciaszczykm: Updated URL regex to the most accurate one from https://mathiasbynens.be/demo/url-regex. It's on MIT license, I've left comment in the code.

maciaszczykm:
https://github.com/kubernetes/dashboard/pull/1685#discussion_r103688290:

> Updated URL regex to the most accurate one from https://mathiasbynens.be/demo/url-regex. It's on MIT license, I've left comment in the code.

cheld:
Hmm...just to make sure: this PR does not create an XSS issue, which is s concern, right?

maciaszczykm:
@cheld It doesn't run any code from annotations. It puts annotation value in `<a href="...">...</a>` if it matches URL regex. XSS shouldn't be a problem here.

Starefossen:
This is so perfect, thanks a lot @maciaszczykm 😄🎉

PR #1693, opened 2017-03-06
Fix requirements installation guide

VeraLyu:
As dashboard now depends on go 1.7+,
instructions on requirements installation can be misleading.
Fix it by substitute 1.6 go installation instructions
to 1.8.


VeraLyu:
I signed it!

PR #1695, opened 2017-03-07
TPR instances list

maciaszczykm:
Added TPR instances list to its details view:

image

image

Things to consider for the future:

- [ ] If there are >3 TPRs perhaps we should display only 3 of them **in the menu** and in the next line put a link to list of all of them (something like `... and X more`)
- [ ] How to handle long TPR names? At the moment full TPR names are displayes, perhaps we should use only short name (i.e. `CronTabs`/`Cron Tabs` instead of `cron-tab.stable.example.com`)?

Related to #1504.


src/app/backend/resource/thirdpartyresource/objects.go
```diff
@@ -0,0 +1,113 @@
...
+
+// ThirdPartyResourceObject is a single instance of third party resource.
+type ThirdPartyResourceObject struct {
+	metav1.TypeMeta `json:",inline"`
+	Metadata        metav1.ObjectMeta `json:"metadata,omitempty"`
+}
+
+// ThirdPartyResourceObjectList is a list of third party resource instances.
+type ThirdPartyResourceObjectList struct {
+	ListMeta        common.ListMeta `json:"listMeta"`
```
bryk: Hmm... Why do we have 3 types of meta here? Do we need all of them?

src/app/frontend/thirdpartyresourcedetail/detail.html
```diff
@@ -14,4 +14,21 @@
...
+    <kd-content>
+        <kd-third-party-resource-objects object-list="$ctrl.tprDetail.objects"
+                                         object-list-resource="$ctrl.tprObjectsResource"
+                                         with-statuses="false">
+            <kd-zerostate>
+              <div class="kd-zerostate-title">
+                  [[There is nothing to display here|Title for objects card zerostate in third party resource details page.]]
+              </div>
+              <div class="kd-zerostate-text">
+                  [[There are currently no instances of third Third Party Resource.|Text for objects card zerostate in third party resource details page.]]
```
bryk: third Third?

src/app/frontend/thirdpartyresourcedetail/objectcard_component.js
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @final
+ */
+export class ObjectCardController {
+  /**
+   * @ngInject,
```
bryk: Comma at the end?

maciaszczykm: Removed.

src/app/frontend/thirdpartyresourcedetail/detail.html
maciaszczykm: Fixed.

src/app/backend/resource/thirdpartyresource/objects.go
maciaszczykm: Removed unnecessary one.

maciaszczykm:
Applied fixes, PTAL.

danielromlein:
Since TPR's are not Namespaced, might there be a better location in the information architecture than where they are currently? The navigation scheme was to have only namespace-dependent objects below the Namespace dropdown...
/ @maciaszczykm @floreks @bryk

PR #1721, opened 2017-03-13
Fix frontend tests

floreks:
Fixes frontend tests that are failing after updating angular to 1.6. Travis is green because we haven't removed cached dependencies.

Default [$locationProvider](https://docs.angularjs.org/api/ng/provider/$locationProvider) prefix has changed to `!` now and unhandled promise rejections cause errors so they have to be handled by us now.


maciaszczykm:
There are `!` missing in integration tests too, in: `getElementByAppName()` funtions and `    browser.get('#/deploy/app');`.

PR #1749, opened 2017-03-17
Add storage group view

### LINKED ISSUE #1733: Storage menu group cannot be viewed
It has only one element at the moment, but to be consistent we should implement it:

image

maciaszczykm:
Closes #1733.


maciaszczykm:
@floreks PTAL

Issue #371, opened 2016-02-12
Introduce HTML formatter to the project


bryk:
Like this one: https://www.npmjs.com/package/gulp-prettify

The formatter would save us code review time and produce consistent code. 

The proposal is pretty similar to what we have with JS now. In `gulp lint` check that HTML formatter does not change the code.

@maciaszczykm Can you take it? :)

maciaszczykm:
@bryk Sure, I can.

maciaszczykm:
I think, that `gulp-prettify` might not be suitable for us, because of missing configuration options.

First of all it doesn't allow to define max line length, which makes some lines 200-300 characters long. It's too much and there is no option to fix it using `gulp-prettify`. I've made research and there are not too much HTML formatters for Gulp.

Second thing is formatting style, which seems to be strange, from this:

``` html
<md-input-container class="md-block" layout-wrap>
  <label>Image pull secret data</label>
  <textarea name="data" ng-model="ctrl.data" class="kd-secret-data"
            columns="1" rows="5"  ng-pattern="ctrl.dataPattern" required></textarea>
  <div ng-messages="ctrl.secretForm.data.$error" role="alert" multiple>
      <div ng-message="required">Data is required</div>
      <div ng-message="pattern">Data must be Base64 encoded</div>
  </div>
</md-input-container>
```

To this:

``` html
<md-input-container class="md-block" layout-wrap> <label>Image pull secret data</label> <textarea name="data" ng-model="ctrl.data" class="kd-secret-data" columns="1" rows="5" ng-pattern="ctrl.dataPattern" required></textarea>
   <div ng-messages="ctrl.secretForm.data.$error" role="alert" multiple>
      <div ng-message="required">Data is required</div>
      <div ng-message="pattern">Data must be Base64 encoded</div>
   </div>
</md-input-container>
```

Here is Gulp task, that I've used:

``` javascript
gulp.task('prettify', function() {
  gulp.src([path.join(conf.paths.frontendSrc, '**/*.html')])
      .pipe(prettify({
        indent_size: 2,
      }))
      .pipe(gulp.dest(conf.paths.frontendSrc))
});
```

Other possible configuration options are:

```
indent_handlebars: true,
indent_inner_html: true,
preserve_newlines: false,
max_preserve_newlines: 1,
brace_style: 'expand',
indent_char: ' ',
indent_size: 2
```

But none of them seems to solve mentioned issues. I'll do more research and let you know if I'll find anything useful :)

[...]

PR #1760, opened 2017-03-20
Add basic html/go formatting and common gulp format script

floreks:
Added go/html formatting. HTML formatter will preserve bigger part of the code. It will only split and keep all tag attributes aligned in new lines, keep line indentation set to 2, and try to preserve max line length which is now 120. Additionally it will add new lines on end of each file so we don't have to worry about that anymore.

I've also added task for go formatting (`gulp format-go`) and exposed one common `gulp format` task to format JS, HTML and Go code at once.


maciaszczykm:
Closes #371.

PR #1787, opened 2017-03-27
Update menu

### LINKED ISSUE #1778: Menu entries should be sorted
Just like at https://kubernetes.io/docs/resources-reference/v1.5/.

maciaszczykm:
image

I'll continue if you will like it @kubernetes/dashboard-maintainers @danielromlein. Done to align with https://kubernetes.io/docs/resources-reference/v1.5/.

Closes #1778.


bryk:
LGTM

danielromlein:
Seems like the nav items should be Title Case? e.g. `Discovery and Load Balancing` WDYT?

maciaszczykm:
PTAL

floreks:
Looking at [docs](https://kubernetes.io/docs/resources-reference/v1.6/#storageclass-v1-storage) and our menu I am a bit confused now. We should decide whether we want to group nav menu on our own or be more consistent with the docs.

Docs do not group items based on namespace relation but rather on their role, so i.e. under `Config and Storage` section there is `Storage Class` resource that is non-namespaced.

If we want to keep our current grouping then we have to think how to display Roles because there is `ClusterRole` resource under Cluster section that is non-namespaced and there is also `Role` resource under Cluster section that is namespaced and showing them currently under Cluster does not make sense to me.

I am leaning towards grouping that is introduced in docs and rethinking how  and where to show namespace selector. IMHO it would be more consistent and we do not have to think about custom grouping of resources.

@danielromlein @maciaszczykm @bryk WDYT?

maciaszczykm:
@floreks Pull request title might be confusing, but what I wanted to achieve here is sorting of menu entries and making it more consistent with current state (that's why I've renamed `Access Control` to `Role` and moving few other entries).

What you're saying has bigger impact and needs further discussion. Currently we tried to keep all non-namespaced resources above namespace selector, but then contribution from @mlamina came and we decided to merge and iterate on it. This pull requests keep `Roles` in non-namespaced resource group, but we need to split it to `Roles` and `Cluster Roles` if we want to be consistent and have one menu entry per one resource.

I like the idea of using similar grouping as in the docs, because it makes things easier for us. However, I'm not sure if we can do it, because we're implementing user interface which has to be easy to understand and use.

I'd suggest reviewing it now, because it should be step forward no matter which decision we'll take and move this discussion to some issue where we'll decide what to do next.

floreks:
I'm fine with merging this and continuing the discussion in related issue.

PR #1792, opened 2017-03-29
Add support for sorting and fix minor issues

### LINKED ISSUE #966: Resource sorting in the UI
We display list of resources in many places. The tables are not sortable, though. We should make them so.

@maciaszczykm What do you think?

floreks:
It required some major changes in frontend. Currently sorting is supported for 3 types of columns: name, status and age. 

In order to add support for sorting to the existing list open the cardlist.html template and add i.e.: `sortable="true"` and `sord-id=$root.SortableProperties.NAME` to the `kd-resource-card-header-column` representing name column. It will enable sorting for this column and use corresponding backend method for sorting.

## Important changes
 - Created `DataSelectService` instead of `PaginationService` that handles all kinds of data selection, including: sorting, filtering, paging. Currently sorting and paging is implemented.
 - Moved `list`, `list-resource` and `data-select-id` dependency from `kd-resource-card-list-pagination` to `kd-resource-card-list`. It will be shared with child components thanks to parent - child relationship between components.
 - Moved data select ids to corresponding component controller `getSelectId` method. In case if it's empty, data select logic will not be applied to the list.
 - By default all resources will be sorted by age in the ascending order. From youngest to oldest one.
 - In order to share supported SortableProperties I have added run config to DataSelect module and injected them into the rootScope that is by default available in all templates by executing `$root.SortableProperties.NAME`. I'm not sure if there is a better way to do that. They are const values so there should be no risk with exposing them in root scope.

## Additional changes
 - Add missing graph cards to cluster view.
 - Update closure compiler externs to version 1.6
 - Enable console log usage in karma tests
 - Add missing externs to `backendapi.js`
 - Update `foreground-3` css variable to match material design spec

UI:

Default sorting by age
zrzut ekranu z 2017-03-29 14-33-46

Sorted by name
zrzut ekranu z 2017-03-29 14-33-59

TODO:
 - [x] Add tests

Closes #966.


floreks:
@bryk @maciaszczykm @danielromlein PTAL. Also WDYT about using arrows to indicate sorting?

danielromlein:
The [Material Spec ](https://material.io/guidelines/components/data-tables.html#data-tables-tables-within-cards)is to use the [Down Arrow Icon](https://material.io/icons/#ic_arrow_downward). I notice GCP uses the [Keyboard Arrow Down Icon](https://material.io/icons/#ic_keyboard_arrow_down), which they may be doing because of space constraints.
I think we have room though for the Down Arrow Icon.
This is preferable to the screenshot, because those arrows are similar to the icon used for a dropdown and could be confusing to users.

danielromlein:
Pagination controls should be at the [bottom of the card](https://material.io/guidelines/components/data-tables.html#data-tables-tables-within-cards), so we have room to add card-specific controls later if we want.

floreks:
I'll try to use arrow icons and we will see how it works. 

As for the pagination controls it was moved to the top of the card list because there were plans to make list headers sticky. Similar to what inbox does when you open a mail. This way it would be easier to control the list (filter, sort, paginate) instead of having to always scroll to the bottom of the list in order to select some data. Also I think that with longer lists i.e. 50 items per page (we are planning to reintroduce items per page selector eventually) having to scroll down on every page change would be inconvenient for users.

floreks:
I see one potential issue with using arrows. We will probably not support sorting for every column and if we decide to follow material spec then user won't know which columns are actually sortable. On click arrow would be displayed for column that is sorted but some visual indicator that this column is sortable would be nice. That's why I have used these small arrows and opacity to show which columns are sortable and which column is currently sorted in what direction.

WDYT?

danielromlein:
@floreks 
>As for the pagination controls it was moved to the top of the card list because there were plans to make list headers sticky.

Ah, OK. Well I still think it's important to consider if there were any table-relevant actions we want to include, as this would be made impossible by putting pagination up there. I'll try to do some research into pros / cons of having pagination at the top vs. the bottom. 
screen shot 2017-03-30 at 3 06 05 pm

>We will probably not support sorting for every column and if we decide to follow material spec then user won't know which columns are actually sortable. 

That's a good point. Drive handles this simply by using a hover state on columns that are sortable; no hover state is present on columns that aren't sortable. WDYT about that solution?

floreks:
>That's a good point. Drive handles this simply by using a hover state on columns that are sortable; no hover state is present on columns that aren't sortable. WDYT about that solution?

That was actually first thing I thought about and implemented but from user perspective it's also not that nice because he can't visually recognize which columns are sortable but rather has to hover over every column on every list and check for cursor change. Additionally on touch devices user won't be able to do that and has to click on every column to check if it's sortable.

floreks:
I have tried to use something from both ideas and come up with that:

zrzut ekranu z 2017-03-31 11-19-47

zrzut ekranu z 2017-03-31 11-19-57

According to spec sorted column should be 87% black so I have changed that. To indicate that column is sortable I have left default icon with default black level (54%). 

@danielromlein WDYT?

danielromlein:
Mm, I hadn't considered that the hover-state requires a user to hover / click any column to understand its sortability. 

The arrows in your most recent version I think would be confusing – Material only shows arrows on the currently sorted column.

All that to say, I think the up / down caret is a well-enough established pattern that it makes sense to use. In addition to the caret, I would add a hover state of 87% to sortable column headers on `hover` and `active`.

floreks:
Ok, I have applied some changes.

Default view (sorted by age, asc):
zrzut ekranu z 2017-04-03 10-44-08

Sorted by name, desc with hover on status column:
pods - kubernetes dashboard - google chrome_003

danielromlein:
I think it would be a better experience to have the icon _after_ the `th` value – that way the text of the `th` is horizontally aligned with the text below it. Otherwise I think this solution looks good!

build/script.js
```diff
@@ -136,19 +136,19 @@ function createCompileTask(translation) {
  */
 function compileES6(translation) {
   let externs = [
-    path.join(conf.paths.nodeModules, 'google-closure-compiler/contrib/externs/angular-1.5.js'),
+    path.join(conf.paths.nodeModules, 'google-closure-compiler/contrib/externs/angular-1.6.js'),
```
bryk: Do we need 1.6 for sorting?

src/app/frontend/cluster/cluster.html
```diff
@@ -14,6 +14,19 @@
 limitations under the License.
 -->
 
+<div layout="row">
+  <kd-graph-card graph-title="[[CPU usage|Title for graph card displaying CPU metric of one all resources.]]"
+                 metrics="::$ctrl.cluster.nodeList.cumulativeMetrics"
```
bryk: I think that this PR adds not only sorting. Can you outline all changes in the PR message?

bryk:
selection_183

Re two arrows: I'd just show one. It is hard to differentiate where do I sort, because the arrows are so similar in color.

build/script.js
floreks: Not really, I have just noticed that this wasn't updated. Since it's small I haven't considered moving it into separate PR.

src/app/frontend/cluster/cluster.html
floreks: Ye. I will outline all changes once everything is finished. Usually if I come across small issue/missing code I'm trying to add/fix it. If it's something bigger then I am creating new PR.

floreks:
> Re two arrows: I'd just show one. It is hard to differentiate where do I sort, because the arrows are so similar in color.

I've got similar feeling, however as discussed above if we choose to show only one arrow then how should we behave when column is not sorted but sortable? It would be nice to just look at all columns and be able to tell which are sortable, sorted and not sortable without having to hover or click on them.

We could make the difference between them more visible by increasing black level for sorted column and decreasing for not sorted.

floreks:
Latest UI change. When not sorted both icons are shown are default text (54% black) is shown. On hover text is changed to 87% black (inline with spec). On click only one icon is shown.

@bryk @maciaszczykm @danielromlein WDYT?

Default sorted by age (asc):
zrzut ekranu z 2017-04-10 14-23-19

Sorted by name (desc):
zrzut ekranu z 2017-04-10 14-23-30

danielromlein:
>We could differentiate them better by increasing black level for sorted column and decreasing for not sorted.

@floreks I like this approach better. I think the current design looks like a dropdown, which is misleading. @bryk WDYT?

Sorry for making you revise this so much!

floreks:
> Sorry for making you revise this so much!
That's ok :) We all want to make it look good so this is inevitable.

Material spec has few black levels: 12%, 38%, 54% and 87%. Normally for table headers we should use 54% black and 38% should be used for disabled text. IMO even if i increase 87% to 99 or 100 there is no much visual difference.

Personally I like this https://github.com/kubernetes/dashboard/pull/1792#issuecomment-291085365 option just with icons moved to the right.

zrzut ekranu z 2017-04-12 10-24-46

We have to decide on something because I have ran out of options here :)

danielromlein:
Ok, how's this sound: on currently sorted row, the carets should be 20% and 87%. Increase the sorted column header font weight to medium (still at 87% black) to make it stand out more. Have all other carets be at 38% opacity.
sorting

floreks:
IMHO having 38% for all non active carets looks nice. Even on sorted column it stand out nicely (87% vs 38%). WDYT?

zrzut ekranu z 2017-04-12 16-25-02

floreks:
@bryk @danielromlein @maciaszczykm can we move forward with that? I think only thing missing is to agree on UI part.

maciaszczykm:
We can always update opacity in the future. Let's merge, because users really wanted to have it :-)

PR #1797, opened 2017-03-30
Allow default setttings override for serve and serve:prod

floreks:
Currently when using `serve` or `serve:prod` dashboard could only connect to local cluster. In order to change that we've had to change build files. I have changed that to first check environment variables, then use predefined ones.

This allows to easily configure build pipeline to use external cluster for `serve` and `serve:prod`, i.e. I'm using kubeconfig env variable to connect to minikube cluster that by default is secured and does not expose insecure api server port.

Environment variables:
 - `KUBE_DASHBOARD_APISERVER_HOST` - do `export KUBE_DASHBOARD_APISERVER_HOST=<HOST>` to override default api server url
 - `KUBE_DASHBOARD_KUBECONFIG` - do `export KUBE_DASHBOARD_KUBECONFIG=<PATH>` to use kubeconfig file instead of api server host param. Kubeconfig env variable has higher priority than apiserver host.

~I will also adjust the documentation.~ Done.

### DIFF
--- a/build/conf.js
+++ b/build/conf.js
@@ -101,11 +101,24 @@ export default {
      */
     apiServerHost: 'http://localhost:8080',
     /**
+     * Env variable with address for the Kubernetes API server.
+     */
+    envApiServerHost: process.env.KUBE_DASHBOARD_APISERVER_HOST,
+    /**
+     * Env variable with path to kubeconfig file.
+     */
+    envKubeconfig: process.env.KUBE_DASHBOARD_KUBECONFIG,
+    /**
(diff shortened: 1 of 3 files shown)


cheld:
You could have used kubectl proxy, also

floreks:
I know that it's possible to use kubectl proxy but with this change user does not need to have kubectl. Also he doesn't need to remember to always start a proxy tunnel before working on dashboard. User can just add export command to his `.bashrc` and new config will be picked up and used.

IMO it's more convenient to use and doesn't require any external dependencies. That's why I have introduced this change.

Issue #1745, opened 2017-03-16
Percentage values for maxUnavailable/maxSurge for Deployment not rendered properly


ahmetb:
##### Environment

```
Dashboard version: v1.5.0
Kubernetes version: Server Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.3", GitCommit:"029c3a408176b55c30846f0faedf56aae5992e9b", GitTreeState:"clean", BuildDate:"2017-02-15T06:34:56Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}
```

##### Steps to reproduce

Create a Deployment with percent values:

```
spec:
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 40%
  ...
```

##### Observed result

Percent value is shown as `0` in the Dashboard.

![image](https://cloud.githubusercontent.com/assets/159209/24018911/1c7d0dd6-0a53-11e7-8ef8-aed62edc5aa9.png)

PR #1799, opened 2017-03-30
Deployment strategy MaxSurge and MaxUnavailability can actually be strings too.

rf232:
fixes: #1745

### DIFF
--- a/build/gocommand.js
+++ b/build/gocommand.js
@@ -109,8 +109,6 @@ function checkGoVersion() {
         env: env,
       },
       function(error, stdout) {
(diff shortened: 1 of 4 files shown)


build/gocommand.js
```diff
@@ -109,6 +109,8 @@ function checkGoVersion() {
         env: env,
       },
       function(error, stdout) {
+        deferred.resolve();
+        return;
```
maciaszczykm: Should it be here?

rf232: yes, local go has a borked version number :)

PR #1804, opened 2017-03-31
Redirect after namespace change if on details of resource from different namespace

### LINKED ISSUE #816: Namespace selection dropdown
#### Issue details

I think the default choice behaviour is not obvious.
By default, the drop down is on "namespace not selected". You would expect from this, to display all the resources from every namespace. But actually it displays the resources from whatever namespace is configured as the `default` namespace.

I see two possibilities personally:
- Remove this entry, and by default, select whatever is the default namespace
- Keep the entry but display everything from every namespace

Thoughts ?
##### Environment

```
Dashboard version: v1.1.0-beta2
Kubernetes version: v1.2.4
Operating system: CoreOS
```

maciaszczykm:
Closes #816.

### DIFF
--- a/src/app/frontend/common/namespace/component.js
+++ b/src/app/frontend/common/namespace/component.js
@@ -14,11 +14,22 @@
 
 import {namespaceParam} from 'chrome/state';
 
-/** Internal key for empty selection. To differentiate empty string from nulls. */
+/**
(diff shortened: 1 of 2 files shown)


maciaszczykm:
PTAL

maciaszczykm:
Tests will follow.

Issue #1546, opened 2016-12-29
dashboard 1.5 unknown logs prefix appears 


Slahser:
##### Environment

```
Dashboard version: 1.5
Kubernetes version: 1.5.1 
Operating system: Centos 7.2

```

##### Steps to reproduce

compare dashboard log panel with `kubectl logs xxx`

##### Observed result

i.e. 2017-03-15T10:02:38.48196648Z appears in dashboard log panel.

##### Expected result

maybe there is some ways to disable the prefix display?

##### Comments

IanLewis:
@Slahser The screenshot of the dashboard doesn't seem to be viewable. Could you either re-upload your screenshot or expand on what the issue is?

maciaszczykm:
@Slahser What is unknown prefix? By default Dashboard displays log message time, i.e. `2017-03-15T10:02:38.48196648Z`. We cannot see the screenshots. Please reopen issue if it still occurs.

Slahser:
oh sorry i ignored this issue... 

just like @maciaszczykm say ,it maybe `2017-03-15T10:02:38.48196648Z` , can we have a feature toggle here?

maciaszczykm:
Okay, we'll check.

PR #1825, opened 2017-04-06
Make timestamp toggable 

cheld:
In Dashboard Logs-View we currently show the timestamp, always. In kubectl the timestamp is not shown by default, but can be enabled (--timestamp=true)

The timestamp is required in the backend for pagination. I refactored the code to be able to strip off the timestamp. Seems to work. Maybe I should do some more refactoring for beautification, because there a several similar objects. I will check. 

The toggle button in the UI is also missing

Will fix: https://github.com/kubernetes/dashboard/issues/1546

### DIFF
--- a/src/app/backend/resource/logs/logs.go
+++ b/src/app/backend/resource/logs/logs.go
@@ -14,28 +14,10 @@
 
 package logs
 
-import "strings"
+import (
+	"strings"
+)
 
-// Logs is a representation of logs response structure.
-type Logs struct {
-	// Pod name.
-	PodId string `json:"podId"`
-
-	// Logs string lines.
-	LogLines `json:"logs"`
-
-	// The name of the container the logs are for.
-	Container string `json:"container"`
-
-	// Reference of the first log line in LogLines
-	FirstLogLineReference LogLineId `json:"firstLogLineReference"`
-
-	// Reference of the last log line in LogLines
-	LastLogLineReference LogLineId `json:"lastLogLineReference"`
-
-	// Structure holding information about current log view
-	LogViewInfo `json:"logViewInfo"`
-}
 
 // Number that is returned if requested line could not be found
 var LINE_INDEX_NOT_FOUND = -1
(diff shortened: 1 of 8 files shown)


cheld:
@floreks can you have a quick look?

src/app/backend/resource/logs/logs.go
```diff
@@ -122,11 +124,16 @@ var OldestLogLineId = LogLineId{
 // is equal to half of the log retention time. Therefore line in the middle of logs would serve as a good reference point.
 // LogLines allows to get ID of any line - this ID later allows to uniquely identify this line. Also it allows to get any
 // slice of logs relatively to certain reference line ID.
-type LogLines []string
+type LogLines []LogLine
+
+type LogLine struct {
+	Timestamp LogTimestamp
```
floreks: Please add json annotations.

cheld:
@floreks I added the button, but I had some problems to refresh the screen.

floreks:
> @floreks I added the button, but I had some problems to refresh the screen.

I have looked at the code and beside missing annotations it looks fine. To reload the logs you could add method to reload logs

```js
  reloadLogs() {
    this.loadLogs(this.podLogs);
  }
```

and just execute it in `onShowTimestamp` method after switching show timestamp variable.

Other two buttons only rely on css update and do not tamper with actual logs so they do not require reloading the view.

cheld:
hmm...of course I tried that. Ok I do it again. In my understanding the method name loadLogs is misleading as no access to the backend is done, right? It just parses and reformats the loaded data again - which is exactly what I want

floreks:
It should work since `loadView` method calls `loadLogs` after getting new portion of logs from backend to update the view. I have actually tried that and it worked fine.

cheld:
Ok works...my code contained a compile error :o) sorry. 

(BTW: the name reload or load are still missleading as these method are not loading any data - they just update the UI model. I will rename)

cheld:
PTAL

I did not refactor a lot, yet. I just moved code around to reduce vertical scope (sorry for review). There are still too many object types (structs), e.g. LogViewSelector and LogViewInfo is effectively the same type. I guess the autor lost track himself :) I will continue on logs and clean up in subsequent PRs. (BTW: Object type names should be singular. e.g. LogData is better than LogLine**s**)

PR #1837, opened 2017-04-10
Switch to Material Design Icons library

### LINKED ISSUE #1460: Use richer material design icons library
Currently we use the official material design icon library: https://material.io/icons/. It is, however, very limited in the number of icons there. We should evaluate other libraries, such as https://materialdesignicons.com, which have much more icons. 

cc @batikanu

maciaszczykm:
image

Old icon set: https://material.io/icons/
New icon set: https://materialdesignicons.com/

**If we will agree on this change I will refactor other icons across the Dashboard. Currently did it only in few places.**

Moreover, while I was doing it I refactored a bit status code on resource cards. We can use `ng-switch` there to make it easier to maintain.

Fixes #1460.


maciaszczykm:
Ping @kubernetes/dashboard-maintainers.

bryk:
This is complicated.

1. Did you read the license they have?
2. Do we comply with it?
3. What do we need to do to be compliant? 

@danielromlein WDYT? 

https://opensource.org/licenses/OFL-1.1

maciaszczykm:
@bryk I know. I am not an expert on this topic, so this is just a proposal. I read their licence and in my opinion we do comply with it. Probably more oriented guys have to check it :-)

maciaszczykm:
Plus we probably need to add their license or/and info about usage if it's going to be merged.

cheld:
@maciaszczykm In all Google products (as well as all Fujitsu products) the licenses of all third party libaries are linked to the about page. However, I have not linked any license information yet, because we have no process or tool to track

danielromlein:
I don't have a good understanding of the mechanics of icon usage. I think the expanded set makes sense to use from a UI perspective. Took a quick look at the license but I don't really know what to look for there...

For reference, #1460 was the original issue.

cheld:
FAQ 
http://theleagueof.github.io/licenses/ofl-faq.html

maciaszczykm:
@bryk @rf232 Do you know if there is somebody on core that is responsible for such stuff and can help us with licenses?

maciaszczykm:
Closing as stale.

PR #1847, opened 2017-04-13
Deprecated :modify core to CoreV1

zouyee:
gentle ping @maciaszczykm 
[Core](https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go#L47) deprecated,so we need modify `Core` to `CoreV1` in the file [`resource/event/common`](https://github.com/kubernetes/dashboard/blob/master/src/app/backend/resource/event/common.go#L147)


maciaszczykm:
@zouyee Great, thanks! Could you also check other files? For example: https://github.com/kubernetes/dashboard/blob/master/src/app/backend/resource/common/resourcechannels.go

zouyee:
@maciaszczykm  👌

zouyee:
@maciaszczykm  would you like to spear some time to review it？

PR #1852, opened 2017-04-17
modify code annotations and client version which have been omitted though an oversight

zouyee:
1. modify client version which has been omitted
2. modify code annotations(completely)
3. modify code style(vet check)
4. modify ` type ScaleTargetRef`
@maciaszczykm  sorry to trouble you!


src/app/backend/resource/logs/logs.go
```diff
@@ -37,7 +37,7 @@ type Logs struct {
 	LogViewInfo `json:"logViewInfo"`
 }
 
-// Number that is returned if requested line could not be found
+// LINE_INDEX_NOT_FOUND that is returned if requested line could not be found
```
cheld: Could you please remove the first 'that':

LINE_INDEX_NOT_FOUND is returned if requested line could not be found

thanks

zouyee: 👌

src/app/backend/resource/horizontalpodautoscaler/common.go
```diff
@@ -14,8 +14,7 @@
...
 
-// Simple mapping of an autoscaling.CrossVersionObjectReference
-type ScaleTargetRef struct {
-	Kind string `json:"kind"`
-	Name string `json:"name"`
-}
+import "k8s.io/client-go/pkg/apis/autoscaling/v1"
+
+// autoscaling.CrossVersionObjectReference
+type ScaleTargetRef v1.CrossVersionObjectReference
```
maciaszczykm: If we decide to use `v1.CrossVersionObjectReference` we can do it without `ScaleTargetRef` alias. That means we can delete whole `common.go` file.

zouyee:
@maciaszczykm  hi,I have reverted the commit about `common.go`,would like to spent some time  to take a look? thanks

src/app/backend/resource/horizontalpodautoscaler/common.go
maciaszczykm: Sure, it is. You can go ahead and delete file. We want to keep our structures only for types where we filter data or add new fields.

PR #1859, opened 2017-04-18
Add sorting support to remaining list resources

### LINKED ISSUE #1677: Feature request: Jobs should be sorted by date instead of name
#### Issue details
It appears that jobs are sorted by name.  I think sorting by date would make more sense for batch type work.  i.e.  Newest jobs show up at the top of the list.  

Ideally this issue gets addressed across all resources in a flexible way, see https://github.com/kubernetes/dashboard/issues/966

floreks:
## Additional changes
 - Add age column to all resources as by default we are sorting by age
 - Update backend documentation
 - `gofmt` backend
 - Clean up horizontal pod autoscalers frontend code to correctly support sorting
 - Disable sorting in case list resource is not provided instead of throwing an error

Last change is done due to i.e. deployment's new replica sets not having it's own endpoint for updates. As there is always only one new replica set and we are using list to display it I am not sure if having separate endpoint is good. We don't need sorting nor paging on such lists so we can disable it instead of throwing an error.

Closes #1677.


maciaszczykm:
Closes #1677.

PR #1868, opened 2017-04-20
confighandler: remove the redundant http.Error call

zhouhaibing089:
The removed line is actually duplicated with https://github.com/kubernetes/dashboard/blob/master/src/app/backend/handler/confighandler.go#L44.


zhouhaibing089:
@maciaszczykm @floreks Can one of you take a look at this?

floreks:
Good catch. Thanks!

PR #1875, opened 2017-04-21
Add filtering (search) support

floreks:
We can already start discussion on the UI side of this feature as I will be finishing everything.

## Changes
 - Refactored list footer
 - Moved pagination controls to the bottom of the list (https://github.com/kubernetes/dashboard/pull/1875/commits/f8e6a3def34f1aeea5e6907b8adec67a1b442560)
 - Adapted backend to filter based on partial match (https://github.com/kubernetes/dashboard/pull/1875/commits/c0639c04e823e75add9d350f4d8e281e87e754ef)
 - Added filtering option to ~pod~ all resource lists (https://github.com/kubernetes/dashboard/pull/1875/commits/0f8767be4f2e20e50bfcfeab33f401d4777a9f2d, https://github.com/kubernetes/dashboard/pull/1875/commits/91228ee97b52de3eb7676babaeb75016e4498c14)
 - Refactored list header to correctly support overriding (https://github.com/kubernetes/dashboard/pull/1875/commits/dccd834641c4e269af8dcd1ed6fc23b217beefbb)
 - Refactored zero state and reused component logic in all places (https://github.com/kubernetes/dashboard/pull/1875/commits/13395f5f50c209424f51a4aa09e8c320276e3721)

## TODO
 - [x] Fix existing tests (https://github.com/kubernetes/dashboard/pull/1875/commits/3288d9dba515d8892cf903d949b43754df906d68)
 - [x] Add tests

Should we also update graphs when filtering is applied?

### Old version
gifrecord_2017-04-21_203141

### Updated (https://github.com/kubernetes/dashboard/pull/1875/commits/d8f8958bb5734be7ea8d0ddc00802961c7b51f22)
dashboard


floreks:
Calling @kubernetes/dashboard-maintainers to start discussion.

src/app/backend/resource/dataselect/dataselect.go
```diff
@@ -122,7 +124,7 @@ func (self *DataSelector) Filter() *DataSelector {
 				matches = false
 				continue
 			}
-			if filterBy.Value.Compare(v) != 0 {
+			if !filterBy.Value.StartsWith(v) {
```
maciaszczykm: Shouldn't we use `contains` instead of `starts with`?

floreks: I don't think so. For example contains would return true in case of object `heapster` and search string such as `ster`. I suppose we do not want that.

cheld:
Cool

src/app/backend/resource/dataselect/dataselect.go
maciaszczykm: Still, I disagree, but maybe it is only my impression. Check this example:

https://docs.angularjs.org/api/ng/filter/filter

As you can see they check if value contains searched string. It seems easier to use IMO. If I would like to find a pod with `kubernetes-dashboard-n32iurn23` name I can type just `n32` and it will be found. Now I need to write almost whole name.

maciaszczykm: Let's see what others think.

@kubernetes/dashboard-maintainers

maciaszczykm:
Besides what I have mentioned looks really good. Perhaps @danielromlein can help us out to findout layout there, but it already looks good. Perhaps input style should be like everywhere else?

floreks:
Yes I have found that while developing but I was thinking about actual use cases. What is more common for people that want to search objects on the list?

Still, contains is a superset of startswith so if we need more detailed search then we can change that. My first thought while testing was grouping elements and easy removal of the ones we don't want to see.

Contains adds option to quickly search for single item with partial match which i guess is also good in some cases.

floreks:
As for the UI i followed the html5 style for input with type search: https://css-tricks.com/webkit-html5-search-inputs/

I have just styled it a bit to match our style better.

floreks:
@danielromlein @bryk WDYT about the UI part?

cheld:
FYI: Piotr is on holidays

Issue #1869, opened 2017-04-21
dashboard should support https


zhouhaibing089:
Per https://github.com/kubernetes/dashboard/blob/master/src/app/backend/dashboard.go#L89, it only serves a http server, a better practice would be using https instead.

To achieve that, we need to introduce three more flags:

*   `--secure-port`: the port that https will listen on.
*   `--tls-cert-file`: the certificate.
*   `--tls-key-file`: the private key.

Issue #1744, opened 2017-03-16
Why is NodePort exposed?


colemickens:
Why does the `kubernetes-dashboard` service exposing via NodePort by default? https://github.com/kubernetes/dashboard/blob/master/src/deploy/kubernetes-dashboard.yaml

floreks:
I'm not sure why it was added there. One reason might be that it makes dashboard easily accessible on local clusters such as minikube(they are even exposing it using NodePort 30000). Usually kubelet is bound to local ip anyway so there should not be any security issues there. 

However if there are any security issues related to NodePort beeing default way of exposing dashboard then we might consider changing that.

@maciaszczykm @rf232 @bryk WDYT?

bryk:
We could be ClusterIP by default and do NodePort in minikube and other environments that need this. I like @colemickens proposal.

cheld:
I did a bit of research:

* it was introduced a while ago: https://github.com/kubernetes/dashboard/commit/8d249ecaf9cdc3487aa7b008c26027003ae47d8d

* minikube is using nodeport, but it uses own deployment descriptors

* the descriptor is most likely used by kubeadm users

So, I don't see any reason to keep nodeport

PR #1877, opened 2017-04-22
https: tls configuration and run on https

zhouhaibing089:
Fixes https://github.com/kubernetes/dashboard/issues/1869.

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -29,9 +29,13 @@ import (
 )
 
 var (
-	argPort          = pflag.Int("port", 9090, "The port to listen to for incoming HTTP requests")
-	argBindAddress   = pflag.IP("bind-address", net.IPv4(0, 0, 0, 0), "The IP address on which to serve the --port (set to 0.0.0.0 for all interfaces).")
-	argApiserverHost = pflag.String("apiserver-host", "", "The address of the Kubernetes Apiserver "+
+	argInsecurePort        = pflag.Int("insecure-port", 9090, "The port to listen to for incoming HTTP requests.")
+	argPort                = pflag.Int("port", 8443, "The secure port to listen to for incoming HTTPS requests.")
+	argInsecureBindAddress = pflag.IP("insecure-bind-address", net.IPv4(127, 0, 0, 1), "The IP address on which to serve the --port (set to 0.0.0.0 for all interfaces).")
+	argBindAddress         = pflag.IP("bind-address", net.IPv4(0, 0, 0, 0), "The IP address on which to serve the --secure-port (set to 0.0.0.0 for all interfaces).")
+	argCertFile            = pflag.String("tls-cert-file", "", "File containing the default x509 Certificate for HTTPS.")
+	argKeyFile             = pflag.String("tls-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
+	argApiserverHost       = pflag.String("apiserver-host", "", "The address of the Kubernetes Apiserver "+
 		"to connect to in the format of protocol://address:port, e.g., "+
 		"http://localhost:8080. If not specified, the assumption is that the binary runs inside a "+
 		"Kubernetes cluster and local discovery is attempted.")
@@ -86,7 +90,15 @@ func main() {
 	// TODO(maciaszczykm): Move to /appConfig.json as it was discussed in #640.
 	http.Handle("/api/appConfig.json", handler.AppHandler(handler.ConfigHandler))
 	http.Handle("/metrics", prometheus.Handler())
-	log.Print(http.ListenAndServe(fmt.Sprintf("%s:%d", *argBindAddress, *argPort), nil))
+
+	// Listen for http and https
+	addr := fmt.Sprintf("%s:%d", *argInsecureBindAddress, *argInsecurePort)
+	go log.Fatal(http.ListenAndServe(addr, nil))
+	secureAddr := fmt.Sprintf("%s:%d", *argBindAddress, *argPort)
+	if len(*argCertFile) != 0 && len(*argKeyFile) != 0 {
+		go log.Fatal(http.ListenAndServeTLS(secureAddr, *argCertFile, *argKeyFile, nil))
+	}
+	select {}
 }
 
 /**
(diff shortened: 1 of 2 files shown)


cheld:
Most users access dashboard through API server (https://apiserver/ui). This has the benefit that user have to setup certificates and authentication only once. Isn't this working for you?

src/app/backend/dashboard.go
```diff
@@ -86,7 +89,15 @@ func main() {
 	// TODO(maciaszczykm): Move to /appConfig.json as it was discussed in #640.
 	http.Handle("/api/appConfig.json", handler.AppHandler(handler.ConfigHandler))
 	http.Handle("/metrics", prometheus.Handler())
-	log.Print(http.ListenAndServe(fmt.Sprintf("%s:%d", *argBindAddress, *argPort), nil))
+
+	// Listen for http and https
+	addr := fmt.Sprintf("%s:%d", *argBindAddress, *argPort)
+	go log.Fatal(http.ListenAndServe(addr, nil))
```
cheld: In case of secure setup, you probably want to limit the bind andress to localhost for the insecure port

zhouhaibing089:
> Most users access dashboard through API server (https://apiserver/ui).

I assume you are mentioning about service proxy, right?

Even with that, there is no way to stop user from accessing the http port directly, the nature of http plain text makes it vulnerable actually.

src/app/backend/dashboard.go
```diff
@@ -86,7 +89,15 @@ func main() {
...
+	secureAddr := fmt.Sprintf("%s:%d", *argBindAddress, *argSecurePort)
+	if len(*argCertFile) != 0 && len(*argKeyFile) != 0 {
+		go log.Fatal(http.ListenAndServeTLS(secureAddr, *argCertFile, *argKeyFile, nil))
```
cheld: The certificates should be reloaded in case they are updated.

cheld:
How could users access the http port from outside?

zhouhaibing089:
@cheld well, this is how we deploy kube-dashboard: pod => svc => dns.

As dashboard only opens http port, so will the service spec, currently kubernetes only supports TCP load balancer(thus a passthrough) which means user outside can access dashboard via the load balancer ip plus http port.

cheld:
Are you refering to this?

https://github.com/kubernetes/dashboard/issues/1744

I think NodePort is a bug. It should be ClusterIP. Not sure why it was introduced. But even with NodePort it is still behind firewall and not accessible from outside. Only type LoadBalancer is accessible from outside.

BTW: this is the file used for production deployments

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dashboard/dashboard-service.yaml

As you can see we do not expose dashboard

zhouhaibing089:
@cheld well, I see your point, in kubernetes, the dashboard is only a cluster-ip service, thus user outside can only access this service via proxy way, but could you please help me understand:

1.  does it make sense to export a `type: LoadBalancer` service for dashboard, if not, then why.
1.  does it make sense to secure the traffic even inside kubernetes cluster?

src/app/backend/dashboard.go
zhouhaibing089: @cheld I did not find a trivial way to implement it, it might be overkill to listen for a file change here.

cheld:
> does it make sense to export a type: LoadBalancer service for dashboard, if not, then why.

Rather no, because it would complicate setup. Some might want to, but they must also handle authentication with some kind of proxy in front of dashboard. So also in this case Dashboard is not the external endpoint

> does it make sense to secure the traffic even inside kubernetes cluster?
hmm...good question. In general, kubernetes tries to secure traffic inside the cluster as well, I guess

zhouhaibing089:
@cheld it does not save setup effort with a `cluster-ip` service, user can still discover the dashboard service, and then run a pod in the cluster, and access the `ClusterIP` directly, and yeah, guys may argue:

1.  dashboard is mostly for UI access, so mostly, it is useful when the client is browser. however, the dashboard itself has apis as well, I can still access the api and behave on behalf of another issue without any issue.
1.  user can setup `NetworkPolicy` to only allow a subset of user(source ips, label selected pods, etc) to access the dashboard service. and thus I think it does not save setup effort, there is no difference in nature when you set up a cluster-ip service or load-balancer one.

>  kubernetes tries to secure traffic inside the cluster as well, I guess.

Well, no, as the service itself only serves for http, the connection between user to apiserver is secured(as apiserver is running on https), but the communication between apiserver and dashboard is still in plain text.

cheld:
There are many different ways people setup a kubernetes cluster and many different ways people deploy Dashboard. We do not want to restrict in any way. Just wanted to understand the usecase.

> the communication between apiserver and dashboard is still in plain text.

yes, I know. In my understanding the Kubernetes Community tries to secure traffic inside the cluster  e.g. as far as I remember the API server access the container logs through a secure connection (the kubelet certificates have been self-signed, though)

zhouhaibing089:
@cheld anyway, thanks for your review.

src/app/backend/dashboard.go
```diff
@@ -86,7 +89,17 @@ func main() {
 	// TODO(maciaszczykm): Move to /appConfig.json as it was discussed in #640.
 	http.Handle("/api/appConfig.json", handler.AppHandler(handler.ConfigHandler))
 	http.Handle("/metrics", prometheus.Handler())
-	log.Print(http.ListenAndServe(fmt.Sprintf("%s:%d", *argBindAddress, *argPort), nil))
+
+	// Listen for http and https
+	addr := fmt.Sprintf("%s:%d", *argBindAddress, *argPort)
+	if argBindAddress.Equal(net.IPv4(127, 0, 0, 1)) {
```
cheld: hmm...my suggestion was basically:

if secure setup (argCertFile != null)
    http server binds to 127.0.0.0
    https server binds to given address (0.0.0.0 by default)

Otherwise (no secure setup)
    http server binds to given address (0.0.0.0 by default)

cheld: or make two variables --insecure-bind-address and --bind-address

similar to apiserver:
https://kubernetes.io/docs/admin/kube-apiserver/

zhouhaibing089: @cheld thanks for pointing it our, I've done changes like this:

1.  `--port` => `--insecure-port`, and the new `--port` will be used for secure port.
1.  `--bind-address` => `--insecure-bind-address`, and the new `--bind-address` will be used for secure address.
1.  update the `serve` task to use `--insecure-port` to keep compatibility.

cheld:
Sorry, for the delay. I thought someone else might want to review and merge the PR.
LGTM

PR #1879, opened 2017-04-23
Cleanup of logs page

cheld:
- Renamed a lot
- Structed the result object in 3 parts (ui-part, selection slice, and logs)
- Remove duplicated selection object

### DIFF
--- a/src/app/backend/resource/logs/logs.go
+++ b/src/app/backend/resource/logs/logs.go
@@ -58,13 +58,13 @@ var AllSelection = &Selection{
 	ReferencePoint: NewestLogLineId,
 }
 
-// Representation of one page of logs
-type LogPage struct {
+// Representation of log lines
+type LogDetails struct {
 
-	// Meta information for this page
-	Info LogPageInfo `json:"info"`
+	// Additional information of the logs e.g. container name, dates,...
+	Info LogInfo `json:"info"`
 
-	// Meta information about page position in the logs
+	// Reference point to keep track of the position of all the logs
 	Selection `json:"selection"`
 
 	// Actual log lines of this page
@@ -72,7 +72,7 @@ type LogPage struct {
 }
 
 // Meta information about the selected log lines
-type LogPageInfo struct {
+type LogInfo struct {
 
 	// Pod name.
 	PodName string `json:"podName"`
--- a/src/app/backend/resource/logs/logs.go
+++ b/src/app/backend/resource/logs/logs.go
@@ -18,7 +18,6 @@ import (
 	"strings"
 )
 
-
 // LINE_INDEX_NOT_FOUND is returned if requested line could not be found
 var LINE_INDEX_NOT_FOUND = -1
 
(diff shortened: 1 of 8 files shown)


src/app/externs/backendapi.js
```diff
@@ -918,12 +917,20 @@ backendApi.LogLine;
 
 /**
  * @typedef {{
- *   referenceLogLineId: !backendApi.LogLineReference,
- *   relativeFrom: number,
- *   relativeTo: number
+ *   referencePoint: !backendApi.LogLineReference,
```
floreks: Isn't type of this `LogPageInfo`?

src/app/backend/resource/logs/logs.go
```diff
@@ -46,60 +45,71 @@ var OldestLogLineId = LogLineId{
...
-type LogViewSelector struct {
-	// ReferenceLogLineId is the ID of a line which should serve as a reference point for this selector.
-	// You can set it to last or first line if needed. Setting to the first line will result in standard slicing.
-	ReferenceLogLineId LogLineId
-	// First index of the slice relatively to the reference line(this one will be included).
-	RelativeFrom int
-	// Last index of the slice relatively to the reference line (this one will not be included).
-	RelativeTo int
+// Representation of one page of logs
+type LogPage struct {
```
floreks: I'm struggling a bit with using `Page` as part of backend api structure. For me it suggests that it is in some correlation with web page that shows logs, however it is only part of our API and it is returned as JSON object.

WDYT about something like `LogChunk` `LogSlice` `LogPart`, something that suggests that it is not full log but also doesn't suggest that it has to be related to web view.

src/app/externs/backendapi.js
cheld: The current structure is:

```
LogPage
      LogPageInfo (podname, containername, timestamps )
      LogSelection (reference point to keep track of paging)
      LogLines  (actual logs)
```
LogViewInfo -> LogSelection is syntactically correct. Still confusing?

src/app/backend/resource/logs/logs.go
cheld: Ok, I am thinking about it

src/app/externs/backendapi.js
floreks: Ok, It makes sense now. LogSelection name works for me.

src/app/backend/resource/logs/logs.go
cheld: I was thinking about trying to make it consistent with other resources. 

Typically we name things *list or *details (podlist, poddetails). I was wondering if the following structure would make sense (or lets say indicates how it is to be used):

```
LogDetails {
     LogInfo (LogMeta?)
     LogSelection
     LogLines
} 
```

floreks: Sounds good to me.

cheld:
PTAL

cheld:
@floreks can I merge?

PR #1889, opened 2017-04-26
Update Go deps

maciaszczykm:


zhouhaibing089:
As far as I know, govendor does not have much adoption as other tools(like `glide`).

And also to be honest, govendor do have some killer feature(name remapping, subfolder vendoring) that other tools can not support.

Do we have any plan to evaluate other tools? `glide`, `dep`, etc.

maciaszczykm:
@zhouhaibing089 Yes, `govendor` seems to be not the best choice at the moment. There were no clear candidate when the choice was made, maybe we should evaluate it again. `dep` seems like a good choice as this is already official Golang organization repo.

Migration could take a lot time though...

zhouhaibing089:
`dep` is still at alpha phase which might not be a good choice at this stage. I think it is worth to give a try on `glide` though. 😄

maciaszczykm:
@zhouhaibing089 We can wait if it is worth it. Let's see how soon it will be released.

PR #1892, opened 2017-04-28
Increase items per page in list views

woopstar:
Using the dashboard for quite a while now and it has become very annoying that the list views are only showing 10 items per page. I think we should increase the default.

If someone is able to, it would be even more awesome if you could have a drop down selector for each list view and change between some default values like 10, 20, 30, 40, 50, 100 etc.

### DIFF
--- a/src/app/frontend/common/dataselect/builder.js
+++ b/src/app/frontend/common/dataselect/builder.js
@@ -23,7 +23,7 @@ export const SortableProperties = {
 };
 
 /** @const {number} **/
-export const ItemsPerPage = 10;
+export const ItemsPerPage = 30;
 
 /**
  * @final
--- a/src/app/frontend/common/dataselect/builder.js
+++ b/src/app/frontend/common/dataselect/builder.js
@@ -23,7 +23,7 @@ export const SortableProperties = {
 };
 
 /** @const {number} **/
-export const ItemsPerPage = 30;
+export const ItemsPerPage = 15;
 
 /**
  * @final


floreks:
We could probably increase it to 15 but IMO 30 is too much. Workloads page could get overloaded with so many objects. Currently we have 7 resources x 30 items per resource would give us 210 resources on single page.

@maciaszczykm WDYT?

maciaszczykm:
@floreks I agree, we can switch to 15, but 30 is too much.

maciaszczykm:
Also you need to sign the CLA.

floreks:
@woopstar could you address our comments?

woopstar:
Sorry. I will get the CLA signed and edit the pull request to be 15 items. I am just very busy at work, so will do it a bit later today.

maciaszczykm:
@woopstar Will you be able to push it forward?

bryk:
Yep, +1 :)

maciaszczykm:
I have created pull request with requested changes, because CLA is missing here and there is no progress since a month.

woopstar:
I have signed the CLA. What to do?

maciaszczykm:
Sometimes it takes time to accept it, let's wait few minutes. Also, you need to change 30 to 15.

woopstar:
Hmm. Now it says I have to sign the document, but when I go to the document from the e-mail, it says 404.

woopstar:
Got it signed now at HelloSign. Lets see when the bot picks it up

PR #1895, opened 2017-04-28
Introduce global search

### LINKED ISSUE #1829: Introduce global search

maciaszczykm:
Basic global search implementation. Does not include any kind of advanced features or cache yet.

- [x] add backend
- [x] update toolbar
- [x] fix empty query search
- [x] handle all resources and sort them alphabetically
- [x] adjust search breadcrumb label
- [x] make filtering work on search results page
- [x] move search logic to the frontend
- [x] remove `!important`
- [x] update tests
- [x] test on `serve:prod`

Closes #1829.

### DIFF
--- a/build/script.js
+++ b/build/script.js
@@ -162,6 +162,7 @@ function compileES6(translation) {
     path.join(conf.paths.externs, 'uirouter.js'),
     path.join(conf.paths.externs, 'dataselect.js'),
     path.join(conf.paths.externs, 'dirPagination.js'),
+    path.join(conf.paths.externs, 'searchapi.js'),
   ];
 
   let closureCompilerConfig = {
--- a/i18n/messages-en.xtb
+++ b/i18n/messages-en.xtb
@@ -39,6 +39,7 @@
   <translation id="7569244915796303702" key="MSG_BREADCRUMBS_RC_LABEL" desc="Label 'Replication Controllers' that appears as a breadcrumbs on the action bar.">Replication Controllers</translation>
   <translation id="837518421741773666" key="MSG_BREADCRUMBS_REPLICA_SETS_LABEL" desc="Label 'Replica Sets' that appears as a breadcrumbs on the action bar.">Replica Sets</translation>
   <translation id="5794147661362074597" key="MSG_BREADCRUMBS_ROLES_LABEL" desc="Label 'Roles' that appears as a breadcrumbs on the action bar.">Roles</translation>
+  <translation id="2851304790534826779" key="MSG_BREADCRUMBS_SEARCH_LABEL" desc="Label 'Search' that appears as a breadcrumbs on the action bar.">Search for</translation>
   <translation id="391192137756114804" key="MSG_BREADCRUMBS_SECRETS_LABEL" desc="Label 'Secrets' that appears as a breadcrumbs on the action bar.">Secrets</translation>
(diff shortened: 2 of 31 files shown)


maciaszczykm:
@danielromlein Is [this](https://projects.invisionapp.com/share/3C8BDK9ZD#/screens/182037314) mockup up to date?

danielromlein:
Not seeing your comment anymore for some reason @maciaszczykm, but here's my thinking: since ultimately the direction we're heading with the chrome is this [app bar + light top bar](https://invis.io/XZ7RITOQD#/183042254_Light_Top_Bar_1280_Tabs), even though we don't have IAM at the moment it may make sense to introduce the top bar to accommodate search. Otherwise search could collide with the breadcrumb in certain situations and start to make the app bar cluttered.
WDYT?

cheld:
Nice, looks similar to Inbox. (I assume we are not implementing the hero transition, though)

maciaszczykm:
Current state:

image

Missing some styling tweaks and zoom icon in search bar.

maciaszczykm:
Current state:

image

danielromlein:
@maciaszczykm is there an environment where I could check this out? I have a few recommendations I might recommend and would like to click around with a prototype to get you precise feedback.

maciaszczykm:
@danielromlein Not yet... But I will try to provide something for you.

Sorry for missing SIG UI again, we were busy with our current assignments and totally forgot to call in.

danielromlein:
@maciaszczykm no worries! Thanks for the update.

maciaszczykm:
It's ready for a review.

maciaszczykm:
> Also personally I like what inbox does when we empty search bar and click on the page. It returns to the previous state. Our search behavior is a bit different but we could think about something similar.

It is yet another feature and I would like to merge this first and then work on enhancements. Search is pretty complex topic and we will not be able to do everything at once. This pull request is already big and I don't want it to grow anymore.

Regarding the idea itself I do not agree with it. I do not want to cause confusion and redirect user in described scenario. In my opinion such redirect would make sense only if we would have live updates of search results just like in Inbox. Currently this search is more like GitHub one and it does not have such feature.

> How should it behave on empty search submit? Different pages deal with that differently. Github tries to do a search and shows page that "This search could not be executed". Inbox just stays on current state and does nothing.

Search input is marked as `required`, so it will not allow to submit empty search query. When user will try to do it he will be notified with specific message, that he needs to fill query.

I am looking forward to @danielromlein suggestions, now or after it will be merged. Just like we did with filtering implemented by you.

floreks:
> Regarding the idea itself I do not agree with it. I do not want to cause confusion and redirect user in described scenario. In my opinion such redirect would make sense only if we would have live updates of search results just like in Inbox. Currently this search is more like GitHub one and it does not have such feature.

I didn't mean to have it exactly the same because we do not have live search as u mentioned. Maybe something like `X` clear button that would get user back to previous page when used on search page. Arrow back is also an option. GitHub also is a bit different because it doesn't have to follow material spec guideliness like we should.

> Search input is marked as required, so it will not allow to submit empty search query. When user will try to do it he will be notified with specific message, that he needs to fill query.

For not it's ok. Maybe we'll think about something better later on.

We can merge it and update later after we receive feedback from @danielromlein.

Issue #974, opened 2016-06-29
Localization of text messages coming from api-server


taimir:
#### Issue details

Currently text messages returned by the kubernetes core (in particular `api-server`) are not localized. This is because the api-server is an external system component, and the dashboard localization framework does not apply to it.

A decision needs to be made about how to display these messages, and whether to treat them in any particular way.
##### Environment

```
Dashboard version: v1.1.0
Kubernetes version: v1.2.5
```
##### Steps to reproduce

Clone dashboard. Start with `gulp serve:prod`. Use `ja` as an example locale (different than `en`) for your browser.
##### Observed result

In general, error messages and status messages that come directly from the k8s API-server are still displayed in plain English. This also applies to short status strings, e.g. `Running` for a Pod.
##### Expected result

The expected behavior is a matter of further discussion.

taimir:
Off the top of my head:
* Complex error messages will be extremely hard to localize on the side of `dashboard`. To my knowledge, currently the k8s api-server does not support any type of localization on its own. Correct me if I'm wrong.

* It might be a good idea to not display long status / error / warning messages from the api-server directly, but to hide them behind some sort of generic "proxy" text messages. This would have the added benefit of better control of the text flow on dashboard.

* For short status text, such as the status of a pod that's returned by the api-server, localization on dashboard's side might be feasible. However, some sort of mappings will need to be defined. Currently it is not clear to me how many such status "enum values" exist in k8s (regarding all resources), so I cannot assess the actual overhead of this additional change.

taimir:
@cheld @bryk @floreks @maciaszczykm @digitalfishpond

bryk:
> It might be a good idea to not display long status / error / warning messages from the api-server directly, but to hide them behind some sort of generic "proxy" text messages. This would have the added benefit of better control of the text flow on dashboard.

Statuses, types, etc. can be easily translated by the "proxy". They are mostly enums, so we can just hardcode translations in the UI. I've seen projects that do this.

bryk:
Regarding free form text messages, like descritpion of an error event. We can either ignore that or provide our own messages for errors and events, rather than taking them directly from the api. It is more work, but certainly doable.

GavinSong123:
How to change 'ja' on dashboard? Is there a config file?

cheld:
It is defined by the browser locale. Change the browser to Japanese

GavinSong123:
@cheld ok, thx.

zouyee:
[...]

PR #1896, opened 2017-04-28
Fix deploy from file not respecting namespaces and introduce localizer for backend error messages

### LINKED ISSUE #1374: YAML upload should respect the selected namespace
#### Issue details

The upload of a YAML file does not respect the selected namespace in the dashboard.
##### Environment

```
Dashboard version: n/a
Kubernetes version: 1.4 (GKE hosted)
Operating system: debian
Node.js version: n/a
Go version: n/a
```
##### Steps to reproduce
- Select a namespace other than default.
- Build a simple YAML of any resource (example a secret), upload it through the YAML upload dialog.
##### Observed result

If no namespace is provided in the YAML the resource is created on the 'default namespace' even when "All namespaces" or "default" are not selected.
##### Expected result

It should react like the CLI commands. If in the dashboard: 
- a namespace is selected and no namespace is present in the YAML file, use the selected one.
- a namespace is selected but different than the one in the YAML file, show an error that the selected namespace differs from the one enforced in the file.
- no namespace is selected and the YAML enforces a namespace, create the resource in the proper namespace.
- no namespace or 'default' is selected and the YAML doesn't enforce a namespace, create the resource in the 'default' namespace.
##### Comments

At the moment the selected namespace is only shown when the menu is open, which I guess is there for smaller screens. Because of this, I think,  we should also have a reminder of the current namespace next to each screen name on the top as a secondary information.

floreks:
This change fixes #1374 and #974.

## Changes
 - Introduced error localization mechanism
 - Fixed #1374

## TODO
 - [x] Clean up the code

## Error localizer
I have introduced a mechanism to localize messages coming from backend. Basically it is based on 2 maps that have to be synchronized on backend and frontend side.

`backend[partial -> code] -> frontend[code -> msg]`

Backend map structure:
 - key: unique partial string used to differentate message coming from api server. Some errors have their own api and are exposed as const objects so we can also use them.
 - value: unique error code string that can be recognized by frontend and used for localization

Frontend map structure:
 - key: basically it is backend's map value
 - value: message localized with `goog.getMsg`.

Additionally I have created `LocalizerService` that will take error string and localize it or return original string if it is not recognized.

## Main fix description
Basically we have 5 cases to consider (i'm assuming that namespace is always correct and exists):

1. Namespace is set in both file and dashboard and
   1. they are equal - **OK**
   1. they are different - **NOK** error is thrown that there is namespace mismatch between dashboard and yaml file.
1. Namespace is set in file and all namespaces are set in dashboard - **OK**
1. Namespace is set in dashboard but not set in file - **OK**
1. No namespace in both file and dashboard (all namespaces) - **NOK** error is thrown that namespace is not selected in both file and in dashboard

### UI

Case 1.2
zrzut ekranu z 2017-04-28 15-37-53

Case 4
zrzut ekranu z 2017-04-28 15-37-31


floreks:
@maciaszczykm @rf232 PTAL

cheld:
Sorry, a bit late. Anyway, looking at the code, the logic is as follows:

1 check if localized error message exists => return localized message
2. otherwise return original engilsh message

right? 

The plan is to localize only very few messages in situations when user actions lead directly to errors, correct?

BTW: sometimes an alternative approach is to show a generic message and show the original message optional behind a "show details" button. But in this case not a very good fit.

A bit unfortunate that the server messages have no error code...

PR #1912, opened 2017-05-04
Replace the app label in deploy yaml to k8s-app

billy2180:
replace app label to k8s-app, to keep it uniform with other k8s components app label format,and let dashboard e2e test passed


maciaszczykm:
Could you point us to e2e tests so we can confirm it?

billy2180:
go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=Kubernetes\sDashboard"

floreks:
Ok I have checked existing test and you are right. We need to update our labels to make it pass.

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dashboard.go

PR #1914, opened 2017-05-05
add resourcekind componentstatus

zouyee:
add resource kind :componentstatus


maciaszczykm:
Why do we need that?

zouyee:
@maciaszczykm  using this api, we could inspect the cluster‘s situation，like that

maciaszczykm:
Yeah, but Dashboard is the UI and not Kubernetes proxy. We don't need it if we won't display it in the UI. To make it useful you need to implement frontend part.

zouyee:
I will fix it which will take a few times, because i  am weak in angularjs2

maciaszczykm:
Closing as stale. Please reopen when rest of needed logic will be implemented.

PR #1930, opened 2017-05-10
Fix head image

### LINKED ISSUE #1929: Dashboard head image is broken
Currently we have an issue with our head images. It appears that dashboard tries to start and hangs. Only available logs are:
```
2017-05-10T13:22:25.091480549Z Using HTTP port: 9090
2017-05-10T13:22:25.091563183Z Creating API server client for https://10.96.0.1:443
2017-05-10T13:22:25.110026516Z Successful initial request to the apiserver, version: v1.6.2
2017-05-10T13:22:25.110041548Z Creating in-cluster Heapster client
2017-05-10T13:22:25.110045875Z Using service account token for csrf signing
```

After that it hangs and restarts. Marking as P0 as this is a major blocker.

floreks:
Translations updated during the build.

Fixes #1929. Due to introduced recently https support for dashboard docker image stopped working. It was caused by default binding for insecure address to `127.0.0.1`. Inside docker container we have to bind to `0.0.0.0` in order to access dashboard.

### DIFF
--- a/i18n/messages-zh.xtb
+++ b/i18n/messages-zh.xtb
@@ -1,17 +1,29 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE translationbundle><translationbundle lang="zh">
+  <translation id="961242340990264399" key="MSG_ABOUT_ABOUT_0" desc="This is a label for button displayed on about page used to open documentaion in browser">Documentation</translation>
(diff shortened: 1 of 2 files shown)


floreks:
Small update. `port` was actually overriding secure port binding and it would cause port clash together with insecure port.

PR #1938, opened 2017-05-11
Feature 1657

kgignatyev-inspur:
Hi, 

I have implemented the feature with using cookies as store, please have a look

image


maciaszczykm:
@kgignatyev-inspur Hi! Thanks for spending your time on it. I have few remarks. 

General idea behind color adjustments was to make it more visible, i.e. change primary color of whole template. At the moment colored area is too small and it is not styled properly. Maybe @danielromlein can give us some tips about it (users would like to have possibility to differ dashboards by changing their colors). I would also like to have generic color picker instead of defined colors if this is possible and it should be localed elsewhere (add settings page?).

I have not checked implementation yet, so no comments anymore at this stage. You would also need to sign the CLA if we want to push this forward.

kgignatyev-inspur:
I have signed the CLA as employee I see this message:
image
Is there anything else I need to do or it is just a delay in a processing?

As for making it more prominently visible, since it defines 'style' as color, and background combination, it can be used to colorize top bar, by simply injecting service and using ng-style. But that might need some code refactoring depending on desired architecture, for example, keeping it small like this or try to lay foundation for general customer defined 'skins' for dashboard to include general color scheme and logos and per-namespace customization.  What do you think?

floreks:
> Is there anything else I need to do or it is just a delay in a processing?

Make sure that your commits are signed with correct email address `git config user.(e)mail`. The one that you used to sign the CLA.

> As for making it more prominently visible, since it defines 'style' as color, and background combination, it can be used to colorize top bar, by simply injecting service and using ng-style. But that might need some code refactoring depending on desired architecture, for example, keeping it small like this or try to lay foundation for general customer defined 'skins' for dashboard to include general color scheme and logos and per-namespace customization. What do you think?

To be honest we were thinking about something like user settings page where user could change primary color of the whole dashboard or assign colors to specific namespaces. Angular itself has some theme management services so I'd try to reuse that at first.

kgignatyev-inspur:
Email seems correct:
git config user.email
konstantin@inspur.com

maciaszczykm:
I did not read full conversation, but the issue with CLA is connected to the fact, that second commit author did not sign it:

image

image

You must fix it too to make check green.

kgignatyev-inspur:
Ahh, thanks, did not notice that some of commits were by me as individual, signed.

maciaszczykm:
Closing as stale, please reopen after signing CLA, we can conttinue discussions how to push it forward then.

PR #1939, opened 2017-05-12
Pod attach/exec proto

lenartj:
**Prototype** -- please comment extensively. This is building on code/ideas/comments from #1455 and #1345, so some of the problems/suggestions mentioned there are addressed.

Task list:
- [x] Builds prod
- [x] You can actually execute a shell and get a terminal in the containers
- [x] Support trying different shells (tries `bash` than `sh` for now)
- [x] Using hterm. So copy-paste, colors, mouse, etc.
- [x] Resizing
- [x] Moon buggy!!
- [x] Message on disconnect
- [x] SockJS transport (kubectl proxy does not support websockets (kubernetes/kubernetes#32026) which is a major use case. So if websockets do not work it falls back on different kinds of http streaming. It is jerkier and much less efficient than websockets but at least it works.)
- [ ] Ability to attach as in kubectl attach
- [x] Come up with a better name than "Attach" for the shell exec link
- [x] Make code look not horrible. For example, find better names for all the `*attach*` functions and structures. Dependencies for gorilla/websocket should be removed.
- [ ] Implement locking for session map (while the current code is _bad_, actually triggering the problem is unlikely)
- [ ] Add links leading to exec view (something like logs button)
- [ ] Serve `SockJSURL` from dashboard
- [ ] utf-8 _input_ (needs hterm needs configuration)
- [ ] Add some tests
- [ ] Some solution to respawn if the user exits the shell (?)
- [ ] Some padding on the left (?)

I have been running it like this:
`DOCKER_RUN_OPTS="-v $HOME/.kube/config:/kubeconfig.yaml:ro -e KUBE_DASHBOARD_KUBECONFIG=/kubeconfig.yaml" build/run-gulp-in-docker.sh serve
`

moon-buggy

### DIFF
--- a/build/serve.js
+++ b/build/serve.js
@@ -19,9 +19,8 @@ import browserSync from 'browser-sync';
 import browserSyncSpa from 'browser-sync-spa';
 import child from 'child_process';
 import gulp from 'gulp';
+import proxyMiddleware from 'http-proxy-middleware';
 import path from 'path';
-import proxyMiddleware from 'proxy-middleware';
-import url from 'url';
 import conf from './conf';
 
 /**
@@ -77,9 +76,11 @@ function browserSyncInit(baseDir, includeBowerComponents) {
   }));
 
   let apiRoute = '/api';
-  let proxyMiddlewareOptions =
-      url.parse(`http://localhost:${conf.backend.devServerPort}${apiRoute}`);
-  proxyMiddlewareOptions.route = apiRoute;
+  let proxyMiddlewareOptions = {
+    target: `http://localhost:${conf.backend.devServerPort}`,
+    // proxy websockets
+    ws: true,
+  };
 
   let config = {
     browser: [],       // Needed so that the browser does not auto-launch.
@@ -87,7 +88,7 @@ function browserSyncInit(baseDir, includeBowerComponents) {
     // TODO(bryk): Add proxy to the backend here.
     server: {
       baseDir: baseDir,
-      middleware: proxyMiddleware(proxyMiddlewareOptions),
+      middleware: proxyMiddleware(apiRoute, proxyMiddlewareOptions),
     },
     port: conf.frontend.serverPort,
     startPath: '/',
--- a/src/app/backend/handler/apihandler.go
+++ b/src/app/backend/handler/apihandler.go
@@ -60,6 +60,7 @@ import (
 	"golang.org/x/net/xsrftoken"
 	errorsK8s "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubernetes/pkg/client/unversioned/remotecommand"
 )
 
 const (
@@ -74,6 +75,12 @@ const (
 type APIHandler struct {
 	heapsterClient heapster.HeapsterClient
 	manager        client.ClientManager
+}
+
+// TerminalResponse is sent by handleExecShell. The Id is a random session id that binds the original REST request and the SockJS connection.
+// Any client in possession of this Id can hijack the terminal session.
+type TerminalResponse struct {
+	Id string `json:"id"`
 }
 
 // CreateHTTPAPIHandler creates a new HTTP handler that handles all requests to the API of the backend.
@@ -238,6 +245,10 @@ func CreateHTTPAPIHandler(heapsterClient heapster.HeapsterClient, manager client
 		apiV1Ws.GET("/pod/{namespace}/{pod}/event").
 			To(apiHandler.handleGetPodEvents).
 			Writes(common.EventList{}))
+	apiV1Ws.Route(
+		apiV1Ws.GET("/pod/{namespace}/{pod}/shell/{container}").
+			To(apiHandler.handleExecShell).
+			Writes(TerminalResponse{}))
 
 	apiV1Ws.Route(
 		apiV1Ws.GET("/deployment").
@@ -1146,6 +1157,35 @@ func (apiHandler *APIHandler) handleGetPodEvents(request *restful.Request, respo
 		return
 	}
 	response.WriteHeaderAndEntity(http.StatusOK, result)
+}
+
+// Handles execute shell API call
+func (apiHandler *APIHandler) handleExecShell(request *restful.Request, response *restful.Response) {
+	sessionId, err := genTerminalSessionId()
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	k8sClient, err := apiHandler.manager.Client(request)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	cfg, err := apiHandler.manager.Config(request)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
+
+	terminalSessions[sessionId] = TerminalSession{
+		id:       sessionId,
+		bound:    make(chan error),
+		sizeChan: make(chan remotecommand.TerminalSize),
+	}
+	go WaitForTerminal(k8sClient, cfg, request, sessionId)
+	response.WriteHeaderAndEntity(http.StatusOK, TerminalResponse{Id: sessionId})
 }
 
 func (apiHandler *APIHandler) handleGetDeployments(request *restful.Request, response *restful.Response) {
--- a/src/app/backend/handler/terminal.go
+++ b/src/app/backend/handler/terminal.go
@@ -1,0 +1,268 @@
+package handler
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+
+	restful "github.com/emicklei/go-restful"
+	"gopkg.in/igm/sockjs-go.v2/sockjs"
+	remotecommandconsts "k8s.io/apimachinery/pkg/util/remotecommand"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/client/unversioned/remotecommand"
+)
+
+// PtyHandler is what remotecommand expects from a pty
+type PtyHandler interface {
+	io.Reader
+	io.Writer
+	remotecommand.TerminalSizeQueue
+}
+
+// TerminalSession implements PtyHandler (using a SockJS connection)
+type TerminalSession struct {
+	id            string
+	bound         chan error
+	sockJSSession sockjs.Session
+	sizeChan      chan remotecommand.TerminalSize
+}
+
+// TerminalMessage is the messaging protocol between ShellController and TerminalSession.
+//
+// OP      DIRECTION  FIELD(S) USED  DESCRIPTION
+// ---------------------------------------------------------------------
+// bind    fe->be     SessionID      Id sent back from TerminalReponse
+// stdin   fe->be     Data           Keystrokes/paste buffer
+// resize  fe->be     Rows, Cols     New terminal size
+// stdout  be->fe     Data           Output from the process
+// toast   be->fe     Data           OOB message to be shown to the user
+type TerminalMessage struct {
+	Op, Data, SessionID string
+	Rows, Cols          uint16
+}
+
+// TerminalSize handles pty->process resize events
+// Called in a loop from remotecommand as long as the process is running
+func (t TerminalSession) Next() *remotecommand.TerminalSize {
+	select {
+	case size := <-t.sizeChan:
+		return &size
+	}
+	return nil
+}
+
+// Read handles pty->process messages (stdin, resize)
+// Called in a loop from remotecommand as long as the process is running
+func (t TerminalSession) Read(p []byte) (int, error) {
+	m, err := t.sockJSSession.Recv()
+	if err != nil {
+		return 0, err
+	}
+
+	var msg TerminalMessage
+	if err := json.Unmarshal([]byte(m), &msg); err != nil {
+		return 0, err
+	}
+
+	switch msg.Op {
+	case "stdin":
+		return copy(p, msg.Data), nil
+	case "resize":
+		t.sizeChan <- remotecommand.TerminalSize{msg.Cols, msg.Rows}
+		return 0, nil
+	default:
+		return 0, fmt.Errorf("unknown message type '%s'", msg.Op)
+	}
+}
+
+// Write handles process->pty stdout
+// Called from remotecommand whenever there is any output
+func (t TerminalSession) Write(p []byte) (int, error) {
+	msg, err := json.Marshal(TerminalMessage{
+		Op:   "stdout",
+		Data: string(p),
+	})
+	if err != nil {
+		return 0, err
+	}
+
+	if err = t.sockJSSession.Send(string(msg)); err != nil {
+		return 0, err
+	}
+	return len(p), nil
+}
+
+// Toast can be used to send the user any OOB messages
+// hterm puts these in the center of the terminal
+func (t TerminalSession) Toast(p string) error {
+	msg, err := json.Marshal(TerminalMessage{
+		Op:   "toast",
+		Data: p,
+	})
+	if err != nil {
+		return err
+	}
+
+	if err = t.sockJSSession.Send(string(msg)); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Close shuts down the SockJS connection and sends the status code and reason to the client
+// Can happen if the process exits or if there is an error starting up the process
+// For now the status code is unused and reason is shown to the user (unless "")
+func (t TerminalSession) Close(status uint32, reason string) {
+	t.sockJSSession.Close(status, reason)
+}
+
+// terminalSessions stores a map of all TerminalSession objects
+// FIXME: this structure needs locking
+var terminalSessions = make(map[string]TerminalSession)
+
+// handleTerminalSession is Called by net/http for any new /api/sockjs connections
+func handleTerminalSession(session sockjs.Session) {
+	var (
+		buf             string
+		err             error
+		msg             TerminalMessage
+		terminalSession TerminalSession
+		ok              bool
+	)
+
+	if buf, err = session.Recv(); err != nil {
+		log.Printf("handleTerminalSession: can't Recv: %v", err)
+		return
+	}
+
+	if err = json.Unmarshal([]byte(buf), &msg); err != nil {
+		log.Printf("handleTerminalSession: can't UnMarshal (%v): %s", err, buf)
+		return
+	}
+
+	if msg.Op != "bind" {
+		log.Printf("handleTerminalSession: expected 'bind' message, got: %s", buf)
+		return
+	}
+
+	if terminalSession, ok = terminalSessions[msg.SessionID]; !ok {
+		log.Printf("handleTerminalSession: can't find session '%s'", msg.SessionID)
+		return
+	}
+
+	terminalSession.sockJSSession = session
+	terminalSession.bound <- nil
+	terminalSessions[msg.SessionID] = terminalSession
+}
+
+// CreateAttachHandler is called from main for /api/sockjs
+func CreateAttachHandler(path string) http.Handler {
+	return sockjs.NewHandler(path, sockjs.DefaultOptions, handleTerminalSession)
+}
+
+// startProcess is called by handleAttach
+// Executed cmd in the container specified in request and connects it up with the ptyHandler (a session)
+func startProcess(k8sClient *kubernetes.Clientset, cfg *rest.Config, request *restful.Request, cmd []string, ptyHandler PtyHandler) error {
+	namespace := request.PathParameter("namespace")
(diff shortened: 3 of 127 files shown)


lenartj:
@IanLewis, @floreks, @maciaszczykm, @rf232, @bryk,  @jwforres, @beyondblog, @Dmitry1987, if you have a some time your comments would be much appreciated. I hope I haven't left out anyone who seem to have shown interest previously.. Thanks

cheld:
I just tested it with `gulp serve` and `gulp serve:prod`. Cool. The attach function is a bit hidden, but thats a detail... Had no chance yet to look at the code.

lenartj:
Update: I'm working on adding a [SockJS](https://github.com/sockjs) based transport to circumvent the lack of WebSocket support in the kubectl proxy. (SockJS is using WebSockets if possible and falling back  on a number of different alternatives otherwise. Some of those can pass trough kubectl proxy fine.)

It might be a good idea anyway as not all corp proxies fancy WebSockets.

maciaszczykm:
@lenartj Good idea, I am going to take a look on it after release too.

lenartj:
Update: the SockJS implementation works fine. I will try to push the code tomorrow after some cleanup.

floreks:
I have quickly looked at the code and overall it looks fine. Probably we'll have some comments about where to place the code. You can continue for now. Once it's ready for review just write PTAL.

lenartj:
@floreks, PTAL - also, I have updated the tasks in the first comment, please take a look.

For everyone else: If you want to give this code a try you can do:
`kubectl apply -f https://lenart.io/kubernetes-dashboard-head-attach-proto.yaml
`

If you are running `kubectl proxy` visit the address http://127.0.0.1:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard-head/proxy/

I have tested this using the following setups:
- Chrome on Linux, via kubectl proxy
- Chrome on Linux, trough an ingress (secured with mandatory client certificate)
- Chrome on Android (trough the ingress)

floreks:
Sorry for the delay. I will take a look at it tomorrow.

lenartj:
@floreks, would it help if I cut this PR into 2-3 separate PRs?

src/app/backend/handler/apihandler.go
```diff
@@ -1474,6 +1487,127 @@ func (apiHandler *APIHandler) handleGetPersistentVolumeClaimDetail(request *rest
...
+	io.Writer
+	remotecommand.TerminalSizeQueue
+}
+
+func (apiHandler *APIHandler) handleAttachHelper(request *restful.Request, cmd []string, ptyHandler PtyHandler) error {
+	namespace := request.PathParameter("namespace")
+	podName := request.PathParameter("pod")
+	containerName := request.PathParameter("container")
+
+	req := getApiClient(request).Core().RESTClient().Post().
```
floreks: As this is related to pod I'd move this logic to i.e. `attach.go` or `terminal.go` file under pod resource. As this is actually exec then maybe we can keep it under some generic name like terminal/console.

src/app/backend/handler/attach.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func (t AttachSession) Read(p []byte) (int, error) {
+        m, err := t.sockJSSession.Recv()
+        if err != nil {
+          return 0, err
+        }
+
+	var msg struct {
```
floreks: Refactor structs outside of the functions.

src/app/backend/handler/attach.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (t AttachSession) Close(status uint32, reason string) {
+  t.sockJSSession.Close(status, reason)
+}
+
+// FIXME: this structure needs locking
+var attachSessions = make(map[string]AttachSession)
+
+func handleAttachSession(session sockjs.Session) {
+  if buf, err := session.Recv(); err == nil {
```
floreks: I'd reverse the logic here to decrease complexity.

```
if ...; err != nil {
log(err)
return
}
```

Also in case of any error we can probably just retun immediately, then last part does not have to be in else block.

src/app/frontend/attach/controller.js
```diff
@@ -0,0 +1,163 @@
...
+    let containerName = this.stateParams_.container || '';
+
+
+    // https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm/doc/embed.md
+    hterm.defaultStorage = new lib.Storage.Memory();
+
+    // https://github.com/sockjs/sockjs-client
+
+    this.term = new hterm.Terminal();
+    this.term.onTerminalReady = function() {
```
floreks: I'd extract anonymous functions to class functions. It will be easier to document and understand the code.

By the way instead of `function() {}` you can use ES6 fat arrow syntax `() => {}`.

floreks:
@maciaszczykm could you also take a closer look at it? I may have missed something.

lenartj:
Very useful comments. I will get to it in the coming few days.

lenartj:
@floreks, PTAL. Read the commit message for info. Mostly just cosmetical changes but touches a lot of lines.

maciaszczykm:
I have just checked out how it works and I really like it.

One minor thing for quick fix related to stylesheet:

image

We should apply padding here, just like in logs view:

image

I will take a look on the code soon.

maciaszczykm:
We should also add links leading to exec view (something like logs button), but we can work on it later on:

image

build/serve.js
```diff
@@ -19,9 +19,8 @@ import browserSync from 'browser-sync';
 import browserSyncSpa from 'browser-sync-spa';
 import child from 'child_process';
 import gulp from 'gulp';
+import proxyMiddleware from 'http-proxy-middleware';
```
maciaszczykm: You should run `gulp format`.

src/app/backend/handler/apihandler.go
```diff
@@ -95,6 +96,10 @@ type CsrfToken struct {
 	Token string `json:"token"`
 }
 
+type TerminalResponse struct {
```
maciaszczykm: Please add comments on big-case types and functions.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// Called from apihandler.handleAttach as a goroutine
+// Waits for the SockJS connection to be opened by the client the session to be bound in handleTerminalSession
+func TerminalWaiter(request *restful.Request, sessionId string) {
```
maciaszczykm: Function name should be verb, not a noun.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+	"log"
+
+	restful "github.com/emicklei/go-restful"
+        "gopkg.in/igm/sockjs-go.v2/sockjs"
+	"k8s.io/kubernetes/pkg/client/unversioned/remotecommand"
+	remotecommandconsts "k8s.io/apimachinery/pkg/util/remotecommand"
+	"k8s.io/kubernetes/pkg/api"
+)
+
+// What remotecommand expects from a pty
```
maciaszczykm: Please start comments with type/method name. In this case `PtyHandler ...`.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+// Implements PtyHandler
+type TerminalSession struct {
+  id string
+  bound chan error
+  sockJSSession sockjs.Session
+  sizeChan chan remotecommand.TerminalSize
+}
+
+type TerminalMessage struct {
+  Op, Data, SessionID string
```
maciaszczykm: Can you explain it a little? What is `Op` etc.?

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+
+  terminalSession.sockJSSession = session
+  terminalSession.bound <- nil
+  terminalSessions[msg.SessionID] = terminalSession
+}
+
+// Called from main for /api/sockjs
+func CreateAttachHandler(path string) (http.Handler) {
+  options := sockjs.DefaultOptions
+  // options.Websocket = false
```
maciaszczykm: Do we need it?

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+	if _, err := rand.Read(bytes); err != nil {
+		return "", err
+	}
+	id := make([]byte, hex.EncodedLen(len(bytes)))
+	hex.Encode(id, bytes)
+	return string(id), nil
+}
+
+// Checks if the shell is an allowed one
+func isValidShell(validShells, shell []string) bool {
```
maciaszczykm: Why is shell an array? Before call you do `[]string{request.QueryParameter("shell")}` which is redundant.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+		}
+	}
+	return false
+}
+
+// Called from apihandler.handleAttach as a goroutine
+// Waits for the SockJS connection to be opened by the client the session to be bound in handleTerminalSession
+func TerminalWaiter(request *restful.Request, sessionId string) {
+	shell := []string{request.QueryParameter("shell")}
```
maciaszczykm: Use `shell := request.QueryParameter("shell")` instead.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+	select {
+	case <-terminalSessions[sessionId].bound:
+		close(terminalSessions[sessionId].bound)
+
+		var err error
+		validShells := []string{"bash", "sh"}
+
+		if shell[0] != "" {
+			if isValidShell(validShells, shell) {
+				err = execHelper(request, shell, terminalSessions[sessionId])
```
maciaszczykm: If you need to pass an array do it here instead of using array everywhere.

maciaszczykm: Also, use different name, because using `shell` everywhere is confusing.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,264 @@
...
+	shell := []string{request.QueryParameter("shell")}
+
+	select {
+	case <-terminalSessions[sessionId].bound:
+		close(terminalSessions[sessionId].bound)
+
+		var err error
+		validShells := []string{"bash", "sh"}
+
+		if shell[0] != "" {
```
maciaszczykm: Instead of doing this check check if `isValidShell(validShells, shell)`, else `try some shells until one succeeds or all fail`. Then, we don't need to throw `err = fmt.Errorf("Invalid shell")`.

src/app/backend/handler/terminal.go
lenartj: I am not sure if if you are asking about `options` or the whole function. I've removed the variable; it was just a leftover from experiments.

If you meant the function itself: I think it's a good idea to abstract this implementation detail out of the main package (dashboard.go). It is also how the other handlers are setup:

>         http.Handle("/", handler.MakeGzipHandler(handler.CreateLocaleHandler()))
>         http.Handle("/api/", apiHandler)
>         // TODO(maciaszczykm): Move to /appConfig.json as it was discussed in #640.
>         http.Handle("/api/appConfig.json", handler.AppHandler(handler.ConfigHandler))
>         http.Handle("/api/sockjs/", handler.CreateAttachHandler("/api/sockjs"))
>         http.Handle("/metrics", prometheus.Handler())

src/app/backend/handler/terminal.go
lenartj: I am thinking of commands like `tmux att`. Maybe over-engineered?

src/app/backend/handler/terminal.go
lenartj: Done

build/serve.js
lenartj: Done

src/app/backend/handler/apihandler.go
lenartj: Done

src/app/backend/handler/terminal.go
lenartj: Done

src/app/backend/handler/terminal.go
lenartj: Done

src/app/backend/handler/terminal.go
lenartj: Done

src/app/backend/handler/terminal.go
lenartj: Yes using `shell` in the loop as well is pretty confusing. I've changed the latter.. Is it any better?

lenartj:
I've experimented with a couple of different paddings:
- Around: padding on the right of the scrollbar looks really odd
image

- padding-left only, black: the actual background color of hterm is not quite black, but [#f0f0f0](https://www.google.ie/#q=%23101010) so black looks odd next to it
image

- padding-left only, same as hterm
image
... but now any process that changes the background color will have an almost black line next to it:
image

- padding-left only, color inherited ([whiteish](https://www.google.ie/#q=%23FAFAFA): this looks the least odd, but doesn't accomplish much in my opinion:
image

All in all, I am not a fan of any of the paddings above so I haven't added any for now.

lenartj:
Didn't mean to close it, sorry

lenartj:
The [error](https://travis-ci.org/kubernetes/dashboard/jobs/237212992) is unrelated to the code (`Response: {"error": "Too many active user tunnels: 5 >= 5"}.`)

lenartj:
(I now have a version that has attach functionality too. I will submit a separate PR after this one is merged)

src/app/backend/handler/terminal.go
maciaszczykm: Just about the single line :-) It's good.

maciaszczykm:
I will take another look tomorrow.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,267 @@
...
+		return 0, err
+	}
+
+	if err = t.sockJSSession.Send(string(msg)); err != nil {
+		return 0, err
+	}
+	return len(p), nil
+}
+
+// Toas can be used to send the user any OOB messages
```
maciaszczykm: Typo here.

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,267 @@
...
+
+// Close shuts down the SockJS connection and sends the status code and reason to the client
+// Can happen if the process exits or if there is an error starting up the process
+// For now the status code is unused and reason is shown to the user (unless "")
+func (t TerminalSession) Close(status uint32, reason string) {
+	t.sockJSSession.Close(status, reason)
+}
+
+// terminalSessions stores a map of all TerminalSession objects
+// FIXME: this structure needs locking
```
maciaszczykm: Do you plan to fix it now?

maciaszczykm:
> padding-left only, same as hterm

It looks good, but only with black background. I think we should apply it anyways, but we can do it in next pull request to avoid long discussion here.

@floreks @danielromlein Could you take a look on provided styling options and tell which one is the best? Do you have any other ideas?

vendor/vendor.json
```diff
@@ -925,6 +931,12 @@
 			"revisionTime": "2016-10-03T13:09:00Z"
 		},
 		{
+			"checksumSHA1": "1quR5pQNYFEgMLp987my6ZXh3vU=",
+			"path": "github.com/igm/sockjs-go/sockjs",
```
lenartj: Thanks for these comments!

- We are (`terminal.go`) only using the gopkg.in version. The github version creeps in due to an import in `gopkg.in/igm/sockjs-go.v2/sockjs/example_handler_test.go` but is unused otherwise.
- Can you (or perhaps, have you already) compiled dashboard with the github sockjs to see if it resolves the problem? If it does I will do some debugging to find out why :)
- I decided to depend on the gopkg.in version to play nice with the release model for sockjs. Unfortunately I haven't done extensive tests _yet_
- The `SockJSURL` problem I've spotted before, I want to serve that file from the dashboard, just haven't got around to do that yet. In my tests the transport that uses that file is not being used so it wasn't a priority. I have added this to the task list now.

src/app/backend/handler/terminal.go
lenartj: Not unless you this as a blocker. Otherwise let's create an issue (which I plan to do from the other items left on the task list), assign it to me, and I get to them one by one.

src/app/backend/handler/terminal.go
lenartj: Fixed.

lenartj:
I've added .5em padding-left (hterm-black), instead of 1em. This looks tolerable even with non-black background, like mc.

image

image

And again, I don't feel so strongly about it..

floreks:
I'd keep padding only if we can make it the same on left and right side (without moving scrollbar) otherwise imho it looks strange when background color is different than black.

lenartj:
@floreks, are you referring to the small "black" line on the right between the scrollbar and the console itself? I don't know how to increase that one, but we could make the left one even smaller, like 5px

floreks:
> @floreks, are you referring to the small "black" line on the right between the scrollbar and the console itself?

Yes. We can try to do that later. It's not that important for now.

> ... we could make the left one even smaller, like 5px

We want to follow material design spec so it should be 8 or some divisor of 8 (2, 4). For now I'd just remove it and work on that later.

lenartj:
I've removed the padding. It turns out that the width of the black vertical bar on the right depends on the exact the width. Something like: rightpaddingwidth = totalwidth%charwidth, so it can be 0 depending on the width.

cheld:
FYI: the old one, I almost forgot https://github.com/kubernetes-ui/container-terminal

src/app/backend/handler/terminal.go
```diff
@@ -0,0 +1,267 @@
...
+		return err
+	}
+	return nil
+}
+
+// Close shuts down the SockJS connection and sends the status code and reason to the client
+// Can happen if the process exits or if there is an error starting up the process
+// For now the status code is unused and reason is shown to the user (unless "")
+func (t TerminalSession) Close(status uint32, reason string) {
+	t.sockJSSession.Close(status, reason)
```
adrianchifor: `session.Close()` can return an error which is currently unhandled, crashing the dashboard. Might be good for this Close function to pass the error down to usages if it's not nil, like in line 261 and 265.

lenartj: Yes, this could/should be done, but the won't fix a crash. The crash happens when a Read() happens on the closed session. I will look into this, thanks

floreks:
I think you have missed @maciaszczykm [comment](https://github.com/kubernetes/dashboard/pull/1939/files/699bd1c976488bff2c9887e2a7792670a1c051a1..ac4249f057cef8e9b291bb74531b4950e2703add#r118936161).

Also you need to rebase code. Rest LGTM. Last changes and we are ready to merge.

lenartj:
@maciaszczykm, @floreks, I must be blind, I can't find the comment I've missed (I have followed the link too)

Do you mean this one?

> Use shell := request.QueryParameter("shell") instead.

But to this I have replied/explained and have received no further comment (or maybe just can't find it?)

floreks:
hmm... I don't see any reply to this comment. Link is valid and I can see a comment from @maciaszczykm (scroll a bit up if you can't see it).

maciaszczykm:
I may also missed one of the replies, but generally `shell := []string{request.QueryParameter("shell")}` doesn't look good especially when in `isValidShell` you need to do `shell[0]`. In my opinion `isValidShell` should accept single string, not array and extract last one. You should change that. I think @floreks will agree with me.

lenartj:
Loud and clear. I fix that.

floreks:
It's already fixed on master. Will be available in 1.6.2 release.

Duplicate of #2025.

moon03432:
@lenartj Thanks for your great feature! I deployed it in k8s 1.6.1 with 
kubectl apply -f https://lenart.io/kubernetes-dashboard-head-attach-proto.yaml

And when I want to exec in the container "kubernetes-dashboard-head" itself, an error occurred and the dashboard restarted. Is that expected?

exec-command-in-dashboard-head
kube-dashboard-error

floreks:
Not every container supports exec into it. It depends on what base image it was created. Dashboard does not support it. When you try to exec into it, it will crash.

_Sent from my LGE Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

cheld:
Dashboard container has no shell and as @floreks explain exec into it will not work. However, the running Dashboard instance should not crash and it isn't in my environment. Are you using roket? Anything else special in your environment?

lenartj:
Sorry I only noticed this comment now. I will be able to take a look on Wednesday, I am busy until that :/

moon03432:
I'm just using docker 1.12.6, nothing special

Kaijun:
Hmm, it seems that the dashboard pod will exit with error if you're executing a not existed command.

floreks:
This is a different issue related to session handling. We are working on a fix.

PR #1945, opened 2017-05-15
Fix ImagePullPolicy for versioned and :head releases.

bryk:
This is in response to https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-dev/EL8rglJrSUw/vdZiV7RmBQAJ


cheld:
I quickly checked history, it seems it was set to always from the beginning - most likely because image was not tagged properly in the beginning.

BTW: in general there is also a security aspect to it: a non-priviliged user could create a malicous image with the same name....using policy 'always' you can make sure that the original image is deployed.

bryk:
@cheld Interesting point about security. I think, however, that the intention of the imagePullPolicy is not to increase security. If someone can pull images on the machine, they can already break things in other ways.

PR #1948, opened 2017-05-15
Missing controller declaration.

kenan435:
Line is present on my pr branch for scaling but somehow missing on the actual master branch after merge.


cheld:
After merge, I think we should add a simple integration/click test. WDYT?

kenan435:
Agree. Will do it as part of the testing pr. that I need to submit.

Issue #1577, opened 2017-01-18
Proposal: experimentally embed Weave Scope


lukemarsden:
[Weave Scope](http://github.com/weaveworks/scope) provides orchestrator independent troubleshooting & debugging through visualisation. It has table and graph views. It includes Kubernetes integration to show pods, replica sets, deployments, services etc, including showing the live network connections between components in a graph, and also includes a terminal-in-the-browser to let you exec into a pod. Scope is 100% open source and this proposal is made in the spirit of a community effort.

Embedding Weave Scope into Dashboard addresses this topic from your [blog post](http://blog.kubernetes.io/2017/01/kubernetes-ux-survey-infographic.html).

> Troubleshooting & Debugging [...] is now our top area of investment. Currently in-progress is the ability to be able to exec into a Pod [...] One of a UI’s strengths over a CLI is its ability to show things, and the troubleshooting and debugging feature area is a prime application of this capability.
– http://blog.kubernetes.io/2017/01/kubernetes-ux-survey-infographic.html

In the SIG call today we discussed bringing these features to Dashboard via a Scope integration, and @bryk, @danielromlein and I provisionally agreed that an experiment to add Scope to Dashboard via a lightweight iframe integration would be a great way to gather user feedback. If this is successful (i.e. if we get good user feedback) we could later look at a deeper app-level integration.

Specifically, therefore I'm proposing that we, for this Phase I experiment:
1. Identify a location in the existing Dashboard UI for a "Scope (experimental)" or similar nav item.
2. When a user clicks that nav item, Dashboard determines whether the Scope probes and UI service are installed on the Kubernetes cluster. If they are not, it provides the user with a button to deploy these API objects into the kube-system namespace.
3. If they are, it simply includes the Scope UI within the "frame" of the Dashboard UI.

cc @bryk @danielromlein please correct me if I got any of this wrong :)

danielromlein:
@lukemarsden FYI it's @bryk you want to mention 😉 
And technically "Kubedash" is a deprecated UI; the official UI for K8s is "Dashboard" – I should have clarified that on our call.
Otherwise this looks great! Thanks for tackling this so promptly and I'm excited to see this in the product!
/ cc @rf232

bryk:
> Identify a location in the existing Dashboard UI for a "Scope (experimental)" or similar nav item.

@danielromlein I'm calling out you here. Do you have any idea of how/where we may place this?

IanLewis:
[...]

PR #1949, opened 2017-05-15
Feature scope

vlal:
* Add scope to k8s dashboard

As per #1577


src/app/backend/resource/scope/scope.go
```diff
@@ -0,0 +1,97 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Workloads structure contains all resource lists grouped into the workloads category.
+type Scope struct {
+	Deployed bool   `json:"deployed"`
+	Address  string `json:"address"`
+}
+
+// GetScope returns a populated Scope struct.
```
olekzabl: Copy-pasted comment?

src/app/frontend/scope/controller.js
```diff
@@ -0,0 +1,139 @@
...
+      });
+    });
+  }
+
+  /**
+   * Pings until scope is up and ready.
+   * @export
+   */
+  pingScope(tries) {
+    if (tries < 5) {
```
olekzabl: Calling pingScope(5) will do nothing, right? I initially was a little confused by this.
Can you refactor this argument (e.g. to decrease until 0), or add a comment about its meaning?

src/app/backend/resource/scope/scope.go
```diff
@@ -0,0 +1,99 @@
...
+type Scope struct {
+	Deployed bool   `json:"deployed"`
+	Address  string `json:"address"`
+}
+
+// PostScope deploys Scope onto the cluster.
+func PostScope(client client.Interface) (*Scope, error) {
+	log.Print("Installing scope on the cluster")
+
+	response, err := http.Get("https://cloud.weave.works/launch/k8s/weavescope.yaml?k8s-service-type=NodePort")
```
bryk: This is the part I'm worried about. It allows cloud.weave.works to hack user's cluster. Can we embed the YAML in repository here and versionize it? It'd be a bit safer at least.

errordeveloper: @bryk I believe that's been resolved now. Could you please review the rest of this PR and let @vlal and myself know if you broadly approve of it before we rebase and update? Thanks!

errordeveloper:
Let's get back to this!

/remove-lifecycle stale

cc @danielromlein @bryk

maciaszczykm:
@errordeveloper I think it is not the best time. We are migrating the whole app to new Angular and there is no sense in implementing it for AngularJS. Let's wait for #2727 and return to this discussion then.

errordeveloper:
Thanks, Marcin! We shall wait.

PR #1952, opened 2017-05-15
Removed NodePort from deployment descriptor

cheld:
Tested one file (kubernetes-dashboard.yaml)

fixes https://github.com/kubernetes/dashboard/issues/1744


cheld:
Codecov a bit confused?

maciaszczykm:
@cheld Yes, it happens since we have it configured. We have switched to integration and there is a red check, but it shouldn't block merging. Also, having this icon is great.

cheld:
@maciaszczykm well...I removed code and the coverage decreased in an area I did not touch...

maciaszczykm:
@cheld Yes, I noticed. When pull requests did not change anything then it is a common problem.

PR #1957, opened 2017-05-16
Scaling for Statefulset resource.

kenan435:
Implements scaling for statefulsets.


maciaszczykm:
I am going to check it now.

PR #1964, opened 2017-05-17
WIP: create testcluster with kubeadm

cheld:
I have cut out the existing build and just implemented enough to create a cluster with kubeadm

Status
It seems that kubeadm requires Ubuntu 16.04 while Travis supports only Ubuntu 14

I did not try to upgrade package manually, so far

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,22 +31,7 @@ cache:
     - bower_components
     - .tools
 
-addons:
-  # Run tests that require a browser on SauceLabs CI provider. Learn more at: https://saucelabs.com
-  sauce_connect: true
-  # Enables sauce labs for pull requests
-  jwt:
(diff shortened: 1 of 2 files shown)


cheld:
An alternative approach might be https://github.com/Mirantis/kubeadm-dind-cluster

Tested, works quite nice. Not sure if we should use it

cheld:
@floreks , @maciaszczykm WDYT

cheld:
Any opinion how we should continue?

floreks:
I think that for now we do not need to migrate from hyperkube. Dashboard uses cluster for few simple click tests only. Most of tests (frontend and backend) do not require cluster to be present. Once travis upgrades to ubuntu 16 we can think about kubeadm again.

cheld:
Ok, lets postpone until we must migrate...

PR #1977, opened 2017-05-19
Show node allocated resources in list view

### LINKED ISSUE #1259: A single page/view to see overview of nodes allocation
I think it would be useful to have a page with overview of all nodes and % of ram, cpu allocated and available per node.

Is there any plans for this?

maciaszczykm:
To spare space in table I will use format similar to `kubectl`:

zrzut ekranu 2017-05-19 o 12 35 03

On details page it will look like before:

zrzut ekranu 2017-05-19 o 12 36 12

Checklist:

- [x] backend refactoring
- [x] showing data in frontend
- [x] tests

Closes #1259.

### DIFF
--- a/src/app/backend/resource/node/list.go
+++ b/src/app/backend/resource/node/list.go
@@ -30,9 +30,7 @@ import (
 
 // NodeList contains a list of nodes in the cluster.
 type NodeList struct {
-	ListMeta common.ListMeta `json:"listMeta"`
-
-	// Unordered list of Nodes.
+	ListMeta          common.ListMeta `json:"listMeta"`
 	Nodes             []Node          `json:"nodes"`
 	CumulativeMetrics []metric.Metric `json:"cumulativeMetrics"`
 }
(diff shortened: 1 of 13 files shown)


src/app/backend/resource/node/list.go
```diff
@@ -100,11 +104,17 @@ func toNodeList(nodes []api.Node, dsQuery *dataselect.DataSelectQuery, heapsterC
 	return nodeList
 }
 
-func toNode(node api.Node) Node {
+func toNode(node api.Node, pods *api.PodList) Node {
+	allocatedResources, err := getNodeAllocatedResources(node, pods)
+	if err != nil {
+		log.Printf("Couldn't get allocated resources of %s node\n", node.Name)
```
floreks: Same as above.

Will frontend hide additional columns on the list if allocatedResources are empty? They might be if there is an error which we only print and continue.

src/app/backend/resource/node/list.go
```diff
@@ -73,21 +70,28 @@ func GetNodeList(client client.Interface, dsQuery *dataselect.DataSelectQuery, h
...
+	nodeCells, metricPromises, filteredTotal := dataselect.GenericDataSelectWithFilterAndMetrics(toCells(nodes),
+		dsQuery, dataselect.NoResourceCache, &heapsterClient)
 	nodes = fromCells(nodeCells)
 	nodeList.ListMeta = common.ListMeta{TotalItems: filteredTotal}
 
 	for _, node := range nodes {
-		nodeList.Nodes = append(nodeList.Nodes, toNode(node))
+		pods, err := getNodePods(client, node)
+		if err != nil {
+			log.Printf("Couldn't get pods of %s node\n", node.Name)
```
floreks: I'd also display error message here. It may help user find root cause of issue.

maciaszczykm: Good point.

src/app/backend/resource/node/list.go
maciaszczykm: It uses the same format as `kubectl describe node ...` and very similar logic behind it:

image

If cluster won't be able to return data and it will throw an error our backend will return `nil` to the frontend. Otherwise, if data is here, we should not alter it in any way I think. 

Still few checks in frontend part are missing.

maciaszczykm:
Current state:

image

src/app/backend/resource/node/list.go
maciaszczykm: Added checks.

src/app/backend/resource/node/list.go
maciaszczykm: Done.

Issue #1911, opened 2017-05-03
Don't add application version as default label on create new deploy


tobiasrdm:
#### Issue details
When creating new deploy, application version is added as default label. But when image version is updated, version label stay as the old version.

##### Environment
```
Minikube version: v0.18.0
Dashboard version: v1.6.0
Kubernetes version: v1.6.0
Operating system: Linux Mint 18.1 x64
Node.js version:
Go version: go1.7.1
```

##### Steps to reproduce

1. minikube start
2. minikube dashboard
3. Click in "+ CREATE" button
4. App name= nginx
5. Conteiner Image= nginx:1.7.9
6. Click in "SHOW ADVANCE OPTIONS"
Label version=1.7.9 was automatically added.
7. Click in "DEPLOY" button
8. Click in "Workloads > Deployment" and select action "View/Edit YAML"
9. Change `.spec.template.spec.containers[0].image` from `nginx:1.7.9` to `nginx:1.9.1` and click in "UPDATE"

##### Observed result
Image was updated to nginx:1.9.1 but label stay as version=1.7.9

##### Expected result
Don't add application version as default label on create new deploy

PR #1985, opened 2017-05-23
Don't add application version as default label on create new deploy

### LINKED ISSUE #1991: How to create a different user than admin to access dashboard?
We have a kubernetes cluster up and running. we have added dashboard as file. Currently we are using 'admin' user for authentication. How can I setup a new user for authentication ?

kenan435:
Fixes https://github.com/kubernetes/dashboard/issues/1911

### DIFF
--- a/src/app/frontend/deploy/deployfromsettings_controller.js
+++ b/src/app/frontend/deploy/deployfromsettings_controller.js
@@ -13,9 +13,6 @@
 // limitations under the License.
 
 import {stateName as workloads} from 'workloads/state';
-
-import DockerImageReference from '../common/docker/dockerimagereference';
-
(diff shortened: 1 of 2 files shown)


kenan435:
I just took away functionality in this pr to fix the bug that at one point was a feature. This is what I got from the op's comment: "Don't add application version as default label on create new deploy". Do you guys see it the same way?

tobiasrdm:
@kenan435, I agree with the solution adopted.
In the future also someone should update the [documentation page](https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/) by removing the reference to "version" in the text: "Labels: Default labels to be used for your application are application name and version."

PR #2015, opened 2017-06-06
Increase items per page in list views to 15

maciaszczykm:
It is only temporary improvement. When we will implement storage, we will add setting page where users will be able to pick number of items per page themselves.

Closes #1892.


maciaszczykm:
Closing as @woopstar already answered.

Issue #1310, opened 2016-10-06
Provide support for other metrics sources than Heapster


superdump:
#### Issue details

The new dashboard in 1.4.0 was announced to support metrics graphs that provide a quick look into the recent state of the various assets running on a cluster. However, Heapster is not everyone's choice of monitoring system and indeed Prometheus is an excellent match for Kubernetes. It would be very useful if Prometheus and others were supported as sources for the dashboard metrics.
##### Environment

Dashboard was deployed using the YAML description as noted in the README.md.

```
Dashboard version: 1.4.0
Kubernetes version: 1.4.0
Operating system: Ubuntu 16.04.1 LTS but not relevant.
```
##### Steps to reproduce
1. Roll out a Kubernetes cluster
2. Install the dashboard
3. Observe their are no CPU/RAM graphs as in the announcement
##### Observed result

Only Heapster is supported as a source of metrics for the dashboard graphs.
##### Expected result

An easy way to configure the data source of metrics in the dashboard. Data sources should likely include the most used solutions (Prometheus, InfluxDB, Graphite, ...) and be pluggable so people can implement and easily extend that dashboard with their own data sources.
##### Comments

Not sure what the prioritisation should be for support. Consider this issue a request for Prometheus support.

bryk:
Cool, thanks for this feature request! Looks like something we'd love to have.

Is this something you can contribute to?

cheld:
As far as I understand we have three use cases:

1. One important use case is to provide historic data to Dashboard.

Heapster old-timer is intended to provide an interface for historic data, by reading data from DBs such as Hawkular, InfluxDB, GCM, and OpenTSDB

https://github.com/kubernetes/heapster/blob/master/docs/proposals/old-timer.md

Does this cover part of your request?

2. Another important use case is custom application metrics e.g. active user sessions

There have been suggestions that pods could expose custom metrics in prometheus syntax. Not sure if this is still up to date:

https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/custom-metrics.md

I believe currently only cAdvisor can be used incorporating custom metrics, e.g. to be used by HPA.

3. The third use case is to replace Heapster altogether and make Kubernetes more pluggable.

cheld:
BTW: in my understanding the only way to use InfluxDB or Graphite is to install Heapster with a corresponding sink. How would you suggest to install without Heapster? Also, could you elaborate on your prometheus setup?

puja108:
[...]

PR #2017, opened 2017-06-06
Introduce integration framework backend

floreks:
There are still lots of things to be done. I have decided to take care of config storage and sync in next PRs as this will be already a big change. Any comments are appreciated.

## Major changes
 - Added `IntegrationManager` that will be responsible for providing state of integration app (installed, connected, enabled, etc.) and listing all integrations. 
- Added `MetricManager` as a dependency of `IntegrationManager`. It represents group of metric integrations like heapster, prometheus, etc. It will be responsible for controlling status of all metric integrations and providing active metric client.
- Added `MetricClient` interface as a layer of abstraction between resources and metric api. All metric integrations will have to implement that interface.
- Removed all hard dependencies on heapster from our sources and replaced them with `MetricClient`.

Rules for introducing new integration would be:
1. Implement `Integration` interface to make sure that every integration app has unique ID and can do a health check to make sure if we can enable it.
1. Introduce new group-like interface that will be responsible for handling new integration. I'm thinking about adding interface similar to `Integration` that will make sure that every 'group' can handle integrations, i.e. install/uninstall or enable/disable them.

Later on they should support storage thus implement necessary methods that will make sure that state of integrations is persisted and up to date.

### TODO
 - [x] Agree on the integration framework API
 - [x] Clean up code
 - [x] Finish heapster migration (~graphs are working, sparklines not~)
 - [x] Add documentation
 - [x] Add tests


src/app/backend/integration/api/types.go
```diff
@@ -0,0 +1,23 @@
+package api
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// IntegrationID is a unique identification string that every integrated app has to provide.
```
bryk: Can you provide in-comment example of what an id would be? And where I can find other IDs to avoid conflicts.

src/app/backend/integration/api/types.go
```diff
@@ -0,0 +1,23 @@
...
+	HealthCheck() error
+	ID() IntegrationID
+}
+
+// IntegrationState represents integration application state. Provides information about
+// health (if dashboard can connect to it) of the integrated application.
+// TODO(floreks): Support more information like 'installed' and 'enabled'.
+type IntegrationState struct {
+	Connected   bool    `json:"connected"`
+	LastChecked v1.Time `json:"lastChecked"`
```
bryk: From this I see that our backend code will have state. Can you comment on scalability of this? I.e., how would this work if I have two or twenty dashboard UI replicas?

src/app/backend/integration/handler.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/emicklei/go-restful"
+	"github.com/kubernetes/dashboard/src/app/backend/integration/api"
+)
+
+// IntegrationHandler manages all endpoints related to integrated applications, such as state.
+type IntegrationHandler struct {
+	manager IntegrationManager
+}
+
+// Install creates new endpoints for integrations.
```
bryk: Hmm... I'm not sure this comment explains what the installation means. And this is the interesting part :) Can you comment? What happens after installation?

src/app/backend/integration/api/types.go
```diff
@@ -0,0 +1,23 @@
+package api
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// IntegrationID is a unique identification string that every integrated app has to provide.
+type IntegrationID string
```
bryk: What's the format of integration ID? Can it be any string?

src/app/backend/integration/manager.go
```diff
@@ -0,0 +1,70 @@
...
+type integrationManager struct {
+	metric metric.MetricManager
+}
+
+// List implements integration manager interface. See IntegrationManager for more information.
+func (self *integrationManager) List() []api.Integration {
+	result := make([]api.Integration, 0)
+
+	// Append all types of integrations
+	result = append(result, self.Metric().List()...)
```
bryk: This could be a static list somewhere in a dedicated file. It is hard to find it here. Among business logic code.

src/app/backend/integration/api/types.go
floreks: We could keep all of them as const values in this file. WDYT?

I don't think we have to worry about name collisions here, so simple integration name should be sufficient for the foreseeable future.

src/app/backend/integration/api/types.go
floreks: See previous comment.

src/app/backend/integration/api/types.go
floreks: Until storage sync will be implemented we will not have consistent behavior here with multiple pods since we don't know to which instance kubernetes will proxy the request.

I will implement it next so all data will be kept in a configmap (in the beginning) and backend will be able to sync them across multiple replicas. Backend should be stateless but I didn't want to make this PR bigger by adding storage manager and synchronization.

src/app/backend/integration/handler.go
floreks: I'll describe it better. Right now it's very simple and just exposes integration state but other integrations might want to install their own endpoints also.

src/app/backend/integration/manager.go
floreks: I was thinking about that at some point but this list would be empty in the beginning and would have to be filled by some function. In example to create MetricClient integration for heapster we need host and `ClientManager` args so we can't easily initialize this list. 

Additionally slices in go are not thread-safe so keeping static array and operating on it from multiple places might cause some troubles.

cheld:
We could think about combining our config-map to keep settings with the component config proposal:

https://docs.google.com/document/d/1arP4T9Qkp2SovlJZ_y790sBeiWXDO6SG10pZ_UUU-Lc/edit#

floreks:
Unfortunately I don't have access to see it.

cheld:
hmm...it is public

this is issue tracker: https://github.com/kubernetes/features/issues/115

src/app/backend/integration/api/types.go
cheld: Storage sync is kind of cache to config map?

What happens if addon is removed with kubectl and the variable connected is still set to true? Does the user get a lot of error message until synchronisation? Do we listen on apiserver or poll?

src/app/backend/integration/handler.go
cheld: An addon most likely needs to know the namespace selection. Would this be somthing to add?

src/app/backend/integration/api/types.go
cheld: In case of Weave Scope we will render the menu option depending on the installation of the backend. We cannot query the apiserver everytime we render the menu. In this case the caching would be rather required in the frontend.

src/app/backend/dashboard.go
```diff
@@ -76,14 +77,17 @@ func main() {
 
 	log.Printf("Successful initial request to the apiserver, version: %s", versionInfo.String())
 
-	heapsterRESTClient, err := heapster.CreateHeapsterRESTClient(*argHeapsterHost,
-		apiserverClient)
+	// Init integrations
+	integrationManager := integration.NewIntegrationManager(clientManager)
+	err = integrationManager.Metric().
+		ConfigureHeapster(*argHeapsterHost).
+		Enable(integrationapi.HeapsterIntegrationID)
```
cheld: In the long run, could we move argHeapsterHost into the config map?

cheld: So, what is the intended workflow if a user wants to use Prometheus instead of Heapster. Or maybe has already Prometheus installed and then deploys dashboard?

floreks: Any integration still requires some code to be pushed upstream. We'd need to implement prometheus client that implements `MetricClient` interface. 

Normally with settings page user will be able to refresh state of integrations, i.e. if prometheus is not installed in the cluster then user goes to settings page clicks install near prometheus integration and then enables it. That would automatically disable other metric clients.

cheld: hmm...I wonder if this is the right approach. I have not checked the api or anything lately, but my understanding that we do not have to care about which monitoring system is installed. We just use the api facade from apiserver. Also, installing a monitoring system might not be as easy as installing e.g. tiller. 

The old issue: https://github.com/kubernetes/dashboard/issues/1310

However, we should care about the integration of Graphana.

src/app/backend/integration/api/types.go
floreks: I was thinking about using watch for dashboard's config map. This way it would be always up to date. Some kind of subscribe/notify mechanism will be needed together with watch to make sure that state is always up to date.

cheld:
kubeadm also plans an addon manager:

https://docs.google.com/document/d/1Laov9RCOPIexxTMACG6Ffkko9sFMrrZ2ClWEecjYYVg/edit#

we have to think were we want to draw the line so we do not have redundancy. In gerneral we should focus on frontend integration.

floreks:
Integrations have the benefit of providing better UX experience for users but might require more work and code to be pushed upstream.

On the other hand addons could be created by everyone and be maintained independent of dashboard. Also they seem to be more flexible.

Both options have some advantages and disadvantages. IMO integrations are good for popular tools used by many people, like helm. Plugins could maybe be able to customize and add simple views.

src/app/backend/integration/manager.go
bryk: Yeah. I meant that we could have integrations-list.go file with this array. Filled the same way you do here, but in a separate file so that the place I register integrations is only one.

src/app/backend/integration/api/types.go
bryk: Having all IDs here LGTM.

src/app/backend/integration/manager.go
floreks: Ahh... this makes sense :) I will do that.

floreks: Done

cheld:
I was wondering which extention should be handled by addon-manager and which extention by Dashboard? 
e.g.:
Overlay-network => ok, that is add-on manager. 
Weave scope =>Dashboard. However, if I remember correctly I saw it on a feature list of addon manager

So, should we draw some line somewere or accept duplication? IMHO we should check addon manager concepts and also discuss with other SIGs

cheld:
Kops add-on:
https://github.com/kubernetes/kops/tree/master/addons

cheld:
Kops addon manager:
https://github.com/kubernetes/kops/tree/master/channels

floreks:
I think we can merge this part. It does not change current behavior nor add state. It is mostly refactor and also few small bug fixes for getting heapster metrics. Additionally if heapster is not found or is unhealthy on the backend start then it will be disabled so there won't be any unnecessary requests made.

Issue #910, opened 2016-06-16
The old translation sentences remain


Yamagami-Takuma:
#### Issue details

The old translation sentences remain in the i18n xtb file.

e.g. these three lines is same place translation.
https://github.com/kubernetes/dashboard/blob/f7f5649c54834fc6f7c29326d2de4b4cf1680959/i18n/messages-en.xtb#L143
https://github.com/kubernetes/dashboard/blob/f7f5649c54834fc6f7c29326d2de4b4cf1680959/i18n/messages-en.xtb#L424
https://github.com/kubernetes/dashboard/blob/f7f5649c54834fc6f7c29326d2de4b4cf1680959/i18n/messages-en.xtb#L434

taimir:
This is the expected behavior. The English text for this particular message has changed, this is why in the Japanese bundle you will get a new entry for the very same message with the new version of the text. It is the translator's job to then translate the new message and remove the old translation.

I think the old translation is kept intentionally for reference purposes, so that the translator can compare what the change was and not lose the context of the message from before.

Note that old translations are not directly harmful as they are disregarded during build. It is just that they unnecessarily clutter the messages list.

bryk:
That's true. Should we remove the stale messages?

taimir:
Yes, I think that would be O.K.

PR #2020, opened 2017-06-08
Update npm and bower dependencies

maciaszczykm:
It was unexpected, but this update solves #910 :tada:


maciaszczykm:
#910 was not solved, translations were regenerated due to dep update.

PR #2026, opened 2017-06-13
Add COC

jamiehannaford:


code_of_conduct.md
```diff
@@ -0,0 +1,5 @@
+## Kubernetes Community Code of Conduct
```
floreks: Shouldn't it be `Kubernetes Dashboard...`?

code_of_conduct.md
```diff
@@ -0,0 +1,5 @@
...
+
+Kubernetes follows the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md).
```
floreks: Same here.

code_of_conduct.md
jamiehannaford: Since it's a project-wide CoC I thought `Kubernetes` would be enough, but I can personalise if you think it's better :)

floreks: I guess this doesn't make much difference. I see that other projects in k8s org also use the same sentence. We can keep it this way then.

PR #2027, opened 2017-06-14
Add license header check/update tasks and update existing headers

floreks:
Added new gulp tasks to check and update license header in our source files. I have also updated existing headers. Aditionally `gulp check` will now check if every file has correct license matching the one in `license/header*.txt` file.


cheld:
Copyright Kubernetes Authors

You find templates in core

floreks:
I see. I didn't know it has changed. I will update it.

_Sent from my LGE Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #2033, opened 2017-06-16
Added taints field for node listing/details

### LINKED ISSUE #2024: Feature request: list taints on nodes
##### Environment
I run k8s using kops, and I use a modified dashboard template (the default from kops except with the 1.6.1 Docker image).

```
Dashboard version:
Kubernetes version: 1.6.2
Operating system: Debian 8
```

##### Steps to reproduce
Go to Nodes, click on a node.

##### Observed result
Taints are not listed and can only be seen in the YAML.

##### Expected result
Taints are listed.

lenartj:
In Node list:
image

In Node details:
image

Closes #2024.

### DIFF
--- a/src/app/backend/resource/node/detail.go
+++ b/src/app/backend/resource/node/detail.go
@@ -117,6 +117,9 @@ type NodeDetail struct {
 
 	// Metrics collected for this resource
 	Metrics []metric.Metric `json:"metrics"`
+
+	// Taints
+	Taints []v1.Taint `json:"taints,omitempty"`
 }
 
 // GetNodeDetail gets node details.
@@ -271,5 +274,6 @@ func toNodeDetail(node v1.Node, pods *pod.PodList, eventList *common.EventList,
 		EventList:          *eventList,
 		AllocatedResources: allocatedResources,
 		Metrics:            metrics,
+		Taints:             node.Spec.Taints,
 	}
 }
--- a/src/app/backend/resource/node/list.go
+++ b/src/app/backend/resource/node/list.go
@@ -43,6 +43,7 @@ type Node struct {
 	TypeMeta           api.TypeMeta           `json:"typeMeta"`
 	Ready              v1.ConditionStatus     `json:"ready"`
 	AllocatedResources NodeAllocatedResources `json:"allocatedResources"`
+	Taints             []v1.Taint             `json:"taints"`
 }
 
 // GetNodeListFromChannels returns a list of all Nodes in the cluster.
@@ -118,6 +119,7 @@ func toNode(node v1.Node, pods *v1.PodList) Node {
 		TypeMeta:           api.NewTypeMeta(api.ResourceKindNode),
 		Ready:              getNodeConditionStatus(node, v1.NodeReady),
 		AllocatedResources: allocatedResources,
+		Taints:             node.Spec.Taints,
 	}
 }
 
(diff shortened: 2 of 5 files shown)


lenartj:
Thanks for the review. I am actually not a big fan of this format (`key: value: effect`) either. And yes, `key=value:effect` would be consistent with kubectl.

The reason I haven't done that is because the labels and annotations key-value pairs are separated with a colon too and I tried to be consistent with that. Do you think it might be a good idea to change those to use `=` as well? I've found myself copy-pasting and then changing the `: ` to `=` before so maybe I am not the only one finding it awkward.

floreks:
>Do you think it might be a good idea to change those to use = as well?

I'm not sure. It is how kubectl displays it so if we would want to be consistent with it then we should change `:` to `=`.  I guess if someone uses kubectl to update something then copying already correct format will be better. IMHO you can change that.

@cheld WDYT?

Also adding @danielromlein as this is UX related.

cheld:
IMHO the taints should be removed from the overview, because 
1. the column width is so small that they are useless anyway 
2. I think taints are too technical and too hard to read to help with a drill down. I think labels are sufficient


Taints are cryptic to read. For this reason I would tend to keep the format.

BTW: in my default cluster setup a taint exists without a value

`node-role.kubernetes.io/master=:NoSchedule`

It should not have the equals, I think

floreks:
I guess taints are also used for grouping nodes, i.e. they can be used to group nodes by different architecture to prevent deployment of `arm` app on `amd64` node. They could be also applied dynamically to nodes that are during maintenance to remove all deployed applications and redeploy them on other nodes.

Since we plan to add filtering by more than just name then probably taints on list page also make sense. Labels are rather describing role of the node, i.e. node that is located in Europe where taints may enforce that only specific applications with correct annotation can be deployed on this node.

`node-role.kubernetes.io/master=:NoSchedule`

I think that it still makes sense to show `=:` because then it is clear that this taint's key does not contain any value and effect is set to `NoSchedule`. If we remove `=` then it's not clear if `NoSchedule` is value or effect.

We could also add some information icon with tooltip to shortly explain the format and purpose of taints. Maybe also display tooltip on list page to see whole taint.

lenartj:
I think showing the taints on the listing makes a lot of sense; not less than showing labels.
An idea: what if we merged the column and had it like "Labels/Taints" showing taints with a different background than labels?

floreks:
I'd keep them separate and just think about a way to display tooltip on hover in case they are truncated. It's still a valuable information based on which we can filter results. Merging them might be too confusing and does not save us that much space.

cheld:
I don't really agree here. The UI must be reasonable without tooltips. If my cluster has 100 nodes, I do not want to hover over every annotation. Also, tooltips do not work on mobile.

IMHO, in the current implementation the taints column _as well as_ the labels column is complete useless. I don't see any values, I don't see even the full keys. 

In the screenshot is see `beta.kubernetes.ar...` ...ar what? architecture? which one????? I don't see why this information has any value. 

btw: kubectl displays`node-role.kubernetes.io/master:NoSchedule`

cheld:
`An idea: what if we merged the columns and had it like "Labels/Taints" showing taints with a different background than labels? `

This is going in the right direction. Not sure about the UX, though

floreks:
Then the issue here is how to properly display this data, not if it is useful to display this information. Even on details page labels and taints might get truncated, i.e. on smaller screens or when browser window is resized. We have to improve overall display of our label component, because both labels and taints present useful information.

cheld:
Yes, but it gets worse with taints column. I can imagine that taints have some value. But so does many other information. We must really limit the amount of information shown in the table. Not sure if taints is absolutely required. It seems to be a more a technical fields. Labels are intented to organize stuff

floreks:
We have got to conclusion that for now we should remove taints from node list view and reintroduce that once we improve our chips component.

lenartj:
[CI error](https://travis-ci.org/kubernetes/dashboard/jobs/245672568#L544) is unrelated

floreks:
LGTM

PR #2038, opened 2017-06-20
Fix labels form not show when create an app use advanced options.

Zexi:
The specified labels form is missing when create app use advanced options.

10601497966914_ pic_hd

And the placeholder content is wrong.


Zexi:
I signed it!

Issue #2041, opened 2017-06-21
Deploy from file should use token from rest request


qiujian16:
DeployAppFromFile still use in cluster config. It should also use the token extracted from request to build the config. 

I would like to submit a pr to fix it.

floreks:
That's true. You are welcome to create PR for that. Thanks for creating an issue!

qiujian16:
https://github.com/kubernetes/dashboard/pull/2046

floreks:
Fixed by #2046.

PR #2046, opened 2017-06-22
Use bearer token when deploy from file

qiujian16:


qiujian16:
fix https://github.com/kubernetes/dashboard/issues/2041

src/app/backend/client/manager.go
```diff
@@ -97,6 +100,38 @@ func (self *clientManager) Config(req *restful.Request) (*rest.Config, error) {
...
+// Currently request header is only checked for existence of 'Authentication: BearerToken'
+func (self *clientManager) ClientCmdConfig(req *restful.Request) (clientcmd.ClientConfig, error) {
+	authInfo := self.extractAuthInfo(req)
+
+	cfg, err := self.buildConfigFromFlags(self.apiserverHost, self.kubeConfigPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(authInfo.Token) > 0 {
```
floreks: Shouldn't this check only set `cmdCfg.AuthInfos[DefaultCmdConfigName] = &authInfo` and config be returned anyway similar to what `Config` method does? Otherwise when user is not using authorization header and `apiserver-host` is used instead of in-cluster config deploy will fail because `cmdutil.NewFactory(nil)` assumes that dashboard runs using in-cluster config.

src/app/backend/client/manager.go
```diff
@@ -97,6 +100,38 @@ func (self *clientManager) Config(req *restful.Request) (*rest.Config, error) {
 	return cfg, nil
 }
 
+// ClientCmdConfig creates ClientCmd Config based on authentication information extracted from request.
+// Currently request header is only checked for existence of 'Authentication: BearerToken'
+func (self *clientManager) ClientCmdConfig(req *restful.Request) (clientcmd.ClientConfig, error) {
```
floreks: Maybe we could use this method to simplify `Config` method. `clientcmd.ClientConfig` has method `ClientConfig()` that returns `*rest.Config`. This way `Config` method would look like

```golang
cmdConfig, err := self.ClientCmdConfig(req)
if err ...
cfg, err := cmdConfig.ClientConfig()
if err ...
self.initConfig(cfg)
return cfg, nil
```

I'm not 100% sure if they will be equal but I think they should.

src/app/backend/client/manager.go
qiujian16: you are right, it is only necessary to set token in this check.

src/app/backend/client/manager.go
qiujian16: That will cause that in ```Config``` method, we create ```*rest.Config``` at first, then translate it to ```clientcmd.ClientConfig``` and translate it back to ```*rest.Config```, I thought it would increase some complexity. comments?

floreks: If we look at simplified code then cyclomatic complexity is lower because there is no custom logic involved. Result structure is created by calling other methods. Looking at code complexity it would get worse because indeed we are creating additional structure that normally is not needed, however this will still be O(1) as is only involves reassignment of structures. The advantage here is that we can keep authorization related code strictly in one method and if we'll want to support other authorization options then we have to change only 1 method not 2.

qiujian16: ok, that make sense :)

floreks:
LGTM!

PR #2050, opened 2017-06-22
Hide roles menu entry if RBACs api not enabled or supported

### LINKED ISSUE #2010: "Roles" section is broken
##### Environment

```
Dashboard Version: v1.6.1
Git Commit: dd28d1a01225d441e8d012235e5cf91d23381197
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.3", GitCommit:"0480917b552be33e2dba47386e51decb1a211df6", GitTreeState:"clean", BuildDate:"2017-05-10T15:48:59Z", 
GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.2", GitCommit:"477efc3cbe6a7effca06bd1452fa356e2201e1ee", GitTreeState:"clean", BuildDate:"2017-04-19T20:22:08Z", 
GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
```

##### Steps to reproduce
Clic "Roles" menu item.

##### Observed result
The page responds 404. Logs are:

```
[2017-06-02T21:42:14Z] Incoming HTTP/1.1 GET /api/v1/rbacrole?filterBy=&itemsPerPage=10&namespace=&page=1&sortBy=d,creationTimestamp request from 10.242.0.28:59740: {}
Getting list of RBAC roles
the server could not find the requested resource
[2017-06-02T21:42:14Z] Outcoming response to 10.242.0.28:59740 with 404 status code
Getting application global configuration
```

image

##### Comments
Is this because I'm running on an old cluster? Or have disabled Rbac? How can we prevent this from happening?

floreks:
Fixes #2010.

There is a check now if rbac api group is enabled and its' version is supported by dashboard.
As kubernetes 1.6 added `v1beta1` version then this is what we'll support. In case some older version is enabled or RBACs are disabled menu entry will be hidden.

Additionally I have fixed `Cluster` page to hide resources in case of error. Error log will be available in backend.

### TODO
- [x] Add tests

### DIFF
--- a/src/app/backend/resource/rbacroles/list.go
+++ b/src/app/backend/resource/rbacroles/list.go
@@ -21,7 +21,8 @@ import (
 	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
 	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
 	client "k8s.io/client-go/kubernetes"
-	rbac "k8s.io/client-go/pkg/apis/rbac/v1alpha1"
+	rbac "k8s.io/client-go/pkg/apis/rbac/v1beta1"
+	"fmt"
 )
 
 // RbacRoleList contains a list of Roles and ClusterRoles in the cluster.
@@ -41,6 +42,7 @@ type RbacRole struct {
 
 // GetRbacRoleList returns a list of all RBAC Roles in the cluster.
 func GetRbacRoleList(client *client.Clientset, dsQuery *dataselect.DataSelectQuery) (*RbacRoleList, error) {
+	fmt.Println(client.RbacV1beta1().RESTClient().APIVersion())
 	log.Println("Getting list of RBAC roles")
 	channels := &common.ResourceChannels{
 		RoleList:        common.GetRoleListChannel(client, 1),
@@ -55,11 +57,11 @@ func GetRbacRoleList(client *client.Clientset, dsQuery *dataselect.DataSelectQue
 func GetRbacRoleListFromChannels(channels *common.ResourceChannels, dsQuery *dataselect.DataSelectQuery) (*RbacRoleList, error) {
 	roles := <-channels.RoleList.List
 	if err := <-channels.RoleList.Error; err != nil {
-		return nil, err
+		return &RbacRoleList{Items: []RbacRole{}}, err
 	}
 	clusterRoles := <-channels.ClusterRoleList.List
 	if err := <-channels.ClusterRoleList.Error; err != nil {
-		return nil, err
+		return &RbacRoleList{Items: []RbacRole{}}, err
 	}
 
 	result := SimplifyRbacRoleLists(roles.Items, clusterRoles.Items, dsQuery)
--- a/src/app/backend/resource/rbacroles/list.go
+++ b/src/app/backend/resource/rbacroles/list.go
@@ -22,7 +22,6 @@ import (
 	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
 	client "k8s.io/client-go/kubernetes"
 	rbac "k8s.io/client-go/pkg/apis/rbac/v1beta1"
-	"fmt"
 )
 
 // RbacRoleList contains a list of Roles and ClusterRoles in the cluster.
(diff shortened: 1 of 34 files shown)


src/app/backend/resource/rbacroles/list.go
```diff
@@ -41,6 +42,7 @@ type RbacRole struct {
 
 // GetRbacRoleList returns a list of all RBAC Roles in the cluster.
 func GetRbacRoleList(client *client.Clientset, dsQuery *dataselect.DataSelectQuery) (*RbacRoleList, error) {
+	fmt.Println(client.RbacV1beta1().RESTClient().APIVersion())
```
cheld: Debug left over?

src/app/frontend/chrome/nav/rolenav_component.js
```diff
@@ -0,0 +1,64 @@
...
+    this.states;
+  }
+
+  /**
+   * Check status of rbacs in the cluster to show/hide Role menu entry.
+   *
+   * @export
+   */
+  $onInit() {
+    this.resource_('api/v1/rbac/status').get().$promise.then((result) => {
```
cheld: I assume this method is called only once when the page is initially loaded. I checked with network tab in chrome. currently we make 15 initial requests. In my understanding, this will increase the number to 16. I guess this is ok for the moment. 

From architectual point of view I wonder, if it would be better to have a generic service in the backend for all menu entries and move the business logic more to the backend

What about the other menu entries? Isn't it possible to remove with RBAC any other menu entry?

src/app/backend/validation/validaterbacstatus.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	apiVersions := metav1.ExtractGroupVersions(groupList)
+	return &RbacStatus{
+		Enabled: found(apiVersions, rbac.SchemeGroupVersion.String()),
+	}, nil
+}
+
+// Returns true if element has been found in given array, false otherwise.
+func found(arr []string, str string) bool {
```
cheld: contains() or find()

or maybe

arrayContains()

src/app/backend/resource/rbacroles/list.go
floreks: Yes. I will remove it.

src/app/backend/validation/validaterbacstatus.go
floreks: Forgot to rename it. I think `find` should be ok.

src/app/frontend/chrome/nav/rolenav_component.js
floreks: Only roles entry is related to RBAC. There are only few menu entries that require some logic to be shown/hidden, like third party resources and rbacs. Most resources will always be there and can not be disabled.

cheld: ok, we can refactor when it becomes relevant

src/app/backend/validation/validaterbacstatus.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	apiVersions := metav1.ExtractGroupVersions(groupList)
+	return &RbacStatus{
+		Enabled: contains(apiVersions, rbac.SchemeGroupVersion.String()),
+	}, nil
+}
+
+// Returns true if element has been found in given array, false otherwise.
+func contains(arr []string, str string) bool {
```
floreks: `contains` seems better after all since `find` should return index of object not only boolean value.

src/app/backend/validation/validaterbacstatus_test.go
```diff
@@ -0,0 +1,123 @@
...
+type FakeDiscovery struct {
+	fakediscovery.FakeDiscovery
+	fakeServerGroupsMethod
+}
+
+func (self *FakeDiscovery) ServerGroups() (*metav1.APIGroupList, error) {
+	return self.fakeServerGroupsMethod()
+}
+
+func TestValidateRbacStatus(t *testing.T) {
```
floreks: After all I have found a way to override used method through composition.

PR #2061, opened 2017-06-27
Fix documentation for running local-up-cluster

### LINKED ISSUE #1483: local-up-cluster looks for heapster at the wrong place according to the guide
#### Issue details
After the changes #1351 the default heapster api is set to in-cluster heapster api. If the developer follows the  [Getting-started guide](https://github.com/kubernetes/dashboard/blob/master/docs/devel/getting-started.md), gulp local-up-cluster command doesn't work by default. 

##### Environment

```
Dashboard version: master
```

##### Steps to reproduce
run "gulp local-up-cluster"

##### Observed result
It waits heapster forever.

##### Expected result
It should find working heapster by default. 

##### Comments
Currently the developer has to pass  --heapsterServerHost parameter for "gulp local-up-cluster" and after for "gulp serve" which is not referred in the documentation.
As a solution heapster can be deployed in hyperkube local cluster or documentation can be updated. Any ideas?

floreks:
Closes #1483

### DIFF
--- a/docs/devel/getting-started.md
+++ b/docs/devel/getting-started.md
@@ -41,7 +41,7 @@ task is provided that checks out the latest stable version, and runs it inside a
 Run the following command:
 
 ```shell
-$ gulp local-up-cluster
+$ gulp local-up-cluster --heapsterServerHost 'http://localhost:8082'
 ```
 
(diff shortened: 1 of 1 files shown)


maciaszczykm:
I was doing it right now :)

PR #2068, opened 2017-06-27
Move license files to build/assets directory

maciaszczykm:
We should not keep directory and file with same name (only case is different) in the same location. macOS doesn't handle it well and constantly removes one of them (directory or file). It was the case with `license` directory and `LICENSE` file.


maciaszczykm:
Quoting from https://discussions.apple.com/thread/791721?tstart=0:

> This is all based on the format of the disk you're using.
>
>HFS+ is the default format for Mac OS X. It is case-preserving but not case-sensitive, that is it knows when you use upper and lower case characters in a filename, but file.c and file.C are considered the same.
>
>Many shell-based applications automatically assume case sensitive (where file.c and file.C are completely different files) and break when you don't have a case-sensitive file system, as youv've found.
>
>It is possible to use case-sensitive file systems on Mac OS X - there's even a HFS+ (case sensitive) option, but it can only be defined when you format a drive, and it's not recommended for your boot volume.
>
>Therefore your best option is to create a second volume/partition that is formatted as HFS+ (case sensitive) and run all your case-sensitive work from there.
>
>If you have a second drive, that's easy, otherwise you'll need to either repartition your existing drive or create a disk image to store the files on. /Applications/Utilities/Disk Utility.app can reformat drives and create disk images if that's the way you want to go.

Issue #574, opened 2016-03-24
Authentication guidelines


cescoferraro:
##### Environment

```
Dashboard version: kubernetes-dashboard-canary
Kubernetes version: v1.2
Operating system: CoreOS 899 self hosted on Digital Ocean through Terraform
Go version: go version go1.6 darwin/amd64
```
##### Steps to reproduce

i am deploying this [kubernetes-dashboard-canary](https://github.com/kubernetes/dashboard/blob/master/src/deploy/kubernetes-dashboard-canary.yaml)
##### Observed result

using the insecure port
`- --apiserver-host=http://{{ .Env.MASTER_PRIVATE }}:8080`
it works but I have no restriction to who access the dashboard.

using the secure port
`- --apiserver-host=https://{{ .Env.MASTER_PRIVATE }}:6443`
i get
`Get https://10.134.19.15:6443/api/v1/replicationcontrollers: x509: failed to load system roots and no roots provided`
as expected because the pod ip in not on the certificates I generated as the kube-apiserver confirms
`Mar 24 22:59:32 master kube-apiserver[1298]: I0324 22:59:32.632002    1298 logs.go:41] http: TLS handshake error from 10.100.68.2:45592: remote error: bad certificate`

I am using simple authentication to the api, I was expecting something like this for the dashboard as well. What the dashboard way of doing this?
Do I have to build a new docker-image copying the certificates to /etc/kubernetes/ssl/??

bryk:
@luxas Can you chime in?

bryk:
@cescoferraro When you run the Dashboard UI inside your cluster, the UI should attempt to securely connect to the master using serviceaccounts. This is our default behaviour.

bryk:
Re securing the Dashboard UI itself: at the moment you should secure it by accessing it via service proxy or through firewalled private network. We're working on IAM/auth support in the meantime.

luxas:
@bryk FYI, this PR will make local docker users more happy: kubernetes/kubernetes#23550

In this case @cescoferraro, I think you should set 
```
--root-ca-file=/home/core/ssl/ca.pem
--service-account-private-key-file=/home/core/ssl/kubelet-key.pem
```
as @antoineco pointed out. 
And if you sign `/home/core/ssl/*` certs with `10.100.68.2` as one SAN, I think it should be OK.

cescoferraro:
[...]

PR #2074, opened 2017-06-29
[WIP] Update documentation for single/multi tenant configurations

### LINKED ISSUE #2045: Split setup document in single tenant and multi-tenant
The default setup is a privilige escalation in a multi-tenant setup.

At least, we split the documentation to indicate that the default setup cannot be used in a mult-tenant environment

floreks:
Closes #2045. Related to discussion started in #574.

First we have to state what we will support. What I think we should do is:

1. Provide good documentation for latest supported kubernetes release (currently 1.6).
1. Provide option to deploy on older clusters but leave advanced configuration up to the users. Only basic yaml without RBAC configuration.
1. Split deployment into 2 parts: dashboard, rbac configuration
1. Document details about single/multi tenant approach that our proposed configuration supports.
1. For multi-tenant setup describe how to authorize users using authorization header and tokens but without specific examples as we don't have enough people to maintain them well.
1. Simplify head releases and support only latest kubernetes versions with full access to the cluster (by providing yaml file). Of course we will add information about that to warn potential users.
1. Remove other arch yaml files and just document what architectures we support with example that in order to deploy for different arch user can use this command `curl -sSL <dashboard_url> | sed "s/amd64/arm64/g" | kubectl create -f -`
1. State that `kubectl proxy` is not the right way to authorize users for multi-tenant configurations.

I have updated documentation a bit and if we agree on required changes I will update yamls and remaining docs.

Calling everyone involved in discussion. 

@bryk @cheld @rf232 @liggitt @maciaszczykm

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,48 +13,43 @@ itself.
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
 ## Deployment
-Provided instructions are compliant with kubernetes 1.6+. If you plan to deploy dashboard on older kubernetes version please
-follow [these instructions](#older-kubernetes-versions).
+Provided instructions are compliant with Kubernetes 1.6+. If you plan to deploy Dashboard on older Kubernetes version please
+follow [these instructions](docs/user-guide/deployment-old.md) or if you want to check out latest
+features that the team works on try our HEAD releases [installation guide](docs/devel/head-releases.md).
+
+For security reasons installation has been split into two parts:
+[Deployment](#deployment) and [Configuration](#configuration). Deployment part will deploy Dashboard on your
+cluster without any privileges. It will only work if RBACs are disabled on your cluster.
 
 It is likely that the Dashboard is already installed on your cluster. Check it with the following command:
 ```shell
 $ kubectl get pods --all-namespaces | grep dashboard
 ```
 
-If it is missing please follow installation instructions provided below. For security reasons installation has been split into two parts.
-First part will deploy dashboard on your cluster without any privileges. It will only work if RBACs are disabled on your cluster. Second part
-will grant dashboard necessary privileges based on chosen cluster configuration.
-
-Deploy dashboard by running:
+If it is missing deploy Dashboard by running:
 ```shell
 $ kubectl create -f https://git.io/kube-dashboard
 ```
 
-### Multi-tenant configuration
+## Configuration
+Now that Dashboard is deployed in your cluster you need to  grant it necessary privileges. Please choose one of the following
+configurations that suits you best.
+#### Multi-tenant configuration
 This configuration should be used if cluster will be used by multiple users with different privileges. It grants
 dashboard minimal access to apiserver required to start it. Read more about this configuration in [multi tenant setup](docs/user-guide/multi-tenant.md).
 ```shell
 TODO multi-tenant SA, Role deployment
 ```
 
-### Single-tenant configuration
+#### Single-tenant configuration
 This configuration should be used only if cluster will be used by trusted users and all of them are allowed to access all of its' resources.
 ```shell
 TODO single-tenant SA, Role deployment
 ```
 
-### Older kubernetes versions
-If you are using Kubernetes 1.5 or earlier, you can install the latest stable release by running the following command:
-```shell
-$ kubectl create -f https://git.io/kube-dashboard-no-rbac
-```
-**IMPORATNT:** Please keep in mind that this will only work if RBACs are disabled in your cluster.
-
-### Head releases
-You can also install unstable HEAD builds with the newest features that the team works on by
-following the [development guide](docs/devel/head-releases.md).
-
-### Graphs
+## Addons
+Dashboard can use cluster addons to enhance user experience, i.e. show metrics and graphs.
+#### Graphs
 For the metrics and graphs to be available you need to
 have [Heapster](https://github.com/kubernetes/heapster/) running on your cluster. We require heapster to be deployed in `kube-system` namespace
 together with service named `heapster`. 
--- a/README.md
+++ b/README.md
@@ -13,26 +13,51 @@ itself.
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
 ## Deployment
-It is likely that the Dashboard is already installed on your cluster. Check with the following command:
+Provided instructions are compliant with kubernetes 1.6+. If you plan to deploy dashboard on older kubernetes version please
+follow [these instructions](#older-kubernetes-versions).
+
+It is likely that the Dashboard is already installed on your cluster. Check it with the following command:
 ```shell
 $ kubectl get pods --all-namespaces | grep dashboard
 ```
 
-If it is missing, you can install the latest stable release by running the following command:
+If it is missing please follow installation instructions provided below. For security reasons installation has been split into two parts.
+First part will deploy dashboard on your cluster without any privileges. It will only work if RBACs are disabled on your cluster. Second part
+will grant dashboard necessary privileges based on chosen cluster configuration.
+
+Deploy dashboard by running:
 ```shell
 $ kubectl create -f https://git.io/kube-dashboard
 ```
 
+### Multi-tenant configuration
+This configuration should be used if cluster will be used by multiple users with different privileges. It grants
+dashboard minimal access to apiserver required to start it. Read more about this configuration in [multi tenant setup](docs/user-guide/multi-tenant.md).
+```shell
+TODO multi-tenant SA, Role deployment
+```
+
+### Single-tenant configuration
+This configuration should be used only if cluster will be used by trusted users and all of them are allowed to access all of its' resources.
+```shell
+TODO single-tenant SA, Role deployment
+```
+
+### Older kubernetes versions
 If you are using Kubernetes 1.5 or earlier, you can install the latest stable release by running the following command:
 ```shell
 $ kubectl create -f https://git.io/kube-dashboard-no-rbac
 ```
+**IMPORATNT:** Please keep in mind that this will only work if RBACs are disabled in your cluster.
 
+### Head releases
 You can also install unstable HEAD builds with the newest features that the team works on by
 following the [development guide](docs/devel/head-releases.md).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,26 +13,51 @@ itself.
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
 ## Deployment
-It is likely that the Dashboard is already installed on your cluster. Check with the following command:
+Provided instructions are compliant with kubernetes 1.6+. If you plan to deploy dashboard on older kubernetes version please
```
maciaszczykm: s/dashboard/Dashboard
s/kubernetes/Kubernetes

README.md
```diff
@@ -13,26 +13,51 @@ itself.
...
 ```
 
+### Multi-tenant configuration
+This configuration should be used if cluster will be used by multiple users with different privileges. It grants
+dashboard minimal access to apiserver required to start it. Read more about this configuration in [multi tenant setup](docs/user-guide/multi-tenant.md).
+```shell
+TODO multi-tenant SA, Role deployment
+```
+
+### Single-tenant configuration
```
cheld: The deployment section is the single tenant documentation, right?

README.md
```diff
@@ -13,26 +13,51 @@ itself.
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
 ## Deployment
-It is likely that the Dashboard is already installed on your cluster. Check with the following command:
+Provided instructions are compliant with kubernetes 1.6+. If you plan to deploy dashboard on older kubernetes version please
+follow [these instructions](#older-kubernetes-versions).
```
cheld: I like this idea, but I would move to separate page. The landing page should be as easy as possible
You could do similar with multi-tenant setup

README.md
cheld: move to separate page

README.md
```diff
@@ -13,26 +13,51 @@ itself.
...
+TODO multi-tenant SA, Role deployment
+```
+
+### Single-tenant configuration
+This configuration should be used only if cluster will be used by trusted users and all of them are allowed to access all of its' resources.
+```shell
+TODO single-tenant SA, Role deployment
+```
+
+### Older kubernetes versions
```
cheld: move to separate page

README.md
floreks: I didn't want to force user to open another page just to copy command in order to deploy rbac configuration. Single-tenant description is also very simple. There is not much more to add so keeping 1 sentence and 1 command on separate page seems unnecessary.

Do we want to force users to open another page just to set up RBACs for dashboard? Usually all commands required to run application are present on a landing page.

README.md
floreks: I agree here. We can move this information to a separate page.

README.md
cheld: Landing Page should describe getting started steps. Everything advanced to subpages

cheld:
My comments on the questions:
1. Hope so 
2. We must support the current version and the version before. At least this was the rule in kube-deploy.
3. No, leave as it is. It will become deployment for single tenant.
4. ok
5. Yes, please. In addition, you can also add links to issues with detailed discussions
6. Yes. (Do we need Head at all?  Replace by regex?)
7. I think Dashboard for Raspi is quite popular. But only single tenant
8. Yes, as part of the multi-tenant setup

floreks:
> 3. No, leave as it is. It will become deployment for single tenant.

As we discussed in #574 single-tenant(full dashboard access to the cluster) approach is overall not a secure approach, therefore I'd not make it our default and suggested configuration. I think information about both configurations have to be stated loud and clear on the same level without making one of them more prominent. This way user will have to choose in the beginning configuration he preffers and we will not suggest that he should use one or another.

Either we leave both options with short description on landing page or we only list links to pages with information about second deployment step, i.e.

Deploy dashboard using the following command:
`$ ...`
Set dashboard privileges by choosing one of the configurations:
`link to single-tenant doc`
`link to multi-tenant doc`

maciaszczykm:
`Multi/single-tenant` is not easy to understand for new users. We have to let beginners, what it is (secure setup or a setup with everything unlocked, which is super dangerous). I would change this name or at least add `(secure setup)` and `(insecure setup)` information. 

@bryk

maciaszczykm:
`Multi/single-tenant` is not easy to understand for new users. We have to let beginners, what it is (secure setup or a setup with everything unlocked, which is super dangerous). I would change this name or at least add `(secure setup)` and `(insecure setup)` information. 

@bryk @rf232 @liggitt

cheld:
If we don't want to provide a default kind of deployment and want to force the user to make a decission  than we provide just 2 links on landing page:

``link to single-tenant doc`` 
``link to multi-tenant doc``

In the single tenant doc we provide the current yaml and the current documentation. I don't like to split up the deployment for single tenant in two steps.

maciaszczykm:
@cheld We have to let users know, that *installation process has two steps* to do it like this I believe.

cheld:
In any case the documentation must explain the decision the user has to take. 

The options secure/insecure could work in other scenarios, but in our it does not make sense. So, why would someone want an insecure setup? So every user would follow the secure setup and then fail, because it is far to complicated. Then you have to explain that the insure setup is actually secure for single tenant environments. So, the actual choice for the user is single tenant/multi-tenant. There is no escape. It has to be explained.

cheld:
`We have to let users know, that installation process has two steps to do it like this I believe.`

In the multi-tenant case that certainly is ok. In the single tenant case, I wonder if there is any So, the gain for the user executing 2 commands instead of one.

Maybe you name the yaml something like "grant-full-access-to-dashboard.yaml", but there has to be some benefit...

floreks:
Actually this single-tenant approach could still be insecure even if there is only 1 person that manages the cluster. In example if you deploy ubuntu and decide to give user access to this "virtualized system". Even so user is not aware that underneath kubernetes is used then privilege escalation is possible thanks to pod-to-pod communication because you have deployed single-tenant configuration and dashboard has full access to the cluster.

IMHO single-tenant will always be insecure by default and exposed to potential security breaches. Multi-tenant on the other side will be secure by default.

cheld:
@gertipoppel will help to bring it into proper shape. But please provide all the required steps and yaml

floreks:
I can't provide any steps or yamls until we agree how it should look like. Anyway yamls are not required. I can prepare them once everything is documented.

@bryk @rf232 could you share your opinion?

cheld:
I think we agree that there is no point in documenting a setup for production and defining it as insecure by default and a gate for security breaches. In this case we should not document it at all and only document the secure setup, but then in _all_ details. 

On the other hand this architectual problem is known since the beginning. So far we considered it secure enough for single tenant usage and securing multi tenant usage as an ongoing process.

floreks:
I think that we can still document this setup but we are obligated to warn users about potential security issues and maybe suggest some steps to avoid it, i.e. by setting up network policies and blocking traffic from other pods/namespaces.

I only wouldn't default to single setup but rather present both configurations as equal and let user choose what he needs. We only need to describe pros and cons of both configurations. Then we can just point people to our documentation if there are some issues caused by not reading it and close them.

liggitt:
> On the other hand this architectual problem is known since the beginning. So far we considered it secure enough for single tenant usage

I strongly disagree. Private pod and service networks should not be assumed, even in single-tenant setups. 

> I think that we can still document this setup but we are obligated to warn users about potential security issues and maybe suggest some steps to avoid it, i.e. by setting up network policies and blocking traffic from other pods/namespaces.

Yes, I'd expect a clear explanation of the exposure and a list of the aspects that must be secured to avoid letting unintended actors make use of that exposure

bryk:
@floreks @maciaszczykm 

Can we go with the secure/insecure wording? Or insecure/default? And explain why someone would use insecure (for onboarding and stuff). I believe that multi-single tenant wording hides away the important piece of security information that we want to convey.

README.md
```diff
@@ -13,26 +13,46 @@ itself.
...
 
 ## Deployment
-It is likely that the Dashboard is already installed on your cluster. Check with the following command:
+Provided instructions are compliant with Kubernetes 1.6+. If you plan to deploy Dashboard on older Kubernetes version please
+follow [these instructions](docs/user-guide/deployment-old.md) or if you want to check out latest
+features that the team works on try our HEAD releases [installation guide](docs/devel/head-releases.md).
+
+For security reasons installation has been split into two parts:
+[Deployment](#deployment) and [Configuration](#configuration). Deployment part will deploy Dashboard on your
+cluster without any privileges. It will only work if RBACs are disabled on your cluster.
```
bryk: Without any privileges? I think this is actually "with administrator privileges open to anyone with access to the UI". Is this correct?

maciaszczykm:
>@floreks @maciaszczykm
>
>Can we go with the secure/insecure wording? Or insecure/default? 

I suggested same thing in https://github.com/kubernetes/dashboard/pull/2074#issuecomment-312249032.

README.md
floreks: This is the current behavior but my proposal is to split insallation and privileges configuration into 2 parts so simple `kubectl create -f dashboard.yaml` would deploy dashboard but without cluster role or service account with admin privileges.

Then as a second step we'd deploy access rights configuration based on chosen configuration: minimal or full.

floreks:
> Can we go with the secure/insecure wording? Or insecure/default? 

Yes, I'd also go with this wording. We have to make sure that when dashboard is deployed with kubernetes then its' default configuration will not cause any security issues. That is currently the case.

cheld:
> Or insecure/default?

I am fine with that (or something else). However, it would be bizzar if the "default" installation document only describes the idea for the setup and is only usefull for experts...But lets wait and see how simple we can make the document

bryk:
Any progress here? It'd be great to have this for 1.6.2.

floreks:
It's on hold for now as I'm working on implementing log in mechanism for dashboard. Once it's done we will fallback to secure by default and rely on it.

bryk:
Awesome, thanks!

maciaszczykm:
Already done in a different way.

PR #2081, opened 2017-07-04
Fix cascade delete job and job associated pod

riverzhang:
Cascade delete job and job associated pod
Fix test

Replication Controller, Replica Set, StatefulSet, DaemonSet, and Deployment Cascading deletions are controlled by garbage-collection, but job resource is no Garbage Collection, job's pod is not ownerReferences.
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/
https://github.com/kubernetes/community/blob/548eb7292d45a4af78e0141d33094130dfe6080a/contributors/design-proposals/synchronous-garbage-collection.md

So I refer to the kuberctl delete job:
get job info->get job's pod->delete job's pod->delete job


floreks:
There are build errors.

riverzhang:
@floreks  Could you please reivew my code?

maciaszczykm:
@riverzhang You need to prepare some tests, because overall coverage went down.

src/app/backend/resource/job/detail.go
```diff
@@ -95,3 +97,27 @@ func getJobDetail(job *batch.Job, metricClient metricapi.MetricClient,
 		Completions:     job.Spec.Completions,
 	}
 }
+
+// Deletes job with given name in given namespace and related pods.
```
maciaszczykm: It should start with method name.

src/app/backend/handler/apihandler.go
```diff
@@ -1412,21 +1415,31 @@ func (apiHandler *APIHandler) handlePutResource(
...
-	verber, err := apiHandler.cManager.VerberClient(request)
-	if err != nil {
-		handleInternalError(response, err)
-		return
-	}
-
 	kind := request.PathParameter("kind")
 	namespace, ok := request.PathParameters()["namespace"]
 	name := request.PathParameter("name")
+	if kind == "job" {
```
floreks: Replace "job" with `ResourceKindJob` from `backend/api/types.go`

src/app/backend/handler/apihandler.go
```diff
@@ -1412,21 +1415,31 @@ func (apiHandler *APIHandler) handlePutResource(
 
 func (apiHandler *APIHandler) handleDeleteResource(
```
floreks: I'd move clients creation to the beginning of function since it's very cheap.

Then use something like
```golang
switch kind {
    case ResourceKindJob:
      err = job.DeleteJob(k8sClient, namespace, name)
    default:
      err = verber.Delete(kind, ok, namespace, name)
}

if err != nil {
   // handle it
}

response.WriteHeader(http.StatusOK)
```

It will decrease complexity.

src/app/backend/resource/job/detail.go
```diff
@@ -95,3 +97,27 @@ func getJobDetail(job *batch.Job, metricClient metricapi.MetricClient,
...
+func DeleteJob(client k8sClient.Interface, namespace, name string) error {
+
+	log.Printf("Deleting %s job from %s namespace", name, namespace)
+
+	pods, err := getRawJobPods(client, namespace, name)
+	if err != nil {
+		return err
+	}
+	if err := client.BatchV1().Jobs(namespace).Delete(name, &metaV1.DeleteOptions{}); err != nil {
```
floreks: I think that kubectl first deletes all the pods and then deletes job with specific delete options. I'd keep their logic as they know better how it should be handled.

https://github.com/kubernetes/kubernetes/blob/a92123c5304514c2996f1c7d0f4ed4b7311cd18a/pkg/kubectl/stop.go#L351

src/app/backend/resource/job/detail.go
riverzhang: Thanks! I will add the test code.

src/app/backend/handler/apihandler.go
riverzhang: Thanks! t is good.

src/app/backend/handler/apihandler.go
riverzhang: Thanks !I will remember this usage

src/app/backend/resource/job/detail.go
riverzhang: I agree with this idea, I am here wrong.

src/app/backend/handler/apihandler.go
```diff
@@ -1412,19 +1412,31 @@ func (apiHandler *APIHandler) handlePutResource(
...
-		handleInternalError(response, err)
-		return
+	switch kind {
+	case api.ResourceKindJob:
+		k8sClient, err := apiHandler.cManager.Client(request)
+		if err != nil {
+			handleInternalError(response, err)
+			return
+		}
+		if err := job.DeleteJob(k8sClient, namespace, name); err != nil {
```
floreks: If you want to keep client creation in the switch then at least move this error handling out of switch. Just assign error and handle it outside the switch. 

```golang
var err error
switch kind {
    case api.ResourceKindJob:
        k8sClient, err := apiHandler.cManager.Client(request)
        if err != nil {
            handleInternalError(response, err)
            return
        }

        err = job.DeleteJob(k8sClient, namespace, name)
    default:
        verber, err := apiHandler.cManager.VerberClient(request)
        if err != nil {
            handleInternalError(response, err)
            return
        }

        err = verber.Delete(kind, ok, namespace, name)
    }
}

if err != nil {
    handleInternalError(response, err)
    return
}
```

riverzhang: Thanks!   I would like to write like this.
I did not define the variables err, so I am running code is error.

floreks:
I'm not sure I understand what you mean. Could you give some example or try to describe your issue better?

src/app/backend/resource/job/detail_test.go
```diff
@@ -100,3 +101,60 @@ func TestGetJobDetail(t *testing.T) {
...
+					Template: v1.PodTemplateSpec{
+						ObjectMeta: metaV1.ObjectMeta{
+							Name:      "test-pod",
+							Namespace: "ns-1",
+							Labels:    map[string]string{"app": "test"},
+						},
+					},
+				},
+			},
+		},
```
floreks: Yes but second test will probably not work like this. You need to create pod list and pass it to `fake.NewSimpleClientset` method. Then when `client.Pods().List()` is called it will return pod list passed to this method.

src/app/backend/resource/job/detail_test.go
```diff
@@ -100,3 +101,60 @@ func TestGetJobDetail(t *testing.T) {
...
+		namespace, name string
+		expectedActions []string
+		job             *batch.Job
+	}{
+		{
+			"ns-1", "job-1",
+			[]string{"get", "list", "delete"},
+			createJob("job-1", "ns-1", map[string]string{"app": "test"}),
+		},
+		{
```
floreks: Remove this case as it does not test what it's supposed to test. It does not test deletion of related pods. It is the same as first one essentialy.

riverzhang: ok! I delete second test.  There is no test to deletion of related pods.

floreks:
I'll test it and merge if everything works ok. Code LGTM.

floreks:
It doesn't work. Code does not wait for pods to be deleted and Job recreates new pod before it gets deleted.

riverzhang:
@floreks 
I am sorry,   It is work ok when  the job's pod state is completed,  
So I think delete  job then delete  pod!

floreks:
It has to work in all cases. When jobs' pod is in error state then pod gets deleted, job recreates pod, job gets deleted, recreated pod is not deleted.

riverzhang:
@floreks   It‘s  work fine.
Scale job to 0 -> get related pods -> delete related pods -> delete job with orphanDependants set

src/app/backend/resource/job/detail.go
```diff
@@ -95,3 +99,50 @@ func getJobDetail(job *batch.Job, metricClient metricapi.MetricClient,
...
+	log.Printf("Deleting %s job from %s namespace", name, namespace)
+
+	job, err := client.BatchV1().Jobs(namespace).Get(name, metaV1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	count := 0
+	*job.Spec.Parallelism = int32(count)
+	_, err = client.BatchV1().Jobs(namespace).Update(job)
```
floreks: I'm still not sure if this is the right approach. This is non-blocking operation which means that it will trigger the scale operation and proceed with rest of the code. In case cluster is heavily used this could cause same effect as before. Pods will be deleted and job will have time to recreate some of them before it gets scaled and deleted.

Kubectl executed blocking operation which means that it waits until scale operation is done and only then proceeds.

riverzhang: Thanks， I did not think so well.  I will try and learn to write this code again.

floreks:
@riverzhang are you working on this? I can do this quickly if you have no time. We'd like to have this change before new release. Probably it will happen next week.

riverzhang:
@floreks  I came back to work today. I will fix it, if I need help ,I will ask you about it.

riverzhang:
@floreks

maciaszczykm:
Closing as stale. It should be handled by Kubernetes itself, not us.

Issue #2049, opened 2017-06-22
Aggregate logs across controllers and services


cheld:
The current implementation has a combobox to select container within pod:
![logs](https://user-images.githubusercontent.com/7448799/27423683-b0e25c16-5732-11e7-80de-e8b0c3fb103d.png)

I would suggest to add second combobox to select pod within a replicaset and keep the selected container

Tracking issue: https://github.com/kubernetes/dashboard/issues/1828

PR #2096, opened 2017-07-07
WIP: Added pod selection to logs view

cheld:
Added view logs link to higher controller objects, like replica set. In the logs view the user can select the pod. 

Fixes https://github.com/kubernetes/dashboard/issues/2049.


src/app/backend/resource/generic/controller.go
```diff
@@ -47,6 +56,8 @@ type ResourceController interface {
 	UID() types.UID
 	// Get is a method, that returns ResourceOwner object.
 	Get(allPods []v1.Pod, allEvents []v1.Event) ResourceOwner
+	// Returns all log sources of controlled resource (e.g. a list of containers and pods for a replica set)
+	GetLogSources(k8sClient *client.Clientset) LogSources
```
maciaszczykm: I like the idea of adding more code here. Ideally we would use interfaces for all resources and our code would be more consistent.

src/app/frontend/replicaset/detail/info_component.js
```diff
@@ -34,6 +38,13 @@ export default class ReplicaSetInfoController {
   areDesiredPodsRunning() {
     return this.replicaSet.podInfo.running === this.replicaSet.podInfo.desired;
   }
+
+  getLogsHref() {
+    return this.state_.href(
+        logsStateName,
+        new LogsStateParams(
+            this.replicaSet.objectMeta.namespace, this.replicaSet.objectMeta.name, 'ReplicaSet'));
```
cheld: @maciaszczykm how are constants ('ReplicaSet') in Dashboard handled?

maciaszczykm: We have no specific pattern. If you use `'ReplicaSet'` more than once within a file it would be good to export it to `const` to ensure, that there are no typos etc.

floreks: Backend has resource kind names defined as lower case strings in `api/types.go`.  We should use the same convention here.

src/app/backend/resource/generic/controller.go
```diff
@@ -214,3 +256,26 @@ func (self StatefulSetController) Get(allPods []v1.Pod, allEvents []v1.Event) Re
...
+	allPods, _ := k8sClient.CoreV1().Pods(namespace).List(listEverything)
+	matchingPods := common.FilterPodsByOwnerReference(namespace, uid, allPods.Items)
+	names := make([]string, 0)
+	for _, pod := range matchingPods {
+		names = append(names, pod.Name)
+	}
+	return names
+}
+
+var listEverything = meta.ListOptions{
```
cheld: @maciaszczykm is that correct? Should different selectors be used?

maciaszczykm: @cheld It looks okay for me, maybe you could move this variable to the top of the file to make it more visible.

maciaszczykm: Also, I think we may use this variable somewhere else. You could make it public if it is true and simplify code.

cheld:
PTAL

Maybe some more tests needed

src/app/backend/resource/generic/controller.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package owner
+package generic
```
floreks: I don't think generic is a good name. It doesn't say anything about what can be inside. Since it's about resource controller then `controller` would look better. It's already under resource package so we would have `resource/controller/controller.go`.

src/app/backend/resource/logs/source.go
```diff
@@ -0,0 +1,53 @@
...
+import "github.com/kubernetes/dashboard/src/app/backend/resource/common"
+import client "k8s.io/client-go/kubernetes"
+import meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+import "k8s.io/client-go/pkg/api/v1"
+
+// Returns all log sources for a given resource. A log source identifies a log file through the combination of pod & container
+func GetLogSources(k8sClient *client.Clientset, ns string, resourceName string, resourceType string) (generic.LogSources, error) {
+	if resourceType == "Pod" {
+		return getLogSourcesFromPod(k8sClient, ns, resourceName)
+	} else {
```
floreks: Unnecessary else.

src/app/backend/resource/logs/source.go
```diff
@@ -0,0 +1,53 @@
...
+
+package logs
+
+import "github.com/kubernetes/dashboard/src/app/backend/resource/generic"
+import "github.com/kubernetes/dashboard/src/app/backend/resource/common"
+import client "k8s.io/client-go/kubernetes"
+import meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+import "k8s.io/client-go/pkg/api/v1"
+
+// Returns all log sources for a given resource. A log source identifies a log file through the combination of pod & container
```
floreks: Should start with method name for public methods.

src/app/backend/resource/logs/source.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Returns all containers for a given pod
+func getLogSourcesFromPod(k8sClient *client.Clientset, ns, resourceName string) (generic.LogSources, error) {
+	logSources := generic.LogSources{}
+	pod, err := k8sClient.CoreV1().Pods(ns).Get(resourceName, meta.GetOptions{})
+	if err != nil {
+		return logSources, err
+	}
+	logSources.ContainerNames = common.GetContainerNames(&pod.Spec)
```
floreks: IMHO return with object creation looks better. @maciaszczykm WDYT?

```golang
return generic.LogSource{
    ContainerNames: ...,
    PodNames: ...,
}
```

src/app/backend/resource/logs/source.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Returns all pods and containers for a controller object, such as ReplicaSEt
+func getLogSourcesFromController(k8sClient *client.Clientset, ns, resourceName, resourceType string) (generic.LogSources, error) {
+	logSources := generic.LogSources{}
+	rc, err := generic.NewResourceController(v1.ObjectReference{Kind: resourceType, Name: resourceName, Namespace: ns}, k8sClient)
+	if err != nil {
+		return logSources, err
+	}
+	logSources = rc.GetLogSources(k8sClient)
+	return logSources, nil
```
floreks: Same here. First line is not needed and overall there would be less lines.
```
return rc.GetLogSources(k8sClient), nil
```

src/app/backend/resource/logs/source.go
maciaszczykm: I agree in gerneral, but there is also `logSources := generic.LogSources{}`.

floreks: Yes it could be deleted also and instead return can be changed to `return generic.LogSource{}`.

src/app/backend/resource/generic/controller.go
cheld: In general, it is not a good practice to use names that a biased from a different context. In this case controller is a pattern and also a domain object. Not so nice, but maybe ok since it is an existing domain term and we have no better alternative.

I was wondering, if we want to add some more generic or reflective methods to this package. On the other hand we could rename once this is the case...

Unless you happen to have an alternaive idea, I would agree to go with controller

src/app/backend/resource/logs/source.go
cheld: I clicked randomly around in the code. The only code idiom I always find is  'return nil, err'. I never found 'return <empty object>, err'. 

Should I switch to pointer?

floreks: Pointers should be used if structure can be changed by methods that accept it (without having to return it) and if structure is big and is passed to many methods to avoid copying.

I guess this is not the case here so there is no benefit in using pointer.

cheld:
PTAL

src/app/backend/handler/apihandler.go
```diff
@@ -2096,10 +2096,8 @@ func (apiHandler *APIHandler) handleLogs(request *restful.Request, response *res
 	offsetTo, err2 := strconv.Atoi(request.QueryParameter("offsetTo"))
 	logFilePosition := request.QueryParameter("logFilePosition")
 
-	var logSelector *logs.Selection
-	if err1 != nil || err2 != nil {
-		logSelector = logs.DefaultSelection
-	} else {
+	logSelector := logs.DefaultSelection
+	if err1 == nil || err2 == nil {
```
floreks: ~ (A || B) == (~A) && (~B)

src/app/backend/resource/logs/source.go
```diff
@@ -14,40 +14,37 @@
...
-	} else {
-		return getLogSourcesFromController(k8sClient, ns, resourceName, resourceType)
 	}
+	return getLogSourcesFromController(k8sClient, ns, resourceName, resourceType)
 }
 
-// Returns all containers for a given pod
-func getLogSourcesFromPod(k8sClient *client.Clientset, ns, resourceName string) (generic.LogSources, error) {
-	logSources := generic.LogSources{}
+// GetLogSourcesFromPod returns all containers for a given pod
```
floreks: Private methods do not need to start with method name. This is only required (by go linter) for public methods.

cheld:
this.logSources is loaded but not injected into component.js

Not sure about the problem

PR #2097, opened 2017-07-08
Fix tiny typo on resource limit description for all available languages

kessiler:


cheld:
The typo was also translated - cool :-) 

The englisch file is generated. Please fix also 
https://github.com/kubernetes/dashboard/blob/dbbcb06995321e75f7e67cac83fb8a5f044d1374/src/app/frontend/resourcelimit/detail/detail.html#L37

Thanks for the PR

floreks:
You have to fix HTML file and run `gulp build` it will update our translation files. You should not directly change them only translated strings can be manually updated. Part `xxx` in the `<translation>xxx</translation>` tag.

kessiler:
done @cheld and @floreks

src/app/frontend/resourcelimit/detail/detail.html
```diff
@@ -34,7 +34,7 @@
     </kd-resource-card-header-column>
     <kd-resource-card-header-column size="small"
                                     grow="1">
-      [[Min|]Resource Limits min entry.]]
+      [[Min]Resource Limits min entry.]]
```
maciaszczykm: It should be `[[Min|Resource Limits min entry.]]`

kessiler: done @maciaszczykm

PR #2108, opened 2017-07-12
Update bower dependencies (ui-router migration)

floreks:
`ui-router` has improved a lot since 0.x version and new version offers better handling of states and state transitions. Most of the work is done. I have to check if everything works as intended. Migration was required for my work on log in mechanism.

As a next step we can rewrite all state controllers to components now. There are many new features described in [UI-Router migration guide](https://ui-router.github.io/guide/ng1/migrate-to-1_0).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,23 +2,23 @@
   "name": "kubernetes-dashboard",
   "version": "1.6.1",
   "dependencies": {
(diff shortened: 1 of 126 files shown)


maciaszczykm:
Code looks good, but I need to verify if it works correctly.

PR #2112, opened 2017-07-13
Fix Google Closure Compiler warnings

### LINKED ISSUE #2099: Google Closure Compiler produces lots of warnings
##### Environment
Development

##### Steps to reproduce
gulp serve:prod

##### Observed result
Long list of warnings

##### Comments
It seems that an update of the version of Google Closure Compiler causes this warnings. I did a quick experiment and set the flag module_resolution: 'LEGACY' and it removed all warnings. However, some more time and investigation is needed to solve it properly.

Documentation: https://github.com/google/closure-compiler/wiki/JS-Modules

maciaszczykm:
Left comment in the code, we can solve all the issues later, but at the moment I would go with this solution. Without it even CI is hard to use, because logs are too long.

Closes #2099.

### DIFF
--- a/build/script.js
+++ b/build/script.js
@@ -199,6 +199,11 @@ function compileES6(translation) {
     // ---- JS MODULES ----
     js_module_root: `/${path.relative(conf.paths.base, conf.paths.frontendSrc)}`,
 
+    // Specifies how the compiler locates modules. Default value 'BROWSER' requires all imports to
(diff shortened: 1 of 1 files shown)


floreks:
For now we should do this and refactor the code to use new standard. This is however a time consuming task and travis logs should be clear now.

PR #2113, opened 2017-07-13
Show name of the node that pod in running at

### LINKED ISSUE #1372: Allow to see pods distribution across nodes
#### Enhancement details

It would be useful to be to able to see on which nodes all replicas of the same type are scheduled. 
The pod table currently shows the clusterIP only, if we could add to that column the pod Name (which is available on the pod description page that would help have a quick idea on which node all of our replicas are. 

Eventually having a better distribution dashboard could be nicer where the user could quickly identify if replicas are stacked too much on the same nodes or if replicas are spawning across different DC or Rack Hosts.

maciaszczykm:
image

Closes #1372.

### DIFF
--- a/src/app/backend/resource/pod/list.go
+++ b/src/app/backend/resource/pod/list.go
@@ -46,9 +46,8 @@ type PodStatus struct {
 	ContainerStates []v1.ContainerState `json:"containerStates"`
 }
 
-// Pod is a presentation layer view of Kubernetes Pod resource. This means
-// it is Pod plus additional augmented data we can get from other sources
-// (like services that target it).
+// Pod is a presentation layer view of Kubernetes Pod resource. This means it is Pod plus additional augmented data
+// we can get from other sources (like services that target it).
 type Pod struct {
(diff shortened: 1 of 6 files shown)


src/app/backend/resource/pod/list.go
```diff
@@ -123,6 +126,7 @@ func ToPodList(pods []v1.Pod, events []v1.Event, nonCriticalErrors []error, dsQu
 		warnings := event.GetPodsEventWarnings(events, []v1.Pod{pod})
 		podDetail := ToPod(&pod, metrics, warnings)
 		podDetail.Warnings = warnings
+		podDetail.NodeName = pod.Spec.NodeName
```
floreks: This should be done in `ToPod` method. Also warnings are passed to the function so they can be assigned in there.

maciaszczykm: I know. It is done like this in a few places. Didn't touch it now, as this is not part of pull request, but I can change it if you want.

floreks: For such a small change it's not worth to create another PR. It can be done in this one.

PR #2120, opened 2017-07-17
Aggregate logs

### LINKED ISSUE #2049: Aggregate logs across controllers and services

cheld:
Added view logs link to higher controller objects, like replica set. In the logs view the user can select the pod.

Fixes #2049.

(Created 2nd PR after messing up translations)


cheld:
Injection of logSources is not working with new ui router...

src/app/frontend/logs/stateconfig.js
```diff
@@ -27,16 +27,10 @@ import {stateName} from './state';
  */
 export default function stateConfig($stateProvider) {
   $stateProvider.state(stateName, {
-    url: `${appendDetailParamsToUrl('/log')}/:container`,
-    params: {
-      'container': {
```
cheld: I removed that. Is that ok

src/app/frontend/logs/component.js
```diff
@@ -62,20 +60,29 @@ export class LogsController {
...
+     */
+    this.container;
 
-    /** @export {!backendApi.PodContainerList} - initialized from resolve */
-    this.podContainers;
+    /**
+     * Pods and containers available for selection
+     * @export {!backendApi.LogSources}
+     */
+    this.logSources;
```
cheld: This is not injected, but 'this.podLogs' is set. I don't see any difference. Not sure about the new mechanism...

src/app/frontend/logs/stateconfig.js
floreks: No you have to rewrite that to match `resourceType` param. Logs page won't work correctly without thtat.

src/app/frontend/logs/component.js
floreks: `podLogs` is injected through component definition. Remove `podContainers` from there and add `logSources`.

cheld: Thanks

src/app/frontend/logs/stateconfig.js
cheld: hmmm...container is optional, thus (as far as I understand) default value is null and squash is true. However, resourceType is not optional.

floreks: No. It has changed.

https://ui-router.github.io/guide/ng1/migrate-to-1_0#non-optional-path-parameters

cheld:
PTAL

cheld:
What is the problem with coverage. Any idea?

maciaszczykm:
@cheld Probably some Codecov issue. On our side it seems to be fine: https://travis-ci.org/kubernetes/dashboard/jobs/254430310#L694

floreks:
Master build failed during coverage report creation and there was no coverage to compare against.

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

src/app/backend/handler/apihandler.go
```diff
@@ -61,6 +57,10 @@ import (
 	errorsK8s "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/kubernetes/pkg/client/unversioned/remotecommand"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
```
floreks: Was that formatted by your IDE? Try change it to `goimports` style or run `goimports -w src/app/backend` in dashboard root directory. Previous format is enforced by `goimports`. I have to update `gulp format`.

cheld: @floreks goimports changes about 20 files. Should I change as part of this PR? WDYT?

floreks: Yes. Formatting files is fine.

cheld:
PTAL

src/app/backend/client/manager.go
```diff
@@ -20,7 +20,6 @@ import (
 	"log"
 	"strings"
 
-	"github.com/emicklei/go-restful"
```
floreks: In every package `go-restful` import is removed. Backend is not building now.

cheld: hmm...seems to be a bug if import name has a hyphen (https://github.com/joefitzgerald/go-plus/issues/52)

floreks: Install and use [this](https://godoc.org/golang.org/x/tools/cmd/goimports) `go get golang.org/x/tools/cmd/goimports`. I have used it and it doesn't delete anything.

cheld:
I would suggest to fix goimport in a separate PR, e.g. adding it to gulp format. WDYT?

floreks:
Revert changes and make CI pass. We can format it later.

PR #2130, opened 2017-07-18
Hide double border when there is no footer

### LINKED ISSUE #1893: Card footer's top border should not be shown if footer has no content
Creating this issue to discuss how we can improve filtering UI:

image

Two things come to my mind:

- [x] styling of filtering input
- [ ] footer top border should not be shown if footer has no content

maciaszczykm:
This is workaround for issue described in #1893. Until now when there was no footer bottom border was displayed two times (first list has footer, second does not):

image

After this fix:

image

As you probably noticed it removes bottom border from the card, but it is almost not visible since we are using box shadows for every card and when there are no footer we have one border anyways. To do it in a cleaner way I can add missing border in pagination CSS, but then I would have to turn off option to add custom content to footer (we do not use it anyways). 

Fixes #1893.

### DIFF
--- a/src/app/frontend/common/components/contentcard/contentcard.scss
+++ b/src/app/frontend/common/components/contentcard/contentcard.scss
@@ -16,8 +16,9 @@
 
 .kd-content-card-content {
   background-color: $content-background;
-  border: 1px solid $border;
+  border: solid $border;
   border-radius: 2px;
+  border-width: 1px 1px 0;
   box-shadow: $whiteframe-shadow-1dp;
 }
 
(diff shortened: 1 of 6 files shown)


floreks:
IMHO this change may fix one issue but introduces another. Whether first or second visual issue is more/less prominent is a very subjective opinion. For me both of them are equal. First one never actually bothered me and new one probably wouldn't also.

I don't really have an opinion whether or not we should merge it. One way or another we will end up with some visual issue.

maciaszczykm:
I've replaced workaround with clean solution somehow :-)

maciaszczykm:
image

maciaszczykm:
PTAL

PR #2160, opened 2017-07-21
Improve display of container env variables

### LINKED ISSUE #2153: Secrets in environment variables are displayed in plaintext in the pod details
##### Environment

```
Dashboard version: 1.6.1
Kubernetes version: 1.7.0
```
I am running k8s in GKE, at version 1.7.0 (the latest supported as of writing) using the dashboard exactly as it comes configured out of the box (1.6.1 going by the image described in the pod in kube-system).

##### Steps to reproduce
1) Create a deployment (or just a pod) that uses a secret to populate an environment variable
```yaml
...
spec:
  containers:
    - env:
        name: SOME_NAME
        valueFrom:
          secretKeyRef:
            name: secret-name
            key: key-in-secret
...
```
2) View the pod details page for the pod created in the above step

##### Observed result
In the Environment Variables section, the secret is displayed

##### Expected result
The env var value is redacted in some way (links to the secret, just says "hidden", inlined "eye" toggle like on the secret page, etc)

floreks:
Closes #2153

Secret value will be by default hidden and eye icon button will be shown. Once clicked secret value will be shown and can not be hidden. Additionally I have added correct link to object ref (secret/configmap) and changed a bit overall UI.

We have to improve this view more after the release. Current improvements are not enough.

Before:
zrzut ekranu z 2017-07-21 14-37-49

After:
zrzut ekranu z 2017-07-21 14-34-19
zrzut ekranu z 2017-07-21 14-33-52

### DIFF
--- a/i18n/messages-en.xtb
+++ b/i18n/messages-en.xtb
@@ -602,7 +602,7 @@
   <translation id="3190654996078973226" key="MSG_POD_DETAIL_CONTAINERINFO_0" desc="Subtitle at the top of the container details.">Containers</translation>
   <translation id="1970650270984178962" key="MSG_POD_DETAIL_CONTAINERINFO_1" desc="Label for container image.">Image</translation>
   <translation id="1360847805196042023" key="MSG_POD_DETAIL_CONTAINERINFO_2" desc="Label for container environment variables.">Environment variables</translation>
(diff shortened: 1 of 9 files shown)


zerda:
Maybe we could make config with line breaks in block, but others remain inline. It's more clear to me. @floreks

before
before

after
after

Issue #2180, opened 2017-07-25
Heapster does not work any more out of the box


saschagrunert:
If I am running 1.6.2 the logs give me the output:

```
[restful] 2017/07/25 15:49:13 log.go:26: No metric client provided. Skipping metrics.
[restful] 2017/07/25 15:49:13 log.go:26: No metric client provided. Skipping metrics.
```

And no metrics will be shown on the dashboard pages. I am using K8S 1.7.1 with the following heapster image config:

```
gcr.io/google_containers/heapster-grafana-amd64:v4.4.1
gcr.io/google_containers/heapster-amd64:v1.4.0
gcr.io/google_containers/heapster-influxdb-amd64:v1.1.1
```

Is there any configuration missing within the `yml` file [here](https://git.io/kube-dashboard)?

floreks:
Something had to change on either kubernetes or heapster side because we haven't touched it for quite some time and it just stopped working. The actual reason is listed on backend start: `Could not enable metric client: Health check failed: the server could not find the requested resource (get services heapster). Continuing.`. We'll investigate that and fix ASAP. Probably roll out new release because this functionality is very important.

floreks:
As a workaround for now you can pass `heapster-host` argument to dashboard container in the yaml file. Direct connection still works. There are only problems with using in-cluster configuration.

floreks:
Take a look at #2181. You can use these changes it to make it work for now. It will be pretty much the same as in-cluster config.

saschagrunert:
Alright, thank you!

miramar-labs:
what is the solution exactly?

rbjorklin:
@miramar-labs this is the workaround they mention: https://github.com/coreos/coreos-kubernetes/blob/3e0986915c1e6123c409aab7418656cb83e89ab5/multi-node/generic/controller-install.sh#L822-L830

floreks:
It is not required anymore in `1.6.3` version. In-cluster config works fine again.

rbjorklin:
@floreks I tried this 15 minutes ago with Kubernetes 1.7.4 and Dashboard 1.6.3 and it didn't work without that flag.

floreks:
It usually means networking/cluster issue. You can check dashbord logs. It prints initial heapster health check result.

rbjorklin:
@floreks I'm not sure this is the relevant snippet but here goes:

[...]

PR #2181, opened 2017-07-25
Add workaround to make heapster work with latest release

floreks:
Related to #2180.

This is only a workaround for now. We have to investigate why in-cluster config does not work and heapster service can not be found. This will work the same as in-cluster config as we require `heapster` service to exist in `kube-system` namespace.

As most people is using command from our master readme to deploy dashboard it will work for now. We will do next release with fix ASAP.


miramar-labs:
adding ```- --heapster-host=http://heapster.kube-system:80``` didn't make any difference for me .. what's the solution?

PR #2182, opened 2017-07-25
Create an Overview page for namespaces

cnwalker:
# Create an Overview page for namespaces
screen shot 2017-07-24 at 12 08 09 pm

Often when looking at the main dashboard page, users on my team want to see all of the objects associated with a particular namespace. **Currently it is only possible to see objects in the groups "Workloads", "Discovery and Load Balancing", and "Config and Storage".** This PR implements a button on the left sidebar that shows all the resources under those fields at the same time. This is my first pull request on the dashboard project and I am eager for suggestions and feedback! Please let me know if this diff is too big and I can submit the backend and frontend PRs separately.

## Backend
A `/overview/` endpoint has been added to the api. The endpoint returns all of the objects included in "Workloads", "Discovery and Load Balancing", and "Config and Storage" for a given namespace.

A new endpoint was created to be in keeping with the convention for "Workloads", "Discovery" etc. which all have their own endpoints. The `/search/` endpoint does something similar to `/overview/`, but it is not namespace specific.

## Frontend
The frontend component is implemented using the `workloads` component as a guide. When allobjects is clicked, it hits the `/overview/` endpoint and pulls the desired resources.

Edit: I renamed the branch and accidentally closed the old PR. Sorry for any inconvenience @bryk @maciaszczykm @floreks.


src/app/backend/search/search.go
```diff
@@ -70,6 +62,18 @@ type SearchResult struct {
 	PodList                   pod.PodList                  `json:"podList"`
 	DaemonSetList             daemonset.DaemonSetList      `json:"daemonSetList"`
 	StatefulSetList           statefulset.StatefulSetList  `json:"statefulSetList"`
+}
+
+// SearchResult is a list of resources matching search criteria found in whole cluster.
+type SearchResult struct {
+	ResourcesInNamespace
```
maciaszczykm: I think we should keep `json:...` annotation as it's most common pattern for us.

@floreks Opinion?

src/app/backend/overview/overview.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/discovery"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/workload"
+	"github.com/kubernetes/dashboard/src/app/backend/search"
+	"k8s.io/client-go/kubernetes"
+)
+
+// OverviewObjectList is a list of objects present in a given namespace
+type OverviewObjectList struct {
+	search.ResourcesInNamespace
```
maciaszczykm: Check comment with annotations.

src/app/externs/backendapi.js
```diff
@@ -144,7 +144,24 @@ backendApi.ReplicationControllerList;
  *   podList: !backendApi.PodList,
  *   daemonSetList: !backendApi.DaemonSetList,
  *   statefulSetList: !backendApi.StatefulSetList,
- *   errors: !Array<!backendApi.Error>
```
maciaszczykm: I think it should be here.

cnwalker:
Ready for another review @maciaszczykm :)

maciaszczykm:
Hmm, most probably after updating `json:...` comments you did not align the code to it:

zrzut ekranu 2017-07-26 o 13 12 32
zrzut ekranu 2017-07-26 o 13 10 38
zrzut ekranu 2017-07-26 o 13 09 49

cnwalker:
It looks like adding the annotations wraps the response in a `ResourcesInNamespace` object rather than inheriting the fields as it does now. 

screen shot 2017-07-26 at 10 19 16 am with annotation

vs

screen shot 2017-07-26 at 10 20 32 am without annotation

I put the annotations in comments, is that a good compromise?

src/app/backend/overview/overview.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/kubernetes/dashboard/src/app/backend/resource/workload"
+	"k8s.io/client-go/kubernetes"
+)
+
+// OverviewObjectList is a list of objects present in a given namespace
+type OverviewObjectList struct {
+	// Inherits the fields from the config, discovery, and workloads objects
+	config.Config
+	discovery.Discovery
+	workload.Workloads
```
cnwalker: Hey so revisiting this, I realized there was a way to do this without the `ResourcesInNamespace` object. Instead we can use struct inheritance to directly inherit the fields from config, discovery, and workloads. I'm not sure adding a type annotation would be semantic here, because this syntax is causing the struct to inherit many different types, rather than declaring new types.

maciaszczykm:
> I put the annotations in comments, is that a good compromise?

@cnwalker Check http://godoc.org/labix.org/v2/mgo/bson#Marshal, `inline` flag is what you are looking for.

src/app/backend/overview/overview.go
maciaszczykm: Sounds good to me. Check if you can use `inline` flag to prepare annotations for these fields.

cnwalker:
looks like `inline` flag worked! Ready for another review @maciaszczykm

maciaszczykm:
@cnwalker Try to switch to `All namespaces` while on `Overview` page:

image

src/app/backend/handler/apihandler.go
```diff
@@ -557,6 +558,11 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
 			To(apiHandler.handleLogs).
 			Writes(logs.LogDetails{}))
 
+	apiV1Ws.Route(
+		apiV1Ws.GET("/overview/{namespace}").
```
maciaszczykm: Missing endpoint without namespace param causes 404 error. You need to add it and test like I said.

cnwalker: Tested, should be fixed now!

cnwalker:
I think it's ready for another review once travis tests finish @maciaszczykm

PR #2206, opened 2017-07-28
Add log in mechanism

### LINKED ISSUE #2093: Dashboard log in mechanism
We have had some initial discussion about that with @bryk. I have continued my investigation and came up with an idea how we could provide some log in mechanism in dashboard. 

Here is a sketch of how the data flow would look like (not very detailed though). First let's explain the general flow.

1. For user to log in we would need to handle few options. We would use structure similar to [this](https://github.com/kubernetes/client-go/blob/master/tools/auth/clientauth.go#L76) and at log in screen show this log in options: kubeconfig file, certificates, username & password, token.
2. Based on given data we can fill in this structure, encrypt it and save in a cookie.
3. Backend would get this cookie, decrypt structure and use it to create client.
4. Rest works similar to what we do now with authorization header. API server checks if user that "created client" is authorized to execute operation and returns data/not authorized error.

Note: We need to investigate if there is a way to check if user "exists" first.

zrzut ekranu z 2017-07-06 16-01-15

Few things to consider:
1. We need some symmetric encryption alghorithm such as AES (128/256 bit key) as we need to send whole structure to the backend. Hashing will not work for us.
2. Key needs to be stored outside of our backend (secret/config map) in a secure way. Is it possible to give dashboard access only to single resource? Not whole resource type. If not then we could create separate namespace with configmap/secret just to store the key somewhere. 
3. Only dashboard/cluster admin should be able to see the key. Encrypted data is only stored in user browser as a cookie.
4. Do we need to think of a way to secure cookie from being copied to another computer and used there? We could add some expired flag and roll out new encryption key every 30/60/x min? Is it secure enough if we use HTTPS protocol to encrypt traffic between client and server? We can't force users to expose dashboard through HTTPS only but they have to be aware that they are open for MITM attacks if they use plain HTTP. 

Our backend would remain stateless and encrypted "session" would be stored in users' browser as a cookie. Are there any other problems to consider?

@bryk @rf232 @maciaszczykm @cheld @liggitt WDYT?

floreks:
## Overview
Closes #2093.

This PR adds basic login functionality to dashboard. There are still many things to improve but after this initial version will be merged we will be able to split the work. It is based on [JSON Web Encryption](https://tools.ietf.org/html/rfc7516) (JWE) standard.

### Data flow
Dashboard will generate encryption key during the start and it will be stored in a secret accessible by dashboard (not yet implemented).

1. User opens dashboard
1. Token is not present in a cookie and authorization header is not passed to dashboard
1. User fills in token field and clicks on login button
1. LoginSpec is sent to the backend (we'll make it work only over HTTPS)
1. Backend gets encryption key from a secret.
1. Token is taken and basic health check is made to see if it is recognized by apiserver.
1. Auth structure is encrypted and sent back to the frontend.
1. Frontend stores encrypted token in a cookie and user is redirected to the default state.
1. Cookie is passed to the backend on every request, decrypted there and auth info is used to create client.

This is the basic flow. In case authorization header is sent to the dashboard it will override our log in mechanism and will be used instead.

## Changes
 - Added angular cookies.
 - Added login page and made it independent of chrome state.
 - Added new transition/request hooks to redirect user to login page in case he is not logged in.
 - Added support for token based log in.
 - Adapted nav bar to be able to correctly show/hide it during initial page load. Before that nav component settings were overriding other components so I have moved logic to the nav service.
 - Added `go-jose` as a dependency for managing JWE tokens.

## Current limitations
 - Only token based login is supported
 - Encryption key is generated on backend. Backend crash means that stored token will be invalid
 - Because of previous issue dashboard can not be scaled

## Improvements
List of things to improve will be tracked from #2207.

## TODO
List of things to be done before the merge.

 - [x] Add JSDoc
 - [x] Add documentation
 - [x] Remove hardcoded strings (state, token name, etc.)
 - [x] Refactor the code
 - [x] Add tests
 - [x] Test on serve:prod
 
## UX
We'll rework it after everything is stable. For now it's very simple. There was also basic login option but I have disabled it for now due to https://github.com/kubernetes/kubernetes/issues/49709. 

zrzut ekranu z 2017-07-28 12-56-16

zrzut ekranu z 2017-07-28 12-56-44

Calling @kubernetes/dashboard-maintainers and @liggitt


src/app/frontend/common/auth/service.js
```diff
@@ -0,0 +1,156 @@
...
+  logIn(loginSpec) {
+    let deferred = this.q_.defer();
+
+    this.tokenPromise_.then(
+        token => {
+          let resource = this.resource_('api/v1/login', {}, {
+            save: {
+              method: 'POST',
+              headers: {
+                'X-CSRF-TOKEN': token,
```
liggitt: what are you setting the CSRF token header to here?

liggitt: and where is it being checked server-side?

liggitt:
> Dashboard will generate encryption key during the start and it will be stored in a secret accessible by dashboard.

and

> Encryption key is generated on backend. Backend crash means that stored token will be invalid
> Because of previous issue dashboard can not be scaled

seem to contradict each other

src/app/backend/auth/api/types.go
```diff
@@ -0,0 +1,54 @@
...
+type LoginSpec struct {
+	// Username is the username for basic authentication to the kubernetes cluster.
+	Username string `json:"username"`
+	// Password is the password for basic authentication to the kubernetes cluster.
+	Password string `json:"password"`
+
+	// ClientCert contains PEM-encoded data from a client cert file for TLS. Overrides ClientCertificate
+	ClientCert []byte `json:"clientCert"`
+	// ClientKey contains PEM-encoded data from a client key file for TLS. Overrides ClientKey
+	ClientKey []byte `json:"clientKey"`
```
liggitt: don't include these... having the user send private key material via a cookie is not something the dashboard should enable

src/app/frontend/common/auth/service.js
floreks: `this.tokenPromise_.then` calls our backend that generates and returns token for login action request. Once it is returned from our backend I am passing it again during login request and it is verified on server side.

src/app/backend/auth/jwt/manager.go
```diff
@@ -0,0 +1,88 @@
...
+	"crypto/rsa"
+
+	authApi "github.com/kubernetes/dashboard/src/app/backend/auth/api"
+	"gopkg.in/square/go-jose.v2"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// TODO(floreks): Should be retrieved from a secret
+var tokenSigningKey *rsa.PrivateKey
```
liggitt: would recommend not making this a package var... jwtTokenManager should include the signing/verifying key as an instance var

floreks: I'm planning to store it in a secret and retrieve during backend start. This is just temporary solution to check if everything works.

src/app/backend/client/manager.go
```diff
@@ -188,20 +236,35 @@ func (self *clientManager) buildAuthInfoFromConfig(cfg *rest.Config) api.AuthInf
...
 	if req == nil {
-		log.Print("No request provided. Skipping authorization header")
-		return api.AuthInfo{}
+		log.Print("No request provided. Skipping authorization.")
+		return nil, nil
 	}
 
 	authHeader := req.HeaderParameter("Authorization")
-	token := ""
+	jweToken := req.HeaderParameter("kdToken")
```
liggitt: what is `kdToken`?

src/app/backend/auth/api/types.go
floreks: You're saying that we shouldn't provide option to log in using certificates? That would leave us only token and basic auth.

liggitt:
with credentials sent in a cookie, doesn't the dashboard need csrf protection on all requests, not just login?

src/app/frontend/common/auth/interceptor.js
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export class AuthInterceptor {
+  constructor($cookies) {
+    this.request = config => {
+      // TODO: Better filtering for urls that do not require this token
+      if (config.url.startsWith('api/v1')) {
```
liggitt: can you describe what is going on here, both in terms of the `api/v1` check and the moving of a cookie value to a header value

src/app/backend/auth/jwt/manager.go
liggitt: it still should be in an instance, not a singleton to enable rotation, testing, etc

src/app/backend/auth/api/types.go
liggitt: logging in via certificates should not involve the private key leaving the user's machine

liggitt:
cc @kubernetes/sig-auth-pr-reviews

floreks:
> > Dashboard will generate encryption key during the start and it will be stored in a secret accessible by dashboard.
> and
>
> >Encryption key is generated on backend. Backend crash means that stored token will be invalid
Because of previous issue dashboard can not be scaled
> seem to contradict each other

In first I meant that it will do that but it is not yet implemented. Second one states the current behavior. #2207 lists all the things that need to be done.

> with credentials sent in a cookie, doesn't the dashboard need csrf protection on all requests, not just login?

I'm not sure but this is probably due to my incomplete knowledge in this matter. JWE cookie is sent by the frontend to the backend only in request header. Response (other than login) from the server does not contain it. 

If you embed some request to the dashboard on other page what exactly happens? Is code from our frontend executed, cookie retrieved, added to header and sent to the backend or is it more like a curl?

src/app/frontend/common/auth/interceptor.js
floreks: I will add documentation this is still very early.

This code intercepts all the requests to our backend (done with `$resource('api/v1...')`) and adds our encrypted token to it. Token is then decrypted on backend side and based on auth information stored in it apiserver client is created.

src/app/backend/client/manager.go
floreks: It is encrypted token returned by our backend during login request.

src/app/frontend/common/auth/service.js
liggitt: ah, too many vars named `token`. rename `tokenPromise_` to `csrfTokenPromise_`, and rename `token` here to `csrfToken` for clarity.

src/app/backend/auth/api/types.go
floreks: I understand. How it is done in kubectl then? I mean if I have in my kubeconfig `certKey` and `certData` set how is it used? Client and `AuthInfo` structure have to be created based on it.

src/app/backend/auth/jwt/manager.go
floreks: Definitely. I agree. There will be some manager to handle token rotation, refreshing, etc. I wasn't planning on leaving it like that. This is only to be able to test it locally.

src/app/frontend/common/auth/service.js
floreks: Refactor is on my "to be done" list of things . I will rename it and add proper documentation.

src/app/backend/auth/api/types.go
liggitt: kubectl uses the client cert and private key to establish the TLS connection. The server verifies the client has possession of the private key during the TLS handshake. That is something the browser would have to do, since the cert is used at the transport layer, it's not something you would do in client application logic.

ericchiang:
> LoginSpec is sent to the backend (we'll make it work only over HTTPS)

What does "we'll make it work only over HTTPS" mean? It seems like a lot of this proposal is trying to create a secure channel over HTTP to encrypt the token. If you already have HTTPS, why do you need this specialized encryption scheme?

floreks:
> What does "we'll make it work only over HTTPS" mean? It seems like a lot of this proposal is trying to create a secure channel over HTTP to encrypt the token. If you already have HTTPS, why do you need this specialized encryption scheme?

Login request that goes to the backend contains plain auth data (token, username/password). I'm taking this data encrypting it in JWE token payload and sending back to the browser where it is stored in a cookie. 

Making login work only over HTTPS is done to prevent MITM attacks right? I'm not sure if sending user credentials/token over plain HTTP would be a good idea.

ericchiang:
> Making login work only over HTTPS is done to prevent MITM attacks right?

It's used to encrypt traffic too.

> I'm not sure if sending user credentials/token over plain HTTP would be a good idea.

Sending encrypted credentials over HTTP isn't much better. It's trivial to sniff the user's cookie and use the encrypted token to impersonate the user to the backend.

src/app/backend/auth/api/types.go
floreks: Makes sense. Thanks for explanation. This leaves us with token and basic auth only for now.

floreks:
> Sending encrypted credentials over HTTP isn't much better. It's trivial to sniff the user's cookie and use the encrypted token to impersonate the user to the backend.

That is why I want to enable login option only if dashboard is using HTTPS (sorry if I wasn't clear about that).  Cookie is encrypted in order to make sure that real auth data is not compromised (easily). Additionally there will be a token expiration and encryption key rotation mechanism added later on.

If you have any ideas how to improve it further then feel free to comment. I have some basic understanding of this topic but some things are still a bit vague to me. That is why I appreciate any reviews and comments that would help to make it better.

ericchiang:
In my opinion, getting the dashboard to use HTTPS and have CSRF mitigations would be a more impactful change. Store the credentials in a cookie plain text, then harden later with added encryption. This JWE scheme is a nice added layer, but it's not extremely meaningful without ensuring those first.

floreks:
Dashboard already supports HTTPS so it's just the matter of making sure that it is enabled before enabling login option.

As for the CSRF tokens, we are already using them. Is it required to secure all the requests (even GET) after user is "logged in"? I'm not familiar with what information exactly can be exposed during this attack. For sure attacker can force you to execute some request to dashboard but can he sniff the request in order to hijack the cookie? Isn't request made only from my browser to dashboard server without exposing it to third party page?

ericchiang:
> As for the CSRF tokens, we are already using them. Is it required to secure all the requests (even GET) after user is "logged in"? I'm not familiar with what information exactly can be exposed during this attack. For sure attacker can force you to execute some request to dashboard but can he sniff the request in order to hijack the cookie? Isn't request made only from my browser to dashboard server without exposing it to third party page?

It's required to secure all mutating requests. CSRFs doesn't actually expose information to the attacker, it can only cause the attacker to be able to make the user perform an arbitrary request, in inspect the response.

floreks:
All our GET requests are idempotent and all POST/DELETE requests are already using csrf tokens.  Response requests do not contain our encrypted token (beside login which also requires csrf token). 

I guess only thing is HTTPS that we have to deal with here. Anything else to remember?

ericchiang:
> I guess only thing is HTTPS that we have to deal with here. Anything else to remember?

Thanks. No immediate concerns come to mind.

i18n/messages-ja.xtb
```diff
@@ -1,929 +1,936 @@
...
-  <translation id="961242340990264399" key="MSG_ABOUT_ABOUT_0" desc="This is a label for button displayed on about page used to open documentaion in browser">ドキュメント</translation>
-  <translation id="5634646746226609179" key="MSG_ABOUT_ABOUT_1" desc="Label \'Dashboard Version\' for the about page.">Dashboard バージョン</translation>
-  <translation id="4718120324314749596" key="MSG_ABOUT_ABOUT_2" desc="Label \'Git Commit\' for the about page.">Gitコミット</translation>
-  <translation id="5788105450172886273" key="MSG_ABOUT_ABOUT_3" desc="This is a label for button displayed on about page used to send feedback to GitHub new issue.">フィードバックを送信</translation>
+  <translation id="961242340990264399" key="MSG_ABOUT_ABOUT_0" desc="This is a label for button displayed on about page used to open documentaion in browser">??????????????????</translation>
+  <translation id="5634646746226609179" key="MSG_ABOUT_ABOUT_1" desc="Label \'Dashboard Version\' for the about page.">Dashboard ???????????????</translation>
+  <translation id="4718120324314749596" key="MSG_ABOUT_ABOUT_2" desc="Label \'Git Commit\' for the about page.">Git????????????</translation>
```
maciaszczykm: It's broken.

src/app/backend/auth/jwe/manager.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Decrypt provided token and return AuthInfo structure saved in a token payload.
+func (self *jweTokenManager) Decrypt(jweToken string) (*api.AuthInfo, error) {
+	jweTokenObject, err := self.validate(jweToken)
+	if err != nil {
+		return nil, err
+	}
+
+	decrypted, err := jweTokenObject.Decrypt(self.tokenEncryptionKey)
+	// TODO(floreks): Check for decryption error and handle it
```
maciaszczykm: ?

src/app/backend/auth/manager.go
```diff
@@ -0,0 +1,84 @@
...
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// Implements AuthManager interface
+type authManager struct {
+	tokenManager  authApi.TokenManager
+	clientManager client.ClientManager
+}
+
+// Implements auth manager. See AuthManager interface for more information.
```
maciaszczykm: Should start with `Login`.

src/app/backend/client/manager.go
```diff
@@ -188,20 +204,57 @@ func (self *clientManager) buildAuthInfoFromConfig(cfg *rest.Config) api.AuthInf
 	}
 }
 
-// Extracts authentication information from request header
-func (self *clientManager) extractAuthInfo(req *restful.Request) api.AuthInfo {
+// Based on auth info and rest config creates client cmd config.
+func (self *clientManager) buildCmdConfig(authInfo *api.AuthInfo, cfg *rest.Config) clientcmd.ClientConfig {
```
maciaszczykm: To be consistent it's good to start these comments with method names too.

src/app/backend/validation/validateloginstatus.go
```diff
@@ -0,0 +1,44 @@
...
+	// True when token header indicating logged in user is found in request.
+	TokenPresent bool `json:"tokenPresent"`
+	// True when authorization header indicating logged in user is found in request.
+	HeaderPresent bool `json:"headerPresent"`
+	// True if dashboard is configured to use HTTPS connection. It is required for secure
+	// data exchange during login operation.
+	HTTPSMode bool `json:"httpsMode"`
+}
+
+// ValidateLoginStatus TODO
```
maciaszczykm: Review your TODOs :)

src/app/backend/auth/jwe/manager.go
floreks: I have left todos in the code to remember where improvements should be put.

src/app/backend/validation/validateloginstatus.go
maciaszczykm: And keep only these marked in separate issue.

src/app/backend/client/manager.go
floreks: To be consistent I should remove comment because most of our private functions are not commented :smile:. If we don't have to add method name I wouldn't do that as it doesn't improve documentation and usually makes it harder to read.

src/app/frontend/common/errorhandling/errors.js
```diff
@@ -24,4 +24,7 @@ export const kdErrors = {
   /** @export {string} @desc Text shown on in error dialog when there is no namespace provided selected in both dashboard and yaml file. */
   MSG_DEPLOY_EMPTY_NAMESPACE_ERROR: goog.getMsg(
       'Dashboard and your file do not specify any namespace to deploy a resource. Please select a specific namespace in dashboard or add one in yaml file.'),
+
+  /** @export {string} @desc Text shown when unauthorized user tries to log in. */
+  MSG_LOGIN_UNAUTHORIZED_ERROR: goog.getMsg('Authentication Failed. Please try again.'),
```
maciaszczykm: s/Failed/failed

floreks:
>  we should hide search and menu

This is already done. I haven't updated screenshots in first post.

ericchiang:
How did this merge? There are giant TODOs, unimplemented code paths, and form what I can see none of this is feature flagged.

floreks:
In order to split the work we've had to merge initial PR. There is a tracking issue with list of improvements that have to be done before release. Our HEAD images are for testing only. It was already very big PR and we do not want to review such big changes. Master in any way is not meant to be used on real cluster.

lxlxok:
Hey @floreks, Thanks for this great feature. I am very interested in enhancing this feature, but I can't use gulp serve to debug this log in feature? Is there any instruction about access control in development mode with https?

PR #2219, opened 2017-08-01
update hyperkube info

zouyee:
update hyperkube info

### DIFF
--- a/build/hyperkube.sh
+++ b/build/hyperkube.sh
@@ -14,32 +14,32 @@
 # limitations under the License.
 
 # Starts local hypercube cluster in a Docker container.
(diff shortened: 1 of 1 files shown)


cheld:
1. Hyperkube is deprecated. Updates are not accepted. 1.5.4 is the latest supported version.

2. You removed nsenter. This was introduced as a work around of 'shared mount propagation' problems in Docker. Maybe this problem is fixed by now. But it has to be checked carefully. (nsenter also causes problems, e.g. socat is not working. But the problems are not relevant for Dashboard)

Overall, I would not change anything, except replacing with kubeadm or kubeadm-dind-cluster.

zouyee:
Got it.

PR #2220, opened 2017-08-01
Add compatibility matrix to readme

### LINKED ISSUE #355: Document which versions of K8s masters are compatible
Per #349, the UI does not work on 1.0.x masters.

maciaszczykm:
Closes #355.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ itself.
 
 |                     | Kubernetes 1.4 | Kubernetes 1.5 | Kubernetes 1.6 | Kubernetes 1.7 |
 |---------------------|----------------|----------------|----------------|----------------|
-| Dashboard 1.4.x     | ✓              | ✕              | ✕              | ✕              |
-| Dashboard 1.5.x     | ✕              | ✓              | ✕              | ✕              |
-| Dashboard 1.6.x     | ✕              | ✕              | ✓              | ?              |
+| Dashboard 1.4       | ✓              | ✕              | ✕              | ✕              |
+| Dashboard 1.5       | ✕              | ✓              | ✕              | ✕              |
+| Dashboard 1.6       | ✕              | ✕              | ✓              | ?              |
 | Dashboard HEAD      | ✕              | ✕              | ?              | ✓              |
 
 - `✓` All the functionalities are supported with given version.
--- a/README.md
+++ b/README.md
@@ -59,8 +59,8 @@ If the username and password is configured but unknown to you, then use `kubectl
 | **Dashboard 1.6**   | ✕              | ✕              | ✓              | ?              |
 | **Dashboard HEAD**  | ✕              | ✕              | ?              | ✓              |
 
-- `✓` Supported version range.
-- `?` Some of the functionalities might not be working.
+- `✓` Fully supported version range.
+- `?` Due to breaking changes between Kubernetes API versions, some features might not work in Dashboard (logs, search etc.).
 - `✕` Not supported version range.
 
 ## Documentation
--- a/README.md
+++ b/README.md
@@ -12,19 +12,6 @@ itself.
 
 ![Dashboard UI workloads page](docs/dashboard-ui.png)
 
-## Compatibility matrix
-
-|                     | Kubernetes 1.4 | Kubernetes 1.5 | Kubernetes 1.6 | Kubernetes 1.7 |
-|---------------------|----------------|----------------|----------------|----------------|
-| **Dashboard 1.4**   | ✓              | ✕              | ✕              | ✕              |
-| **Dashboard 1.5**   | ✕              | ✓              | ✕              | ✕              |
-| **Dashboard 1.6**   | ✕              | ✕              | ✓              | ?              |
-| **Dashboard HEAD**  | ✕              | ✕              | ?              | ✓              |
-
-- `✓` All the functionalities are supported with given version.
-- `?` Some of the functionalities might not be working.
-- `✕` Most probably will not work with given version.
-
 ## Deployment
 It is likely that the Dashboard is already installed on your cluster. Check with the following command:
 ```shell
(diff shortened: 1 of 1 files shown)


cheld:
I like the matrix, but overall the document feels strange, now: First we describe a lot of dashboard versions in the matrix and then nothing is mentioned in the deployment section. I think both sections should fit together.

We could add something like this:
```
$ kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
```
or maybe just state that Dashboard 1.6 works only with Kubernetes 1.6 and put the matrix to the bottom

README.md
```diff
@@ -12,6 +12,19 @@ itself.
...
+
+|                     | Kubernetes 1.4 | Kubernetes 1.5 | Kubernetes 1.6 | Kubernetes 1.7 |
+|---------------------|----------------|----------------|----------------|----------------|
+| **Dashboard 1.4**   | ✓              | ✕              | ✕              | ✕              |
+| **Dashboard 1.5**   | ✕              | ✓              | ✕              | ✕              |
+| **Dashboard 1.6**   | ✕              | ✕              | ✓              | ?              |
+| **Dashboard HEAD**  | ✕              | ✕              | ?              | ✓              |
+
+- `✓` All the functionalities are supported with given version.
+- `?` Some of the functionalities might not be working.
```
cheld: Is there a big benefit for the destiction between '?' and 'x'. For me it seems to be more or less the same. Maybe '?' should be defined as "should work but untested"?

cheld:
I think it is ok if we update the deployment section a bit and use concrete version numbers here,too.

If you are done, I would like to ask Gerti for a review

README.md
floreks: The main difference is that first declares that it is supported (by us) version of kubernetes core and we guarantee that it will work and if anything does not work then we will fix it.

`?` on the other hand states that due to version differences some functionalities might not work correctly in dashboard and we will not fix that.

cheld: 'x' known not to work. bugs will not be fixed
'?' maybe working. bugs will not be fixed

floreks: `?` dashboard is known to be working but some functionalities might not work

cheld:
@gertipoppel it would be nice if you can have a look

gertipoppel:
I would like to suggest the following: 
1. Move the matrix below the "Alternative Usage" section
2. Remove the ? from the matrix and just state "not tested". 
3. Remove the "? Some of the ... " sentence altogether
4. Rephrase "All the functionalities are supported with given version." to "The entire functionality is supported."
6. Rephrase x to "Not supported". 

The current wording suggests that the Dashboard has not been tested and we do not know whether some version works or not. This makes an unprofessional impression.

maciaszczykm:
@gertipoppel Thanks for your input. Done almost all besides 2. and 3., because I am not convinced here. It was partly tested and most of the features are working, but not all of them (i.e. logs). I think we should rephrase `Some of the functionalities might not be working.` to something better and keep `?` in the table. Check https://github.com/kubernetes/client-go#compatibility-matrix as it is Kubernetes project and in my opinion we can follow it.

> The current wording suggests that the Dashboard has not been tested and we do not know whether some version works or not. This makes an unprofessional impression.

It contradicts:

> Remove the ? from the matrix and just state "not tested".

gertipoppel:
@maciaszczykm Thanks for reworking the readme. What do you think of the following: 
Add an introductory sentence like "The available versions of Dashboard are compatible with the following releases of Kubernetes:"

    ✓ Fully supported.
    ? Kubernetes has features the Dashboard can't use (logs, etc).
    ✕ Not supported.

floreks:
```
✓ Fully supported. 
? Kubernetes has features the Dashboard can't use (logs, etc).
✕ Not supported.
```

1 and 3 are ok but I'd still add `... version range` at the end as their versioning system is `1.x.y` to indicate that full version range is supported.

As for the second option this is not correct. It's more like: "Due to breaking changes between kubernetes versions API some features will not work in Dashboard (logs, search, etc.)"

gertipoppel:
What about then: 
"Due to breaking changes between Kubernetes API versions, some features will not work in Dashboard (logs, search, etc.)".

floreks:
It's closer to the truth. I have used `will` but it suggests that we are certain something will not work. Unfortunately we are not and we do not have enough people to test it properly. Because of that I'd change `will` to `might not work`. Dashboard will work but some features might not.

gertipoppel:
OK for me. Thanks

maciaszczykm:
Thanks for comments.

PTAL

README.md
```diff
@@ -23,7 +23,7 @@ If it is missing, you can install the latest stable release by running the follo
 $ kubectl create -f https://git.io/kube-dashboard
 ```
 
-If you are using Kubernetes 1.5 or earlier, you can install the latest stable release by running the following command:
+If you are using the Dashboard on cluster without authorization setup, you can install the latest stable release by running the following command:
```
gertipoppel: Should be: 
If you are using Dashboard on a cluster ...

README.md
```diff
@@ -50,6 +50,19 @@ Please note that this works only if the apiserver is set up to allow authenticat
...
+|                     | Kubernetes 1.4 | Kubernetes 1.5 | Kubernetes 1.6 | Kubernetes 1.7 |
+|---------------------|----------------|----------------|----------------|----------------|
+| **Dashboard 1.4**   | ✓              | ✕              | ✕              | ✕              |
+| **Dashboard 1.5**   | ✕              | ✓              | ✕              | ✕              |
+| **Dashboard 1.6**   | ✕              | ✕              | ✓              | ?              |
+| **Dashboard HEAD**  | ✕              | ✕              | ?              | ✓              |
+
+- `✓` Fully supported version range.
+- `?` Due to breaking changes between Kubernetes API versions, some features might not work in Dashboard (logs, search etc.).
+- `✕` Not supported version range.
```
gertipoppel: If you include the version range, it should read "Unsupported version range"

README.md
floreks: I'd also change `without authorization setup` to `without enabled authorization` or `with disabled authorization`.

gertipoppel: OK for me

maciaszczykm:
PTAL

Issue #1849, opened 2017-04-14
What happened to the cluster-service annotation?


colemickens:
https://github.com/kubernetes/dashboard/blob/b98d167dadaafb665a28091d1e975cf74eb31c94/src/deploy/kubernetes-dashboard.yaml

Why do the yamls in this repo not feature the `cluster-service` tag that causes it to appear in `kubectl cluster-info` ?

From some dashboard yaml that we imported ages ago and have been updating, for example, we had these:
```
   labels:
     kubernetes.io/cluster-service: "true"
     k8s-app: kubernetes-dashboard
    version: v1.5.1
```

cheld:
hmmm...I don't remember the full problem out of my head. The behavior you describe ist just one part. The other part is the addon manager. It uses this annotation to synchronizes the cluster state with static manifest files. The behavior was something like this:
- addon manager reads a yaml from disk -> deploys the contents
- addon manager reads all deployments from api server with annotation cluster-service:true -> deletes all that do not exist as files

As a result, if you add this annotation, addon manager will remove dashboard after a minute or so. 

So, 
- dashboard is deployed after cluster creation -> annotation should not be set:
https://github.com/kubernetes/dashboard/blob/b98d167dadaafb665a28091d1e975cf74eb31c94/src/deploy/kubernetes-dashboard.yaml

- dashboard is deployed part of cluster creation -> annotation should be set:
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dashboard/dashboard-controller.yaml

At least this was the behavior some time ago. I think kubeadm does not use addon-manager. But it is still part of kube-up script.

colemickens:
Thanks for the very comprehensive answer. It looks like kubeadm doesn't use addon-manager. It deploys kube-dns by itself and users are expected to deploy dashboard (for now).

We're deploying dashboard with addon-manager, so we'll still need those. Thanks again for clarifying with such detail!

cheld:
I have not tested it yet, but it seems that addon manager has been refactored:

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/kube-addons.sh

I think we can use `kubernetes.io/cluster-service: "true"` by now

maciaszczykm:
@cheld Addon manager is deprecating this label:

https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/addon-manager#addon-manager

However, cluster-info depends on it, so perhaps we can use it.

[...]

PR #2227, opened 2017-08-02
Update kubernetes-dashboard.yaml

4406arthur:
show with  kubectl cluster-info

### DIFF
--- a/src/deploy/kubernetes-dashboard.yaml
+++ b/src/deploy/kubernetes-dashboard.yaml
@@ -86,6 +86,7 @@ apiVersion: v1
 metadata:
   labels:
     k8s-app: kubernetes-dashboard
+    kubernetes.io/cluster-service: "true"
   name: kubernetes-dashboard
   namespace: kube-system
 spec:


cheld:
I think we cannot merge this because of the described behavior: https://github.com/kubernetes/dashboard/issues/1849

But maybe the behavior changed. Can you check?

4406arthur:
@cheld , Label kubernetes.io/cluster-service=true is deprecated (only for Addon Manager). In future release (after one year), Addon Manager may not respect it anymore. Addons have this label but without addonmanager.kubernetes.io/mode=EnsureExists will be treated as "reconcile class addons" for now. Ref from https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/README.md

floreks:
So if I understand correctly addons without addon manager specific label and with `kubernetes.io/cluster-service=true` are still treated as "reconcile class addons". Last change of this file is from february so I believe this is still the case and people with addon manager enabled would have a problem with dashboard.

PR #2237, opened 2017-08-04
Add browsers support section to readme

maciaszczykm:
Open to discussion here. Do we need it? Are versions correct?


maciaszczykm:
Why was it closed? I remember we agreed to introduce something like this in the past.

floreks:
Closed unintentionally during cleaning up branches on upstream repo.

PR #2243, opened 2017-08-07
Add encryption key storing mechanism

floreks:
~I have to refactor the code and cleanup some things but it already shows the basic idea and works fine from what I've tested.~ It might need some improvements when key rotation mechanism will be added.

## Changes
 - Added secret synchronizer that is mainly responsible for watching on a single resource changes.
 - Added encryption key import/export option in order to be able to save it in a secret. Thanks to https://stackoverflow.com/a/44688503.
 - Adapted token manager to use key holder which uses secret synchronizer in order to import/export encryption key.
 - Added `Overwatch` that is started automatically with backend and is responsible for starting and watching over every registered synchronizer. Every synchronizer with restart policy `Always` will be restarted in case of error.

## TODO
 - [x] Refactor code
 - [x] Add documentation
 - [x] Add tests

Backend will always generate encryption key on start and try to push it to a secret. After it is created synchronizer basically watches for secret updates. On every update, handler registered by key holder is fired and encryption key is updated. As all backends will always be watching for the same resource, key from last event will be used.

In case of decryption error we can always force single update from secret and try once again.

```
2017/08/21 13:48:08 Using kubeconfig file: /home/floreks/.kube/config
2017/08/21 13:48:08 Skipping in-cluster config
2017/08/21 13:48:08 Using random key for csrf signing
2017/08/21 13:48:08 No request provided. Skipping authorization
2017/08/21 13:48:08 Starting overwatch
2017/08/21 13:48:08 Successful initial request to the apiserver, version: v1.7.2
2017/08/21 13:48:08 New synchronizer has been registered: kubernetes-dashboard-key-holder-kube-system. Starting
2017/08/21 13:48:08 Starting secret synchronizer for kubernetes-dashboard-key-holder in namespace kube-system
2017/08/21 13:48:08 Initializing secret synchronizer synchronously using secret kubernetes-dashboard-key-holder from namespace kube-system
2017/08/21 13:48:08 Initializing JWE encryption key from synchronized object
2017/08/21 13:48:08 Trying to update secret with same object. Skipping
2017/08/21 13:48:08 Creating remote Heapster client for http://192.168.30.243:32113
2017/08/21 13:48:08 Successful initial request to heapster
2017/08/21 13:48:08 Serving insecurely on HTTP port: 9090
2017/08/21 13:48:08 Serving securely on HTTPS port: 8443
```


zouyee:
I think it is helpful which has finished something you want.[certificate-manager](https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/certificate)

floreks:
It will be useful for key rotation and token refreshing mechanism. This PR is just for synchronization between multiple replicas. Thanks anyway.

src/app/backend/auth/api/types.go
```diff
@@ -16,6 +16,11 @@ package api
 
 import "k8s.io/client-go/tools/clientcmd/api"
 
+const (
+	EncryptionKeyHolderName      = "kubernetes-dashboard-key-holder"
+	EncryptionKeyHolderNamespace = "kube-system"
```
floreks: It doesn't really matter as long as dashboard will have access to this secret. It can access it from any namespace. Our yaml files will contain correct permissions. If user decides to change them then it's his responsibility to make sure they are valid.

I'd only think how to configure multiple dashboard deployments to use different secrets (not replicas of single deployment). We might add some argument to dashboard as suffix that will be used for secret creation. Autogeneration and sync between multiple replicas might be hard.

floreks:
PTAL

redbaron:
I stumbled upon this, and became curious why this cookie encryption scheme is needed when HTTPS is already a requirement? This is not to undermine all the work done there, but more for my education.

floreks:
@redbaron This layer of protection were added to increase security as we do not have "typical" log in routine. Normally when you log in using credentials server responds with some token. Even if token gets in possession of attacker, original credentials used to log in will not be compromised and usually token can be revoked (blacklisted).

In our case we can only proxy your credentials (token, username/password) to API server and it decides whether or not grant access. This forces us to save actual credentials in a token. JWT token would not be safe as it does not encrypt sensitive data, so I have decided to use JWE token where payload is encrypted. Encryption key is generated by backend on start and stored in a secret. It is then used to encrypt/decrypt user credentials and JWE token is saved in your browser's cookie storage. In case attacker gets in possession of our token he will not be able to get (easily) in possession of actual credentials you used to log in.

PR #2246, opened 2017-08-09
Use the client-go deployment method instead of the kubectl deployment

liuan1986:
Use the client-go deployment method instead of the kubectl deployment method, because the “k8s.io/kubernetes/pkg/kubectl” depend on too many kunernetes packages.
So that dashboard and kubernetes are tightly coupled .
In this way, dashboard  and kubernetes  can be loosely coupled, and the next step we can remove most of  the kubernetes packages.

### DIFF
--- a/src/app/backend/resource/deployment/deploy.go
+++ b/src/app/backend/resource/deployment/deploy.go
@@ -22,10 +22,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	client "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


floreks:
This is good. We've been waiting for such functionality in `client-go` to migrate it. As this is last or one of the last things that forces us to keep kubernetes core dependencies.

I'll try to check it later if it works the same as before. Also you need to sign the CLA first and make build pass.

maciaszczykm:
This is great!

maciaszczykm:
Could you also run `dep prune` after these changes?

maciaszczykm:
Your last commit is not signed by email that is recognizable by CNCF, you must fix that.

liuan1986:
Yes I forget to set my email, and how to fix that?

floreks:
Amend commit and push again.

maciaszczykm:
You should rebase.

liuan1986:
This request has moved to [https://github.com/kubernetes/dashboard/pull/2252](url)

PR #2249, opened 2017-08-09
Support for User Links add through annotations for Pods and Services

kenan435:
I added support for user custom links that can be added as annotations to Pods and Services. 
For example: absolute vs relative links
alpha.dashboard.kubernetes.io/links: |
 {
     "debug":"{{apiserver-proxy-url}}/debug",   // relative path using the apiserver proxy urls
     "request":"http://www.someurl.com/debug/requests"   // absolute path since its includes http://
}

This pr is just covers two use cases and we plan on introducing more in subsequent pr's, especially after we get the global settings feature merged so that we can replace placeholder vars with global settings var values at run-time.

We also want to extend flexibility by constructing URLs at run-time based on:
- pod dns names
- labels?
- etc


cheld:
From UX perspective, I don't like URL `{{apiserver-proxy-url}}/debug`. IMHO it is not fully obvious what kind of URL it should be. Maybe better alternatives:
`http://{{apiserver-proxy-url}}/debug`
`http://{{proxy-url}}/debug`
`http://{{pod-proxy-url}}/debug` and `http://{{service-proxy-url}}/debug`

cheld:
This feature request is related:

https://github.com/kubernetes/dashboard/issues/2332

maciaszczykm:
It seems that translation files are corrupted. You should revert them. I will take a closer look later.

src/app/backend/userlinks/userlinks.go
```diff
@@ -0,0 +1,142 @@
...
+// getPodLinks get userlinks for links
+func getPodLinks(client k8sClient.Interface, namespace, name, host string) ([]UserLink, error) {
+	userLinks := []UserLink{}
+	pod, err := client.CoreV1().Pods(namespace).Get(name, metaV1.GetOptions{})
+	if err != nil {
+		return userLinks, err
+	}
+	if len(pod.Annotations[annotationObj]) == 0 {
+		return userLinks, err
+	}
```
maciaszczykm: Merge both ifs to `err != nil || len(pod.Annotations[annotationObj]) == 0 `.

src/app/backend/userlinks/userlinks.go
```diff
@@ -0,0 +1,142 @@
...
+// getServiceLinks get userlinks for services
+func getServiceLinks(client k8sClient.Interface, namespace, name, host string) ([]UserLink, error) {
+	userLinks := []UserLink{}
+	service, err := client.CoreV1().Services(namespace).Get(name, metaV1.GetOptions{})
+	if err != nil {
+		return userLinks, err
+	}
+	if len(service.Annotations[annotationObj]) == 0 {
+		return userLinks, err
+	}
```
maciaszczykm: Merge both ifs to err != nil || len(pod.Annotations[annotationObj]) == 0.

maciaszczykm: `getServiceLinks` and `getPodLinks` have a lot in common. Perhaps you could avoid such code duplication?

src/app/backend/userlinks/userlinks.go
```diff
@@ -0,0 +1,142 @@
...
+
+	for key, val := range m {
+		userLink := new(UserLink)
+		userLink.Description = strings.ToUpper(key)
+		//use api server proxy url
+		if strings.Contains(val, apiserverProxyURL) {
+			//strip the url of apiserver-proxy-url
+			userLink.Link = host + "/api/" + "v1" + "/namespaces/" + pod.ObjectMeta.Namespace +
```
maciaszczykm: `userLink.Link = host + "/api/v1/namespaces/" + pod.ObjectMeta.Namespace +`

src/app/backend/handler/apihandler.go
```diff
@@ -1391,10 +1400,15 @@ func (apiHandler *APIHandler) handleGetPodDetail(request *restful.Request, respo
...
+	config, err := apiHandler.cManager.Config(nil)
+	if err != nil {
+		handleInternalError(response, err)
+		return
+	}
 
 	namespace := request.PathParameter("namespace")
 	name := request.PathParameter("pod")
-	result, err := pod.GetPodDetail(k8sClient, apiHandler.iManager.Metric().Client(), namespace, name)
+	result, err := pod.GetPodDetail(k8sClient, apiHandler.iManager.Metric().Client(), namespace, name, config.Host)
```
maciaszczykm: Were you able to test it not on a localhost?

kenan435: not sure I understand the question... Are you asking if I was not able to test it on localhost?

src/app/backend/userlinks/userlinks.go
kenan435: I tried but they differ enough so that it makes it not worth refactoring further.

src/app/frontend/pod/detail/info.html
```diff
@@ -32,4 +31,10 @@
       {{::$ctrl.pod.podIP}}
     </kd-info-card-entry>
   </kd-info-card-section>
+  <kd-info-card-section name="[[User links|Subtitle 'User links' at the top of the column about user links attached in resource objects annotations, (right) at the pod detail view.]]"
```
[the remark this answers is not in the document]
kenan435: As of now only these two resources.

src/app/backend/handler/apihandler.go
maciaszczykm: I meant setup different than localhost, i.e. AWS, GCE etc.

src/app/frontend/common/components/endpoint/endpoint.scss
```diff
@@ -17,4 +17,5 @@
   height: inherit;
   margin: 0;
   vertical-align: middle;
+  padding: 5px;
```
maciaszczykm: Use `0.5 * $baseline-grid` as we always try to be dependent of this variable.

src/app/backend/handler/apihandler.go
kenan435: Tested running master on a vagrant vm. Dashboard did generated correct links.

kenan435:
PTAO I tested it on an env. other then localhost. If it looks good I am going to increase the test coverage.

maciaszczykm:
One more thing to test is an in-cluster config. Could you check it on Docker image?

floreks:
Revert broken i18n files and rebase.

src/app/backend/handler/apihandler.go
cheld: I think it will be difficult to retrive the correct IP that works on-premise and in the cloud. Actually the node has an external ip field (as far as I remember it is set in GKE), however in many cases it is probably not set (local cluster). 


```
 kubectl get node kube-master -o wide
NAME          STATUS    AGE       VERSION   EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION
kube-master   Ready     11d       v1.7.3    <none>        Ubuntu 16.04.3 LTS   4.4.0-93-generic

```

So, I think it is better not to use it and rather cut of the host in the front end

src/app/frontend/service/detail/controller.js
```diff
@@ -37,4 +41,35 @@ export class ServiceDetailController {
...
+   * default way of deploying/running dashboard is in a container we concluded that, for the time
+   * being, we will parse the host:port address from the current browser window used to access the
+   * k8's api server and inject it into the userlink proxy address so that its accessible from the
+   * browser running dashboard.
+   *
+   * @param {!backendApi.ServiceDetail} serviceDetail
+   * @param {!angular.$location} $location
+   * @return {!backendApi.ServiceDetail} serviceDetail
+   */
+  checkUserlinksForProxyURL(serviceDetail, $location) {
```
cheld: This function will always create links - however the links are only correct if Dashboard is accessed through proxy. A maybe better alternative might be not to render the user links at all in case Dashboard is not accessed through proxy
However, as we consider this featrue alpha anyway it might be ok for now

src/app/externs/backendapi.js
```diff
@@ -874,6 +875,16 @@ backendApi.RoleList;
 
 /**
  * @typedef {{
+ *  description: string,
+ *  link: string,
+ *  isURLValid: string,
+ *  isProxyURL: string
```
cheld: why not boolean type?

src/app/backend/userlinks/userlinks.go
```diff
@@ -0,0 +1,146 @@
...
+	for key, uri := range m {
+		userLink := new(UserLink)
+		userLink.Description = key
+		if strings.Contains(uri, apiserverProxyURL) {
+			userLink.Link = host + "/api/v1/namespaces/" + pod.ObjectMeta.Namespace +
+				"/pods/" + pod.ObjectMeta.Name + "/proxy" + strings.TrimPrefix(uri, apiserverProxyURL)
+			userLink.IsURLValid = true
+			userLink.IsProxyURL = true
+		} else if strings.Contains(uri, podDNSName) {
+			//10-192-3-8.default.pod.cluster.localhttp://{{pod.dns_name}}:9090/debug
```
cheld: is this an intended comment? Seems like a good idea, but add some extra words and spaces

src/app/backend/userlinks/userlinks.go
```diff
@@ -0,0 +1,146 @@
...
+	"strings"
+
+	"github.com/kubernetes/dashboard/src/app/backend/api"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sClient "k8s.io/client-go/kubernetes"
+)
+
+const (
+	annotationObj     = "alpha.dashboard.kubernetes.io/links"
+	apiserverProxyURL = "http://{{apiserver-proxy-url}}"
```
cheld: we should remove the 'http' prefix from the pattern. we do not know if http or https is used. 

comment: removing http might makes the pattern less recognizable as URL, but easier to handle.

The protocol prefix seems to be cut off in the code, anyway

src/app/backend/userlinks/userlinks.go
```diff
@@ -0,0 +1,146 @@
...
+	err = json.Unmarshal([]byte(service.Annotations[annotationObj]), &m)
+	if err != nil {
+		return userLinks, err
+	}
+
+	for key, uri := range m {
+		userLink := new(UserLink)
+		userLink.Description = key
+		if strings.Contains(uri, apiserverProxyURL) {
+			userLink.Link = host + "/api/v1/namespaces/" + service.ObjectMeta.Namespace +
```
cheld: in my understanding the host is replaced later on in the frontend. So could we use a static placeholder, like localhost?

cheld:
One more comment: a malformed annotation is accepted, but will cause non-graceful error during rendering.

cheld:
Small bug: list order is changed after every refresh. Other than that it works with the known limitation. Anyway, not visible to end users until we document. So we could fix in subsequent PRs

Yaml for testing:
```
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: hello
spec:
  replicas: 1
  template:
    metadata:
      annotations:
        alpha.dashboard.kubernetes.io/links: |
          {
              "debug": "https://blog.golang.org/error-handling-and-go",   
              "proxy": "http://{{apiserver-proxy-url}}/debug",
          }        
    spec:
      containers:
        - name: helloo
          image: "gcr.io/google-samples/hello-go-gke:1.0"
          ports:
            - name: http
              containerPort: 80
```

kmova:
This is a great feature to have and will provide a solution for managing the openebs storage from within kubernetes dashboard. 

openebs - is a cotainerized storage for containers. so a openebs PV is actually served by a set of pods - that expose an iscsi target that does sync writes to multiple replicas (distributed on different nodes). The replicas will write the data to the local storage attached to the nodes. 

For openebs, when a PV is launched, I am planning to annotate it with the associated OpenEBS pods and services that expose the iSCSI block storage. 

This will help the user start with a pod (application) -- click on the volumes associated with it (PV) and then also drill further down to the storage.

cheld:
@kmova 

thanks for sharing your use case. I am not sure if the implementation matches your needs. Currently, we have only an annotation on pod and services. One could use it as follows: a devops team deploys a microservice and dashboard automatically renders links to all backend pages like status or metrics. 

Could you elaborate on which resources you would like to see links to which target addresses?

kmova:
@cheld 

I was hoping to get the links added to the PV. The links on this PV will be:
- dashboard links to pods and services within the same cluster - that are related to serving the PV (hyper converged storage) 
- prometheus / grafana - that can provide detailed metrics on the storage (till we get the PV metrics integrated into metrics-server)

The links can be added to the PV, by the dynamic volume provisioners which create PV objects.

PR #2252, opened 2017-08-10
Use the client-go deployment method instead of the kubectl deployment…

liuan1986:
Use the client-go deployment method instead of the kubectl deployment method, because the “k8s.io/kubernetes/pkg/kubectl” depend on too many kunernetes packages.
So that dashboard and kubernetes are tightly coupled .
In this way, dashboard and kubernetes can be loosely coupled, and the next step we can remove most of the kubernetes packages.
Remove  unnecessary packages by running  dep prune .
This is the previous  pull request ：
[https://github.com/kubernetes/dashboard/pull/2246](url)


maciaszczykm:
Could you actually deploy from files on `gulp serve:prod`? I see this info while using it:

image

And it works from `kubectl create -f ...`.

maciaszczykm:
Deploy without file worked fine.

liuan1986:
It's really a problem and I need to take some time to fix it.

maciaszczykm:
@liuan1986 Sure, take your time.

maciaszczykm:
Just a tip - you can use `gulp format` if you want to format files :)

maciaszczykm:
Ping us when it will be ready for a review.

liuan1986:
Thanks for your tip，and It's ready for review.

maciaszczykm:
Works on `gulp serve:prod` with one minor issue. It does not redirect from deploy page after success.

Got to test it on Docker image.

maciaszczykm:
> Works on gulp serve:prod with one minor issue. It does not redirect from deploy page after success.

@liuan1986 Can you solve it?

src/app/backend/resource/deployment/deploy.go
```diff
@@ -285,55 +290,64 @@ func getLabelsMap(labels []Label) map[string]string {
...
-	validate := spec.Validate
+func DeployAppFromFile(cfg *rest.Config, spec *AppDeploymentFromFileSpec) (bool, error) {
+	reader := strings.NewReader(spec.Content)
+	log.Printf("Namespace for deploy from file: %s\n", spec.Namespace)
+	d := yaml.NewYAMLOrJSONDecoder(reader, 4096)
+	for {
+		data := unstructured.Unstructured{}
+		if err := d.Decode(&data); err != nil {
+			if err == io.EOF {
+				return true, nil
```
maciaszczykm: If I get it correctly `false` is returned in case of error and `true` if deployment was succesful. Why it is `true` in this case?

liuan1986: because data may be a list ,so it need read by  for cycle,when it read at the end of file, it will throw io.EOF, That's just what I expected and I need return true .this idea Refer to kubernetes cli

maciaszczykm: What about redirect?

liuan1986: when i test ,it  redirect to workload page ,and i did not changed frontend code. I don't understand why this happened

maciaszczykm: Deploy from file on `gulp serve:prod`?

maciaszczykm: Okay, it's fine. Redirect happens but to previous state which is always deploy app.

maciaszczykm:
image

It should not happen. Previously we had check if no namespace is selected in file and selector (`All namespaces`). Custom message should be displayed.

src/app/backend/resource/deployment/deploy.go
```diff
@@ -285,55 +290,64 @@ func getLabelsMap(labels []Label) map[string]string {
...
-		NamespaceParam(spec.Namespace).
-		Stream(reader, spec.Name).
-		Flatten()
+		discoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
+		if err != nil {
+			return false, err
+		}
 
-	if strings.Compare(spec.Namespace, "_all") != 0 {
-		builder.RequireNamespace()
```
maciaszczykm: It has to be used to display custom message.

liuan1986:
I've fixed the problem and my way of dealing with it is as follows:
1. request set namespace=default, and deployment file set default ,Deploy success (to default namespace)
2. request set namespace=default, deployment file did not set ,Deploy success (to default namespace)
3. request set namespace=_all, deployment file set default, Deploy success (to default namespace) 
4. request set namespace=_all, deployment file did not set, Deploy  failed (an empty namespace may not be set during creation)
5. request set namespace=kube-public ,and deployment file set default, Deploy failed (the namespace of the provided object does not match the namespace sent on the request)

liuan1986:
It's my pleasure, and I'd be happy to make a contribution to the project.
About the reditect to overview page question, I tested it,and There is a real problem with the discovery.
Maybe my partner and I can try to fix it, if need be.
We are only too glad to do such a thing.

src/app/backend/resource/deployment/deploy.go
```diff
@@ -285,55 +290,68 @@ func getLabelsMap(labels []Label) map[string]string {
...
-	err = errors.LocalizeError(err)
-	return deployedResourcesCount > 0, err
+		if strings.Compare(spec.Namespace, "_all") == 0 {
+			_, err = dynamicClient.Resource(resource, data.GetNamespace()).Create(&data)
+		} else {
+			_, err = dynamicClient.Resource(resource, spec.Namespace).Create(&data)
+		}
+
+		if err != nil {
+			return false, err
```
floreks: `return false, errors.LocalizeError(err)` here as it was done before. We need that to correctly localize error in frontend.

floreks:
1 small comment. Beside that everything looks great. This is a big step forward, thanks!

floreks:
It should be ready for merge. Could you just rerun gulp format?

maciaszczykm:
Great!

PR #2262, opened 2017-08-15
Fix scale RC RS StatefulSet resource error

riverzhang:
Fix scale RC RS StatefulSet resource error
screenshot from 2017-08-15 12-45-00

### DIFF
--- a/src/app/backend/scaling/scale.go
+++ b/src/app/backend/scaling/scale.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/kubernetes/dashboard/src/app/backend/api"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	client "k8s.io/client-go/kubernetes"
 )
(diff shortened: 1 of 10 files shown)


riverzhang:
@maciaszczykm

src/app/backend/scaling/scale.go
```diff
@@ -46,9 +47,10 @@ func GetScaleSpec(client client.Interface, kind, namespace, name string) (rc *Re
 // method since the client scale method does not provide one for the job.
 func ScaleResource(client client.Interface, kind, namespace, name, count string) (rc *ReplicaCounts, err error) {
 	rc = new(ReplicaCounts)
-	if strings.ToLower(kind) == "job" {
+        kind = strings.Replace(kind, " ", "", -1)
```
floreks: This is not a proper fix. You should fix it in frontend. Only changing hardcoded strings is ok.

floreks:
There are couple of fixes I've had in mind.

1. Requires major refactoring and splitting `resourceKindName` into display name and name used to call backend.
2. In [here](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/scaling/controller.js#L105) you can transform display string to proper format. Replace whitespaces and use to lower method.

riverzhang:
@floreks   At first I thought it was better to modify the background ,because the front end needed to separate the ReplicationController spelling.
I was the first to submit the method, Are you think there is a problem?

floreks:
Endpoint for scaling is `/scale/{kind}/{namespace}/{name}/` so kind has to be in correct format already. We shouldn't fix that on backend side.

src/app/frontend/replicationcontroller/detail/actionbar.html
```diff
@@ -18,7 +18,7 @@
                    object-meta="$ctrl.details.objectMeta">
 </kd-actionbar-logs>
 
-<kd-scale-button resource-kind-name="[[Replication Controller|Label 'Replication Controller' which appears at the top of the scale dialog, opened from a replication controller details page.]]"
+<kd-scale-button resource-kind-name="[[ReplicationController|Label 'Replication Controller' which appears at the top of the scale dialog, opened from a replication controller details page.]]"
```
maciaszczykm: It is not a solution.

maciaszczykm: @kenan435 Is this value used in calls to backend? It cannot be done like that, because what if somebody uses Japanese translations?

floreks:
I can take that as I know what needs to be fixed.

PR #2273, opened 2017-08-16
Support Cron Jobs

### LINKED ISSUE #1562: Support cronjobs
Would be great if Dashboard would support cronjobs in the same way it supports jobs (cronjobs being a level above jobs).

maciaszczykm:
It is still in alpha, so let's wait until it will be fully functional before going forward with it.

As https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/ says:

> You need a working Kubernetes cluster at version >= 1.4 (for ScheduledJob), >= 1.5 (for CronJob), with batch/v2alpha1 API turned on by passing `--runtime-config=batch/v2alpha1=true` while bringing up the API server (see [Turn on or off an API version for your cluster for more](https://kubernetes.io/docs/admin/cluster-management/#turn-on-or-off-an-api-version-for-your-cluster)).

Closes #1562.


zouyee:
So I think we could using `client.Discovery` to make resource added dynamically.

zouyee:
in kubernetes 1.8, The CronJob object is now enabled by default at v1beta1

maciaszczykm:
Yes, after deps update I will finish implementation.

maciaszczykm:
I have taken a quick look at it and we have to wait until client-go will release a new version, that is aligned with 1.8 core repo.  It makes no sense to jump on the master branch now because it can cause a lot of conflicts.

PR #2280, opened 2017-08-17
Fix RC's podstatus

riverzhang:
Fix RC's podstatus


maciaszczykm:
Why did you change Chinese and Japanese translations? You said, that you don't know Japanese.

riverzhang:
I can modify the Chinese translation, Japanese unchanged.

maciaszczykm:
@riverzhang I think you can leave translations are they are and modify only thing, that you actually fix.

maciaszczykm:
image

It should be: `{{::$ctrl.replicationController.podInfo.current}} 作成済`

riverzhang:
@maciaszczykm

PR #2283, opened 2017-08-17
Do not use CreatedByAnnotation anymore

### LINKED ISSUE #2261: Dashboard should not use CreatedByAnnotation
Based on this [announcement](https://groups.google.com/forum/#!msg/kubernetes-dev/juMOsdaCml0/FwVNJA9uBAAJ), `CreatedByAnnotation` will be deprecated in 1.8 in favor of `ControllerRef`. However, the annotation is still used in this repo: [here](https://github.com/kubernetes/dashboard/blob/master/src/app/backend/resource/pod/detail.go#L131) and [here](https://github.com/kubernetes/dashboard/blob/master/src/app/backend/resource/pod/detail.go#L204)

There is a necessary change to use `ControllerRef`. Does anyone have bandwidth to work on this?

maciaszczykm:
Closes #2261.

### DIFF
--- a/src/app/backend/resource/common/controller.go
+++ b/src/app/backend/resource/common/controller.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Copied from: https://github.com/kubernetes/kubernetes/blob/master/staging/src/
+// k8s.io/apimachinery/pkg/apis/meta/v1/controller_ref.go
+
 package common
 
 import "k8s.io/apimachinery/pkg/apis/meta/v1"
--- a/src/app/backend/resource/common/controller.go
+++ b/src/app/backend/resource/common/controller.go
@@ -1,16 +1,19 @@
-// Copyright 2017 The Kubernetes Dashboard Authors.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


src/app/backend/resource/pod/detail.go
```diff
@@ -165,14 +158,8 @@ func GetPodDetail(client kubernetes.Interface, metricClient metricapi.MetricClie
...
 
-func getPodCreator(client kubernetes.Interface, creatorAnnotation string, nsQuery *common.NamespaceQuery,
-	pod *v1.Pod) (*controller.ResourceOwner, error) {
-
-	var serializedReference v1.SerializedReference
-	err := json.Unmarshal([]byte(creatorAnnotation), &serializedReference)
-	if err != nil {
-		return nil, err
-	}
+func getPodCreator(client kubernetes.Interface, nsQuery *common.NamespaceQuery, pod *v1.Pod) (
```
crimsonfaith91: Will `getPodController` be a better name? The controller that the function returns may not be the creator (the parent may change).

src/app/backend/resource/common/controller.go
```diff
@@ -0,0 +1,36 @@
...
+func IsControlledBy(obj v1.Object, owner v1.Object) bool {
+	ref := GetControllerOf(obj)
+	if ref == nil {
+		return false
+	}
+	return ref.UID == owner.GetUID()
+}
+
+// GetControllerOf returns a pointer to a copy of the controllerRef if controllee has a controller.
+func GetControllerOf(controllee v1.Object) *v1.OwnerReference {
```
crimsonfaith91: It seems that dashboard repo removes main repo (`/kubernetes`) dependencies a few hours ago. Otherwise, we are able to reuse the function declared in the main repo. Any reason for the dependency removal?

maciaszczykm: Since `client-go` is stable and has almost all functionalities we tend to avoid using `kubernetes` dependencies directly, as seems to be "cleaner" way. Few weeks ago our repository `vendor` directory had like 60-65 MB, now it's around 20 MB. Sadly, we have to copy some functions from core from time to time. We try to replace them with calls to available API when it becomes possible. At the moment we have 2-3 places where we copied core methods, I think it is not that bad result.

src/app/backend/resource/pod/detail.go
maciaszczykm: Good point. Solved.

src/app/backend/resource/common/controller.go
crimsonfaith91: Thanks for the help! :)

PR #2303, opened 2017-08-23
Update dependencies

floreks:
Closes #2253 and closes #2244.

Updated manually `google-closure-compiler` and `webpack-stream` to latest versions.

## Important changes
 - Module resolution `LEGACY` has been removed from `google-closure-compiler`. I have changed it to `NODE` and fixed all errors and warnings.
 - Updated ECMAScript preset to `env` which contains all features from ES2015 together with new features from ES2016 and ES2017. We'll be slowly migrating to use new [async and await](http://babeljs.io/docs/plugins/transform-async-to-generator/) syntax and more. 
 - Updated closure compiler input language to `ECMASCRIPT_2017`.
 - Updated `webpack` configuration to match webpack 3.
 - Updated minor versions of other dependencies, i.e. angular to `1.6.6`
 - Reenabled compiler check `analyzerChecks`.

### **NOTE: All our imports have to start with `/`, `./` or `../` now, however I'd prefer to use options 2 and 3.**

I have left one new option disabled for now `new_type_inf` as it requires many more changes and is more strict when it comes to type checks.


floreks:
As a small kickoff I have refactored one method to use new async/await syntax: [createnamespace_controller.js](https://github.com/kubernetes/dashboard/pull/2303/files#diff-cbcbbd14394ec868e3ec5fd21a2ea7a9L107)

PR #2307, opened 2017-08-26
Added support for filtering nodes by labels as well as by name. This …

amigold:
…changes the filter logic to be OR instead of AND


floreks:
One thing that worries me is that filtering by multiple fields at once might be misleading for users. What I mean is that usually you get some advanced filter options where you can select fields you want to filter by or apply some predefined filters.

I think that UI should also reflect this change and give user a chance to choose fields he wants to filter by. Currently we only filter by name and that is what popup says, however supporting more fields should be carefully reflected in the UI.

amigold:
@floreks I agree however it appears that making it more generic involves a fairly large refactoring both in the UI (perhaps a filter text box in every column?) code and also in the server side code. Should probably be discussed.

Currently I can change the popup to reflect that the filter works on labels as well. I believe this change is more urgently needed than a full refactoring since nodes often have useless names.

floreks:
I'm not sure if as user I'd want to be forced to filter by more than 1 field by default without option to change it. Let's wait for others to share their opinion.

@maciaszczykm @cheld

maciaszczykm:
This would have to be implemented for all resources. Otherwise, we are not consistent. Also, files are not formatted.

> For sure it would be quite big change, however I am not sure if as user I'd want to be forced to filter by more than 1 field by default without option to change it. Let's wait for others to share their opinion.

Yes, this is not good. 

@danielromlein Any idea how we can handle UX here?

maciaszczykm:
Closing as stale.

Issue #2297, opened 2017-08-22
[Feature Request] Traditional Chinese translation support


kairen:
Hi all, Have any plans for Chinese traditional translation?

maciaszczykm:
@kairen Are you willing to provide it? I am asking because within our team nobody knows it, so we can rely only on external contributions. If you are willing to help, take a look at https://github.com/kubernetes/dashboard/blob/master/docs/devel/translations.md.

kairen:
@maciaszczykm Yes, I willing to provide this support. The work is in progress, if work is done, I will provide a preview.

maciaszczykm:
@kairen Amazing :+1:

PR #2308, opened 2017-08-26
Add Support for Traditional Chinese translation 

kairen:
Feature request from #2297.

The commit added support for Traditional Chinese translation(zh-TW), I have completed most of the translation. And this feature is already testing in Chrome, Firefox, Safari browser.The preview show as follow link:
https://drive.google.com/drive/folders/0B75zg6bt8JyKSTA3RUViRTJGUjQ?usp=sharing

Additional, in order to parsing Traditional Chinese from `Accept-Language` header, I modified the `localehandler.go` to mapping full tag name. 

cc @maciaszczykm.

### DIFF
--- a/src/app/backend/handler/localehandler.go
+++ b/src/app/backend/handler/localehandler.go
@@ -96,8 +96,7 @@ func (handler *LocaleHandler) determineLocalizedDir(locale string) string {
 	for _, tag := range tags {
 		matchedLocale := ""
 		for _, l := range handler.SupportedLocales {
-			base, _ := tag.Base()
-			if l == base.String() {
+			if l == tag.String() {
 				matchedLocale = l
 				break
 			}
--- a/src/app/backend/handler/localehandler.go
+++ b/src/app/backend/handler/localehandler.go
@@ -19,6 +19,8 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strings"
+	"sort"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


src/app/backend/handler/localehandler.go
```diff
@@ -96,8 +96,7 @@ func (handler *LocaleHandler) determineLocalizedDir(locale string) string {
 	for _, tag := range tags {
 		matchedLocale := ""
 		for _, l := range handler.SupportedLocales {
-			base, _ := tag.Base()
-			if l == base.String() {
+			if l == tag.String() {
```
floreks: This comparison is case sensitive and according to http://www.metamodpro.com/browser-language-codes code for Chinese (Taiwan) is `zh-tw`. I'd prefer to keep it lower case just to be safe. 

Additionally without this change  if someone had browser locale set to `zh-xx` then it would match our `zh` category. With this change if i.e. someone will have `zh-hk` then it will fallback to english. I'd extend this:

1. Check for exact match
2. If exact match does not work then check base
3. If both checks fail fallback to english

floreks:
Great change. Thank you! Small comment from me.

src/app/backend/handler/localehandler.go
kairen: @floreks  Thanks you for your reply! These suggestions are very helpful to me, I can add this extension in the next commit that to make sure the comparison is safe.

kairen:
This request has moved to #2309

PR #2309, opened 2017-08-27
Add support for Traditional Chinese translation

kairen:
Feature request from #2297.
Fixed #2308. 

The commit added support for Traditional Chinese translation(zh-TW), I have completed most of the translation. And this feature is already testing in Chrome, Firefox, Safari browser.The preview show as follow link:
https://drive.google.com/drive/folders/0B75zg6bt8JyKSTA3RUViRTJGUjQ?usp=sharing

Additional, in order to parsing Traditional Chinese from `Accept-Language` header, I modified the `localehandler.go` to mapping full tag name. 

/cc @maciaszczykm.

### DIFF
--- a/src/app/backend/handler/localehandler.go
+++ b/src/app/backend/handler/localehandler.go
@@ -41,7 +41,7 @@ type Translation struct {
 // LocaleHandler serves different localized versions of the frontend application
 // based on the Accept-Language header.
 type LocaleHandler struct {
-	SupportedLocales []string
+	SupportedLocales []language.Tag
 }
 
 // CreateLocaleHandler loads the localization configuration and constructs a LocaleHandler.
@@ -49,16 +49,16 @@ func CreateLocaleHandler() *LocaleHandler {
 	locales, err := getSupportedLocales("./locale_conf.json")
 	if err != nil {
 		glog.Warningf("Error when loading the localization configuration. Dashboard will not be localized. %s", err)
-		locales = []string{}
+		locales = []language.Tag{}
 	}
 	return &LocaleHandler{SupportedLocales: locales}
 }
 
-func getSupportedLocales(configFile string) ([]string, error) {
+func getSupportedLocales(configFile string) ([]language.Tag, error) {
 	// read config file
 	localesFile, err := ioutil.ReadFile(configFile)
 	if err != nil {
-		return []string{}, err
+		return []language.Tag{}, err
 	}
 
 	// unmarshall
@@ -69,9 +69,9 @@ func getSupportedLocales(configFile string) ([]string, error) {
 	}
 
 	// filter locale keys
-	result := []string{}
+	result := []language.Tag{}
 	for _, translation := range localization.Translations {
-		result = append(result, translation.Key)
+		result = append(result, language.Make(translation.Key))
 	}
 	return result, nil
 }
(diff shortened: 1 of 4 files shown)


src/app/backend/handler/localehandler.go
```diff
@@ -110,6 +110,16 @@ func (handler *LocaleHandler) determineLocalizedDir(locale string) string {
 	return defaultDir
 }
 
+func isMatchedLocale(locale string, tag language.Tag) bool {
+	base, _ := tag.Base()
+	if locale != strings.ToLower(tag.String()) {
+		if locale != base.String() {
```
floreks: To check base I'd extract actual base from locale as we don't know if locale is `zh` or `zh-tw`. You can use `strings.Split(locale, "-")[0]`. For both `zh` and `zh-tw` it would return `zh` as base.

kairen: @floreks The following code is use `language. NewMatcher` to handle locale:
https://github.com/kairen/dashboard/commit/8787a35448df6c17913ab3f3fcbbba5020ea6b94

kairen: But this way need to modify `localehandler_test.go` too.

floreks: I'm not fully sure how NewMatcher actually works. We need behavior such as:

Given supported languages `en, zh-tw, zh-hk, zh`:
 - when required language is `en` return `en`
 - when required language is `zh` return `zh`
 - when required language is `zh-hk` return `zh-hk`
 - when required language is `zh-cn` return `zh`

1. For any required language that is a regional dialect try to find exact match. In case not found: 
   1. return base language if supported
   1. otherwise return default language

kairen: @floreks I implemented a simple code to check the locale. In the code, You can see that the matcher returns a match tag and confidence value. If the confidence value is `Exact`, It is the best locale. If the confidence value is not `Exact`, we could check the base language, otherwise return default language.

The code as below:
https://gist.github.com/kairen/5ed330dd203f23bca8a29b224a55cf1c

floreks: I have checked it. Sounds good to me.

kairen: @floreks The following is a test branch, I have completed the above function changes.
https://github.com/kairen/dashboard/commit/a67118b1e78f239bc3c1b87db511c5f06fc17c29

The test results are as follows:
[Backend test result](https://gist.github.com/kairen/98b578524c2e47dddccbae82c04f4291)

floreks: Looks good! 

PS. You can work on branch you have used to create PR. There is no need to use other branch.

kairen:
@floreks I updated the new changes on this branch.

src/app/backend/handler/localehandler.go
kairen: @floreks I have updated to the current branch, Could you check the branch changes?

floreks:
Travis test fails: https://travis-ci.org/kubernetes/dashboard/jobs/270252336#L775

kairen:
@floreks Okay, I check now.

kairen:
@floreks The test error occurs in the `dirExists` condition of `localehandler.go`, because the environment is not currently supported these locales.

kairen:
@floreks I fixed the problem, could you check again?

floreks:
LGTM. Great work! Thanks.

kairen:
@floreks Thanks you for your help!!

PR #2318, opened 2017-08-29
Link to overview from namespace detail

### LINKED ISSUE #2214: Add link to switch namespace in the list view
Since the overview page is supposed to describe a namespace, I think it would make sense to have some of the information on the namespace itself like that found in the namespace detail page there. Particularly events and annotations.

cnwalker:
Puts a "view overview" button on the namespace detail page . Closes #2214.

- This PR primarily adds the ability to pass a link to a namespace info card, which creates a button on the card allowing the user to redirect.

screen shot 2017-08-29 at 3 39 50 pm


cnwalker:
@maciaszczykm

maciaszczykm:
I am not convinced by the location of this button. I think we can put it in action bar with a slightly different label.

cnwalker:
Hmm what label are you thinking of for the action bar @maciaszczykm? I put the button on the detail card because I was thinking that visiting the overview page another way to get more "details" about the namespace.

maciaszczykm:
> Hmm what label are you thinking of for the action bar @maciaszczykm? 

Just like logs or exec buttons on pod detail page. Something like `Overview`. No strong opinion here.

@danielromlein Any ideas?

Kaijun:
hmm, my opinion on the code. does it necessary to extract these two props instead of letting them exist straightly in the component `kd-namespace-info`?
```
+    'link': '<',
+    'buttonText': '<',
```

maciaszczykm:
@Kaijun Good point.

@cnwalker I have applied fixes and I am going to merge it.

PR #2323, opened 2017-08-30
Add browsers support section to readme

maciaszczykm:
Open to the discussion here. Do we need it? Are versions correct?

Reopened #2237.


floreks:
I agree that we need it but first we need to test if they are working as expected. I have added target browsers to our [`.babelrc`](https://github.com/kubernetes/dashboard/blob/master/.babelrc#L6) so all polyfills should be included already to make sure that our code runs on all targeted browsers.

More information how it works can be found [here](https://github.com/babel/babel-preset-env).

Issue #2207, opened 2017-07-28
Critical login improvements


floreks:
List of things to be done before dashboard log in mechanism will be ready for official release. P0 are must have.
Template: [Feature] - [Size] - [Priority]

 - [x] Login status on main toolbar - M - P0 - @maciaszczykm - #2242
   There is option to skip login page. This causes dashboard to use service account privileges mounted into the pod. We also need to give option to log in if user chooses to skip it at first.
- [x] Logout option - M - P0 - @maciaszczykm - #2242
- [x] Rework login page (UX) - M - P2 - @maciaszczykm - #2242  
- [x] Encryption key storing mechanism - XL - P0 - @floreks - #2243
   - [x] Make sure that dashboard with minimal privileges can not see/edit/delete key holder secret - @floreks - #2311
- [x] Token expiration/refresh feature - L - P0 - @floreks - #2316
  [K8S - Certificate manager](https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/certificate)
- [x] Restrict login functionality to HTTPS - M - P0 - @maciaszczykm - #2250
- [x] Add more login options: basic, kubeconfig-based - M - P1 - @floreks - #2325
- [x] Add `--authentication-mode` argument that will be responsible for enabling/disabling certain login options in the frontend: M - P1 - @floreks - #2325
- [x] Improve error handling - M - P2 - Done in #2316, #2325
   In case of unexpected error during login status check or token refresh we should logout user and redirect to login page.
- [x] https://github.com/kubernetes/dashboard/issues/2207#issuecomment-321777203
   Fixed by #2316. User will be logged out and redirected to login page if existing token is invalid.

Moved all non-critical features to #2353.

maciaszczykm:
I will implement logout.

maciaszczykm:
There is an issue with login, when I am using one cluster instance and I do have token set in cookies (it means I am logged in) and I switch to another cluster. Token will be still available in cookies, but it won't match any secret and page will crash. We have to check not only if token is available, but if it is valid to avoid this issue.

floreks:
This is on the list. Improved error handling will deal with that. Also cookie is saved for a particular page address so unless 2 daahboards on 2 clusters have same address we are fine.

maciaszczykm:
Yes, I am aware of that, but assuming somebody uses proxy errors could happen.

rf232:
[...]

PR #2325, opened 2017-08-31
Add authentication mode flag and kubeconfig based login option

floreks:
Related to #2207.

### Changes
- Add `--authentication-mode` flag to dashboard. Supported options are: `token`, `basic`. Default: `token`.
- Add kubeconfig based login option. It is automatically enabled in case any auth mode is enabled.
- Add more generic upload directive. Deploy should use this one. We should rewrite it eventually.

Right now all errors returned by backend are forwarded and displayed on login screen in warning popup.

### TODO
- [x] Add documentation
- [x] Add tests

zrzut ekranu z 2017-08-31 13-23-42

zrzut ekranu z 2017-08-31 13-28-38


src/app/backend/auth/kubeconfig.go
```diff
@@ -14,4 +14,105 @@
...
+		Token:    info.Token,
+	}, nil
+}
+
+// NewBasicAuthenticator returns Authenticator based on LoginSpec.
+func NewKubeConfigAuthenticator(spec *authApi.LoginSpec) authApi.Authenticator {
+	return &kubeConfigAuthenticator{
+		fileContent: []byte(spec.KubeConfig),
+	}
+}
```
rf232: I like the look of this file, 2 questions, 
1) might be mostly me being not up to speed. but does this 'authentication method' chose which to use based on the 'enabledness' of Token and Basic authentication? and if so, if both are enabled, does it have a preference?

2) do we want to support some auth-provider type kubeconfig files? e.g. GKE writes something like:
```
- name: AzureDiamond
  user:
    auth-provider:
      config:
        access-token: hunter2
        cmd-args: config config-helper --format=json
        cmd-path: /google/google-cloud-sdk/bin/gcloud
        expiry: 2017-08-31 17:19:02
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
```
and while we ofcourse can't do the token refresh by running gcloud, the access-token in the config is a valid bearer token. 

Both are mostly informational questions, I'm not saying we should do fancy stuff for auth-providers, just wondering if it might be worth the effort

rf232: ow, I should read your description for question 1. still wondering about second part of that question though

src/app/frontend/login/login.html
```diff
@@ -25,11 +25,15 @@
...
-      <kd-login-options>
-        <!--TODO(floreks): enable once kubernetes#49709 is resolved -->
-        <!--<kd-basic-login title="[[Basic|Basic login]]" on-update="$ctrl.update(loginSpec)"></kd-basic-login>-->
-        <kd-token-login title="[[Token|Token option label shown on login page]]"
+      <kd-login-options on-change="$ctrl.onOptionChange()">
+        <kd-token-login ng-if="$ctrl.isAuthenticationModeEnabled($ctrl.supportedAuthenticationModes.TOKEN)"
+                        title="[[Token|Token option label shown on login page]]"
                         on-update="$ctrl.onUpdate(loginSpec)"></kd-token-login>
+        <kd-basic-login ng-if="$ctrl.isAuthenticationModeEnabled($ctrl.supportedAuthenticationModes.BASIC)"
+                        title="[[Basic|Basic login]]"
```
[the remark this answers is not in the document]
floreks: I will. This is still work in progress.

src/app/backend/auth/kubeconfig.go
floreks: 1. I'll add check which auth modes are enabled and allow only user fields that match them. In example if only token mode will be enabled then kubeconfig authenticator won't allow to log in with username/password.

2. We are planning to add support for external identity providers in the future. Right now I am not sure if this will be done through kubeconfig or similar to popular `login with facebook/google` options. For now we'll stick to basic options. Best option for us would be some API exposed by apiserver that would handle most of this for us.

PR #2329, opened 2017-09-02
Update and fix zh-tw translation

kairen:
Reference from #2311 to add missing text, and fix some translation.


floreks:
LGTM. Thanks!

Issue #2286, opened 2017-08-18
See full Image name in Overview


Globegitter:
##### Environment

```
Dashboard version: 1.6.3
Kubernetes version: 1.7.3
Operating system:
Node.js version:
Go version:
```
I am running this yaml file: https://github.com/kubernetes/kops/blob/fd6eb881b4a1901b5171ce07966c67f2b1fbda16/addons/kubernetes-dashboard/v1.6.3.yaml

##### Steps to reproduce

Launch a deployment with an image that has a very long name

##### Observed result

Name on overview dashboard gets cut off and not possible to see the full name.

##### Expected result

Some ability to see the full name, e.g. by hovering over it, to easily see what version is deployed

##### Comments

This is how it looks for basically all of our deplyoments:

![image](https://user-images.githubusercontent.com/2580254/29449545-7366e0c8-83fb-11e7-81e7-97eae8ec8aa3.png)

cheld:
Fundamentally, we have two options:
1. find a better presentation. Make it expandable somehow. (Maybe hover or Inbox style or something else.)
2. Remove information that is not absolutely required. It would be interessting to have some guidance from users which infromation they use and which information is obsolete.

Globegitter:
Both are valid arguments - I like having this information in the overview to quickly check what image version this is using - maybe there is another option to have this customizable to users? But maybe that also just adds unnecessary bloat.

Not sure what the best solution here is but if the information does stay it would definitely be great to have  something along the lines of 1. what you mentioned.

cheld:
Customization is a good point. We can do so after we implemented a settings page. Currently, Dashboard is completely stateless.

So, you are actually interessted in the versions. This makes sense. (Of course, the versions require the image name)

imkmf:
Hi all! I'd like to get started contributing to the k8s dashboard and this ticket stood out to me today–would y'all be amenable if I fixed this via CSS? Modifying the text class `white-space` property on `hover` looks like this, locally:

![out](https://user-images.githubusercontent.com/922353/30399993-5f3fc60c-988a-11e7-8c5c-cf00afdc5d0f.gif)

maciaszczykm:
@imkmf I think there could be overlays if there is less space vertically. Also, this could cause a lot of flickering.

floreks:
There is already PR open for this #2330.

cheld:
@imkmf the video is certainly nice. A few comments:
[...]

PR #2330, opened 2017-09-04
add full image name in overview

shun-miyoshi-com:
This pull request will resolve issue #2286.
I added full image name in overview.
We can see full image name by hovering the mouse over "Images".
But, now I implemented of resource "Deployments" only.

If you accept my this concept, I will implement all of resources.
Thank you.

### DIFF
--- a/src/app/frontend/common/components/middleellipsis/component.js
+++ b/src/app/frontend/common/components/middleellipsis/component.js
@@ -29,7 +29,7 @@ export default class MiddleEllipsisController {
     this.element_ = $element;
   }
 
-  isTrancated() {
+  isTruncated() {
     let c = this.element_;
     c.addClass("kd-middleellipsis_width");
     let cOfsWidth = c[0].offsetWidth;
--- a/src/app/frontend/common/components/middleellipsis/component.js
+++ b/src/app/frontend/common/components/middleellipsis/component.js
@@ -19,10 +19,26 @@ export default class MiddleEllipsisController {
   /**
    * Constructs middle ellipsis controller.
    * @ngInject
+   * @param {!angular.JQLite} $element
    */
-  constructor() {
+  constructor($element) {
     /** @export {string} Initialized from the scope. */
     this.displayString;
+
+    /** @private {!angular.JQLite} */
+    this.element_ = $element;
+  }
+
+  isTrancated() {
+    let c = this.element_;
+    c.addClass("kd-middleellipsis_width");
+    let c_ofs_width = c[0].offsetWidth;
+    c.removeClass("kd-middleellipsis_width");
+    return c_ofs_width > this.element_[0].offsetWidth;
+  }
+
+  getDisplayStringAtTooltip() {
+    return goog.getMsg(this.displayString);
   }
 }
 
--- a/src/app/frontend/common/components/middleellipsis/component.js
+++ b/src/app/frontend/common/components/middleellipsis/component.js
@@ -31,9 +31,9 @@ export default class MiddleEllipsisController {
 
   isTruncated() {
     let c = this.element_;
-    c.addClass("kd-middleellipsis_width");
+    c.addClass("kd-middleellipsis-width");
     let cOfsWidth = c[0].offsetWidth;
-    c.removeClass("kd-middleellipsis_width");
+    c.removeClass("kd-middleellipsis-width");
     return cOfsWidth > this.element_[0].offsetWidth;
   }
 
--- a/src/app/frontend/common/components/middleellipsis/component.js
+++ b/src/app/frontend/common/components/middleellipsis/component.js
@@ -32,9 +32,9 @@ export default class MiddleEllipsisController {
   isTrancated() {
     let c = this.element_;
     c.addClass("kd-middleellipsis_width");
-    let c_ofs_width = c[0].offsetWidth;
+    let cOfsWidth = c[0].offsetWidth;
     c.removeClass("kd-middleellipsis_width");
-    return c_ofs_width > this.element_[0].offsetWidth;
+    return cOfsWidth > this.element_[0].offsetWidth;
   }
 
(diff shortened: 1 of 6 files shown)


shun-miyoshi-com:
I signed a CLA.

floreks:
This is actually what we wanted to avoid. Tooltip should be only displayed in case name is truncated, not always.

shun-miyoshi-com:
I changed my code.
My algorithm is shown as follows.

1. get field width.
  I get field width of Images header column.
  * document.getElementById("deployment_image_header").clientWidth

2. get display length of image full name
  I added hidden <div> tag, and get its width
  * <div style="position:absolute;visibility:hidden" ng-attr-id="{{'deployment_image_'+image}}">{{::image}}</div>
  * document.getElementById("deployment_image_"+imageName).clientWidth

3. compare 1. and 2.
  If 2 .> 1. , an image name may be truncated.

Please review my code.

floreks:
I don't have time to investigate but this does not work in all cases. Some of my images are truncated but tooltip is not displayed.

You can check out this thread: https://stackoverflow.com/questions/7738117/html-text-overflow-ellipsis-detection

Also I'd preffer to keep the logic inside `kdMiddleEllipsis` component so it would always work.

shun-miyoshi-com:
Your suggestion is really interesting and cool.
Unfortunately, I cannot implement this thread's code completely, because angularjs cannot use JQuery(only JQLite).
JQLite cannot use width() method.(please see https://docs.angularjs.org/api/ng/function/angular.element)
So, I use offsetWidth, however this variable don't run by element.clone() method.
Thus, I use addClass()/removeClass() method instead of clone()/remove() method.

src/app/frontend/common/components/middleellipsis/component.js
```diff
@@ -19,10 +19,26 @@ export default class MiddleEllipsisController {
...
     /** @export {string} Initialized from the scope. */
     this.displayString;
+
+    /** @private {!angular.JQLite} */
+    this.element_ = $element;
+  }
+
+  isTrancated() {
+    let c = this.element_;
+    c.addClass("kd-middleellipsis_width");
```
maciaszczykm: Use Angular's `ng-class`.

src/app/frontend/common/components/middleellipsis/component.js
```diff
@@ -19,10 +19,26 @@ export default class MiddleEllipsisController {
...
-  constructor() {
+  constructor($element) {
     /** @export {string} Initialized from the scope. */
     this.displayString;
+
+    /** @private {!angular.JQLite} */
+    this.element_ = $element;
+  }
+
+  isTrancated() {
```
maciaszczykm: Missing annotations. Check our Wiki pages for more info about code conventions.

src/app/frontend/common/components/middleellipsis/component.js
```diff
@@ -19,10 +19,26 @@ export default class MiddleEllipsisController {
...
+    let c = this.element_;
+    c.addClass("kd-middleellipsis_width");
+    let c_ofs_width = c[0].offsetWidth;
```
maciaszczykm: Use camelCase.

maciaszczykm:
Any update?

Please check https://github.com/kubernetes/dashboard/wiki/Code-conventions.

src/app/frontend/common/components/middleellipsis/component.js
shun-miyoshi-com: I'm sorry, but I don't understand you.
What should I do?

src/app/frontend/common/components/middleellipsis/component.js
shun-miyoshi-com: Which annotation is mistaken?(this.element_ or isTrancated or any other else ?)
this.element_ is used in  MiddleEllipsisController only, and isTruncated is used in middleellipsis.html.

maciaszczykm: Closure compiler annotations - read about them in our wiki.

src/app/frontend/common/components/middleellipsis/component.js
maciaszczykm: https://docs.angularjs.org/api/ng/directive/ngClass

src/app/frontend/common/components/middleellipsis/component.js
```diff
@@ -19,10 +19,26 @@ export default class MiddleEllipsisController {
...
     /** @export {string} Initialized from the scope. */
     this.displayString;
+
+    /** @private {!angular.JQLite} */
+    this.element_ = $element;
+  }
+
+  isTruncated() {
+    let c = this.element_;
+    c.addClass("kd-middleellipsis_width");
```
maciaszczykm: Do not use `_` in here. Switch to `-`.

maciaszczykm:
https://github.com/kubernetes/dashboard/pull/2330#discussion_r140028573 and https://github.com/kubernetes/dashboard/pull/2330#discussion_r140028432 are not solved yet.

src/app/frontend/common/components/middleellipsis/component.js
shun-miyoshi-com: I have no idea how to use `ng-class` effectively.
The example of the link is how to add class dynamically, but I use addClass method because of adding class temporarily.
Please help me.

src/app/frontend/common/components/middleellipsis/component.js
shun-miyoshi-com: My added annotations( @ param {!angular.JQLite} $element and @ private {!angular.JQLite}) will same as other components such as (common/components/allocatedresourceschart/component.js)
I'm sorry, but could you tell me about a detail of mistaken annotetions.

maciaszczykm:
Closing as stale.

Issue #2121, opened 2017-07-17
Show same version of zero state on all pages?


maciaszczykm:
We have different versions of zero state on list views for example. All list views should look the same.

cnwalker:
I'd be interested in taking a look at this. What are some examples of list views with different zero states?

maciaszczykm:
It's a bit of refactoring, because we have duplicate zero state definitions in our code. Besides that, we are inconsistent so on node list view I can see following zero state:

![image](https://user-images.githubusercontent.com/2823399/28456237-f55bdcf6-6e01-11e7-9084-4e528dd63778.png)

And for almost all workload resources:

![image](https://user-images.githubusercontent.com/2823399/28456289-2417ceb0-6e02-11e7-8a77-01d911d70e6e.png)

It should be more consistent. If you need our help reach out to us on Slack or here :)

floreks:
Second option should be displayed on pages like: Cluster, Workloads, Config and Storage. First option should be displayed on all resource list (for single resource) and detail pages.

PR #2336, opened 2017-09-05
Make roles zero state consistent with other cluster resources

cnwalker:
The zero state for the roles resource is not consistent with the zero states for the other cluster resources. This PR enforces consistency. refs #2121 

Before:
screen shot 2017-09-05 at 2 48 05 pm

After:
screen shot 2017-09-05 at 2 47 57 pm


cnwalker:
I think this is ready for a review when you get the chance @floreks @maciaszczykm

floreks:
We are out of office and will be back tomorrow.

PR #2342, opened 2017-09-07
Fix allocation graphs

maciaszczykm:
Before:

image

After:

image

### DIFF
--- a/src/app/frontend/common/components/allocatedresourceschart/component.js
+++ b/src/app/frontend/common/components/allocatedresourceschart/component.js
@@ -97,7 +97,7 @@ export class AllocatedResourcesChartController {
    * @param {!Array<string>} colors
    * @param {number} margin
    * @param {number} ratio
-   * @param {function(Object|null)} labelFunc
+   * @param {function(Object)|null} labelFunc
    * @private
    */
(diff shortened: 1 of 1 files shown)


src/app/frontend/common/components/allocatedresourceschart/component.js
```diff
@@ -63,6 +63,29 @@ export class AllocatedResourcesChartController {
   }
 
   /**
+   * Displays label only for allocated resources (with index equal to 0).
+   *
+   * @return {string}
+   * @private
+   */
+  displayOnlyAllocated(d, i) {
```
floreks: Private functions should end with `_` suffix.

maciaszczykm: Yep, done.

src/app/frontend/common/components/allocatedresourceschart/component.js
```diff
@@ -63,6 +63,29 @@ export class AllocatedResourcesChartController {
   }
 
   /**
+   * Displays label only for allocated resources (with index equal to 0).
+   *
```
floreks: Missing params annotations.

PR #2355, opened 2017-09-12
Add periodic check for metric client health

### LINKED ISSUE #2251: Add option to enable/disable heapster integration from dashboard
##### Environment
```
Dashboard version: 1.6.3
Kubernetes version: 1.7.3
Operating system: CoreOS
```

##### Steps to reproduce
When I start the dashboard with the workaround for heapster (it does not work without it): -
```
      - name: kubernetes-dashboard
        image: gcr.io/google_containers/kubernetes-dashboard-amd64:v1.6.3
        args:
        - --heapster-host=http://heapster.kube-system
```
If heapster service comes up after the dashboard pods then the dashboard does not display any graphs.

##### Observed result
Kubernetes dashboard does not retry for heapster availability and so no graphs are displayed within the dashboard.

Forcing a redeployment/killing the dashboard pods once heapster is up brings back the graphs. 

##### Expected result
I don't want to have to start the dashboard after heapster as this is hard to orchestrate.
The dashboard should continue to check for heapster availability and start displaying graphs once it is available rather than checking once.

### LINKED ISSUE #2306: Improve heapster integration
As of heapster `1.4.1` `/healthz` endpoint does not return only `ok` message. It is returned only after some initial data gets scraped and some metrics are available. For now we could add some periodic health check (not too long, so all replicas are roughly in the same state at all times) for heapster that would enable/disable it in case something happens.

floreks:
Resolves #2306 and resolves #2251.

Added new flag `metric-client-check-period` that defines how often health check against configured metric client will be run. By default it is set to 30 seconds.


cheld:
@floreks BTW: there was request from core to replace flags with config map

floreks:
We can not replace them as Dashboard has to work also outside of the cluster. We could only add support for ConfigMap.

cheld:
@floreks 

The main complain about params is that:
- services must be restarted on change
- not versioned
- not managed by kubernetes

I guess it could be as a configuration file outside of kubernetes and a config map inside. If you interessted I might be able to dig out some pointers (issues)

cheld:
@floreks Another important reason was that add-on-manager cannot hanlde params properly - just config maps

floreks:
Updating params stored in ConfigMap and updating Dashboard is basically the same (requires similar steps).

With ConfigMap:
1. Update ConfigMap YAML/Use kubectl edit/Use edit within Dashboard
2. Restart Dashboard to pick up new config

Updating Dashboard:
1. Update Dashboard YAML/Use kubectl edit/Use edit within Dashboard
2. Restart Dashboard/Use kubectl apply to reconfigure Dashboard (this will automatically restart it)

Using `kubectl apply` actually supports simple versioning because there is option to see last used configuration.

To support this k8s "native" way we'd need to completely redesign the way we read and handle params. Currently they are read only once during Dashboard start and can not be updated without restarting Dashboard. IMO that's ok since they are not changed often. Once you set it and start Dashboard it will run without changes for a long time.

PS. Personally I have never seen any complaints about how we handle params support in Dashboard.

cheld:
@floreks 
The discussion was in core directed at core components like master, controller, etc...

Eventually, we might want t consider and align us with core concepts. But, I agree currently not so relavant for Dashboard.

BTW:  I assume it would be possible to avoid the restart with some effort using config map or config file

Trackhe:
@floreks my dashboard rc1 does not show display metrics. Kubernetes runs on Raspberry pi 4 arm with metrix arm

Issue #2358, opened 2017-09-12
Display logs also for init containers


jagregory:
I'm running 1.6.3 and it's great to see Init Containers appearing in Pods now, excellent progress!

Just raising that it'd really useful to be able to view the Logs of Init Containers. I was surprised to see them not appear in the drop-down of containers within the Logs view.

maciaszczykm:
Thanks for this feature request. Makes sense. We will implement that in upcoming versions.

jagregory:
Thanks @maciaszczykm. I had a look through the code and it seemed reasonably straight forward for a first PR, so I gave it a go: #2359

maciaszczykm:
Done by @kenan435 in #2425.

georgecrawford:
Thanks @kenan435!

PR #2359, opened 2017-09-12
Include Init Containers in Logs

jagregory:
Fix for #2358.

This change updates the `GetContainerNames` function to also include Init Containers; the impact of this being you can now view the Logs of Init Containers.

screen shot 2017-09-12 at 9 36 18 pm

(copy-migrations and run-migrations are both init-containers)

### DIFF
--- a/src/app/backend/resource/common/pod.go
+++ b/src/app/backend/resource/common/pod.go
@@ -79,6 +79,9 @@ func GetContainerNames(podTemplate *v1.PodSpec) []string {
 	for _, container := range podTemplate.Containers {
 		containerNames = append(containerNames, container.Name)
 	}
+	for _, container := range podTemplate.InitContainers {
+		containerNames = append(containerNames, container.Name)
+	}
 	return containerNames
 }
 
--- a/src/app/backend/resource/common/pod_test.go
+++ b/src/app/backend/resource/common/pod_test.go
@@ -135,9 +135,10 @@ func TestGetContainerNames(t *testing.T) {
 		{&api.PodSpec{}, nil},
 		{
 			&api.PodSpec{
-				Containers: []api.Container{{Name: "container"}, {Name: "container"}},
+				Containers:     []api.Container{{Name: "container"}, {Name: "container"}},
+				InitContainers: []api.Container{{Name: "init-container-1"}, {Name: "init-container-2"}},
 			},
-			[]string{"container", "container"},
+			[]string{"container", "container", "init-container-1", "init-container-2"},
 		},
 	}
 


maciaszczykm:
Would you be able to mark init containers somehow? Just some label to inform users and perhaps put them at the end of the list.

jagregory:
**Sorting:** Currently the `GetContainerNames` method will return the init containers last, only by virtue of them being appended to the slice separately. I'm not sure if the UI does any sorting in that drop down (it just so happens the containers I have are alphabetical like that anyway). I can have a look.

**Label**: Sounds reasonable. Is there any existing pattern I should follow? Elsewhere in the UI which does something similar?

maciaszczykm:
> **Label**: Sounds reasonable. Is there any existing pattern I should follow? Elsewhere in the UI which does something similar?

@jagregory I am thinking about grey label next to container name or something like we have in namespace selector (group, that would include init container names). 

Now I can see, that we are sending list of strings to the front. You could change that to send a list of objects (with `name` and `isInitContainer` fields). Does it sound okay for you?

jagregory:
That should be possible. It'll require a bit of refactoring, as the `GetContainerNames` function is used in a few places to populate a `LogSources` structure returned as part of the `ResourceController` interface, but it looks like it's exclusively used to drive the `handleLogSource` handler, so yeah should be fine.

I'll take a look :)

cheld:
Hmm...

Currently, GetContainerNames() is almost identical to get GetContainerImages(). Also, it is currently simple logic in common package. From that reason I would not merge init container and normal containers in this method + package.

Currently, GetContainerNames() is only used by logs. If you make this method more log specific you should move the method to logs package during your refactoring, I guess.

src/app/backend/resource/common/pod.go
```diff
@@ -79,6 +79,9 @@ func GetContainerNames(podTemplate *v1.PodSpec) []string {
 	for _, container := range podTemplate.Containers {
 		containerNames = append(containerNames, container.Name)
 	}
+	for _, container := range podTemplate.InitContainers {
```
cheld: Hmm...

Currently, GetContainerNames() is almost identical to get GetContainerImages(). Also, it is currently simple logic in common package. From that reason I would not merge init container and normal containers in this method + package.

Currently, GetContainerNames() is only used by logs. If you make this method more log specific you should move the method to logs package during your refactoring, I guess.

jagregory: Thanks @cheld, I think that's valid.

I noticed the `GetContainerImages` method in common does almost exactly the same thing as `GetContainerNames` (but for images, not names, obviously); the Container Images code will also have the same issue as this PR, that init containers aren't included. Is this something which should be corrected? I'm less familiar with where `GetContainerImages` is used, but presumably if it's presenting a list of Images to a user it would be reasonable to expect init containers to be included too.

With that in mind, an alternative refactoring could be merging `GetContainerNames` and `GetContainerImages` into a single method returning a `ContainerInfo{Name,Image,Type string}` struct. Thoughts?

I'm easy either way; it just seems we have an option of pulling these methods further apart, or pushing them together.

cheld: Yesterday I had the feeling that we have no problem, because we show init containers in the UI. Now, thinking again about it, I think e.g. controller list views like ReplicaSet-List-View will probably miss init containers. 
It would be fantastic if you search for usage of the method and check the UI if we have problems. You can create separate issue for it.

If we have the same issue in GetContainerNames and GetContanerImages I would merge. If not I would rather move appart.

Fundamentally, your proposal is fine. But I guess it is not working - but maybe I am wrong. The goal of the method was to reduce a little bit of boiler plate code in the caller. I think with your proposal we have to iterate in the caller again. So there seems no benefit. We just convert one struct in another :)) . I tink adding parameters would work

maciaszczykm:
Any update?

maciaszczykm:
Closing as stale.

PR #2366, opened 2017-09-14
Prepare for 1.7.0 release

maciaszczykm:
- [x] use secret instead of config maps to store certs
- [x] describe how-to access app
- [x] update docs
- [x] update yamls
- [x] compatibility matrix out to wikis?
- [x] shorter desc of alternative setup (move to wikis)
- [x] https://github.com/kubernetes/kubernetes.github.io/pull/5520
- [x] finish wikis (especially installation guide)
- [x] add run command


README.md
```diff
@@ -6,76 +6,86 @@
...
-```shell
-$ kubectl create -f https://git.io/kube-dashboard
+**IMPORTANT:** As of version 1.7 Dashboard uses minimal privileges setup by default. It means, that some manual steps
+are required to make it fully secure and functional. Before performing any steps read [Access Control](
+https://github.com/kubernetes/dashboard/wiki/Access-control) guide.
+
+### Recommended setup
+
+Full security can be ensured only by accessing Dashboard over HTTPS. In order to enable HTTPS mode certificates need
+to be passed to the application. This can be achieved by providing `--tls-cert-file` and `--tls-cert-key` flags to the
```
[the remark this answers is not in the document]
cheld: Can you please elaborate why we recomend to configure HTTPS? Isn't `kubectl proxy` sufficient? Do we want to protect the traffic inside the cluster?

README.md
```diff
@@ -6,76 +6,86 @@
...
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/src/deploy/recommended/kubernetes-dashboard.yaml
 ```
 
-You can also install unstable HEAD builds with the newest features that the team works on by
-following the [development guide](docs/devel/head-releases.md).
+### Alternative setup
 
-Note that for the metrics and graphs to be available you need to
-have [Heapster](https://github.com/kubernetes/heapster/) running in your cluster.
+This setup is not fully secure, however it does not require any additional steps. In this setup access control can be
```
cheld: This section will be confusing for users. It claims that no additional steps have to be taken. However, from user perspective it is not usable without additional steps (like setup of SSO). I assume this setup version will cause the most effort for user (quite the opposite of the claim). 

Move this page to a seperate wiki page

README.md
```diff
@@ -6,76 +6,86 @@
...
-You may access the UI directly via the apiserver proxy. Open a browser and navigate to `https://<kubernetes-master>/ui`.
+**NOTE:** You need to have [Heapster](https://github.com/kubernetes/heapster/) running in your cluster for the metrics
+and graphs to be available. You can read more about it in [Integrations](
+https://github.com/kubernetes/dashboard/wiki/Integrations) guide.
 
-Please note that this works only if the apiserver is set up to allow authentication with username and password or certificates, however certificates require some manual steps to be installed in the browser. This is not currently the case with the setup tool `kubeadm`. See [documentation](http://kubernetes.io/docs/admin/authentication/) if you want to configure it manually.
+## Usage
 
-If the username and password is configured but unknown to you, then use `kubectl config view` to find it.
+After installing Dashboard check [Accessing Dashboard](https://github.com/kubernetes/dashboard/wiki/Accessing-dashboard)
```
cheld: This is important, but not prominent. Might be overlooked by users

README.md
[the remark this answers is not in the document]
floreks: > Am I supposed to modify the deployment yaml? Couldn't it be detected in case the config map is mounted?

User is only supposed to follow described steps. Nothing else. Only manual step is to generate certificates that will be used by Dashboard to enable HTTPS. They will be used to create configmap (we'll change it to secret). This way certs can be mounted into Dashboard container and used to serve HTTPS.

> Can you please elaborate why we recomend to configure HTTPS? 

Full security can be ensured only by accessing Dashboard over HTTPS.

This is the main reason and in case user wants to know more he can read our wiki.

> Isn't kubectl proxy sufficient? Do we want to protect the traffic inside the cluster?

It's Dashboard that has to ensure security, not people accessing it by using some secure tunnel. In order to even see login view Dashboard has to be accessed over HTTPS. This does not mean that it can be accessed through secure tunnel. User has to use HTTPS endpoints exposed by Dashboard. User credentials can never be sent with plain text between frontend and backend.

All that is described in detail in our wiki guides.

README.md
floreks: In Deployment section we describe only how to install Dashboard. No additional steps are required to install Dashboard. We can add word 'install' to make it more prominent. First setup requires user to generate certificates and create resource that will be used to mount them into the pod. Here user can simply use single `kubectl create -f...` command.

Why would you move it? IMO README should describe all official ways of deploying Dashboard. There is a secure setup and a bit less secure but can installed quickly.

README.md
floreks: I'm not sure. How would you make it more prominent?

README.md
```diff
@@ -6,76 +6,86 @@
...
-kubectl will handle authentication with apiserver and make Dashboard available at [http://localhost:8001/ui](http://localhost:8001/ui)
 
-The UI can _only_ be accessed from the machine where the command is executed. See `kubectl proxy --help` for more options.
+**NOTE:** You can also install latest development builds with the newest features that the team works on by
+following the [Development Release](https://github.com/kubernetes/dashboard/wiki/Installation#development-release)
+guide.
 
-## Alternative Usage
-You may access the UI directly via the apiserver proxy. Open a browser and navigate to `https://<kubernetes-master>/ui`.
+**NOTE:** You need to have [Heapster](https://github.com/kubernetes/heapster/) running in your cluster for the metrics
```
floreks: One thing to these notes. When I actually look at the whole page then they seem to be related to 'Alternative Setup' section only. It does not look like they are related to the whole 'Deloyment' section.

README.md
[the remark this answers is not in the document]
cheld: > User is only supposed to follow described steps. Nothing else. Only manual step is to generate certificates that will be used by Dashboard to enable HTTPS.

Ok fine. But text is misleading.

> It's Dashboard that has to ensure security, not people accessing it by using some secure tunnel. In order to even see login view Dashboard has to be accessed over HTTPS. This does not mean that it can be accessed through secure tunnel. User has to use HTTPS endpoints exposed by Dashboard. User credentials can never be sent with plain text between frontend and backend.

If I understand you correctly you want to prefent the user from misconfiguration. Getting certificates from lets-encrypt or distributing the self-generated root certificate to all clients is some effort. 1. It is effort that (maybe) is not required when accessing through secure tunnel and 2. insecure if not done properly (of course the browser will warn). So, I wonder if making it optional would be better choice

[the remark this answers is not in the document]
floreks: > Ok fine. But text is misleading.

Ok, it might be. We can change "This can be achieved by providing --tls-cert-file and --tls-cert-key flags to the Dashboard." sentence and add short explanation that this is already done and user has to just follow the instructions.

> If I understand you correctly you want to prefent the user from misconfiguration. Getting certificates from lets-encrypt or distributing the self-generated root certificate to all clients is some effort. 1. It is effort that (maybe) is not required when accessing through secure tunnel and 2. insecure if not done properly (of course the browser will warn). So, I wonder if making it optional would be better choice

No. Think of Dashboard like any other regular webapp. Do not think about kubernetes for a second. You are a user that opens a page in his browser. Would you rather pass your credentials to HTTP page and send them with plain text to the server or use HTTPS endpoint and know that traffic will be encrypted?

Dashboard (and any other web page) does not care if user creates some additional tunnel to access it. It has to expose HTTPS endpoints itself. This is the only way to make sure that traffic between frontend and backend is encrypted and not easily sniffed. If user is indeed accessing HTTPS endpoints then and only then we are enabling login view. Otherwise the only way of acting as an user is to pass and use `Authorization: Bearer <token>` feature.

floreks:
@cheld We have updated docs. PTAL

@rf232 @bryk could you also take a look and share your opinion?

floreks:
@cheld We have updated docs. PTAL

@rf232 @bryk @liggitt could you also take a look and share your opinion? Maybe some steps need extra explanation.

floreks:
Quick idea. We could prepare docker container for simplest possible deployment that would take care of generating certificates and deploying resources. User would only run something like:
`docker run kd-deploy --kubeconfig=/path/to/file`

It would then:
* create self-signed certificates
* create secret based on certs
* deploy dashboard

More advanced users could follow installation guide in our wiki pages.

README.md
```diff
@@ -6,76 +6,51 @@
...
-If you are using Dashboard on a cluster with disabled authorization, you can install the latest stable release by running the following command:
-```shell
-$ kubectl create -f https://git.io/kube-dashboard-no-rbac
-```
+To deploy Dashboard execute following command:
 
-You can also install unstable HEAD builds with the newest features that the team works on by
-following the [development guide](docs/devel/head-releases.md).
+```sh
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/src/deploy/recommended/kubernetes-dashboard.yaml
```
cheld: what about the git.io shorthand? (BTW: I think there might be some management problem, because edit rights belong to users and cannot be assigned to the project - I think)

floreks: We will use it but first we need to actually merge it and make sure that shortened link won't change.

floreks: > BTW: I think there might be some management problem, because edit rights belong to users and cannot be assigned to the project - I think)

What do you mean?

src/deploy/recommended/kubernetes-dashboard-head.yaml
```diff
@@ -100,7 +100,7 @@ spec:
     spec:
       initContainers:
       - name: kubernetes-dashboard-init
-        image: kubernetesdashboarddev/kubernetes-dashboard-init
+        image: floreks/kubernetes-dashboard-init
```
maciaszczykm: Should be reverted.

floreks: Already fixed

README.md
cheld: Sorry, forget about it...

PR #2389, opened 2017-09-21
Revert "Prepare for 1.7.0 release"

rf232:
Reverts kubernetes/dashboard#2366

Was trying to create images, but there are some problems with the init container. At the moment it will only work for amd64 arch, since the base ubuntu image is a amd64 image. 

Also I think I'd prefer it if we can move it to gcr too for official release.


floreks:
Makes sense. There is a `Dockerfile` in `src/app/deploy/init` directory. We probably need to find different base images for different architectures to make it work.

rf232:
Sorry for the extra mess, everything looked good, so wanted to go ahead with everything, only then thought about additional archs

floreks:
We've also missed it :) Probably it will take some time to find proper way of building this init container for all archs.

rf232:
yeah, and getting a place to test them all might be interesting too, e.g. do you have a place to run a s390x image :)

floreks:
There was like 1 user that requested s390x support :) We've added it and actually he tested if it works.

floreks:
@rf232 I'm wondering if it's worth creating init container for all architectures. Dashboard still supports other architectures and we keep only YAML files for amd64 arch on repo. This init container is not required. We have added it to simplify deployment procedure described in README.md. Full procedure that does not require this container is described on our wiki pages. We probably could easily add support for arm. I'm not sure about other architectures.

maciaszczykm:
I would go for supporting only AMD64 with quick setup and adding a note to the readme, that on other architectures there is a need to create secret with certs on your own.

rf232:
https://github.com/kubernetes/dashboard/pull/2390#issuecomment-331191443 :) so yes amd64 we need it, arm too, or drop yaml

PR #2390, opened 2017-09-21
Prepare for 1.7.0 release

maciaszczykm:
Reopen of #2366.

We need to fix issues mentioned in #2389 before making the release.


src/deploy/recommended/kubernetes-dashboard-arm.yaml
```diff
@@ -0,0 +1,156 @@
...
+    matchLabels:
+      k8s-app: kubernetes-dashboard
+  template:
+    metadata:
+      labels:
+        k8s-app: kubernetes-dashboard
+    spec:
+      initContainers:
+      - name: kubernetes-dashboard-init
+        image: kubernetesdashboarddev/kubernetes-dashboard-init
```
rf232: For the released version I'd prefer to host this in gcr.io as well, and tag it with a version, so if something changes with a new version we can update this too

rf232:
We need to have the init container working at least for all the arch we provide the sample deployment yaml for (amd64 and arm) and have docs on how to prepare certificates for the other archs (be it with init container, or let them do it theirselves)

floreks:
We should be able to easily provide containers for amd64 and arm. [Certificate management](https://github.com/kubernetes/dashboard/wiki/Certificate-management) guide describes pretty well how to generate certs. It should work on any architecture.

src/deploy/recommended/kubernetes-dashboard-arm.yaml
floreks: I'm assuming that it won't have to be changed. It only generates self-signed certs and creates a secret. It will be better to keep it on gcr though.

floreks:
@rf232 Both init containers (amd64 and arm) are available at our DockerHub repo. https://hub.docker.com/u/kubernetesdashboarddev/

I have build arm container on my RPi. You can just pull them, tag again and push to gcr.

rf232:
Looks good, Although I'd still very much like the init containers to also be version tagged in someway, since if we'd need to generate the certs differently/store them somewhere else we'd have to update the image and can't use the :latest version anymore, So I'm thinking having a version tag now would be good.

floreks:
@rf232 makes sense. I will add version tag to them. We can start from version `1.0.0` as it is independent of Dashboard and should work well also with later versions.

floreks:
I think we are ready. Wiki is updated. Images are up. You can start the release process.

maciaszczykm:
@rf232 Can you merge after checking once more?

rf232:
will push images as soon as head image is up to date

rf232:
Images are pushed

Issue #1995, opened 2017-05-26
Feature Request: Show crashed containers and/or pods for logs and events in dashboard


betterman9394:
Hi, when I was investigating a production issue due to OOMing, i realize that there is no easy way for me to view/show the killed/crashed containers/pods in dashboard. It will be super helpful for user to monitor/debug the system

floreks:
As a quick solution we could provide logs from last crashed container. This way user could check the cuase of crash in container logs. WDYT?

Issue #1828, opened 2017-04-07
Improve logs viewer


maciaszczykm:
We should improve logs viewer. There is a list of action items:

- [x] ~Add since time selector (#265)~
- [x] Add option to hide timestamps (#1546 & #1825)
- [x] Aggregate logs across controllers and services
- [x] ~Show logs from time >2 hours~
- [x] Avoid out of memory errors (#1834, possibly #1925 and #1816)
- [x] Support ~tailing and~ following logs (#1083)
- [ ] Buttons in logs view disappear after decreasing window width (#1548)
- [x] How to view logs of crashed pods? #1393

cheld:
https://github.com/kubernetes/dashboard/pull/2007

cheld:
@floreks 

`Add since time selector` does not make much sense any more as we load only part of the log file. We would need better (paging) api support

`Show logs from time >2 hours`

Same for this issue. WDYT?

floreks:
`Show logs from time >2 hours` 

Does not make sense to me also, however `Add since time selector` option should not be that hard even with current implementation and still might be beneficial. In example if user finds that around 2 AM container had very high resource usage then it's nice to have option to easily check logs at this point in time.

`SinceTime` and `LimitBytes` are working together so we should be able to load few pages near provided time frame.

First approach might be limited to just given option to download logs, but in the end such feature might be nice to have.

Globegitter:
It would be nice to be able to increase the height of the log window/the number of lines shown on the screen. Could just be a simple dropdown on the top 50/100/500 lines or something like that.

Globegitter:
Another request that just came up, being able to search through the logs via a search bar - that would be great.

maciaszczykm:
> Another request that just came up, being able to search through the logs via a search bar - that would be great.

Until Kubernetes will provide API for that, we cannot do it on our own. It would kill the whole app. Logs are too big.

cheld:
Once we have implemented the extention framework, we could integrate with ELK

Globegitter:
Just upgraded to 1.7.0 and am not able to download logs, just getting a 404:

```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "the server could not find the requested resource",
  "reason": "NotFound",
  "details": {},
  "code": 404
}
```

cheld:
OK I have a look tomorrow

cheld:
[...]

PR #2392, opened 2017-09-22
Show previous container logs

cheld:
Fixes https://github.com/kubernetes/dashboard/issues/1995 (logs part)

Currently, there is no way to see logs of a crashed container. In other words, after crash a new container instance will be created with a new log file, but for debugging the previous logs are more interessting.

Kubectl (and API) provide a previous flag to see the previous logs. Added button to logs view to see logs of previous container instance

### DIFF
--- a/src/app/backend/handler/apihandler.go
+++ b/src/app/backend/handler/apihandler.go
@@ -2143,8 +2143,7 @@ func (apiHandler *APIHandler) handleLogs(request *restful.Request, response *res
 	if err != nil {
 		refLineNum = 0
 	}
-	previous := request.QueryParameter("previous") == "true"
-
+	usePreviousLogs := request.QueryParameter("previous") == "true"
 	offsetFrom, err1 := strconv.Atoi(request.QueryParameter("offsetFrom"))
 	offsetTo, err2 := strconv.Atoi(request.QueryParameter("offsetTo"))
 	logFilePosition := request.QueryParameter("logFilePosition")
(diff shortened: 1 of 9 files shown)


cheld:
BTW: Works for download as well

cheld:
New UI with latest log file:
logs0

Result of previous log file for a container that never crashed (kubectl returns the same message):
logs-1

cheld:
I searched a while for a "previous" icon. I am kind of proud of myself :)

src/app/backend/handler/apihandler.go
```diff
@@ -2143,6 +2143,7 @@ func (apiHandler *APIHandler) handleLogs(request *restful.Request, response *res
 	if err != nil {
 		refLineNum = 0
 	}
+	previous := request.QueryParameter("previous") == "true"
```
maciaszczykm: Could you explain meaning of `previous` flag more?

cheld: I hestitated to write a long comment in the apihandler. Instead, I made the variable a bit more verbose and added comment in handler function.

cheld:
PTAL

src/app/backend/handler/apihandler.go
```diff
@@ -2143,7 +2143,7 @@ func (apiHandler *APIHandler) handleLogs(request *restful.Request, response *res
 	if err != nil {
 		refLineNum = 0
 	}
-
+	usePreviousLogs := request.QueryParameter("previous") == "true"
```
maciaszczykm: It looks good right now. Anybody should be able to understand it.

maciaszczykm:
I am waiting for Travis to finish and then I will merge if you do not have anything to add.

cheld:
Improves #1828

Issue #2337, opened 2017-09-06
Update licenses to 2017


cnwalker:
There are a number of places that are "copyright 2015" like `src/app/backend/resource/pod/fieldpath.go`. Should these be updated to 2017?

There are also instances where there are multiple licenses in one file like in `src/app/frontend/chrome/nav/module.js`. Are both licenses really correct, or should there only be one?

floreks:
First is excluded as wrotten in comment. This code is copied from core so their license note is applied.

Every file should only have 1 license note. License script had to double them. They have to be manually removed.

cnwalker:
Which license should be removed? I'll try to find as many as I can and submit a PR.

floreks:
Older one.

maciaszczykm:
Copyrights should include files' publication/commit date.

floreks:
So actually all our files should probably include year range `2015-2017`. This would allow us to keep single template for all of them. Normally copyright notice should contain year of creation.

cheld:
In my understanding year of creation is enough

PR #2394, opened 2017-09-22
clean up headers

cheld:
Did some manual testing and grep searches.

1. Removed 2 double licenses
2. Replaced a couple of old Google licenses with current template
3. Removed unused favicon originally generated by yeoman

I noticed we are using Kubernetes Dashboard Autors. Other projects use Kubernetes Autors e.g. kops, minikube. Not sure if it matters. I think we should merge to make it consistent and then make a quick replace with all headers to be on the save side. I can do so.


maciaszczykm:
> I noticed we are using Kubernetes Dashboard Autors. Other projects use Kubernetes Autors e.g. kops, minikube. Not sure if it matters. I think we should merge to make it consistent and then make a quick replace with all headers to be on the save side. I can do so.

+1

maciaszczykm:
We should update Wiki footer too:

https://github.com/kubernetes/dashboard/wiki

build/assets/license/header_yaml.txt
```diff
@@ -0,0 +1,13 @@
+# Copyright 2017 The Kubernetes Dashboard Authors.
```
floreks: We keep here license template files used by gulp tasks to automatically update file headers. You would need to update `check-license-headers` and `update-license-headers` gulp tasks. Right now it is not used.

cheld: Yea, I just needed a place for copy&paste manually a couple of files

cheld: Should the file be removed?

floreks: If it won't be used by our build pipeline then yes.

cheld:
PTAL

cheld:
Fixes https://github.com/kubernetes/dashboard/issues/2337

cheld:
I thought I do it seperately. Find a way to automate, anounce, and then we do a quick merge. So we have less merge conflicts.

maciaszczykm:
@cheld Sure.

Issue #2240, opened 2017-08-07
Default Language can not been changed in dashbaord


PeiKevin:
##### Environment
```
Installation method: 
Kubernetes version:
Dashboard version: v1.6.3
Commit: 5b3e0858b3c713052307f2ec83b06853b61fde84
```

##### Observed result
Default Language has been set up in Chinese, I think this is related to browser or OS environment. 

##### Comments
But how  can I change to English,  is there any setting option ?

cheld:
Dashboard takes the browser locale. Currently, Dashboard is completely statless and has no user authentication, therefore it is not possible to configure anything per user. We could provide a command line flag to force all users to english

PeiKevin:
Great. It will be nice if there is the force flage setting, So many thanks for your support.

floreks:
If you want to enforce language for dashboard just download some browser extensions like `Simple locale switcher` or `Quick language switcher` and when on dashboard page set it to either english/japanese/chinese. 

We will eventually provide some settings page but this is rather low priority for now.

maciaszczykm:
Will be tracked from #2417.

XinhaiWang:
Go to chrome://settings/, and at the Languages section, move your favorite language to the top.

PR #2402, opened 2017-09-25
support force set language in ENV 

joway:
Sometimes, some teams prefer to use english version dashboard when the translation are incomplete or confusing . We need a ENV to force change the language to native english version .


joway:
enable cla

floreks:
It's a bit too hidden. We need a runtime option that can be changed while using Dashboard without the need to exec into container. While this would work, I believe only very few people would actually use it.

We could probably use cookies to store user preferred language and add some UI option to select supported language.

maciaszczykm:
I think we could accept it only as a temp solution.

cheld:
Maybe an even simpler alternative is to use environment variable (like in PR) to completely disable translations. (Until we have settings page)

cheld:
Similar request https://github.com/kubernetes/dashboard/issues/2240

cheld:
Settings page https://github.com/kubernetes/dashboard/issues/2417

PR #2421, opened 2017-09-28
Remove bower dependencies

floreks:
This is a preparation for angular migration. First we need to get rid of bower as new versions of angular are not exported to bower registry.

## Changes
* Update required NPM version in package.json.
* Patch `wiredep` during `npm install` to allow using it with `npm` dependencies instead of bower.
* Remove bower dependencies from the code and build pipeline.
* Update edit resource (angular-jsoneditor) dialog to use `code` mode instead of `tree`.
* Update edit resource UI scaling. It will resize based on `@media` screen size.

I have to perform some more tests, but `serve` and `serve:prod` are working fine now.


src/app/frontend/index.html
```diff
@@ -59,6 +59,7 @@
              ng-if="!$ctrl.isLoginState()">
     <!-- Application content is inserted here. -->
   </kd-chrome>
+
   <!-- build:js static/vendor.js -->
   <!-- bower:js -->
   <!-- Bower JS dependencies are populated here (dev build) and compiled
```
maciaszczykm: Search for `bower` occurrences inside codebase to remove leftovers like this.

floreks: I did. I must have missed this one comment. `bower:js` comment has to stay this way.

maciaszczykm:
https://github.com/kubernetes/dashboard/blob/master/.travis.yml#L45

maciaszczykm:
Works just as expected. A good change towards Angular migration.

Issue #2029, opened 2017-06-16
Pod attach/exec enhancements


floreks:
Moved task list from #1939 to an issue to keep track of them:
- [ ] Ability to attach as in kubectl attach
- [ ] Implement locking for session map (while the current code is bad, actually triggering the problem is unlikely)
- [ ] Add links leading to exec view (something like logs button)
- [ ] Serve SockJSURL from dashboard
- [ ] utf-8 input (needs hterm needs configuration)
- [ ] Add some tests
- [ ] Some solution to respawn if the user exits the shell (?)
- [ ] Some padding on the left (?)

Zexi:
If I run 'while true; do echo 1; done' command in shell, the browser will be stucked.
![image](https://user-images.githubusercontent.com/10767027/27340837-1817a3bc-560e-11e7-811e-6e0e051ac907.png)

adrianchifor:
When I leave a shell and refresh, the dashboard crashes with the following error: 

![dash-crash](https://user-images.githubusercontent.com/6370064/27427580-6a628e64-5737-11e7-84f2-919b8daa8104.png)

Is this related to the `Some solution to respawn if the user exits the shell (?)` point ?

floreks:
> I assume the problem is that it tries to read data from a closed socket.

I think you are right. We probably should have some button to open terminal and when session is terminated or there is a timeout automatically close it. Refreshing the page would close the session and open a new one.

IanLewis:
FYI exec/attach should work w/ kubectl proxy after https://github.com/kubernetes/kubernetes/pull/49534 gets merged.

maciaszczykm:
@IanLewis Great.

IanLewis:
@maciaszczykm Yah. Not sure if updating to use WebSockets is in scope of this issue or not though.

kfox1111:
A couple more issues for the list:

1. Most of my pods are multi container. I need a way to specify which container to exec into.
2. Quite a few containers these days don't have /bin/bash but only /bin/sh. Being able to specify would let it work in many more cases.

floreks:
1. There is a combobox on exec page where you can select a container.
2. If `/bin/bash` is not found it falls back to `/bin/sh` already. If both are not present you'll get an error.

shamil:
@floreks the second part never works, when it fails back to `/bin/sh` the console closes right after shell started. 

![](http://i.imgur.com/lAyn6iU.png)

If `/bin/bash` present, everything works well!

[...]

PR #2430, opened 2017-10-02
Fix Dashboard crash after exec into pod when using proxy

floreks:
Related to #2029 

Dashboard was crashing when trying to exec into pod. This was caused by missing `tmp` directory in the dashboard container. I have added `/tmp` mount that will be stored ~in memory~ on disk. 

Tested on k8s `1.7.7`. Exec works fine when accessing Dashboard using NodePort and kubectl proxy.


adrianchifor:
@floreks How can we view exec logs in the `/tmp` dir if we can't exec into the dashboard pod (`FROM scratch`) ?

floreks:
It is only required to make exec feature work in dashboard. These files are created when you exec into pod (any pod) and deleted when exec session is closed.

PR #2468, opened 2017-10-12
Fix creating empty secrets by kubernetes-dashboard-init

camilb:
Sometimes, the "kubernetes-dashboard-certs" secret is created by kubernetes-dashboard-init and it's empty, crashing the kubernetes-dashboard. The issue it's usually fixed by erasing several times the kubernetes-dashboard pod.

This update checks  the presence of the certificates before creating the secrets, then checks the existence of the certificates inside the "kubernetes-dashboard-certs" secret.

### DIFF
--- a/src/deploy/init/Dockerfile-amd64
+++ b/src/deploy/init/Dockerfile-amd64
@@ -33,12 +33,22 @@ RUN echo '#!/bin/bash\n\
     if [ -z "$(ls -A /certs)" ]; then\n\
       echo "Creating self-signed certificates"\n\
       cd /certs\n\
-      openssl req -nodes -newkey rsa:2048 -keyout dashboard.key -out dashboard.csr -subj "/C=/ST=/L=/O=/OU=/CN=kubernetes-dashboard"\n\
-      openssl x509 -req -sha256 -days 365 -in dashboard.csr -signkey dashboard.key -out dashboard.crt\n\
-      export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\n\
-      rm dashboard.csr\n\
-      kubectl --token $TOKEN -n kube-system delete secret kubernetes-dashboard-certs\n\
-      kubectl --token $TOKEN -n kube-system create secret generic kubernetes-dashboard-certs --from-file=/certs\n\
+      while true; do\n\
+        if [[ -f dashboard.key && -f dashboard.crt ]]; then\n\
+          export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\n\
+          if ! kubectl --token $TOKEN -n kube-system get secret kubernetes-dashboard-certs -o=custom-columns=NAME:.data >/dev/null; then\n\
+            kubectl --token $TOKEN -n kube-system delete secret kubernetes-dashboard-certs\n\
+            kubectl --token $TOKEN -n kube-system create secret generic kubernetes-dashboard-certs --from-file=/certs\n\
+          else\n\
+            break\n\
+          fi\n\
+        else\n\
+          openssl req -nodes -newkey rsa:2048 -keyout dashboard.key -out dashboard.csr -subj "/C=/ST=/L=/O=/OU=/CN=kubernetes-dashboard"\n\
+          openssl x509 -req -sha256 -days 365 -in dashboard.csr -signkey dashboard.key -out dashboard.crt\n\
+          rm dashboard.csr\n\
+          echo $(ls -lah /certs)\n\
+        fi\n\
+      done\n\
     fi\n\' >> /bootstrap.sh
 
 CMD ["/bin/bash", "bootstrap.sh"]
--- a/src/deploy/init/Dockerfile-amd64
+++ b/src/deploy/init/Dockerfile-amd64
@@ -32,23 +32,11 @@ RUN apt-get clean && rm -rf /var/lib/apt/lists/*
 RUN echo '#!/bin/bash\n\
     if [ -z "$(ls -A /certs)" ]; then\n\
       echo "Creating self-signed certificates"\n\
+      export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\n\
+      kubectl --token $TOKEN -n kube-system delete secret kubernetes-dashboard-certs\n\
       cd /certs\n\
-      while true; do\n\
-        if [[ -f dashboard.key && -f dashboard.crt ]]; then\n\
-          export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\n\
-          if ! kubectl --token $TOKEN -n kube-system get secret kubernetes-dashboard-certs -o=custom-columns=NAME:.data >/dev/null; then\n\
-            kubectl --token $TOKEN -n kube-system delete secret kubernetes-dashboard-certs\n\
-            kubectl --token $TOKEN -n kube-system create secret generic kubernetes-dashboard-certs --from-file=/certs\n\
-          else\n\
-            break\n\
-          fi\n\
-        else\n\
-          openssl req -nodes -newkey rsa:2048 -keyout dashboard.key -out dashboard.csr -subj "/C=/ST=/L=/O=/OU=/CN=kubernetes-dashboard"\n\
-          openssl x509 -req -sha256 -days 365 -in dashboard.csr -signkey dashboard.key -out dashboard.crt\n\
-          rm dashboard.csr\n\
-          echo $(ls -lah /certs)\n\
-        fi\n\
-      done\n\
+      openssl req -x509 -nodes -newkey rsa:2048 -keyout dashboard.key -out dashboard.crt -days 365 -subj "/C=/ST=/L=/O=/OU=/CN=kubernetes-dashboard"\n\
+      kubectl --token $TOKEN -n kube-system create secret generic kubernetes-dashboard-certs --from-file=/certs\n\
     fi\n\' >> /bootstrap.sh
 
 CMD ["/bin/bash", "bootstrap.sh"]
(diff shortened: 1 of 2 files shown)


floreks:
This doesn't fix the real issue which is in the core. Original code already checks if `certs` dir is not empty
```bash
if [ -z "$(ls -A /certs)" ]; then\n\
```
and only then it tries to generate certificates and store them in a secret. There should never be a situation where you create secret with `--from-file` flag and secret ends up empty in kubernetes. I have already reported an issue in the core https://github.com/kubernetes/kubernetes/issues/53382. We have similar issue with tests in the core and it can be always reproduced when resource limits are too low. I would not change current logic as according to documentation it is correct and should work.

cheld:
I guess it wouldn't hurt to have a temporary work-around until the root cause is fixed - assuming the retry helps

floreks:
It won't help when resource limits are low. I have tried that on my local environment and no matter how many times I've redeployed dashboard init container could not recreate secret with certs. The real solution is to set high resource limits for Dashboard so init container actually will be able to create secret correctly.

camilb:
Hi @floreks I'm running kubernetes v1.7.8 and tested with resources as low as: 

        resources:
          limits:
            memory: 50Mi
            cpu: 10m
          requests:
            cpu: 5m
            memory: 30Mi

The secret is not created on the first attempt, but successfully created on the second attempt. From what I see it's not a resource limit problem in my case. Then, I think it helps having a method to reattempt the creation of secret in case it's empty.

Here I have a screen recording from my environment: https://youtu.be/w4e44wGCs9A

floreks:
@camilb It's a bit weird as for you it worked after first attempt actually, even with low resource limits. I have checked it on k8s v1.7.7 and with such limits it was always empty.

camilb:
@floreks 
```bash  
if [ -z "$(ls -A /certs)" ]; then\n\
```
works well only to check the presence of the `/certs` directory

Here it's what happens when the directory is empty:
screen shot 2017-10-13 at 12 24 32

floreks:
@camilb 
```bash
floreks@floreks-CELSIUS-W530-Power:~$ ls
certs                   dashboard.yaml  Muzyka    Projects   Szablony        Wideo
dashboard-admin.yaml    Dokumenty       Obrazy    Publiczny  test
dashboard-certs.yaml    job.yaml        Pobrane   Pulpit     test.yaml
dashboard-ingress.yaml  kubeconfig      Programs  script.sh  VirtualBox VMs
floreks@floreks-CELSIUS-W530-Power:~$ ls -la test
razem 16
drwxrwxr-x  2 floreks floreks  4096 paź 13 12:56 .
drwx------ 46 floreks floreks 12288 paź 13 12:59 ..
floreks@floreks-CELSIUS-W530-Power:~$ cat script.sh 
#!/bin/bash

if [ -z "$(ls -A test)" ];then
  echo empty
else
  echo not empty
fi
floreks@floreks-CELSIUS-W530-Power:~$ ./script.sh 
empty
floreks@floreks-CELSIUS-W530-Power:~$ touch test/a
floreks@floreks-CELSIUS-W530-Power:~$ ./script.sh 
not empty
```

Check is correct. Certs are only generated if tested directory is empty.

camilb:
@floreks yes, you're right, sorry. I was thinking on checking for the presence of certificates before creating the secret.

Got confused by your previous comment:

> This doesn't fix the real issue which is in the core. Original code already checks if certs dir is not empty

Another fix that worked in my case was to add `sleep 5` after the openssl generates the certificates. I think the secret it's created  before openssl finishes. I'll try to debug this later and test it with more versions of kubernetes. Currently running 1.7.6, 1.7.8 and 1.8.1.

floreks:
> I think the secret it's created before openssl finishes

I was also thinking about this case but it should never happen. Script is executed synchronously. I don't see any reason why commands would be executed in parallel.

floreks:
@camilb Try this yaml. It always fails for me. Empty secret is recreated. With very low limits it actually works. It looks like mid range fails.

https://pastebin.com/raw/6ZiVzum1

camilb:
@floreks It fails for me also

camilb:
Increasing the limits didn't help.

         resources:
          limits:
            cpu: 2000m
            memory: 4000Mi
          requests:
            cpu: 1000m
            memory: 1000Mi

Still fails, with an empty secret.

floreks:
This is the reason:
```bash
Creating self-signed certificates
Generating a 2048 bit RSA private key
.............................+++
.......................................................................................+++
writing new private key to 'dashboard.key'
-----
No value provided for Subject Attribute C, skipped
No value provided for Subject Attribute ST, skipped
No value provided for Subject Attribute L, skipped
No value provided for Subject Attribute O, skipped
No value provided for Subject Attribute OU, skipped
Signature ok
subject=/CN=kubernetes-dashboard
Getting Private key
Error opening Private key dashboard.key
140576381150872:error:02001002:system library:fopen:No such file or directory:bss_file.c:398:fopen('dashboard.key','r')
140576381150872:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:400:
unable to load Private key
9810_13_10_15_02_30.281218451
-rw-r--r--  1 root root    0 Oct 13 13:02 dashboard.crt
total 4
drwxrwxrwt  3 root root  100 Oct 13 13:02 .
drwxr-xr-x 49 root root 4096 Oct 13 13:02 ..
drwxr-xr-x  2 root root   40 Oct 13 13:02 ..109810_13_10_15_02_30.281218451
lrwxrwxrwx  1 root root   33 Oct 13 13:02 ..data -> ..109810_13_10_15_02_30.281218451
-rw-r--r--  1 root root    0 Oct 13 13:02 dashboard.crt
rm: cannot remove 'dashboard.csr': No such file or directory
secret "kubernetes-dashboard-certs" deleted
secret "kubernetes-dashboard-certs" created
```

camilb:
Tested several times and depending on which node it's scheduled, sometimes it works, sometimes it doesn't. I think openssl generates faster the certificates with more resources and they are available before the secret is created. Which should not happen, but got this issue with bash several times, when trying to execute commands that depend on each-other in a row.

floreks:
I have simplified generation to single command and behavior got even weirder.

Script
```bash
echo "Creating self-signed certificates"\n\
cd /certs\n\
openssl req -x509 -nodes -newkey rsa:2048 -keyout dashboard.key -out dashboard.crt -days 365 -subj "/C=/ST=/L=/O=/OU=/CN=kubernetes-dashboard"\n\
ls -la \n\
export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\n\
ls -la \n\
kubectl --token $TOKEN -n kube-system delete secret kubernetes-dashboard-certs\n\
ls -la \n\
kubectl --token $TOKEN -n kube-system create secret generic kubernetes-dashboard-certs --from-file=/certs\n\
```

Output
```bash
Creating self-signed certificates
Generating a 2048 bit RSA private key
...............................+++
..................................................................+++
writing new private key to 'dashboard.key'
-----
No value provided for Subject Attribute C, skipped
No value provided for Subject Attribute ST, skipped
No value provided for Subject Attribute L, skipped
No value provided for Subject Attribute O, skipped
No value provided for Subject Attribute OU, skipped
total 12
drwxrwxrwt  2 root root   80 Oct 13 13:20 .
drwxr-xr-x 49 root root 4096 Oct 13 13:20 ..
-rw-r--r--  1 root root 1123 Oct 13 13:20 dashboard.crt
-rw-r--r--  1 root root 1708 Oct 13 13:20 dashboard.key
total 12
drwxrwxrwt  2 root root   80 Oct 13 13:20 .
drwxr-xr-x 49 root root 4096 Oct 13 13:20 ..
-rw-r--r--  1 root root 1123 Oct 13 13:20 dashboard.crt
-rw-r--r--  1 root root 1708 Oct 13 13:20 dashboard.key
secret "kubernetes-dashboard-certs" deleted
total 4
drwxrwxrwt  3 root root   80 Oct 13 13:20 .
drwxr-xr-x 49 root root 4096 Oct 13 13:20 ..
drwxr-xr-x  2 root root   40 Oct 13 13:20 ..109810_13_10_15_20_28.459255233
lrwxrwxrwx  1 root root   33 Oct 13 13:20 ..data -> ..109810_13_10_15_20_28.459255233
```
They are there and suddenly not...

floreks:
I may have found fix... No idea why it works but seems to be working fine. Check my init container from dockerhub `floreks/init`.

camilb:
@floreks  Can you please try this yaml: https://pastebin.com/raw/wGRJfL9Q

Or create a build with from the modified Dockerfile from this PR.
For me it never fails. Just tested on all my clusters with very low limits.

camilb:
@floreks didn't see the last comment, I'll test it.

floreks:
> Can you please try this yaml: https://pastebin.com/raw/wGRJfL9Q

It is not available.

floreks:
Ok, you can update dockerfiles or I can do it. We will build and push new images then.

camilb:
@floreks Done, thanks for support!

floreks:
@camilb also thanks for help!

huzhengchuan:
I meet the same question. I think the reason is like.
Becase the /certs directory is sercert kubernetes-dashboard-certs volumeMouted. and when your delete secert kubernetes-dashboard-certs. the dashboard.key/dashboard.crt in /certs will loss. so when your create secert again. the /certs directory is empty.

camilb:
@huzhengchuan Just checked the deployment and you are right. I think we can create another `emptyDir` volume mounted only in `initContainers` to store the certificates generated by openssl, then create the  secret from there. This will definitely solve the issue.  Thanks!

floreks:
@camilb this is not necessarily the case. It makes sense but it worked with very low resource limits and for me it also worked without them. There has to be something else also related to this. @huzhengchuan's description might be only partly true.

PR #2496, opened 2017-10-18
Introduce logs autorefresh/follow

### LINKED ISSUE #2083: Add option to refresh logs
##### Environment
```
Dashboard version: 1.6.1
Kubernetes version: 1.5.2
Operating system: Centos 7.3
```

I'm running the latest kube-dashboard, and my devs have been complaining that the logs viewer is clunky, especially how logs are grouped into pages.
Is it possible to enable endless scrolling of logs somehow, or have (finer) control over the log time period?

maciaszczykm:
image

Related to #1828.

### DIFF
--- a/src/app/frontend/logs/component.js
+++ b/src/app/frontend/logs/component.js
@@ -145,6 +145,17 @@ export class LogsController {
   }
 
   /**
+   * Refreshes logs view.
+   * @export
+   */
+  refresh() {
+    this.loadView(
+        this.currentSelection.logFilePosition, this.currentSelection.referencePoint.timestamp,
+        this.currentSelection.referencePoint.lineNum, this.currentSelection.offsetFrom,
+        this.currentSelection.offsetTo);
+  }
+
+  /**
    * Downloads and loads slice of logs as specified by offsetFrom and offsetTo.
    * It works just like normal slicing, but indices are referenced relatively to certain reference
    * line.
--- a/src/app/frontend/logs/component.js
+++ b/src/app/frontend/logs/component.js
@@ -35,10 +35,11 @@ export class LogsController {
    * @param {!angular.$document} $document
    * @param {!angular.$resource} $resource
    * @param {!angular.$interval} $interval
+   * @param {!angular.$log} $log
    * @param {!../common/errorhandling/dialog.ErrorDialog} errorDialog
    * @ngInject
    */
-  constructor(logsService, $sce, $document, $resource, $interval, errorDialog) {
+  constructor(logsService, $sce, $document, $resource, $interval, $log, errorDialog) {
     /** @private {!angular.$sce} */
     this.sce_ = $sce;
 
@@ -50,6 +51,9 @@ export class LogsController {
 
     /** @private {!angular.$interval} */
     this.interval_ = $interval;
+
+    /** @private {!angular.$log} */
+    this.log_ = $log;
 
     /** @export {!./service.LogsService} */
     this.logsService = logsService;
@@ -99,16 +103,13 @@ export class LogsController {
     /** @export {!kdUiRouter.$transition$} */
     this.$transition$;
 
-    /** @type {boolean} */
+    /** @export {boolean} */
     this.isFollowing = false;
 
-    // TODO(maciaszczykm): Add time interval to settings page.
-    this.interval_(() => {
-      if (this.isFollowing) {
-        // TODO Add log.
-        this.loadNewest();
-      }
-    }, 5000);
+    /** @export {number} Refresh interval in miliseconds. */
+    this.refreshInterval = 5000;
+
+    this.registerIntervalFunction_();
   }
 
 
@@ -118,6 +119,20 @@ export class LogsController {
     this.stateParams_ = this.$transition$.params();
     this.updateUiModel(this.podLogs);
     this.topIndex = this.podLogs.logs.length;
+  }
+
+  /**
+   * Registers interval function used to automatically refresh logs.
+   *
+   * @private
+   */
+  registerIntervalFunction_() {
+    this.interval_(() => {
+      if (this.isFollowing) {
+        this.loadNewest();
+        this.log_.info('Automatically refreshed logs');
+      }
+    }, this.refreshInterval);
   }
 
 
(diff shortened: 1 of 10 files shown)


maciaszczykm:
I will switch it to a basic version of follow feature.

src/app/frontend/logs/component.js
```diff
@@ -92,8 +96,19 @@ export class LogsController {
...
+    /** @type {boolean} */
+    this.isFollowing = false;
+
+    // TODO(maciaszczykm): Add time interval to settings page.
+    this.interval_(() => {
+      if (this.isFollowing) {
+        // TODO Add log.
+        this.loadNewest();
+      }
+    }, 5000);
```
cheld: Actually, I had implemented this and @floreks did not like it :)

Anyway, I would reduce the time to at least one second. In addition it would be possible to increase the performance and and reduce the load of the apiserver by limiting the api call to 100 lines.

floreks: True :), I don't like polling very much and would prefer exposing watch endpoint directly from apiserver. It would allow "live refresh" without causing too much load on our backend and on apiserver.

If we proceed with that as a temporary solution I'd rather keep refresh number rather high (5s+) not to cause too much load. We can make this option global when settings page will be introduced and give option to choose from few predefined intervals.

maciaszczykm: I was not aware. I consider is a temporary solution too. First I have started with just a refresh button, but this is a bit better and we had few requests for it. Next step will be adding refresh interval to the settings page, then we can think about a more clever solution.

cheld: The watch semantic seems not to be fully correct. If a user presses the refresh button we need not only all updates, but also all updates that happend between loading and pressing buttion. Anyway, the API provides a follow flag that keeps the stream open. I was thinking to pass this stream to the frontend. However, to make this work it would require to return a stream of json objects (or text directly) to the frontend instead of one. 

From UX perspective I am not fully convinced if 5s or settings page is reasonable. Lets imagine a use case:  You are devops, start a server and wait for a log output... I would be bothered with 5s. I guess in such a case a refresh button without auto-refresh is better. So a user just click multiple times himself to refresh. Certainly not perfect, but maybe least cumbersome.

maciaszczykm: I think both make sense, but I would keep this minimal auto-refresh at the moment. It makes first call to the API on click and then every 5 secs, so the user should be able to see what happens. I will add time interval value to the tooltip, so it will be better to understand.

floreks: > The watch semantic seems not to be fully correct. If a user presses the refresh button we need not only all updates, but also all updates that happend between loading and pressing button.

In our case since we are using pagination, we do not really care about data in the middle during refresh. Refresh means for us "load last logs page and show it to the user". 

> From UX perspective I am not fully convinced if 5s or settings page is reasonable. Lets imagine a use case: You are devops, start a server and wait for a log output... I would be bothered with 5s.

As a dev I'm more concerned about putting too much load on backend and apiserver by decreasing this number. From UX perspective it is still an improvement since there is no option to auto-refresh logs right now. We need some compromise between performance and usability. Even though it might not be perfect for containers that produce logs very quickly, it will still be ok in many cases.

> I guess in such a case a refresh button without auto-refresh is better. So a user just click multiple times himself to refresh. Certainly not perfect, but maybe least cumbersome.

Refresh normally loads last portion of logs and button to do this is already there so adding yet another button that does the same thing is not that nice. User can simply click on last page button multiple times and see how logs are "refreshing". That's what I'm usually doing.

cheld: >In our case since we are using pagination, we do not really care about data in the middle during refresh. Refresh means for us "load last logs page and show it to the user".

Long term, clicking the button should do 'refresh + follow'. Luckily, the api server is exactly doing that. We just have to refactor and expose a similar behavior to the front end.

src/app/frontend/logs/component.js
```diff
@@ -92,8 +100,16 @@ export class LogsController {
...
+    /** @export {!kdUiRouter.$transition$} */
     this.$transition$;
+
+    /** @export {boolean} */
+    this.isFollowing = false;
+
+    /** @export {number} Refresh interval in miliseconds. */
+    this.refreshInterval = 5000;
+
+    this.registerIntervalFunction_();
```
floreks: Move to `onInit`

src/app/frontend/logs/logs.html
```diff
@@ -71,6 +71,25 @@
         </md-icon>
       </md-button>
       <md-button class="kd-logs-toolbar-button"
+                 ng-click="ctrl.toggleLogFollow()">
+        <md-icon md-font-library="material-icons"
```
floreks: `md-font-library` is obsolete as we are only using `material-icons` set and it is our default.

src/app/frontend/logs/component.js
```diff
@@ -92,8 +100,16 @@ export class LogsController {
     /** @private {!ui.router.$stateParams} */
     this.stateParams_;
 
-    /** @export {!kdUiRouter.$transition$} - initialized from resolve */
+    /** @export {!kdUiRouter.$transition$} */
     this.$transition$;
+
+    /** @export {boolean} */
+    this.isFollowing = false;
```
cheld: All other buttons store the state in the logsService. Not sure if this service is very beneficial, but until we refactor I would suggest to keep it consistent

src/app/frontend/logs/logs.html
```diff
@@ -51,47 +51,58 @@
...
+               width="32px"
+               height="32px"
+               viewBox="0 0 32 32">
+            <line class="kd-cross-line-white" x1="9" x2="27" y1="9" y2="27"/>
+            <line class="kd-cross-line-black" x1="9" x2="27" y1="10" y2="28"/>
+          </svg>
         </md-icon>
+        <md-tooltip md-delay="500"
+                    md-autohide>
+          [[Toggle auto-refresh (every {{ctrl.refreshInterval / 1000}} seconds)|]]
```
cheld: Description missing?

maciaszczykm: Oh, I did not notice it before merging (I was already on the page). We do not need description everywhere. Translators will see English entry anyways.

PR #2499, opened 2017-10-19
change maintainer to label

ben-st:
i changed maintainer to label, since the former is deprecated.
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

All the best,
Benjamin


floreks:
Thanks! LGTM.

PR #2497, opened 2017-10-18
Fix overview button and center toolbar buttons

floreks:
- Center toolbar icon buttons
- Fix glitch with black text color for a second during page load

I'd also update overview icon. Here are few proposals.

### Current
zrzut ekranu z 2017-10-18 17-29-58

### Proposals
zrzut ekranu z 2017-10-18 17-30-11
zrzut ekranu z 2017-10-18 17-32-17
zrzut ekranu z 2017-10-18 17-34-49

### Before
zrzut ekranu z 2017-10-18 17-36-22

### After
zrzut ekranu z 2017-10-18 17-36-13

PR #2512, opened 2017-10-24
Change namespace overview icon

maciaszczykm:
https://github.com/kubernetes/dashboard/pull/2497#issuecomment-338814003

### DIFF
--- a/src/app/frontend/common/components/actionbar/actionbarnamespaceoverview.html
+++ b/src/app/frontend/common/components/actionbar/actionbarnamespaceoverview.html
@@ -16,7 +16,7 @@ limitations under the License.
 
 <md-button ng-click="$ctrl.goToNamespaceOverview()"
            class="kd-actionbar-button">
-  <md-icon>public</md-icon>
+  <md-icon>description</md-icon>
   <md-tooltip md-delay="500"
               md-autohide>
     [[Go to namespace overview page|]]


maciaszczykm:
@cheld Please take a look.

cheld:
So it now this icon

https://material.io/icons/#ic_description

cheld:
hm...I will also have a look

cheld:
https://material.io/icons/#ic_explore

cheld:
In case we consider it like a bookmark (I have not tested the code)
https://material.io/icons/#ic_bookmark_border

maciaszczykm:
I like compass icon, bookmark not so much. However, I still think that we should follow @rahuldhide suggestions here. Mainly because we should try to be consistent with other pages around the web.

cheld:
@rahuldhide suggested a view-kind-of-icon, while description is a document.

cheld:
Ok, I tested the functionality. It actually switches the namespace selector to the current namespaces and shows all resources of that namespace. So my proposal would be

Icon: https://material.io/icons/#ic_open_in_browser
Label: Show
Tooltip: "Show all resources of this namespace"

maciaszczykm:
@cheld `Show` is too generic.

cheld:
Thats complicated :-(
overview1
overview2
overview3
overview4

src/app/frontend/common/components/actionbar/actionbarnamespaceoverview.html
```diff
@@ -16,7 +16,7 @@
 
 <md-button ng-click="$ctrl.goToNamespaceOverview()"
            class="kd-actionbar-button">
-  <md-icon>public</md-icon>
+  <md-icon>description</md-icon>
   <md-tooltip md-delay="500"
               md-autohide>
     [[Go to namespace overview page|]]
```
cheld: Please add descriptions

maciaszczykm:
`Select` should only select namespace, currently we redirect to overview.

`Show` is too generic.

`Resources` is a bit more like `Overview`, I like second more as it is visible in the menu and user can expect what will happen.

`Open` is a interesting idea.

src/app/frontend/common/components/actionbar/actionbarnamespaceoverview.html
maciaszczykm: I will.

cheld:
@maciaszczykm please try 'select' and 'open' a couple of times in the application. The experience is a bit different when using than just looking at static content.

PR #2532, opened 2017-10-31
Introduce global settings

maciaszczykm:
- [x] add settings manager
- [x] add global settings endpoint
- [x] add settings view
- [x] add settings menu entry
- [x] handle concurrent settings change error (dialog with save anyways and refresh options)
- [x] merge settings during save to avoid any data loss
- [x] fix tests
- [x] add tests
- [x] create service to retrieve specific settings values and use it across the app
- [x] generate translations
- [x] service refresh after settings save
- [x] handle cluster name setting
- [x] handle items per page setting
- [x] remove bower leftovers
- [x] fix caching problem


maciaszczykm:
Current progress:

image

maciaszczykm:
Current progress:

image

maciaszczykm:
It is ready for an initial review. I have to fix few things in tests/annotations etc.

src/app/backend/settings/manager.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return s
+}
+
+// GetGlobalSettings implements SettingsManager interface. Check it for more information.
+func (sm *SettingsManager) SaveGlobalSettings(client kubernetes.Interface, s *api.Settings) error {
+	cm, isDiff := sm.load(client)
+	if isDiff {
+		return errors.New("settings changed since last reload")
```
floreks: Use `LocalizerError` method here so we can localize it in frontend.

maciaszczykm: It is not displayed to the user but used to display "save anyways" dialog.

maciaszczykm: I will comment it in the code.

floreks:
Overall LGTM.

src/app/backend/settings/manager.go
maciaszczykm: Done.

maciaszczykm:
PTAL

PR #2537, opened 2017-11-02
Added note that URL shortcut is not working

cheld:


README.md
```diff
@@ -29,12 +29,16 @@ To start using Dashboard run following command:
 ```sh
 $ kubectl proxy
 ```
+This command creates a secure channel with your Kubernetes cluster and provides local access
```
floreks: This should be moved up, before user executes `kubectl proxy` command.

cheld:
PTAL

PR #2544, opened 2017-11-06
Support for User Links add through annotations for PersistentVolumes

harisanker1780:
The commit on this PR add the user customizable links support for persistentvolume.
This is an extenstion on the PR https://github.com/kubernetes/dashboard/pull/2249.

Commit https://github.com/kubernetes/dashboard/pull/2544/commits/78c625d69a64cf4ae23346c41b780866bf22a9a2  has the changes i made on top of parent PR.

preview of the change
screenshot from 2017-11-02 15-46-27

Any suggestions will be greatly appreciated.


maciaszczykm:
I agree with @cheld that it is not flexible enough. Plus the UI needs some improvements because it is not consistent with anything else.

harisanker1780:
What we are trying to achieve with this PR is mainly to show links to Pods and Services, https://github.com/kmova will comment on this more.

btw.. can you tell me what changes i need to make so that the UI consistent with other pages(assuming it is userlinks in Pod)
one change i tried is moved the userlinks section to right side, see the preview of the change with and without userlinks 
screenshot from 2017-11-08 12-40-58
screenshot from 2017-11-08 12-41-19

harisanker1780:
Updated getPersistentVolumeLinks method-- removed check of apiserverProxyURL and pvDNSName
In current changes user has to provide complete uri that he wants to display in userlinks section.

This is useful for us in scenarios like show link to prometheus/grafana - that can provide detailed metrics on the storage.

PR #2559, opened 2017-11-09
Set user-agent for client-go and add --version flag

willstudy:
Set user-agent for client-go.

before
```
client="dashboard/v1.7.1 (linux/amd64) kubernetes/$Format"
```
after
```
client="dashboard/v1.7.1"
```
And add `--version` flag
```
root@vv-node1:~# docker exec -it 956c8bc970cd /dashboard --version
2017/11/09 08:42:28 Starting overwatch
dashboard v1.7.1
```

### DIFF
--- a/build/conf.js
+++ b/build/conf.js
@@ -76,6 +76,12 @@ const imageNameBase = 'kubernetes-dashboard';
  */
 export default {
   /**
+   * the expression of recording version info into src/app/backend/client/manager.go
+   */
+  recordVersionExpression:
+      `-X github.com/kubernetes/dashboard/src/app/backend/client.Version=${version.release}`,
+
+  /**
    * Configuration for container registry to push images to.
    */
   containerRegistry: containerRegistry,
--- a/build/backend.js
+++ b/build/backend.js
@@ -35,6 +35,9 @@ gulp.task('backend', ['package-backend'], function(doneFn) {
         'build',
         // Install dependencies to speed up subsequent compilations.
         '-i',
+        // record version info into src/version/version.go
+        '-ldflags',
+        conf.recordVersionExpression,
         '-o',
         path.join(conf.paths.serve, conf.backend.binaryName),
         conf.backend.mainPackageName,
@@ -116,6 +119,9 @@ function backendProd(outputBinaryPathsAndArchs) {
             '-a',
             '-installsuffix',
             'cgo',
+            // record version info into src/version/version.go
+            '-ldflags',
+            conf.recordVersionExpression,
             '-o',
             path,
             conf.backend.mainPackageName,
--- a/src/app/backend/client/manager.go
+++ b/src/app/backend/client/manager.go
@@ -42,7 +42,12 @@ const (
 	DefaultCmdConfigName = "kubernetes"
 	// Header name that contains token used for authorization. See TokenManager for more information.
 	JWETokenHeader = "jweToken"
+	// Default http header for user-agent
+	DefaultUserAgent = "dashboard"
 )
+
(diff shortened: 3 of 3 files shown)


floreks:
@willstudy just use `git commit --amend` if you want to restart travis.

floreks:
I'm wondering if we actually need version flag. Dashboard is not intended to be used as CLI app. It is a webapp and version information can be found on about page. I think that vast majority of users never used this binary directly and only used docker container to run it. I'd leave user-agent change and remove version flag.

@maciaszczykm WDYT?

willstudy:
@maciaszczykm

maciaszczykm:
I agree with @floreks.

src/app/backend/version/version.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"fmt"
+	"os"
+	"strconv"
+
+	flag "github.com/spf13/pflag"
+)
+
+// VERSION of this binary
+var VERSION = "UNKNOWN"
```
floreks: This can be moved to main `dashboard.go` and `ldflags` adjusted. Rest of the code can be removed.

willstudy: How about `Version` defined in the `src/app/backend/client/manager.go`?

floreks: `client/manager` holds the code for creation of Kubernetes API client. It's not a place for a version of our binary. `dashboard.go` is our main file and is better suited for this.

willstudy: If `Version` defined in the `dashboard.go`, how to reference this `Version` in `src/app/backend/client/manager.go` ? Should I add a `userAgent` field for `clientManager`, just like 
```
type clientManager struct {
	// Autogenerated key on backend start used to secure requests from csrf attacks
	csrfKey string
	// Path to kubeconfig file. If both kubeConfigPath and apiserverHost are empty
	// inClusterConfig will be used
	kubeConfigPath string
	// Address of apiserver host in format 'protocol://address:port'
	apiserverHost string
	// Initialized on clientManager creation and used if kubeconfigPath and apiserverHost are
	// empty
	inClusterConfig *rest.Config
	// Responsible for decrypting tokens coming in request header. Used for authentication.
	tokenManager authApi.TokenManager
	// Kubernetes client created without providing auth info. It uses permissions granted to
	// service account used by dashboard or kubeconfig file if it was passed during dashboard init.
	insecureClient kubernetes.Interface
       // add user agent
       userAgent string
}
```
And rewrite the `NewClientManager` function, just like 
```
func NewClientManager(kubeConfigPath, apiserverHost, userAgent string) ClientManager {
	result := &clientManager{
		kubeConfigPath: kubeConfigPath,
		apiserverHost:  apiserverHost,
                userAgent: userAgent,
	}

	result.init()
	return result
}
```
So we can pass the `Version` defined in `dashboard.go` to the `src/app/backend/client/manager.go`.

floreks: Ahh, ok there would be circular dependency if we define version in `dashboard.go` and import it there. I am ok with moving it to `client/manager.go` then.

willstudy: What's the next? Should I modify the code again?

build/conf.js
```diff
@@ -76,6 +76,12 @@ const imageNameBase = 'kubernetes-dashboard';
  */
 export default {
   /**
+   * the expression of recording version info into src/app/version/version.go
```
floreks: `src/app/version/version.go` wrong path.

floreks:
Last comment and we can merge.

PR #2520, opened 2017-10-25
Link to PersistentVolume and StorageClass from PersistentVolumeClaim

harisanker1780:
The commit on this PR add a link to PersistentVolume details and StorageClass details which are associated with given PersistentVolumeClaim.

Any suggestions will be greatly appreciated.

PR #2560, opened 2017-11-09
Link to PersistentVolumeClaim in PersistentVolume List.

harisanker1780:
The commit on this PR add a link to PersistentVolumeClaim which are associated with given PersistentVolumes in the PersistentVolume list.

preview of the change
screenshot from 2017-11-09 17-45-26


harisanker1780:
#2520

floreks:
I prefer #2520 change. There might be multiple claims associated with PV and displaying such list on list page might be too much. We do not have that much space to do it nicely. I'd rather keep it on details page only.

PR #2565, opened 2017-11-10
Add node ip to node detail page

### LINKED ISSUE #2555: add hostip in node detail page
hostname is can't ssh directly ,so can you add hostip in node detail page

floreks:
Closes #2555 

zrzut ekranu z 2017-11-10 13-02-25

Also removed external id as it is deprecated.
zrzut ekranu z 2017-11-10 13-03-21


maciaszczykm:
In my opionion we should change names in the fly to add spaces and translation or change it to `Adresses` and display it in same format as labels, i.e. `internalIP: 127.0.0.1`.

floreks:
Changed (addresses string can be translated):
zrzut ekranu z 2017-11-10 14-40-48

PR #2581, opened 2017-11-15
Add auto-refresh time interval setting

maciaszczykm:


src/app/frontend/logs/component.js
```diff
@@ -123,12 +131,15 @@ export class LogsController {
...
-      if (this.logsService.getFollowing()) {
-        this.loadNewest();
-        this.log_.info('Automatically refreshed logs');
-      }
-    }, this.refreshInterval);
+    if (!this.isIntervalRegistered_) {
+      this.interval_(() => {
+        if (this.logsService.getFollowing()) {
+          this.loadNewest();
+          this.log_.info('Automatically refreshed logs');
```
floreks: I'd remove that to avoid unnecessary spam in a console.

src/app/frontend/logs/component.js
```diff
@@ -176,6 +187,7 @@ export class LogsController {
    * @export
    */
   toggleLogFollow() {
+    this.registerIntervalFunction_();
```
floreks: Move to `$onInit` and check if it was registered won't be needed.

src/app/frontend/logs/component.js
```diff
@@ -123,12 +131,15 @@ export class LogsController {
    * @private
    */
   registerIntervalFunction_() {
```
floreks: After starting it, interval function will be running in the background forever. I'd start/stop whole interval function on button click. It returns interval id and `interval.cancel(id)` can be used to stop it.

src/app/frontend/logs/component.js
```diff
@@ -176,6 +183,7 @@ export class LogsController {
    * @export
    */
   toggleLogFollow() {
+    this.toggleIntervalFunction_();
     this.logsService.setFollowing();
```
floreks: By swapping this line with toggling interval we can get rid of below if statement. Logs will be reloaded immediately after first interval start.

PR #2598, opened 2017-11-20
Improve workloads statuses graphs on overview page

### LINKED ISSUE #2593: the resource stauts chart of the overvie page should show all running pods
##### Environment
```
Dashboard version:  google_containers/kubernetes-dashboard-amd64:v1.7.1
Kubernetes version:  1.7.8
Operating system: coreos 1298.5.0
```

##### Steps to reproduce
1. Have dashboard running in the non-SSL mode.
2. Config service point to dashboard pod.
3. Config ingress point to the dashboard service.
4. Access dashboard through nginx-ingress-controller external ip.

##### Observed result
The overview page displays the status of the default namespace, but it only shows 'Running  15' pods at the 'Resource status' chart, while we actually see 19 pods at the Pod part. 

##### Expected result
The overview page should show 'Running  19' pod at the "Resource Status" chart.

floreks:
Fixes #2593 

### Changes
- Reorganized resources on overview page (sorted alphabetically per group)
- Added resource status ratio charts for all resource on overview page.
- Show information about status and number of related resources on chart hover, i.e. hovering over green color on pods chart can show `Running: 4`.
- Removed unnecessary `IsTypeFilled` calls on backend.
- Added `Status` struct to all resource list objects that contain basic info about the state of resources on the list.
- Made resource charts easier to customize.

There is an issue (visible on below ss) with graph not being fully filled. It's related to how nvd3 handles showing charts. There is not much we can do right now.

### TODO
- [ ] Add tests

### Before
zrzut ekranu z 2017-11-17 14-18-21

### After
zrzut ekranu z 2017-11-20 15-59-38


cheld:
Great!

src/app/backend/resource/common/status.go
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+// Status provides basic information about resources status on the list.
+type Status struct {
```
maciaszczykm: Change to `ResourceStatus`?

src/app/backend/resource/common/status.go
```diff
@@ -0,0 +1,30 @@
...
+	Running int32 `json:"running"`
+
+	// Number of resources that are currently in pending state.
+	Pending int32 `json:"pending"`
+
+	// Number of resources that are in failed state.
+	Failed int32 `json:"failed"`
+
+	// Number of resources that are in succeeded state.
+	Succeeded int32 `json:"succeeded"`
```
maciaszczykm: I would prefer `map[string]int32` here, because it may be misleading in case of Cron Jobs for example, which are Active or Suspended only. And these labels are not translated anyways.

src/app/backend/resource/cronjob/common.go
```diff
@@ -63,3 +64,17 @@ func FromCells(cells []dataselect.DataCell) []batch2.CronJob {
 	}
 	return std
 }
+
+func getStatus(cronJobList *batch2.CronJobList) common.Status {
+	info := common.Status{}
+
+	for _, cronJob := range cronJobList.Items {
+		if cronJob.Spec.Suspend != nil && !(*cronJob.Spec.Suspend) {
+			info.Running++
```
maciaszczykm: This is what I have mentioned in my earlier comment.

src/app/backend/resource/deployment/common.go
```diff
@@ -64,3 +67,23 @@ func fromCells(cells []dataselect.DataCell) []apps.Deployment {
...
+		podInfo := common.GetPodInfo(deployment.Status.Replicas, deployment.Spec.Replicas, matchingPods)
+		warnings := event.GetPodsEventWarnings(events, matchingPods)
+
+		if len(warnings) > 0 {
+			info.Failed++
+		} else if podInfo.Pending > 0 {
+			info.Pending++
+		} else {
+			info.Running++
+		}
```
maciaszczykm: Export to a function if it repeats somewhere?

src/app/backend/resource/common/status.go
maciaszczykm: There might be issue with coloring graph slices... Perhaps some constant values for common names like active, failed, running, pending etc. would make it better.

src/app/backend/resource/deployment/common.go
floreks: It does not

src/app/backend/resource/common/status.go
floreks: I was considering map but it requires introducing yet another portions of strings that would need to be hardcoded in backend and in frontend. Using generic names like that and mapping them to specific resources seems like better idea. Map would complicate the code even more.

maciaszczykm: Syncing frontend with the backend is required at the moment too. You need to keep in mind, that for Cron Job active = running and suspended = failed during development. Both have their own drawbacks. We can stick to the current implementation for now and make the release.

PR #2636, opened 2017-12-01
Update deploy view and allow upload via text input

### LINKED ISSUE #2530: Allow upload via text input
##### Comments

It would be nice if the dashboard allowed a text input field for yaml/json files. I could simply edit a pod, copy the contents into my clipboard, navigate to the CREATE page, and then paste the json into a text field. With a few modifications, I could then save the newly created app. This would be a huge convenience to have.

maciaszczykm:
- Use `md-tabs` instead of `ui-view` to display different deploy options.
- Refactor deploy from file component to use external component to upload used (it was used on login page).
- Update styling of deploy page.
- Add third deploy option - deploy from text input (closes #2530).
- Use `ui-ace` as a input field, it accepts both YAML and JSON formats and provides syntax highlighting.
- Export depoy logic to a service to avoid code duplication.
- Move translations from `.js` to `.html` files.
- Refactor deploy from settings component.
- Disable deploy button until required fields are filled.
- Fix Closure annotations.
- Remove redundant code and components.

image

image

image


maciaszczykm:
I will add some tests now. You can review it now.

src/app/frontend/deploy/deploy.html
```diff
@@ -15,26 +15,19 @@
...
+    <md-tabs md-dynamic-height
+             md-border-bottom
+             class="kd-deploy-tabs">
+      <md-tab label="[[Upload from input|Upload from input tab on deploy page.]]">
+        <kd-deploy-from-input></kd-deploy-from-input>
+      </md-tab>
+      <md-tab label="[[Upload from file|Upload from file tab on deploy page.]]">
+        <kd-deploy-from-file></kd-deploy-from-file>
+      </md-tab>
+      <md-tab label="[[Specify deployment details|Specify deployment details tab on deploy page.]]">
```
floreks: I'd consider some shorter and simpler title here i.e. 'Create deployment', 'Create new resource/deployment'

maciaszczykm: You can create deployments from a file and from text input, so I think it is too generic to put it here.

src/app/frontend/deploy/deploy.scss
```diff
@@ -12,19 +12,10 @@
...
+@import '../variables';
 
-.kd-upload-file-picker {
-  height: 0;
-  visibility: hidden;
-  width: 0;
+.kd-deploy-tabs {
+  md-tab-content {
+    padding: 2.5 * $baseline-grid;
+  }
```
floreks: I'd remove transition effect when switching tabs. It's not that fluent. It can be removed by using:
```css
[role="tabpanel"] {
  transition: none;
}
```

maciaszczykm: Sure, it looks even better.

src/app/frontend/deploy/deploy.html
floreks: Material writing guidelines say that used language should be concise, simple and direct. In this case first two tabs are fine. They simply say `Upload` which means that we are creating a new resource and `from x` to say that new resource is created based on some provide data.

The last tabs description is a bit misleading and IMO does not use simple and direct language. It says to specify details but does not stress out that it is a creation of a new resource based on those "specified details".

I have provided only sample descriptions. We can think of something else here.

src/app/frontend/deploy/deployfromsettings/deploylabel/component.js
```diff
@@ -134,20 +134,6 @@ export default class DeployLabelController {
   isFilled_(label) {
     return label.key.length !== 0 && label.value().length !== 0;
   }
-
-  /**
-   * @export
-   * @return {string}
-   */
-  getLabelKeyUniqueWarning() {
```
floreks: Is it not used anymore?

floreks:
Did you test deploy behavior when user has no privileges to create some resource in some namespace?

src/app/frontend/deploy/deploy.html
maciaszczykm: `Specify deployment details` is more clear to me than `Create deployment`. I considered a change to `Specify app details` but I suppose that you will not like it. I am not sure what is not simple or concise in `Specify deployment details` phase, so I will let you propose something else.

> It says to specify details but does not stress out that it is a creation of a new resource based on those "specified details".

It is a tab on create page. I think users will be aware that this is resource creation. Besides that upload is not equal to create too.

maciaszczykm:
@floreks I have not touched backend. This PR is already too big to apply more changes in it.

src/app/frontend/deploy/deploy.html
floreks: > It is a tab on create page. I think users will be aware that this is resource creation. Besides that upload is not equal to create too.

In this case `upload == create`. Usually, upload means that you upload file or some data somewhere where you can read it from later. In our case upload means create from file/yaml/json. After all, it is the same as `kubectl create -f xxx`. 

I'd rather try to be persistent here and try to use same template in all cases.
`Create from file`
`Create from input` or `Create from yaml/json`
`Create deployment`

There is also option to move creation info higher so on the form there would be title i.e. `Resource creator` and then show some options, `File`, `YAML/JSON`, `Deployment` or `App`(app is actually better because we are creating multiple resources, not only single Deployment).

src/app/frontend/deploy/deployfromsettings/deploylabel/component.js
maciaszczykm: Yes:

```
<ng-message when="unique">
        [[{{$ctrl.label.key}} is not unique|Label is not unique warning on deploy label component.]]
      </ng-message>
```

floreks:
I am just wondering if this works now correctly. I am not sure if we have actually tested it.

src/app/frontend/deploy/deploy.html
maciaszczykm: > There is also option to move creation info higher so on the form there would be title i.e. Resource creator and then show some options

Would need even more design work.

> I'd rather try to be persistent here and try to use same template in all cases.
>Create from file
>Create from input or Create from yaml/json
>Create deployment or Create an app

I have applied your suggestions, but I do not agree with them. I liked the previous version more. We can change it later.

floreks: @rahuldhide @danielromlein could you share your thoughts on this?

floreks:
@danielromlein @rahuldhide we could use your opinion on the new design.

Issue #2641, opened 2017-12-04
Login form compatibility with password managers


kachkaev:
After upgrading to dashboard 1.8.0 and making it available at https://dashboard.example.com/, I've decided to save a few tokens into my password manager to then quickly authenticate in the UI with different privileges. However, I've noticed that the dashboard's login form is incomparable with the password managers I tried ([1password](https://1password.com/) and [LastPass](https://www.lastpass.com/)). Usually password managers work fine with any website without any effort from its developers, so it seems like there's something trivial missing in the angular-controlled login form. I'd like to open this issue to start a discussion.

##### Environment
```
Dashboard version: 1.8.0
Kubernetes version:
Operating system: macos Firefox / Chrome with 1password and LastPass extensions
```

##### Steps to reproduce
0. Make the dashboard available at https://dashboard.example.com/

_scenario 1_
1. Open the dashboard and see the login screen
1. Paste a token into the form and click "sign in"

_scenario 2_
1. Open your password manager browser extension and manually add a new password entry for https://dashboard.example.com/ (I tried [1password](https://1password.com/) and [LastPass](https://www.lastpass.com/)).
1. Open the k8s dashboard and see the login screen
1. Click on the password manager's icon, select the existing entry with the just-created token and click on it to paste the data into the form.

##### Observed result
_scenario 1:_ the password manager does not suggest to save the token as a new password entry

_scenario 2:_ the token (which is saved as password) does not get pasted into the _token_ form field

##### Expected result
_scenario 1:_ the password manager suggests to save the token as a new password entry

_scenario 2:_ the token gets pasted into the _token_ form field of the login form and "sign in" is automatically pressed

##### Comments

Password managers like 1password and LastPass are widely used both by individuals and teams. They are recommended in quite a few security articles, so it is likely that their prevalence will grow in the forthcoming months. Keeping k8s tokens in a password manager seems like a right thing to do, because this way cluster admins can share access with their colleagues quite easily while keeping things secure and transparent.

It'd be great to have the dashboard login form compatible with the password managers, which probably just means that its elements should get a few extra HTML attributes. If anyone has any experience with making password managers happy, please share your thoughts!

maciaszczykm:
Thanks for reporting this issue. We will take a look on it.

floreks:
Great feature request. Simple, yet very helpful. We already have an idea why it is not working and should be easy to fix.

floreks:
[...]

PR #2642, opened 2017-12-04
Change the token login input field to password

farcaller:
This prevents the browsers from caching the value.

### DIFF
--- a/src/app/frontend/login/token.html
+++ b/src/app/frontend/login/token.html
@@ -19,6 +19,7 @@ limitations under the License.
     <label>[[Enter token|'Token' field label shown on login page for token auth]]</label>
     <input ng-model="$ctrl.token"
            ng-change="$ctrl.onTokenUpdate()"
-           required>
+           required
+           type="password">
   </md-input-container>
 </div>


maciaszczykm:
Nice change. I wondering about one thing - masking the token value. Is it needed?

I think we can also set `autocomplete` attribute to solve https://github.com/kubernetes/dashboard/issues/2641. Can you do it?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete

farcaller:
Done (although #2641 would be solved by the previous commit alone—that's the default behavior).

farcaller:
> I wondering about one thing - masking the token value. Is it needed?

I don't expect anyone to type the auth token manually, to be honest. There's very little value in showing a huge "random-looking" string to the end user as they won't be able to parse it in their mind either way.

maciaszczykm:
Fine, thanks. Can you also set input types to other login forms?

farcaller:
Basic auth form is already correct, as is kubeconfig one.

floreks:
From what I have checked autocomplete won't help us with password managers. In my case browser cached last password and used it to autocomplete the field. It can be removed. Password managers integration might be not easily doable here. It will work only partially.

farcaller:
> In my case browser cached last password and used it to autocomplete the field. It can be removed

That was my point too. I'll remove the second commit as the first one will do the job.

kachkaev:
Thanks for this PR @farcaller! Keen to check if password managers work a bit better with the form now 😃 Looking forward to for docker image `v1.8.1` or `v1.9.0`!

maciaszczykm:
@kachkaev Check dev release. You can find more info in our wikis.

kachkaev:
Good point @maciaszczykm! See my thoughts on the new experience here: https://github.com/kubernetes/dashboard/issues/2641#issuecomment-349607302.

PR #2663, opened 2017-12-12
chore(package): update clang-format to version 1.2.1

maciaszczykm:
Closes #2612


floreks:
There has been some changes in `1.2` version causing `gulp-clang-format` wrapper to fail. I think it was not adjusted to new changes in core `clang-format` package. We can close until wrapper will be updated.

PR #2671, opened 2017-12-13
Add imagePullPolicy for deployment yaml

riverzhang:
Add imagePullPolicy for deployment yaml


riverzhang:
@floreks @maciaszczykm

floreks:
Taken from kubernetes documentation:
> The default pull policy is IfNotPresent which causes the Kubelet to skip pulling an image if it already exists...

Closing as this is not needed.

PR #2679, opened 2017-12-14
Fix create an app

floreks:
There was an issue with creating an app when user had no permission to check if object exists but was able to create new objects.


src/app/frontend/deploy/deployfromsettings/deployfromsettings.html
```diff
@@ -178,7 +178,7 @@
       </div>
 
       <kd-user-help>
-        [[The specified labels will be applied to the created Replication Controller, Service (if any) and Pods. Common labels include release, environment, tier, partition and track.|User help for the labels section on the deploy from settings page.]]
+        [[The specified labels will be applied to the created Deployment, Service (if any) and Pods. Common labels include release, environment, tier, partition and track.|User help for the labels section on the deploy from settings page.]]
```
maciaszczykm: Generate translations.

src/app/backend/errors/handler.go
```diff
@@ -18,6 +18,7 @@ import (
 	"log"
 	"net/http"
 
+	"github.com/emicklei/go-restful"
```
maciaszczykm: Using `restful "github.com/emicklei/go-restful"` will let us avoid problems with `gulp format`.

floreks: For me it didn't really make any difference. Sometimes import gets still deleted. Maybe IntelliJ plugin removes aliases and then gulp format removes import.

maciaszczykm: My IntelliJ leaves them prefixed, let's use this format.

PR #2739, opened 2018-01-06
Fixing asset path resolution

rbwsam:
- When determining the path of the localized frontend assets, do
not look in current working directory, use the directory of the
executable. (Resolves symlinks to determine the directory of
the executable.)


floreks:
Nice work. Looks good. I will test that and merge on monday. We will be doing `v1.8.2` release next week so this should go in.

PR #2741, opened 2018-01-08
Remove symbol table and debug info from go binary to reduce image size.

oilbeater:


build/backend.js
```diff
@@ -37,7 +37,7 @@ gulp.task('backend', ['package-backend'], function(doneFn) {
         '-i',
         // record version info into src/version/version.go
         '-ldflags',
-        conf.recordVersionExpression,
+        conf.recordVersionExpression + ' -w -s',
```
cheld: This code is only executed during development. Not to build a release binary or docker image. I think we can keep debug information in this case.

oilbeater: That's right, we should keep them for development

floreks:
Fix error and we can merge.

https://travis-ci.org/kubernetes/dashboard/jobs/326317609

oilbeater:
error fixed @floreks

PR #2743, opened 2018-01-08
User customizeable links for persistentvolume

harisanker1780:
The commit on this PR add the user customizable links support for persistentvolume.

Closed the old PR #2544 which I created for the same functionality,
This PR adds only external links for persistentvolume, this does not have any userlinks support for pod and service like PR #2544 does. 

This is useful for us in scenarios like show link to prometheus/grafana - that can provide detailed metrics on the storage. 

preview of the change
screenshot from 2018-01-08 14-48-32

Any suggestions will be greatly appreciated.


floreks:
We'll revisit all new features after angular migration is done. We are currently on soft code freeze.

PR #2753, opened 2018-01-10
Update to v1.8.2

floreks:
https://github.com/kubernetes/dashboard/wiki/Release-procedures


rf232:
If i understand this correctly, tide should do automerge shouldn't it?

floreks:
I think that it should, however we can still do it manually if needed.

rf232:
my patience is limited (to approx 2.5 hours apparently)

Issue #2784, opened 2018-01-21
Change namespace dialog actions reversed


dzoeteman:
##### Environment

```
Dashboard version: master
Kubernetes version: 1.8.0
Operating system: Linux Mint
Node.js version: v6.11.2
Go version: 1.9.2
```

##### Steps to reproduce

- Open resource in namespace that is not in a different namespace
- Select a namespace from the select menu on the left
- Click "Yes" in the dialog asking you if you want to change your namespace.

##### Observed result
Clicking "Yes" leaves the user on the page. Clicking "No" switches the user to the different namespace.
The text "Do you want to change the namespace from x to y" has the namespace order reversed.

##### Expected result
Clicking "No" leaves the user on the page. Clicking "Yes" switches the user to the different namespace.
The text "Do you want to change the namespace from x to y" has the namespace order correctly.

floreks:
It is actually correct. The check is fired after changing namespace, not before. If you check namespace on the navbar and in URL then you can see that it was changed and you could see view reload animation during the change.

1. You are on a resource page that is in `kube-system` namespace.
2. Change namespace to `default`. Namespace is changed.
3. Check is fired and dialog is shown saying that currently selected resource is in a different namespace than currently selected (`default`).
4. You can either stay on this resource page and change namespace back to `kube-system` or keep namespace unchanged and get redirected to `overview` page.

![zrzut ekranu z 2018-01-22 09-18-30](https://user-images.githubusercontent.com/2285385/35211124-69be69a8-ff55-11e7-8095-72763d6ed3b7.png)
![zrzut ekranu z 2018-01-22 09-18-37](https://user-images.githubusercontent.com/2285385/35211126-6c0e6014-ff55-11e7-8aee-225595d28b42.png)

dzoeteman:
@floreks I understand it makes sense from a technical POV, but I don't see how that is useful from a user's POV. 
It would seem logical to me that if I change the namespace in the select box, that I want to change the namespace. When I then get a dialog asking me "Change namespace?" I expect that to be a confirmation asking if I'm sure to change the namespace to the namespace I selected. I was quite confused when I first saw this dialog, pressed "Yes" and nothing happened. 

Please see my PR #2785 so you can see what I changed in the behaviour, it might explain the issue a little better.

floreks:
You have changed the logic, but it is not true. The most important indicator is that state actually reloads when you change namespace. You can see that. While being on a new namespace you get message.
[...]

PR #2785, opened 2018-01-21
Fix namespace change dialog actions

dzoeteman:
The change namespace dialog actions seemed to be reversed for the user-expected behaviour (#2784). That is fixed in this PR.


dzoeteman:
With the discussion in #2784, I'll close this PR.

PR #2795, opened 2018-01-25
Auto-generate certs in-memory

liggitt:
The dashboard is currently writing autogenerated cert files to a directory managed by the kubelet secret mounter.

At best, those files are auto-removed shortly after being written and the server startup races with the removal.

At worst, the location is mounted readonly and the dashboard fails to start.

There's no reason to write files to a location where they will just be cleaned up later (and doing so is breaking the dashboard when the location is read-only... see https://github.com/kubernetes/kubernetes/pull/58720/files#diff-7ab88bb7654d97946d6328f11f29d177)

This PR changes the dashboard to keep auto-generated certs in memory


floreks:
/lgtm

This definitely helps, however we have bigger issue here. I thought that if we'd mount secret in RW mode then it would sync secret data in both ways. Since this is not the case then every replica will use different autogenerated certificates. We have to work on that and try to sync them across replicas somehow.

PR #2858, opened 2018-02-24
hide nodePort in internal endpoints when nodePort is not set

jqlu:
### What this PR does

When nodePort is not set for a service, hide the nodePort endpoint rather than display a <service>:0 endpoint.

### How to reproduce it

1. deploy a ClusterIP type service
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-service1
  labels:
    app: nginx-svc
spec:
  selector:
    app: nginx
  ports:
  - protocol: TCP
    name: http
    port: 30080
    targetPort: 80
  type: ClusterIP
```

2. open Dashboard - Service list
3. check Internal endpoints column of the service above
```
my-service1:30080 TCP
my-service1:0 TCP
```

### How to fix it

Add a ng-if directive for Internal endpoints fields


jqlu:
any update? @bryk

denverdino:
@bryk Any update?

PR #2883, opened 2018-03-09
fix some translation typo errors.

youyongsong:
fix some translation typo errors.

### DIFF
--- a/i18n/messages-en.xtb
+++ b/i18n/messages-en.xtb
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE translationbundle><translationbundle lang="en">
-  <translation id="961242340990264399" key="MSG_ABOUT_ABOUT_0" desc="This is a label for button displayed on about page used to open documentaion in browser">Documentation</translation>
+  <translation id="961242340990264399" key="MSG_ABOUT_ABOUT_0" desc="This is a label for button displayed on about page used to open documentation in browser">Documentation</translation>
   <translation id="5634646746226609179" key="MSG_ABOUT_ABOUT_1" desc="Label \'Dashboard Version\' for the about page.">Dashboard Version</translation>
   <translation id="4718120324314749596" key="MSG_ABOUT_ABOUT_2" desc="Label \'Git Commit\' for the about page.">Git Commit</translation>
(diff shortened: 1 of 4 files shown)


i18n/messages-en.xtb
```diff
@@ -97,7 +97,7 @@
   <translation id="3386572067498503433" key="MSG_COMMON_COMPONENTS_CONDITIONS_CONDITIONSLIST_11" desc="Label when there is no data.">-</translation>
   <translation id="6464751813964841186" key="MSG_COMMON_COMPONENTS_CONDITIONS_CONDITIONSLIST_12" desc="Label when there is no data.">-</translation>
   <translation id="8142136741637546639" key="MSG_COMMON_COMPONENTS_CONDITIONS_CONDITIONSLIST_13" desc="Tooltip heartbeat column of pod conditions">Probed at </translation>
-  <translation id="6128317173209217017" key="MSG_COMMON_COMPONENTS_CONDITIONS_CONDITIONSLIST_14" desc="Tooltip transistion column of pod conditions">Probed at </translation>
+  <translation id="6128317173209217017" key="MSG_COMMON_COMPONENTS_CONDITIONS_CONDITIONSLIST_14" desc="Tooltip transition column of pod conditions">Probed at </translation>
```
maciaszczykm: It has to be fixed directly in the template if it is English.

youyongsong: Thanks for commenting, but this typo happens in the `desc` field not in the translation content field.

floreks: All `xtb` files are autogenerated based on HTML template files. `desc` attributes are also taken from HTML templates. This  change would be overridden after running `gulp build:prod` anyway so it's not persistent.

youyongsong: Got it, many thanks, the corresponding HTML template has been corrected before but the `xtb` files are not synced with it, what should I do now?

maciaszczykm: It's a bug of the current system, that we use. You can edit them manually. In the new app version, we will work on new translation system.

youyongsong: @maciaszczykm Does this PR need to be merged? If no needs to be merged, I will close this PR.

maciaszczykm: At the moment it gives us no value. HTML is still not modified.

youyongsong: Yes, I will close this issue.

PR #2907, opened 2018-03-20
add history service

pengx17:
Migrate as a dependency of create view.
Also fixes an issue that TextInput does not update text input correctly.

### DIFF
--- a/src/app/frontend/common/services/global/history.ts
+++ b/src/app/frontend/common/services/global/history.ts
@@ -34,10 +34,9 @@ export class HistoryService {
   }
 
   /**
-   * Goes back to previous state or to the provided default if none set.
-   * @param {string} defaultState
+   * Goes back to previous state or to the provided defaultState if none set.
    */
-  back(defaultState?: string): TransitionPromise {
+  goBack(defaultState: string): TransitionPromise {
     let targetState = this.previousStateName || defaultState;
 
     // If previous state is same as current state then go to default state to avoid loop.
--- a/src/app/frontend/common/services/global/history.ts
+++ b/src/app/frontend/common/services/global/history.ts
@@ -36,7 +36,7 @@ export class HistoryService {
   /**
    * Goes back to previous state or to the provided defaultState if none set.
    */
-  goBack(defaultState: string): TransitionPromise {
+  goToPreviousState(defaultState: string): TransitionPromise {
     let targetState = this.previousStateName || defaultState;
 
     // If previous state is same as current state then go to default state to avoid loop.
--- a/src/app/frontend/common/services/global/history.ts
+++ b/src/app/frontend/common/services/global/history.ts
@@ -1,0 +1,49 @@
+// Copyright 2017 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Injectable, transition} from '@angular/core';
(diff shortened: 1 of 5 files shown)


src/app/frontend/common/services/global/history.ts
```diff
@@ -0,0 +1,47 @@
...
+export class HistoryService {
+  private previousStateName: string;
+  private previousStateParams: RawParams;
+
+  constructor(
+      private readonly state: StateService, private readonly transitions: TransitionService) {}
+
+  /** Initializes the service. Must be called before use. */
+  init(): void {
+    this.transitions.onSuccess({}, (transition) => {
```
floreks: You can subscribe to one of the wrappers exposed in [KdStateService](https://github.com/pengx17/dashboard/blob/b5f84015da58e752832953883a92da4b2be0045f/src/app/frontend/common/services/global/state.ts#L21). It's done this way so we don't have to use this global initialization.

src/app/frontend/common/services/global/history.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Goes back to previous state or to the provided default if none set.
+   * @param {string} defaultState
+   */
+  back(defaultState?: string): TransitionPromise {
+    let targetState = this.previousStateName || defaultState;
+
+    // If previous state is same as current state then go to default state to avoid loop.
+    if (this.state.current.name === this.previousStateName) {
```
floreks: As you have made `defaultState` an optional parameter then you have to handle a case where previous state is same as current and `defaultState` is not set (is undefined).

src/app/frontend/common/services/global/history.ts
pengx17: OK. I guess we should still rely on ui-router's `StateService` to do navigations (except href) like `go`?

src/app/frontend/common/services/global/history.ts
pengx17: Will revert `defaultState` back to optional params. The logic behind determining the "correct" state in navigations is too ambiguous and is subject to features. I guess it's better leave this to the user of this function.

src/app/frontend/common/services/global/history.ts
floreks: In this case, yes. It is required to navigate through lazy loaded modules as a workaround for issue reported on uirouter's repo. In case where module was already loaded, which is the case for previous state, then uirouter will correctly handle transitions.

pengx17: @floreks I see `KdStateService` does not expose transition object (of ui-router). Is this by design?

floreks: It exposes concrete transitions like `onBefore` and `onSuccess`. It is enough for 90% of cases. I don't think we will need more.

src/app/frontend/common/services/global/history.ts
```diff
@@ -0,0 +1,49 @@
...
+  private previousStateName: string;
+  private previousStateParams: RawParams;
+
+  constructor(private readonly kdState: KdStateService, private readonly state: StateService) {
+    this.init();
+  }
+
+  /** Initializes the service. **/
+  init(): void {
+    this.kdState.onSuccess.asObservable().subscribe(transition => {
```
floreks: `asObservable` is not needed here, you can directly use `subscribe` method.

pengx17: Angular's EventEmitter extends subscribe and strips off the generic type info. The asObservable call is to bring the type back.

floreks: Just for typing you can simply use `onSuccess.subscribe((transition: Transition) => {...});`.

pengx17:
History service will be merged with #2915

pengx17:
@floreks could you review #2915 since create from input and file is done. Create from settings (form) will take more time to migrate and I would make a new PR for that.

floreks:
First we need to merge this one. Then I can review #2915.

floreks:
Please rebase to fix travis.

pengx17:
@floreks Travis is failing because of a failure in the original angular-migration branch. Do I need to fix it in this PR？

floreks:
Backend tests are failing. This has nothing to do with this change.

/approve

PR #2909, opened 2018-03-21
Secret Volume Mode

djsd123:
The pod will error when using the `--auto-generate-certificates` flag
while trying to write the cert file. This is because a secret's volume mode is set to `0400` by default.  See
[docs](https://kubernetes.io/docs/concepts/configuration/secret/)


floreks:
Auto generate certs option behavior was changed to keep them in-memory in #2795 PR. Secret is only used as read-only in case user wants to override certificates with custom ones by creating/editing this secret.

djsd123:
Fair enough @floreks. Wierd that the `defaultMode` was required when deploying master last week.  Thanks anyway..

floreks:
Possibly you were using Dashboard `1.8.2` instead of `1.8.3` or older version was taken from cache.

PR #2934, opened 2018-04-02
fix: deployment to rs reference

oilbeater:
Previously template.ObjectMeta will always be empty, deployment will wrongly reference pod with same template.Spec.

Also
```
 	for _, rs := range allRS {	
		if EqualIgnoreHash(rs.Spec.Template, rsTemplate) {	
			**matchingPods = FilterPodsByControllerRef(&rs, allPods)**	
		}	
	}
```
if two rs has same spec only the last rs will be considered.


bryk:
LGTM

Issue #2937, opened 2018-04-04
target node version


zhangjianwei-inspur:
Dashboard version:1.8.2
config .babelrc

    "development": {
      "presets": [["env", {"targets": {"node": 8}, "useBuiltIns": true}]],
    },

target node verison set a fixed number is not reasonable.the version number should be set to fit platform.nodejs high version compatible with low version.

zhangjianwei-inspur:
https://babeljs.io/docs/plugins/preset-env/

PR #2938, opened 2018-04-04
target node version

zhangjianwei-inspur:
https://github.com/kubernetes/dashboard/issues/2937

target node verison set a fixed number is not reasonable.the version number should be set to fit platform.nodejs high version compatible with low version.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -16,7 +16,7 @@
       ]
     },
     "development": {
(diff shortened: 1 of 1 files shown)


maciaszczykm:
It is not required in our setup after Angular migration: https://github.com/kubernetes/dashboard/blob/ng-migration/.babelrc

Closing.

/close

PR #3062, opened 2018-05-29
fix bug: dashboard crashed by terminal with the panic "invalid memory address or nil pointer dereference"

richardxz:
the variable "terminalSessions[msg.SessionID]" maybe still nil when use it in function "WaitForTerminal",
 because sometime it's used before assigned, then it will trigger panic "invalid memory address or nil pointer dereference"


imcsk8:
LGTM

bryk:
Thanks @imcsk8 for review!

Issue #2672, opened 2017-12-13
Remove skip login option


floreks:
As per request from https://github.com/kubernetes/dashboard/issues/2668#issuecomment-351192466.

dohnto:
Our company would like to see this feature implemented since we use several clusters across different teams and we would like to disable even **read-only** access to some clusters.

onitake:
Enforcing a read-only mode for unauthenticated users may be useful, but is unnecessary as this can be modelled easily with proper RBAC rules.

However, under very restrictive rules (i.e. no permissions at all), the SKIP button is useless, so an option to disable it would improve the user experience.

maciaszczykm:
@jessfraz @liggitt @floreks I would like to restart our discussion here as we still have this issue in our backlog and its priority is critical. What do you think about making a new release as soon as possible to include the current fix in it (#3289) and waiting with completely removing the skip option until the migration finish (#3152)? I would say that it should take around a month to get migration on master and starting work on it. Would it be okay for you or you think that the current solution is not enough and we should focus on it right now? What would be your requirements for it then?

bryk:
> and waiting with completely removing the skip option until the migration finish

I think we should not wait and remove the skip option immediately. Making this project secure should always be highest priority.

ekevoo:
Because googling my problem landed in this bug report for the top result:

If you want to **keep** the skip login button, you need to edit the yaml file to include `--enable-skip-login` as one of the args of the deployment.

floreks:
@ekevoo Thanks for this. We've missed that one on the release notes. I have updated [changelog](https://github.com/kubernetes/dashboard/releases/tag/v1.10.1) and [wiki](https://github.com/kubernetes/dashboard/wiki/Dashboard-arguments) has been also updated to contain new argument.

joekohlsdorf:
Sad to see that such a breaking change was done in a minor release and was originally not even in the changelog.

floreks:
@joekohlsdorf It was in the changelog from the beginning but it was not so prominent and obvious. You would need to check linked issues.

> CVE-2018-18264: Fix security issue related to using Dashboard's Service Account (#3400 and #3289)

PR #3077, opened 2018-06-04
Allow disabling the skip button

onitake:
This PR implements the feature requested in #2672 .

It is not fully tested yet, as the dashboard has considerable dependencies on the runtime environment, which are difficult to fulfil on my test system.
The backend part should work as expected, however.

I've split the PR into multiple commits for easier review.


onitake:
Seems like the integration tests have failed.

Sorry, I only have a very basic understanding of Angular and JS.
Would be nice if someone more familiar with the frontend code could look into it.

Thanks!

udit2909:
@floreks Can you please create a release for this pr. Would be very helpful :)

PR #3098, opened 2018-06-18
Fix bug: display job as failed if job backoff limit has reached

### LINKED ISSUE #2788: Jobs are not displayed as failed after reaching backoff limit
##### Environment

```
Dashboard version: 1.8.0
Kubernetes version: 1.8.0
```

##### Steps to reproduce
- Add job with a container that will fail
- Wait until the job has reached the backoff limit

##### Observed result
Dashboard shows job in list with a success icon.

##### Expected result
Dashboard shows job in list with a warning icon (like when a pod fails) and displays the status message.

alexmt:
Pull request resolves #2788: job list API returns an inferred job status; the job card component renders error icon if job status is failed.

### DIFF
--- a/src/app/backend/resource/job/list.go
+++ b/src/app/backend/resource/job/list.go
@@ -43,6 +43,24 @@ type JobList struct {
 	Errors []error `json:"errors"`
 }
 
+type JobStatusType string
+
+const (
+	// JobRunning means the job is still running.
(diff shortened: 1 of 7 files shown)


alexmt:
Can someone please take a look? @konryd , @rf232

rf232:
Will look at it this Monday if no one else did before then.

PR #3102, opened 2018-06-20
Angular migration service detail

yinhaijiao:


yinhaijiao:
@austbot 
Hi, the task is assigned by Marcin. #2727 
Display more infomation for service detail page.

src/app/frontend/common/components/endpoint/cardlist/component.ts
```diff
@@ -0,0 +1,38 @@
...
+@Component({
+  selector: 'kd-endpoint-card-list',
+  templateUrl: './template.html',
+  styleUrls: ['./style.scss']
+})
+export class EndpointListComponent {
+  @Input() initialized: boolean;
+  @Input() endpoints: Endpoint[];
+
+  getEndpointsColumns(): string[] {
```
austbot: you could make this 
```
get endpointsColumns() {
...
}
```
then it can be memoized
or even better to avoid rerun make it an instance variable declared above the constructor.

yinhaijiao:
@austbot Now all the new code has not been tested yet. Can we do it at the end?

Issue #2786, opened 2018-01-21
History of jobs in Cron Jobs view


dzoeteman:
##### Environment

```
Dashboard version: master
Kubernetes version: 1.8.0
```
As was discussed in #2438, it would be great to have a history of jobs available in the Cron Job view, not just the active ones.

Sai-Adarsh:
Is it still open? If so, could I work on this?

floreks:
You are welcome to work on this.

maciaszczykm:
It is already done on master.

maciaszczykm:
https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/resource/workloads/cronjob/detail/template.html#L81

/close

PR #3103, opened 2018-06-20
Support for viewing completed/inactive CronJob jobs 

jeefy:
Hi all! This directly addresses #2786

I created an additional function that pulls only completed Jobs for a given CronJob.
Next I shimmed in "Inactive Jobs" in the CronJobDetail Type as well as in the Angular page under the "Active Jobs" card.

Tested locally via minikube (v1.10, v1.9, v1.8)

Let me know if there's anything I need to change or fix!

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -41,4 +41,5 @@ WORKDIR /dashboard
 COPY ./ ./
 
 # Install dependencies. This will take a while.
+RUN npm install -g npm@latest gulp
 RUN npm install --unsafe-perm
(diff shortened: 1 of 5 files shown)


bryk:
@pengx17 Can you also review this and LGTM/comment?

pengx17:
/LGTM, btw @jeefy I do not find changes related to regular Jobs. Do you think it is better to include these to Job detail pages as well?

jeefy:
Thanks for the feedback/merge!

@pengx17 This is to display jobs owned by a CronJob specifically. It wouldn't apply to a Job itself. If there are more use cases of jobs spawning other jobs (a la HPC workflow) it might make sense. 

It might be a good idea to eventually have the ability to display objects that have an OwnerRef of the currently-viewed object. That would probably solve this across the board. :)

PR #2975, opened 2018-04-20
fix: namespace param should be checked in onEnter hook

pengx17:
Related issue:
https://github.com/kubernetes/dashboard/issues/2974

The issue is related to how app is initialized:

1. user hits `k8s-ui.com`
2. app load default state, with namespace = `''`
3. `kubernetesDashboard.common.namespace` module 's `run` method checks `ensureNamespaceParamPresent`. Since no namespace presents in the url, the app then reload with namespace = `'default'`

By changing the hook from onSuccess to onEnter, overview state (and any other state) will always success with a non empty namespace.

Issue #2974, opened 2018-04-19
Default Dashboard URL is overview for all namespaces, slow to load with many services


davidtseng:
##### Environment

```
Dashboard version: 1.8.3
Kubernetes version: 1.9.2
Operating system: coreos
Node.js version:
Go version:
```

##### Steps to reproduce

Load the default Dashboard URL:
/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/#!/overview
##### Observed result

With a few services this loads fast but with on a larger cluster with 1000+ services, this can take over 1 minute to load. Using the login screen, after logging in the page is not functional until the default page loads.
##### Expected result

Dashboard loads and is functional quickly.
##### Comments

Specifying a namespace dramatically improves load time, for example using the 'default' namespace (Dashboard version 1.6.0 appears to use this):
/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/#!/overview?namespace=default

davidtseng:
Currently with the head version of Kube Dashboard with fix from @pengx17 hitting 'skip' on the login screen hangs, appears to redirect to the login screen.

pengx17:
@davidtseng I will have a look, sorry about that

floreks:
First try to increase memory and cpu limits of Dashboard. As API server does not offer pagination, then all this data has to first be loaded into the memory of our backend. It's possible that due to memory limits your instance of Dashboard becomes unresponsive as it eventually goes out of memory.

PR #3106, opened 2018-06-22
fix: fix an issue which causes hang after skip login page

pengx17:
related:
https://github.com/kubernetes/dashboard/issues/2974
https://github.com/kubernetes/dashboard/pull/2975

### DIFF
--- a/src/app/frontend/chrome/component.js
+++ b/src/app/frontend/chrome/component.js
@@ -33,16 +33,14 @@ export class ChromeController {
    */
   constructor($state, $timeout, $transitions, $resource, $sce) {
     /**
-     * By default this is true to show loading spinner for the first page.
      * @export {boolean}
      */
-    this.showLoadingSpinner = true;
+    this.showLoadingSpinner = false;
 
     /**
-     * By default this is true to show loading for the first page.
      * @export {boolean}
      */
-    this.loading = true;
+    this.loading = false;
 
     /** @private {!ui.router.$state} */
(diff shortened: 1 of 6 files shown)


src/app/frontend/chrome/component.js
```diff
@@ -33,16 +33,14 @@ export class ChromeController {
    */
   constructor($state, $timeout, $transitions, $resource, $sce) {
     /**
-     * By default this is true to show loading spinner for the first page.
      * @export {boolean}
      */
-    this.showLoadingSpinner = true;
+    this.showLoadingSpinner = false;
```
pengx17: Since a state change may not happen after chrome component init, we better not to set loading to true by default

src/app/frontend/common/namespace/module.js
```diff
@@ -42,22 +42,21 @@ export default angular
...
+   * Helper function which redirect namespace param when the given namespace
+   * in the transition is undefined.
    * @returns {boolean}
    */
-  function replaceUrlIfNeeded(namespace) {
-    if (namespace === undefined && !!$state.transition &&
-        $state.transition.to().name !== loginState) {
-      $location.search(namespaceParam, DEFAULT_NAMESPACE);
-      $location.replace();
-      return false;
```
pengx17: This line causes the bug since it will abandon the transition.
Also, previously the namespace is set by watching location.search, but a better solution could be redirecting the state by checking if the namespace is giving.

Issue #3080, opened 2018-06-06
cronjob page to allow the creation of jobs from cronjobs


CharlesCara:
The dashboard support for cronjobs is a welcome addition, but please can you enhance the cronjob page to allow the creation of jobs from cronjobs.  

This would be  implemention/replication of the kubectl command: 
`kubectl create job jobname --from=cronjob\cjname`

Also could allow cronjobs be disabled/enabled in the dashboard

maciaszczykm:
It is already done. Tested in on HEAD. It will be also part of ng migration.

/close

PR #3112, opened 2018-06-27
Allow jobs to be manually triggered from cron jobs

EliZucker:
Hi all,

These changes address #3080. I added a trigger button next to cron jobs that manually starts them. My changes were based on a development last year that allowed this feature via the CLI: https://github.com/jingxu97/kubernetes/commit/5a304c021fcf213e087023654c941ac238d1b232

This is my first pull request on this project, so I would appreciate any feedback. Thanks!


src/app/backend/resource/cronjob/jobs.go
```diff
@@ -144,3 +145,41 @@ func filterJobsByState(active bool, jobs []batch.Job) (matchingJobs []batch.Job)
...
+  annotations["cronjob.kubernetes.io/instantiate"] = "manual"
+
+  labels := make(map[string]string)
+  for k, v := range cronJob.Spec.JobTemplate.Labels {
+    labels[k] = v
+  }
+
+  jobToCreate := &batch.Job{
+    ObjectMeta: metaV1.ObjectMeta{
+      // job name cannot exceed DNS1053LabelMaxLength (52 characters)
```
rf232: CronJob name can be long already though, and especially in a UI, we should perhaps help the enduser with that.

EliZucker:
@austbot Thank you for the prompt feedback! I tried a small tweak to increase codecove, but I will do more tomorrow if that doesn't work.

@rf232 The triggered job's name is now automatically shortened if needed. Let me know if I need to explain/change anything else.

EliZucker:
@rf232 Would you mind reviewing this PR again? I think your earlier concern is addressed and it should be mergeable. Thanks!

Issue #3125, opened 2018-07-03
Logs are presented HTML encoded


oholsen:
##### Environment
```
Dashboard version: 1.8.3
Kubernetes version: 1.9.7
Operating system: kops/aws/debian
Node.js version: ?
Go version: 
```

##### Steps to reproduce
1. Deploy a pod logging a message including '&', e.g.

```
kubectl run -i --tty busybox --image=busybox --restart=Never -- sh -c 'echo "Hello &"'
```

##### Observed result

```kubectl logs busybox``` shows  'Hello &'.

Viewing the logs via the dashboard, the '&' is presented as ```&amp;```:

```
Hello &amp;
```

##### Expected result

Viewing the log in the dashboard, the '&' should be presented as '&'.

##### Comments

Deployed dashboard as:
kubectl apply -f https://raw.githubusercontent.com/kubernetes/kops/master/addons/kubernetes-dashboard/v1.8.3.yaml

oholsen:
I am about to  push and issue a PR but are not allowed. Is there some permission required?

oholsen:
https://github.com/kubernetes/dashboard/pull/3129

PR #3129, opened 2018-07-04
Fix HTML encoding in logs (#3125)

oholsen:
Hi, here is a fix for #3125.

### DIFF
--- a/src/app/frontend/logs/component.js
+++ b/src/app/frontend/logs/component.js
@@ -42,9 +42,6 @@ export class LogsController {
     /** @private {!angular.$sce} */
     this.sce_ = $sce;
 
-    /** @private {!HTMLDocument} */
-    this.document_ = $document[0];
-
     /** @private {!angular.$resource} */
     this.resource_ = $resource;
 
--- a/src/app/frontend/logs/component.js
+++ b/src/app/frontend/logs/component.js
@@ -258,29 +258,15 @@ export class LogsController {
    * @private
    */
   formatLine_(line) {
-    // remove html and add colors
-    // We know that trustAsHtml is safe here because escapedLine is escaped to
-    // not contain any HTML markup, and formattedLine is the result of passing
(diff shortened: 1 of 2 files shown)


src/app/frontend/logs/component.js
```diff
@@ -258,31 +255,17 @@ export class LogsController {
...
-    // We know that trustAsHtml is safe here because escapedLine is escaped to
-    // not contain any HTML markup, and formattedLine is the result of passing
-    // ecapedLine to ansi_to_html, which is known to only add span tags.
-    let escapedContent =
-        this.sce_.trustAsHtml(new AnsiUp().ansi_to_html(this.escapeHtml_(line.content)));
+    // Escape html and add colors
+    // We know that trustAsHtml is safe here because line.content is escaped to
+    // not contain any HTML markup in ansi_to_html, which also replaces ANSI terminal
+    // escape codes with span tags.
+    let escapedContent = this.sce_.trustAsHtml(new AnsiUp().ansi_to_html(line.content));
```
austbot: Will this prevent javascript from being executed?

oholsen: ansi_to_html() does "minimal" escaping by default as per the documentation:
https://github.com/drudru/ansi_up#ansi_to_html-txt
https://github.com/drudru/ansi_up#escape_for_html

The referred issue is that logs are double escaped. It should be the responsibility of ansi_to_html() or the removed function, but not both.

oholsen: The commit references issue  #3125 .

konryd: The html is then embedded using ng-bind-html, which does the escaping:

https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/logs/logs.html#L128

konryd:
Thanks @oholsen for the fix! And thanks @austbot for reviewing!

Issue #2809, opened 2018-02-01
Wrong number of pods in a daemonset


lilithfr:
##### Environment
```
Dashboard version: v1.8.2 
Kubernetes version: v1.9.2
Operating system: Debian 9
```

##### Observed result
In the Daemon Sets, the calico-etcd indicates that 1 pods out of 3 is running.
![screenshot from 2018-01-31 17-32-35](https://user-images.githubusercontent.com/5602250/35656664-59c2e718-06ad-11e8-9228-28233b0cfc9b.png)
But the kubectl indicates that only 1 pod is desired
```
$ kubectl get daemonsets -n kube-system
NAME          DESIRED   CURRENT   READY     UP-TO-DATE   AVAILABLE   NODE SELECTOR                     AGE
calico-etcd   1         1         1         1            1           node-role.kubernetes.io/master=   8d
calico-node   3         3         3         3            3           <none>                            8d
kube-proxy    3         3         3         3            3           <none>                            8d
```

##### Expected result
Except if I misunderstand something, the dashboard should indicates "1/1" and not "1/3" in the columns "Pods"

floreks:
Can you also check the details page if it shows the wrong info?

lilithfr:
Here de details page, I don't see anything special:
![screenshot from 2018-02-01 07-10-58](https://user-images.githubusercontent.com/5602250/35685660-287d41b8-071f-11e8-9eb6-cfb999c60b76.png)

floreks:
Ok, so the details page shows pods status correctly but the list page not. We will check that.

eloyekunle:
Fixed by #3131
/close

PR #3131, opened 2018-07-05
Fix incorrect desired number of pods for DaemonSets 

jeefy:
Hi all! This directly addresses #2809 

There may be a better solution for it, but this was a short and sweet one. It seems there's a race condition when looping through the list of DaemonSets. 

I was able to replicate the bug locally and on a separate cluster I have access to. 

Tested locally via minikube (v1.10, v1.9, v1.8)

Let me know if there's anything I need to change or fix!

### DIFF
--- a/src/app/backend/resource/daemonset/list.go
+++ b/src/app/backend/resource/daemonset/list.go
@@ -116,11 +116,10 @@ func toDaemonSetList(daemonSets []apps.DaemonSet, pods []v1.Pod, events []v1.Eve
 	daemonSets = FromCells(dsCells)
 	daemonSetList.ListMeta = api.ListMeta{TotalItems: filteredTotal}
 
-	for _, daemonSet := range daemonSets {
-		desired := daemonSet.Status.DesiredNumberScheduled
+	for i, daemonSet := range daemonSets {
 		matchingPods := common.FilterPodsByControllerRef(&daemonSet, pods)
 		podInfo := common.GetPodInfo(daemonSet.Status.CurrentNumberScheduled,
-			&desired, matchingPods)
+			&daemonSets[i].Status.DesiredNumberScheduled, matchingPods)
 		podInfo.Warnings = event.GetPodsEventWarnings(events, matchingPods)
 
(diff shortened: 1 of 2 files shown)


jeefy:
Good call! I'll try to add that by tonight

On Jul 6, 2018 10:06, "Austin Adams" <notifications@github.com> wrote:

@jeefy <https://github.com/jeefy> Possible to add a test to capture this?
So that we don't regress?
I'm thinking you can augment list_test.go to look at the DaemonSet pod .
info and match against number of pods desired

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/kubernetes/dashboard/pull/3131#issuecomment-403044612>,
or mute the thread

jeefy:
Updated `TestToDaemonSetList` to have a third entry in their Daemonsets, each having a unique desired count.

src/app/backend/resource/daemonset/list.go
```diff
@@ -117,9 +117,10 @@ func toDaemonSetList(daemonSets []apps.DaemonSet, pods []v1.Pod, events []v1.Eve
 	daemonSetList.ListMeta = api.ListMeta{TotalItems: filteredTotal}
 
 	for _, daemonSet := range daemonSets {
+		desired := daemonSet.Status.DesiredNumberScheduled
```
rf232: Looking throught the golang refspec I don't seen anything that the temp local variable in the scope of this for block is not allowed to be reused. So in a way, I believe a smarter go compiler might break us again. How about also capturing the index in the loop and passing `&daemonSets[i].Status.DesiredNumberScheduled`

jeefy: I like that route better as well, thanks for the input. 

I'll try to churn that out today and verify it works. :)

jeefy: Updated and verified. Thanks @rf232!

jeefy:
Could someone trigger another Travis build? Thought /retest would do it in this repo.

The Travis build failed because Saucelabs apparently timed out. The Backend test did show no problems though.

Thanks!

PR #3142, opened 2018-07-14
Merges in master.

austbot:
WIP- DONT MERGE


maciaszczykm:
@austbot I am wondering if it won't desynchronize `master` with `angular-migration`. I think we might need to keep commit history here.

Issue #3137, opened 2018-07-09
Add nodeSelector architecture to prevent amd64 versions getting deployed to arm nodes


drake7707:
I have a mixed architecture cluster consisting of both amd64 and arm nodes. Currently when deploying the dashboard it does not check the architecture label and often deploys the pod to a node with the wrong architecture.

Please add

```
      nodeSelector:
        beta.kubernetes.io/arch: amd64
```
and
```
      nodeSelector:
        beta.kubernetes.io/arch: arm
```
to the deployment yaml so the pod can never be deployed to a node with a different architecture than the image.

(or alternatively, provide a multi-arch image manifest to use in the deployment)

##### Environment
Mixed architecture cluster consisting of both arm and amd64 nodes

```
Kubernetes version: v1.11.0
Operating system: Linux (debian) for the amd64 nodes and Raspbian for the arm nodes
```

##### Steps to reproduce
Set up a mixed architecture cluster, deploy the amd64 version of the dashboard. if it lands on an arm node it will fail to start

##### Observed result
Dashboard amd64 fails when deployed to arm.

##### Expected result
Deployment should have selected an amd64 node for the dashboard to deploy on

jeefy:
I think it would be better to have a multi-arch Dashboard container?

Thoughts? /cc @bryk 

(FYI I'd be willing to take that on)

drake7707:
@jeefy Yeah that would be the best future proof option.

floreks:
We've already had a proposal to use multi-arch image and deployment in https://github.com/kubernetes/dashboard/issues/2811. If someone is willing to work on this then it would be great.

jeefy:
@floreks Dibs! :) I was already looking to take that on.

floreks:
We support multi-arch docker image now.

/close

Issue #2811, opened 2018-02-03
Describe and document installing dashboard on arm64 clusters


vielmetti:
##### Environment

```
Dashboard version: 1.8.2
Kubernetes version: 1.9.2
Operating system: Ubuntu 16.04 LTS on arm64 Packet "Type 2A"
Node.js version:
Go version: 
```

##### Steps to reproduce

Install dashboard according to package directions found in [releases](https://github.com/kubernetes/dashboard/releases) on an `arm64` machine, with
the command

```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/v1.8.2/src/deploy/recommended/kubernetes-dashboard.yaml
```

##### Observed result

Dashboard fails to load, with status CrashLoopBackOff, since the provided recommended deploy file explicitly references amd64 images.

##### Expected result

Dashboard loads successfully on all provided architectures.

##### Comments

Two remedies for this, one better than the other.

The not-better remedy is to have an explicit `arm64` dashboard, along the lines of the `arm` dashboard, with a parallel file then again each for all architectures where there's a binary provided.

The better remedy is a multl-arch manifest for the dashboard, so that a single dashboard works across all architectures, and the manifest pulls the right image.

floreks:
We do not want to create and support distinct yamls each for every architecture. We do have already do have quite many of them as there are 2 ways of deploying Dashboard (recommended + alternative) + there are head + official deployments for both amd64 and arm. This gives us already 8 yaml files.

Second option is also not possible as yaml file has to explicitly define single image that should be used for deployment. If you need to use other supported architecture then you simply need to take our official deployment yaml, download it and replace image with the one required by you. I think this is quite simple and does not need documenting.

vielmetti:
@floreks 

I agree that the first option is untenable, though I disagree that the second is not possible. Perhaps dashboard is not the first to get a manifest list per https://github.com/kubernetes/kubernetes/issues/57869 but I think that's the roadmap.

floreks:
How do you imagine manifest with multiple architectures to work? Kubernetes resource manifest has to define some resource(s) i.e. deployment and each deployment requires image. We can not simply define multiple deployments in single manifest file for different architectures.

vielmetti:
@floreks I imagine as follows:

[...]

PR #3143, opened 2018-07-15
Update to build and push multi-arch manifests

jeefy:
This allows us to publish v2.2 registry manifests allowing end-users to consume a single container image rather than specifying their architecture. This will directly address and resolve several issues (including #3137 and #2811)

Example:
`docker pull kubernetes-dashboard` 
Instead of
`docker pull kubernetes-dashbaord-amd64`

- Update travis-ci.yml to use a newer version of Docker
- Update build/Dockerfile to use a newer version of Docker
- pushToDocker now creates and pushes manifests 

Tested and verified on both [Docker Hub](https://hub.docker.com/r/jeefy/kubernetes-dashboard/) and GCR (Only had access to a private repo sorry)

If/When this is accepted, there will be additional work updating other repos and documentation to use the Manifest image (kubernetes-dashboard) instead of individual images (kubernetes-dashboard-amd64)

Any questions or comments are appreciated!


floreks:
I have one concern regarding this change. As it requires edge docker client with manually enabled experimental flag, it might be a problem for users to use this instead of the current installation methods. At first, we should keep it only as an addition to current deployment methods and not make it the default.

floreks:
Great, if this will work without additional changes to docker then I am ok with it.

jeefy:
/hold

Awesome! I think we're waiting till after the next release to merge this in.

Once the angular migration is done we can merge this and update everything surrounding it in future PRs.

maciaszczykm:
> Once the angular migration is done we can merge this and update everything surrounding it in future PRs.

@jeefy Are we planning to something about it?

jeefy:
I'd like to get this in the next release, yes. I want to revisit this once
metrics server support is solid.

jeefy:
@floreks @maciaszczykm PTAL, I think it's in a good place.

maciaszczykm:
@jeefy Can you solve conflicts?

mkumatag:
what else pending part of this pr? can we merge this?

maciaszczykm:
@mkumatag Nothing :)

/lgtm

Issue #3012, opened 2018-05-10
remove full service account token from system log file


jimangel:
##### Environment
```
Dashboard version: v1.8.2
Kubernetes version: v1.9.6
Operating system: Ubuntu 16.04.4 LTS
Go version: v1.9.3
```

##### Steps to reproduce

1. Login to dashboard using a service account token
2. `cat /var/log/containers/kubernetes-dashboard* | grep token`

##### Observed result
You can see the full token stored in the logs

##### Expected result
The token should be redacted for security or only visible at a certain level of logging (eg: debug / verbose)

##### Comments
We ship our logs out to an external ELK server and support multi-tenancy for both ELK and k8s. As a result, anyone who has ELK access can see our admin service account.

I see value in the logs containing the token if you're troubleshooting, but I don't think it should be the default. A solution could be to accept an [argument ](https://github.com/kubernetes/dashboard/wiki/Dashboard-arguments) for the level of logging or change the log file path destination outside of the container logs getting shipped to ELK.

mhobotpplnet:
I have same issue with this, and exposes greater security for someone with less privileges to be able to elevate the or impersonate admin login. 
Yes there is a way for to lock it down and exclude it from being logged to services such as ELK, Splunk, Sumo etc.. however it is in best practice for any sensitive data such as usernames, passwords etc be retracted from logs.

jeefy:
Fixed in #3180 :)

/close

PR #3175, opened 2018-07-26
hide request info for sensitive urls

jeefy:
Hey all! This directly addresses #3012 amongst other dupes.

I created an array of URLs that we would deem "sensitive". We would therefore not log any request variables for these URLs. I've tentatively added the two login endpoints.

I also updated the tests to:
- More easily add additional cases based on URL / Method
- Ensure those URLs will log "{ contents hidden}" rather than the contents.

Critiques and comments are welcome. :)

Thanks!


jeefy:
Closing in favor of a single PR rather than two staggered PRs.

Issue #3174, opened 2018-07-26
Bearer Token ID Displayed In Logs


mhobotpplnet:
##### Environment

```
Dashboard version:v1.8.3
Kubernetes version:1.10
```

##### Steps to reproduce
Login to kubernetes dashboard with bearer token id, checkout the pod logs for k8 dashboard. 

##### Observed result

Logs show full token ID
```
kubernetes-dashboard-somenumber kubernetes-dashboard 2018/07/26 12:39:10 [2018-07-26T12:39:10Z] Incoming HTTP/2.0 POST /api/v1/login request from 100.111.312.444:3587: {
kubernetes-dashboard-somenumber kubernetes-dashboard   "kubeConfig": "",
kubernetes-dashboard-somenumber kubernetes-dashboard   "password": "",
kubernetes-dashboard-somenumber kubernetes-dashboard   "token": "5122-taken-away-full-token"
```
##### Expected result

Did not expect token to be logged, and if it is , it should be detached. 
##### Comments

If someone had access to k8 dashboard, they could take higher privileges by just parsing out the token ID and loggin with my username. 
I guess one way is to disable logging but that is not suitable when trying to troubleshoot authentications etc.

jimangel:
duplicate of https://github.com/kubernetes/dashboard/issues/3012

PR #3180, opened 2018-07-27
API request logging improvements

jeefy:
Hey all!

This directly addresses #3174, #3012, and potentially others. 

I created a list of addresses that may potentially have sensitive information that should not be logged. Instead it will only log `{ contents hidden }`. 

However, in some cases (like debugging and development) you still want to see what's being passed along. So I added a new CLI argument to the dashboard called `--api-log-level` with three options: INFO,DEBUG,NONE

The default setting is INFO, which will log everything but sensitive URLs. 
NONE will turn off all request/response logging.
DEBUG will log everything including the contents of sensitive URLs.

I added tests for all this, and also updated `TestFormatRequestLog` to handle more diverse test cases.

Thanks!

### DIFF
--- a/src/app/backend/handler/filter.go
+++ b/src/app/backend/handler/filter.go
@@ -102,11 +102,16 @@ func formatResponseLog(response *restful.Response, request *restful.Request) str
 // checkSensitiveUrl checks if a string matches against a sensitive URL
 // true if sensitive. false if not.
 func checkSensitiveURL(url *string) bool {
-	var sensitiveUrls = make(map[string]bool)
-	sensitiveUrls["/api/v1/login"] = true
-	sensitiveUrls["/api/v1/csrftoken/login"] = true
+	var s struct{}
+	var sensitiveUrls = make(map[string]struct{})
+	sensitiveUrls["/api/v1/login"] = s
+	sensitiveUrls["/api/v1/csrftoken/login"] = s
 
-	return sensitiveUrls[*url]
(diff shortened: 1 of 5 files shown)


src/app/backend/handler/filter.go
```diff
@@ -76,6 +83,19 @@ func formatRequestLog(request *restful.Request) string {
 		}
 	}
 
+	// Is DEBUG level logging enabled?
+	if args.Holder.GetAPILogLevel() != "DEBUG" {
+		// Great now let's filter out any content from sensitive URLs
```
maciaszczykm: I would suggest is to export it to another method. You can also use map instead of slice to get rid of the loop:

```
_, isSensitive := sensitiveUrls[key]
```

You can put `struct{}` as a value to save space.

Please remember to use camelCase and switch likes of `sensitive_urls` to `sensitiveUrls`.

jeefy: Addressed and pushed to c820847a9468e26fb5f3baf06950a11da210989a -- Thoughts?

If that looks good, I can squash before the merge.

maciaszczykm:
/lgtm

It seems to work, but I think it should be possible to use this flag also when running `gulp serve` (take a look at `build/conf.js` file). Besides that, please update https://github.com/kubernetes/dashboard/wiki/Dashboard-arguments. You can do this in separate pull request.

Issue #2474, opened 2017-10-13
kubeconfig files cannot login the dashboard


txg1550759:
##### Environment

```
Dashboard version:1.71
Kubernetes version:1.76
Operating system:centos7 
Node.js version:
Go version:
```

##### Steps to reproduce

the dashboard response Not enough data to create auth info structure.

cat kubeconfig
apiVersion: v1
clusters:
- cluster:
[...]

PR #3181, opened 2018-07-28
Improve error message

lvthillo:
Improve error message https://github.com/kubernetes/dashboard/issues/2474


src/app/backend/auth/kubeconfig.go
```diff
@@ -107,7 +107,8 @@ func (self *kubeConfigAuthenticator) getCurrentUserInfo(config kubeConfig) (user
 // Returns auth info structure based on provided user info or error in case not enough data has been provided.
 func (self *kubeConfigAuthenticator) getAuthInfo(info userInfo) (api.AuthInfo, error) {
 	if len(info.Token) == 0 && (len(info.Password) == 0 || len(info.Username) == 0) {
-		return api.AuthInfo{}, errors.New("Not enough data to create auth info structure.")
+		return api.AuthInfo{}, errors.New("Client certificate is not a valid authentication method for the dashboard.")
```
floreks: This is not correct. There are multiple scenarios that might trigger this error and only few of them are related to certificate based auth.

PR #3197, opened 2018-08-04
Add namespace support

FrenchBen:
When deploying the K8s Dashboard, it's possible to scope it to a certain namespace. Yet even within a namespace, other than `kube-system`, the dashboard will still try to add secrets to the `kube-system` ns.

This attempts to open the discussion around how to fix it. In this case, I've added a simple flag to the binary (still needs some tests and url validation).

The following flag was added:
`--namespace string   When non-default namespace is used, create encryption key in the specified namespace. Default: 'kube-system'. (default "kube-system")`

Another idea is for the dashboard to lookup what namespace it's running in, and simply use that. :fire: and :speaking_head:  are welcome :)

/assign @ianlewis


FrenchBen:
Ping @kubernetes/dashboard-maintainers and @liggitt

floreks:
Unfortunately, this change is not enough to allow moving Dashboard to another namespace. It does not only use secret, but also config map that is stored in the `kube-system` namespace to synchronize configuration. Additionally, lookup for heapster is also done only in `kube-system`. These are 3 things that would have to be configurable in order to fully allow moving it to another ns without any issues.

FrenchBen:
@floreks The namespace would be specified in the [YAML](https://github.com/kubernetes/dashboard/blob/master/src/deploy/recommended/kubernetes-dashboard.yaml#L28), for each object, which would actually move it to the proper namespace (verified) - It's true that heapster is an issue at this point, which needs to be removed. 
This change to the backend is to allow secret to match those. The "ultimate" fix is for the backend to lookup what namespace it's deployed into, and create the secrets there (unsure how to currently do this)

How would you suggest I tackle this? Provide a YAML with the custom-namespace as an example?

floreks:
There are a couple of steps that should be taken here:
1. Parametrize all external resources used by Dashboard, that includes:
    1. `kubernetes-dashboard-key-holder` secret that is by default stored in `kube-system` namespace
    1. `kubernetes-dashboard-settings` config map that is by default stored in `kube-system` namespace
    1. Heapster lookup that is by default done only in `kube-system` namespace
2. Tighten the default Dashboard privileges and remove `create` permission for `ConfigMap` and `Secret` in the namespace that Dashboard is running. In case a required resource was not created during Dashboard deployment and does not exist, there should be an error saying that a required resource `xxx` is missing and should be created by the user in a namespace `yyy`.
3. Update the documentation

jeefy:
@FrenchBen I agree with all of @floreks points. Apologies I haven't chimed in sooner. Was out of signal range for a bit. 

Let me know if/when you address them and I'll start a review then. :) If you need to chat or want any help before then, feel free to ping me!

FrenchBen:
@floreks thanks for putting this together - Below are a couple of questions, to clarify some of those points.

1. Parametrize all external resources used by Dashboard, that includes:
  i. kubernetes-dashboard-key-holder secret that is by default stored in kube-system namespace `this was done in this PR`
  ii. kubernetes-dashboard-settings config map that is by default stored in kube-system namespace `this is done in the kubernetes-dashboard.yaml, and thus shouldn't be part of this PR - the user would change this`
  iii. Heapster lookup that is by default done only in kube-system namespace `can heapster run in non kube-system namespace?` 🤔 
2. Tighten the default Dashboard privileges and remove create permission for ConfigMap and Secret in the namespace that Dashboard is running. In case a required resource was not created during Dashboard deployment and does not exist, there should be an error saying that a required resource xxx is missing and should be created by the user in a namespace yyy. `Not sure if I understand this, shouldn't the deployment be able to create configMaps/secrets in the same namespace?`
3. Update the documentation and deployment yamls. `Docs can be updated to show the additional params, the YAML, imho, should remain as-is.`

To clarify some of the above points, this PR aims to allow users to deploy the dashboard in a namespace that isn't the default `kube-system`.
The steps would be similar to what they would try today:
1. Create a namespace
2. Update the `kubernetes-dashboard.yaml` metadata to use the custom namespace they chose
3. Add the additional flag of `--enc-key-namespace` to let the dashboard know where the secret should be created.
4. ...
5. Profit?


As a side conversation, is there a roadmap that moves the dashboard towards {metrics-server](https://github.com/kubernetes-incubator/metrics-server)

jeefy:
Gonna go backwards:

Metrics API - We're going to address this after the Angular migration. There's been some preliminary talk about how and certain trouble we see in the move away from Heapster

I agree on what the steps to implement a non-default ns deployment should be.

re: 3: Everything I've seen, Heapster is in kube-system only. If there's a way to deploy it otherwise, I haven't seen an example of it.

re: 2: The way I'm reading it is making sure the running dashboard cannot create CMs/Secrets. Have those CMs/Secrets be created upon deployment creation and allow read-only access to those resources by the dashboard.

floreks:
> this is done in the kubernetes-dashboard.yaml, and thus shouldn't be part of this PR - the user would change this

It is not true. Even though config map is created in the yaml, changing its namespace will currently result in Dashboard error. As you can see below config map namespace and name are hardcoded. It should be configurable.
https://github.com/kubernetes/dashboard/blob/741bcb15f5809c3990537a007bfe962afc872541/src/app/backend/settings/api/types.go#L26-L30

> can heapster run in non kube-system namespace?

I don't see a reason why it wouldn't be able to run in other namespaces. It is just the matter of giving it required permissions. Different namespace does not block heapster from reading metrics.

> Not sure if I understand this, shouldn't the deployment be able to create configMaps/secrets in the same namespace?

Currently Dashboard has permissions to create secrets and config maps in the `kube-system` namespace. As in some scenarios this could cause security issues, I would remove them and adjust the backend. Required secret and config map should be created during Dashboard deployment and create permissions should be removed. In case someone deletes it or modifies yaml and they won't exist during Dashboard start, error should be shown as mentioned above.

As for the additional flag I would try to keep all Dashboard exclusive resources in the same namespace and add a single argument that will be used both by `kubernetes-dashboard-key-holder` secret and `kubernetes-dashboard-settings` config map. By default this argument should be set to `kube-system`. It could simply be named `--namespace`.

As for the heapster it does not need to be in the same namespace as Dashboard. I think `--heapster-host` argument that we already have should be enough to configure it in case it is deployed in another namespace. Similar to how heapster configures sink `--sink=influxdb:http://monitoring-influxdb.kube-system.svc:8086`.

todoesverso:
This is a really useful feature. I am planning to use dashboard developers to deploy jobs, but I want to make sure they only do it in a dedicated namespace. This instance of the dashboard would be only for them.

FrenchBen:
@floreks thanks for the breakdown - 
It's deceiving, but `SettingsConfigMapNamespace` is used in a few places, and will need a bit of redesign to make it fully dynamic - Admittedly, this PR is a semi-hack, as we look at the flags, and no longer the API types defined. 

Happy to change the flag to be `--namespace`, wasn't sure if it was explicit enough? 

I'll get started on making a proper namespace port, we can then resume some of this discussion.

floreks:
> Happy to change the flag to be --namespace, wasn't sure if it was explicit enough?

IMHO it's fine. @maciaszczykm @jeefy do you think that it will be fine from the user perspective?

jeefy:
I think so. We need to clearly document what the flag _actually_ does but otherwise it sounds good to me.

FrenchBen:
FYI, the following vendors seem to be missing from the core project - Should I go ahead and add them to this PR?
https://gist.github.com/FrenchBen/c45e460fb835972b3e20871595f31b64

FrenchBen:
@floreks @jeefy per the recommendations, I went ahead and update the flag to be a common `--namespace`. I've update the original comment to reflect that. 

Should I go ahead and modify the deployment `YAML` as well?

jeefy:
Yes, I'd modify the deployment so this isn't some hidden flag to the users. If they don't use an updated deployment script it still defaults to the `kube-system` namespace so they won't notice anything once a new release is cut. 

Also: did your last commit mean to modify a bunch of vendor files? 

Once you make the deployment change and confirm the vendor changes, squash all your commits down and I'll review. :)

Thanks!!

floreks:
@FrenchBen I don't see any `.go` files there so I am assuming that you haven't pruned the dependencies. If I remember correctly `dep prune` should do that. It is described somewhere on the wiki in the development section.

FrenchBen:
@floreks From dep: `Pruning is now performed automatically by dep ensure` - When you run `dep ensure` it should also prune; ran it manually and pushed again.

FrenchBen:
@jeefy yaml has been reverted!

FrenchBen:
ping @dashboard-maintainers - could we get this reviewed and merge?

/cc @ianlewis

FrenchBen:
rebased to remove merge conflicts

FrenchBen:
@maciaszczykm happy to open a different PR that caters to the vendoring only.

src/app/backend/settings/api/types.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	// SettingsConfigMapName contains a name of config map, that stores settings.
+	// // SettingsConfigMapName contains a name of config map, that stores settings.
```
ianlewis: This change isn't necessary

src/app/backend/integration/manager_test.go
```diff
@@ -65,7 +65,7 @@ func TestIntegrationManager_GetState(t *testing.T) {
 			"Server provided and using external heapster",
 			"http://127.0.0.1:8080", "http://127.0.0.1:8081", &api.IntegrationState{
 				Connected: false,
-				Error:     errors.New("Get http://127.0.0.1:8081/healthz: dial tcp 127.0.0.1:8081: getsockopt: connection refused"),
+				Error:     errors.New("Get http://127.0.0.1:8081/healthz: dial tcp 127.0.0.1:8081: getsockopt: connection refused"), // Docker4Mac shows this as: connect: connection refused
```
ianlewis: This line is pretty long. Can you put the comment on the line before?

src/app/backend/args/builder.go
```diff
@@ -148,3 +148,9 @@ func (self *holderBuilder) SetDisableSkipButton(disableSkipButton bool) *holderB
 func GetHolderBuilder() *holderBuilder {
 	return builder
 }
+
+// SetNamespace 'namespace' argument of Dashboard binary.
+func (self *holderBuilder) SetNamespace(namespace string) *holderBuilder {
```
ianlewis: Just to be consistent, this could be before the GetHolderBuilder function.

src/app/backend/dashboard.go
```diff
@@ -71,6 +71,7 @@ var (
 	argSystemBannerSeverity      = pflag.String("system-banner-severity", "INFO", "Severity of system banner. Should be one of 'INFO|WARNING|ERROR'. Default: 'INFO'.")
 	argAPILogLevel               = pflag.String("api-log-level", "INFO", "Level of API request logging. Should be one of 'INFO|NONE|DEBUG'. Default: 'INFO'.")
 	argDisableSettingsAuthorizer = pflag.Bool("disable-settings-authorizer", false, "When enabled, Dashboard settings page will not require user to be logged in and authorized to access settings page.")
+	argNamespace                 = pflag.String("namespace", "kube-system", "When non-default namespace is used, create encryption key in the specified namespace. Default: 'kube-system'.")
```
ianlewis: I suppose this should be the namespace that the dashboard is deployed in? Maybe we should allow setting this via an environment variable? Setting a value via an argument using the downward API is a bit wonky.

Maybe just make the default into os.GetEnv("POD_NAMESPACE") and then to 'kube-system' if that's empty?

```
var defNamespace = getEnv("POD_NAMESPACE", "kube-system")
var argNamespace  = pflag.String("namespace", defNamespace, "When non-default namespace is used, create encryption key in the specified namespace. Default: '"+defNamespace+"'.")

func getEnv(name, def string) string {
  val := os.GetEnv(name)
  if val == "" {
    val = def
  }
  return val
}
```

ianlewis: I'm also curious why we don't already set POD_NAME, POD_NAMESPACE, POD_IP in the src/deploy/**/*.yaml files since it seems to be used in src/app/backend/cert/ecdsa/creator.go?

That might be a question for @maciaszczykm though.

src/app/backend/auth/jwe/keyholder.go
```diff
@@ -136,7 +138,7 @@ func (self *rsaKeyHolder) getEncryptionKeyHolder() runtime.Object {
 	priv, pub := ExportRSAKeyOrDie(self.Key())
 	return &v1.Secret{
 		ObjectMeta: metaV1.ObjectMeta{
-			Namespace: authApi.EncryptionKeyHolderNamespace,
```
ianlewis: AFAICT EncryptionKeyHolderNamespace is used in a few other places. Like auth/api/common.go looks like it is preventing dashboard from overwriting it's own config. Have those been fully addressed?

FrenchBen:
@ianlewis thanks for taking a look over the PR and providing some valuable comments - I'll get things patched up, and will confirm that anything I removed didn't affect the deployment any further.

src/app/backend/dashboard.go
kennethredler: Yes please to leveraging environment variable instead of (or at least in addition to) a new flag.
Then YAMLs can simply add:

```yaml
      env:
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
```

FrenchBen:
Updated the tests related to the integration manager; as of k8s version 1.12 these didn't pass for me.

i18n/messages-en.xtb
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE translationbundle><translationbundle lang="en">
+<translationbundle lang="en">
```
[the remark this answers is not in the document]
FrenchBen: Nope - I did a rebase from master - I can rollback on that?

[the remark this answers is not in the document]
floreks: Definitely revert that.

src/app/backend/auth/jwe/keyholder.go
jeefy: @FrenchBen This looks to be the last dangling thing before we can lgtm this. 

Several files under auth still reference EncryptionKeyHolderNamespace that should become dynamic. Happy to work with you to move this along if you need any help. 👍 Thanks!

monotek:
Can somebody please give a short update on the status of this pr?

jeefy:
@monotek Waiting on one final change/discussion point from @FrenchBen before this moves along. Shouldn't be much longer now that a lot of us are back from break and not completely engrossed in larger projects. :)

FrenchBen:
@jeefy it was my understanding that I had completed the latest changes? Is there something I missed?

jeefy:
/lgtm 😄 

I'm going to let this hang for one more day (Sorry!) to let anyone else chime in, but I'll approve it in 24h. The only thing left (that comes to mind) is once this is merged, we will need to update documentation (Will create issue when relevant) in the wiki. 

Thanks @FrenchBen !!

davidkarlsen:
Awesome! Will there be a release soon?

floreks:
Unfortunately, this feature is not 100% ready. Certificates are still stored in a `kube-system` namespace only and in case dashboard will be deployed in a different namespace with different privileges it will fail to start.

maciaszczykm:
Should we create tracker issue for that?

floreks:
Definitely. It's quite important as it breaks dashboard in certain cases.

FrenchBen:
@floreks 🤔 is this true? I've been using the early (and ugly) version of this PR, and have the `kubernetes-dashboard-certs` secrets properly created in my namespace. I know this to be 💯 correct, as I don't have access to the `kube-system` namespace. 

Can you point to where the `kube-system` namespace is enforced when the secrets are created?

floreks:
@FrenchBen by default if you do not alter the yaml, certs are stored in-memory. Only if you choose to use custom certificates, you have to store them in a secret. In such case if Dashboard would be deployed in a different namespace than `kube-system` and it would try to access a secret in a `kube-system` namespace it might fail and crash.

FrenchBen:
I did modify the YAML some to get it deployed - perhaps I got lucky?
@floreks Thanks for the follow up PR that cleans things up quite nicely 👏

floreks:
You would need to pass `--tls-cert-file` and `--tls-key-file` arguments with you custom certificate to the Dashboard. They are normally stored in a `kubernetes-dashboard-certs` secret, but in case in-memory certs are used, this secret is obsolete.

monotek:
It also worked if you deployed the dashboard to kube-syetem first and after that to another namespace.

monotek:
yes, just wanted to give an explanation why it may have worked for some people...

PR #3198, opened 2018-08-04
filter /token/refresh contents from logs

wpalmeri:
Tokens are currently being written to the logs. Add the `/api/v1/token/refresh` url to the map of sensitiveUrls which should be scrubbed from the logs.


wpalmeri:
signed CLA

Issue #3115, opened 2018-06-28
Rbac kube-proxy dashboard


bhack:
Can we document what kind rolebinding we need to create to give an user the access to the dashboard trought the proxy?
As in the `Readme.md` is still the default access to the dashboard and Rbac now is enabled by default in K8s.

I've found [this reply](https://stackoverflow.com/a/47796023) and probably we could put something like this in the docs.

jeefy:
/assign @jimangel 

I do think we should put a small blurb in the base README when it comes to installing and RBAC.

@jimangel You mind taking this on? I think being able to link to https://github.com/kubernetes/dashboard/blob/master/src/deploy/recommended/kubernetes-dashboard.yaml#L43 and saying "default install has minimal permissions" is a good start, especially as we address more security concerns. :)

Thanks!

jimangel:
The README already had a blurb in there about minimal RBAC, but I will submit a PR to beef it up a bit.

PR #3211, opened 2018-08-17
Minor tweaks to README

jimangel:
I've added more emphasis on RBAC (https://github.com/kubernetes/dashboard/issues/3115) and community info at the bottom.

As we move forward with the new release (and migration), I will look to modify the wiki's more inline with RBAC as a default (and compatibility information).


jeefy:
I am a nit picker :) Thanks!

/lgtm

PR #3213, opened 2018-08-17
Update sort-translations.xslt

wangxy518:
the url has been changed.


jeefy:
The URL hasn't changed. The namespace is still http. An excerpt from the contents of the link:

> The namespace http://www.w3.org/1999/XSL/Transform is referred to as "the XSLT namespace". The prefix xsl is conventionally used to refer to this namespace (and is so used both within this document and within the XSLT specification), but it has no special status: any prefix may be used.

If I'm totally off base, please re-open and let me know. :)

/assign
/close

PR #3233, opened 2018-08-25
Don't claim copyright for future years

bmwiedemann:
When building the kubernetes-dashboard openSUSE package
in future years, files differed because they contained
something like
this.latestCopyrightYear="2033"

See https://reproducible-builds.org/ for why this matters.

And https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated suggests that expiry should not be a concern anyway.

https://github.com/kubernetes/kubernetes/pull/59172 fixed this for kubernetes.

### DIFF
--- a/build/script.js
+++ b/build/script.js
@@ -242,7 +242,7 @@ function patchBuildInformation() {
                   patterns: [
                     {match: 'BUILD_GIT_COMMIT', replacement: commit},
                     {match: 'BUILD_DASHBOARD_VERSION', replacement: conf.deploy.version.release},
-                    {match: 'BUILD_YEAR', replacement: new Date().getFullYear()},
+                    {match: 'BUILD_YEAR', replacement: conf.deploy.version.year},
                   ],
                 }));
 }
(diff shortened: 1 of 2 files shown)


bmwiedemann:
/assign @mhenc 

@sttts

build/script.js
```diff
@@ -242,7 +242,7 @@ function patchBuildInformation() {
                   patterns: [
                     {match: 'BUILD_GIT_COMMIT', replacement: commit},
                     {match: 'BUILD_DASHBOARD_VERSION', replacement: conf.deploy.version.release},
-                    {match: 'BUILD_YEAR', replacement: new Date().getFullYear()},
+                    {match: 'BUILD_YEAR', replacement: "2018"},
```
jeefy: Instead of hard-coding the date here, instead I'd put it in build/conf.js and reference the new variable. Since this seems tied to the release/commit, look for the "version" variable and maybe add a "buildDate" attribute?

bmwiedemann: actually, I think the 'build' part in the name can be misleading. It is meant to contain the date/year of the last source change.
Just wanted to keep the patch simple for the start.
Will push an update for using conf.js

PR #3269, opened 2018-09-24
Update docs and add maciaszczykm@ to sig-ui-leads

maciaszczykm:
- [x] add @maciaszczykm to sig-ui-leads
- [x] update `CONTRIBUTING.md` and keep only 1 file
- [x] prepare document describing roles within project community

I am open for suggestions. I am wondering if we should describe roles like member, contributor etc., because at the moment I've described roles only for SIG UI leads.

/cc @bryk @floreks @danielromlein


jeefy:
Two thoughts:
- Do we technically have a reviewer level contributor? If so we should document that here.
- The [WIP charter](https://docs.google.com/document/d/1GjYKVmOpKCxjW8T_iyxqn2XPBup8sVZ1LWfZHh1wXzE/edit?usp=sharing) will need to link to ROLES.md 

Otherwise lgtm!

maciaszczykm:
@bryk @jeefy Thanks!

PR #3272, opened 2018-09-27
Fix typos issues in backend files

mooncak:
Remove duplicate words in files


src/app/backend/auth/api/types.go
```diff
@@ -84,7 +84,7 @@ type AuthManager interface {
 // TokenManager is responsible for generating and decrypting tokens used for authorization. Authorization is handled
 // by K8S apiserver. Token contains AuthInfo structure used to create K8S api client.
 type TokenManager interface {
-	// Generate secure token based on AuthInfo structure and save it it tokens' payload.
+	// Generate secure token based on AuthInfo structure and save it tokens' payload.
```
DStorck: this sentence could actually be cleaned up more, it reads kind of funny. `save it tokens' payload` could be clearer

PR #3296, opened 2018-10-18
Angular migration updates

maciaszczykm:


src/app/frontend/common/components/actionbar/detailactions/exec/template.html
```diff
@@ -14,11 +14,11 @@
...
-        class="kd-toolbar-action"
-        matTooltip="Exec into pod"
-        i18n-matTooltip
-        (click)="onClick()">
+<a mat-icon-button
+   color="accent"
+   class="kd-toolbar-action"
+   matTooltip="Exec into pod"
+   i18n-matTooltip
+   [href]="getHref()">
```
pengx17: Previously the logs view was opened in a new window. Do you think it is better to make this a tag target="_blank?"

src/app/frontend/common/dialogs/download/template.html
```diff
@@ -15,18 +15,16 @@
 -->
 
 <h2 *ngIf="!hasForbiddenError()"
-    mat-dialog-title>Downloading log file</h2>
+    mat-dialog-title>Download logs file</h2>
```
pengx17: i18n?

src/app/frontend/logs/template.html
```diff
@@ -22,13 +22,15 @@
         <mat-select [(ngModel)]="container"
                     class="kd-logs-toolbar-select"
                     (selectionChange)="loadNewest()">
-          <mat-optgroup label="Containers">
+          <mat-optgroup label="Containers"
```
pengx17: i18n?

src/app/frontend/common/dialogs/download/template.html
```diff
@@ -15,18 +15,16 @@
...
   <ng-container *ngIf="!hasForbiddenError()">
-    <div>Please wait until download has finished.</div>
-    <div>
-      Downloaded:{{loaded | kdMemory}}
-    </div>
+    <div>Size: {{loaded | kdMemory}} B</div>
     <mat-progress-bar class="kd-download-progress-bar"
                       [mode]="getDownloadMode()"
                       value="100"></mat-progress-bar>
+    <div *ngIf="!finished">Preparing file to download...</div>
```
pengx17: ditto

src/app/frontend/common/components/actionbar/detailactions/exec/template.html
maciaszczykm: It was done on purpose. We will open in the same window and later on there will be a swich in the settings allowing to choose what do you prefer.

src/app/frontend/common/dialogs/download/template.html
maciaszczykm: Thank you 👍

maciaszczykm:
@jeefy @pengx17 Thank you guys. Let's see if your LGTMs will be enough to override codecov check 😄

PR #3305, opened 2018-10-22
Update info.html

AdamDang:
remove some duplicated words

PR #3306, opened 2018-10-22
Update messages-en.xtb

AdamDang:
remove duplicated words "that that"


maciaszczykm:
Regenerate translations under #3305 as it is the same text.  Use the command line as you have missed some of the languages.

/close

PR #3310, opened 2018-10-24
Support authenticating via auth-provider token.

### LINKED ISSUE #2849: Signin with the kubeconfig created with gcloud container clusters get-credentials doesnt work
##### Environment
I use a cluster which was created by the google container engine and has a preinstalled dashboard.
Since 1.8.x i need to prove a accesstoken or kubeconfig file via login maskl.

```
Dashboard version: v1.8.0 Git Commit:054916b8fbadebd32dc16d32108fdd3c9ccbf400
Kubernetes version1.8.7-gke.1
```

##### Steps to reproduce
I authenticated with ```gcloud container clusters get-credentials``` this creates a kubeconfig file with has this structure (I omitted the value)

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ...
    server: ...
  name: ...
contexts:
- context:
    cluster: ...
    user: ...
  name: ...
current-context: ...
kind: Config
preferences: {}
users:
- name: ...
  user:
    auth-provider:
      config:
        access-token: ...
        cmd-args: config config-helper --format=json
        cmd-path: ...
        expiry: 2018-02-21T07:58:23Z
        expiry-key: ...
        token-key: ..
      name: gcp

```

##### Observed result
In the auth I select the created config file. auth does fail with : Not enough data to create auth info structure.

image

When i choose the accesstoken opton and copy and paste the accestoken from the file it works

##### Expected result
selecting the file that was created from get-credentials should work

bmcstdio:
Fixes #2849 by falling-back to the value of `auth-provider.config.token` in case the current `user` entry's token is empty.

### DIFF
--- a/src/app/backend/auth/kubeconfig.go
+++ b/src/app/backend/auth/kubeconfig.go
@@ -39,10 +39,19 @@ type userEntry struct {
 	User userInfo `yaml:"user"`
 }
 
+type authProviderConfig struct {
+	AccessToken string `yaml:"access-token"`
+}
+
+type authProviderInfo struct {
+	Config authProviderConfig `yaml:"config"`
+}
+
 type userInfo struct {
-	Token    string `yaml:"token"`
-	Username string `yaml:"username"`
-	Password string `yaml:"password"`
+	AuthProvider authProviderInfo `yaml:"auth-provider"`
+	Token        string           `yaml:"token"`
+	Username     string           `yaml:"username"`
+	Password     string           `yaml:"password"`
 }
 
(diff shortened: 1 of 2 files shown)


jeefy:
/lgtm

I'd like to find a way to be able to test it though. There currently aren't any existing tests for the logic in kubeconfig.go though. 🤔

bmcstdio:
@jeefy I've pushed some tests targeting the existing and new functionalities.

jeefy:
/lgtm

Thanks!

PR #3315, opened 2018-10-29
Enable check:backend task on Travis CI and add goimports check

shu-mutou:
`check:backend` task always return success even if linters reported errors now.
And golangci-lint does not work properly for checking source code outside `${GOPATH}`.

Removed `--issue-exit-code=0` option for `golangci-lint` and change directory into `${GOPAHT}/src/github.com/kubernetes/dashboard` before run `golangci-lint` on Travis CI.

Also, added `enable=goimports` option for `golangci-lint` and applied "goimports -w src/app/backend".

### DIFF
--- a/aio/scripts/lint-backend.sh
+++ b/aio/scripts/lint-backend.sh
@@ -25,13 +25,18 @@ if [ ! -f ${GOLINT_BIN} ]; then
     curl -sfL ${GOLINT_URL} | sh -s -- -b ${CACHE_DIR} v1.12.3
 fi
 
+# Need to check source files under GOPATH
+if [ ${TRAVIS} ]; then
+    cd ${GOPATH}/src/github.com/kubernetes/dashboard/src/app/backend/
+fi
+
 # Run checks.
 ${GOLINT_BIN} run ./... \
   --no-config \
-  --issues-exit-code=0 \
   --deadline=30m \
   --disable-all \
   --enable=govet \
   --enable=gocyclo \
   --enable=misspell \
-  --enable=ineffassign
+  --enable=ineffassign \
+  --enable=goimports
--- a/src/app/backend/resource/job/list_test.go
+++ b/src/app/backend/resource/job/list_test.go
@@ -140,7 +140,7 @@ func TestGetJobListFromChannels(t *testing.T) {
 					{
 						ObjectMeta: metaV1.ObjectMeta{
 							Namespace: "rs-namespace",
-							Labels:	   map[string]string{"foo": "bar"},
+							Labels:    map[string]string{"foo": "bar"},
 							OwnerReferences: []metaV1.OwnerReference{
 								{
 									Name:       "rs-name-running-pod",


shu-mutou:
I switched order of commits between this and relevant PR that adds check-go-format task.

shu-mutou:
goimports did not work properly with "trusty" image for travis ci, so I proposed switching the image to "xenial" in another PR (#3340) and rebased this to it.

ianlewis:
@shu-mutou 
Can you update this PR after #3324 is merged?

jeefy:
@shu-mutou Tests are failing because `check-go-format` is catching additional formatting errors

Please see:
```
[00:55:27] Starting 'check-go-format'...
src/app/backend/resource/deployment/detail.go
src/app/backend/resource/deployment/detail_test.go
src/app/backend/resource/horizontalpodautoscaler/list_test.go
src/app/backend/resource/pod/detail.go
src/app/backend/resource/pod/detail_test.go
src/app/backend/resource/pod/fieldpath_test.go
src/app/backend/resource/replicaset/detail.go
src/app/backend/resource/replicaset/detail_test.go

[00:55:28] 'check-go-format' errored after 1.16 s
```

I just tested it locally and got the same errors. Are you getting that locally as well? Should be able to fix those and have the tests pass. :)

shu-mutou:
@jeefy Thanks for reviewing! In my env, check-go-format goes well. I'm investigating difference between my env and ci env...

shu-mutou:
And more files are marked as incorrect in travis-ci env.... :/

shu-mutou:
@jeefy Could you let me know go version in your environment?

shu-mutou:
Removing node_modules and re-installing dashboard, I could reproduce the errors same as travis-ci. And push this again. :)

floreks:
I am not sure if we should proceed with that as `v1.10.1` will be the last release using AngularJS. On the migration branch we have moved away from gulp almost completely. This change would be lost.

shu-mutou:
I see. I will investigate angular migration branch. Then I will seek the way to help the migration branch.

aio/scripts/conf.sh
```diff
@@ -38,6 +38,7 @@ BEAUTIFY_BIN="${ROOT_DIR}/node_modules/.bin/js-beautify"
 GLOB_RUN_BIN="${ROOT_DIR}/node_modules/.bin/glob-run"
 GOLINT_URL="https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh"
 GOLINT_BIN="${CACHE_DIR}/golangci-lint"
+GOIMPORTS_BIN="${TMP_DIR}/bin/goimports"
```
floreks: How did `goimports` get to the temp dir?

maciaszczykm: I agree 👍

shu-mutou: I copied `goimports` into `.cached_tools` 👌

shu-mutou:
Hmm.... on Travis, `goimports` is not copied into `.cached_tools`.

shu-mutou:
> Hmm.... on Travis, `goimports` is not copied into `.cached_tools`.

Got it! Copied successfully from `${GOPATH}/bin/goimorts` instead of `$(which goimports)`.

maciaszczykm:
This while change can be reduced to one line. Take a look at: https://github.com/golangci/golangci-lint#quick-start

```
--enable=goimports \
```

shu-mutou:
I have already tried `--enable=goimports` option, but golangci-lint always replies status code 0 even if go code is unformated. So I think the option is not useful for check on CI.

jeefy:
@shu-mutou I think that the error I caught in #3579 fixes the behavior you saw (of always exiting 0)

When I tested it locally with some poorly formatted go, it didn't catch it when it was set to use `../...` but did catch it when it was set to `./...`

Please check it and see if this resolves things for you as well. :)

Thanks!

shu-mutou:
/hold

@jeefy, thank you for your follow up. I checked behaviors of golangci-lint and report as follows.

As we say, golangci-lint shows message as "File is not `goimports`-ed (goimorts)" if our code is unformatted when we run `goimorts` with `--enable-goimports` option localy. But this returns status code 0 even if it shows such message.

Actually our code has unformatted code as follow.
https://github.com/kubernetes/dashboard/blob/cf6f04d9e24feff4802f360cba5e2ebc1d4b043a/src/app/backend/resource/job/list_test.go#L143
This code has "\t" after "Labels:".

So we should get error on ci when checked #3579 too, but actually not.

In this PR, please confirm https://travis-ci.org/kubernetes/dashboard/jobs/495238705#L301 .

As you guys see above, `golangci-lint --enable-goimports` does not return error code (always 0) even if the code is not unformatted.
So CI does not report error on this PR.

As my conclusion, I added `goimports` execution independent from `golangci-lint`.

If you guys confirm this behavior and agree with I said, I'll modify my PR again. :wink:

shu-mutou:
/hold cancel

I found reason why golangci-lint always returns 0.
1) `--issue-exit-code=0` option
2) golangci-lint does not work for checking source code outside $GOPATH.

maciaszczykm:
Great!

/lgtm
/approve

PR #3323, opened 2018-11-06
Add container for development outside cluster

shu-mutou:
To construct thinner development environment outside cluster
with docker, this adds Dockerfile and shell scripts.

This container will run npm command as you like, or install and run 
dashboard. 

Also, this adds chrome driver as headless for testing in container.

The following commands are reviewed in this:
* npm ci
* npm run check
* npm run test:frontend
* npm run test:backend
* npm run start

To run this container:
1. Copy kubeconfig from your cluster, and confirm URL for API server in it.
2. Set filepath for kubeconfig into K8S_DASHBOARD_KUBECONFIG environment variable.
3. Change directory into your dashboard source directory.
4. To install and run dashboard, simply run `aio/develop/run-npm-on-container.sh`.

If you want to run npm command in this container as you like, run like follows:
`aio/develop/run-npm-on-container.sh run check`
`aio/develop/run-npm-on-container.sh run test:backend`

### DIFF
--- a/aio/karma.conf.js
+++ b/aio/karma.conf.js
@@ -90,5 +90,21 @@ module.exports = function(config) {
     };
   }
 
+  // Use custom browser configuration when running on container.
+  if (!!process.env.K8S_DASHBOARD_CONTAINER) {
+    configuration.browsers = ['ChromeHeadless'];
+    configuration.customLaunchers = {
+      ChromeHeadless: {
+        base: 'Chrome',
+        flags: [
+          '--disable-gpu',
+          '--headless',
+          '--no-sandbox',
+          '--remote-debugging-port=9222',
+        ],
+      },
+    };
+  }
+
   config.set(configuration);
 };
--- a/aio/develop/Dockerfile
+++ b/aio/develop/Dockerfile
@@ -1,0 +1,70 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This is a Dockerfile for running and building Kubernetes dashboard
+# It installs all deps in the container and adds the dashboard source
+# This way it abstracts all required build tools away and lets the user
+# run gulp tasks on the code with only docker installed
+
+# golang is based on debian:jessie
+FROM golang
+
+# Install Node.js. Go is already installed.
+# A small tweak, apt-get update is already run by the nodejs setup script,
+# so there's no need to run it again.
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - \
+  && apt-get install -y --no-install-recommends \
+	nodejs \
+	patch \
+	chromium \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& apt-get clean
+
+# Install dependencies. This will take a while.
+RUN npm install -g npm@latest gulp
+
+# Set environment variable for JavaScript tests.
+ENV CHROME_BIN=/usr/bin/chromium
+
+# Set environment variable for terminal
+ENV TERM=xterm
+
+# Add ${GOPATH}/bin into ${PATH}
+ENV PATH=${GOPATH}/bin:${PATH}
+
+# For testing, etc., to know if this environment is on container.
+ENV K8S_DASHBOARD_CONTAINER=TRUE
+
+# Download a statically linked docker client,
+# so the container is able to build images on the host.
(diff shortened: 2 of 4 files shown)


build/run-dev-env-in-docker.sh
```diff
@@ -0,0 +1,55 @@
...
+	--net=host \
+	--name=${K8S_DASHBOARD_CONTAINER_NAME} \
+	-v /var/run/docker.sock:/var/run/docker.sock \
+	-v ${K8S_DASHBOARD_SRC}:/dashboard \
+	${DOCKER_RUN_OPTS} \
+	${DASHBOARD_IMAGE_NAME} \
+	sh
+
+echo "Install dependencies and dashboard from your source directory."
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} npm install -g npm@latest gulp
```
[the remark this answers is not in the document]
shu-mutou: Yes, you're right. Sorry, I can't really remember why I added duplicated code. This seems unnecessary.

build/run-dev-env-in-docker.sh
```diff
@@ -0,0 +1,55 @@
...
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} npm install --unsafe-perm
+
+echo "Install development tools."
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} go get golang.org/x/tools/cmd/goimports
```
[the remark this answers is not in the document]
shu-mutou: done.

build/run-dev-env-in-docker.sh
```diff
@@ -0,0 +1,55 @@
...
+
+echo "Remove existing container"
+docker rm -f ${K8S_DASHBOARD_CONTAINER_NAME}
+
+# Always test if the image is up-to-date. If nothing has changed since last build,
+# it'll just use the already-built image
+docker build -t ${DASHBOARD_IMAGE_NAME} -f ${DIR}/Dockerfile ${DIR}/../
+
+# Run dashboard container for development and expose necessary ports automatically.
+docker run \
```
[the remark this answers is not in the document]
shu-mutou: Already the script to run a process is there, i.e. build/run-gulp-in-docker.sh, that uses same Dockerfile. But if we change source files, the "docker build" can not use cache for tasks after copying sources. i.e. for installing dependencies, and it takes a while in each execution of a process.

I don't know who uses existing script, so I proposed adding new script as now. If nobody uses the existing script, especially any other scripts for test or deploy does not use it, it may be better to improve the exisiting one.

aio/karma.conf.js
```diff
@@ -66,7 +66,18 @@ module.exports = function(config) {
     colors: true,
     autoWatch: true,
     port: 9876,
-    browsers: ['Chrome'],
+    browsers: ['ChromeHeadless'],
```
floreks: I am not sure if we want to change the default to headless. Headless is used by the CI and for local development it's easier to debug with standard chrome.

shu-mutou: I added new config for container in next commit. Thank you for your reviewing. :)

aio/develop/run-development-container.sh
```diff
@@ -0,0 +1,65 @@
...
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} rm -fr .tmp/*
+
+# Copy backend source for build
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} mkdir -p .tmp/src/github.com/kubernetes/dashboard/src/app
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} cp -r src/app/backend/ .tmp/src/github.com/kubernetes/dashboard/src/app
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} ln -s ${K8S_DASHBOARD_SRC_ON_CONTAINER}/vendor ${K8S_DASHBOARD_SRC_ON_CONTAINER}/.tmp/src/github.com/kubernetes/dashboard/src/app/backend/vendor
+
+# Copy kubeconfig into /root/.kube/config on the container
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} mkdir -p /root/.kube
+docker cp ${K8S_DASHBOARD_KUBECONFIG} ${K8S_DASHBOARD_CONTAINER_NAME}:/root/.kube/config
```
floreks: It would be better to mount kubeconfig dir into the container. In case it changes we don't need to restart the container.

aio/develop/run-development-container.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Install dashboard
+docker exec ${K8S_DASHBOARD_CONTAINER_NAME} npm ci --unsafe-perm
```
floreks: Is `--unsafe-perm` actually required?

floreks:
I agree with @ianlewis here. From the user perspective, it would be easier to use such script as:
`./script.sh run start`
`./script.sh run check`
etc.

Additionally, I'd prefer mounting source dir and kubeconfig dir into the container so all the changes are reflected in real time. This is a good start though.

aio/develop/run-development-container.sh
shu-mutou: Nice catch! I moved this into "docker run" command as -v option.

aio/develop/run-development-container.sh
shu-mutou: No. ;)

shu-mutou:
After all, "- unsafe - perm" was necessary.

maciaszczykm:
I think we can add this and iterate on it. Seems like all review comments were implemented.

/lgtm
/approve

PR #3334, opened 2018-11-10
Fix namespace issue

mooncak:
Fix namespace issue


jeefy:
/lgtm

Not really a "namespace issue" but just fixes a misspelling, tests don't functionally change.

PR #3340, opened 2018-11-13
Switch image for travis ci to xenial

shu-mutou:
trusty is too old to use it now.


shu-mutou:
We can set "bionic", the setting does not causes error, however travis-ci seems to use "xenial" instead !? So I set "xenial" as now.

PR #3347, opened 2018-11-16
Bump go version to 1.10.3

shu-mutou:
We uses Ubuntu 16.04.5 image in CI, and the image uses go 1.10.3.
So we should bump up the go version on prerequisites by changing
minGoVersion constant and "Getting started" wiki.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ matrix:
       - SAUCE_USERNAME=k8s-dashboard-ci
 
 before_script:
-  - eval "$(gimme 1.8.3)"
+  - eval "$(gimme 1.10.3)"
   - docker --version
 
 script:
--- a/build/gocommand.js
+++ b/build/gocommand.js
@@ -35,7 +35,7 @@ const env = lodash.merge(process.env, {GOPATH: sourceGopath, PATH: devPath});
 /**
  * Minimum required Go Version
  */
-const minGoVersion = '1.8.3';
+const minGoVersion = '1.10.3';
 
 /**
  * Spawns a Go process after making sure all Go prerequisites are
--- a/src/app/backend/integration/manager_test.go
+++ b/src/app/backend/integration/manager_test.go
@@ -58,14 +58,14 @@ func TestIntegrationManager_GetState(t *testing.T) {
 			"Server provided and using in-cluster heapster",
 			"http://127.0.0.1:8080", "", &api.IntegrationState{
 				Connected: false,
-				Error:     errors.New("Get http://127.0.0.1:8080/api/v1/namespaces/kube-system/services/heapster/proxy/healthz: dial tcp 127.0.0.1:8080: getsockopt: connection refused"),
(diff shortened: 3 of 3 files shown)


maciaszczykm:
@shu-mutou Update also:

https://github.com/kubernetes/dashboard/blob/5728e0303a742a889d9a25c7ae9c07e426917e90/.travis.yml#L5

shu-mutou:
thanks! I was just seeking the setting for "gimme".

shu-mutou:
I see. thanks. I will update wiki after this PR merged. ;)

maciaszczykm:
Some tests failed due to Go update. You need to fix them :)

shu-mutou:
Yup, it seems to fail connecting to heapster. I need to investigate about heapster. ToT

maciaszczykm:
@shu-mutou Look:

```
--- FAIL: TestIntegrationManager_GetState (0.00s)
	manager_test.go:87: Test Case: Server provided and using in-cluster heapster. Expected state to be: &{false 0001-01-01 00:00:00 +0000 UTC Get http://127.0.0.1:8080/api/v1/namespaces/kube-system/services/heapster/proxy/healthz: dial tcp 127.0.0.1:8080: getsockopt: connection refused}, but got &{false 2018-11-16 09:41:05.575885369 +0000 UTC m=+0.008780309 Get http://127.0.0.1:8080/api/v1/namespaces/kube-system/services/heapster/proxy/healthz: dial tcp 127.0.0.1:8080: connect: connection refused}.
	manager_test.go:87: Test Case: Server provided and using external heapster. Expected state to be: &{false 0001-01-01 00:00:00 +0000 UTC Get http://127.0.0.1:8081/healthz: dial tcp 127.0.0.1:8081: getsockopt: connection refused}, but got &{false 2018-11-16 09:41:05.576628211 +0000 UTC m=+0.009523147 Get http://127.0.0.1:8081/healthz: dial tcp 127.0.0.1:8081: connect: connection refused}.
```

It is connected to Heapster but look at the timestamps. That seems to be the real reason.

shu-mutou:
Yes, it seemed to be logged with timestamp. But actually checked with only message as follow: https://github.com/kubernetes/dashboard/blob/master/src/app/backend/integration/manager_test.go#L85-L89
We should fix the log also.

jeefy:
The tests in manager_test.go actually ignore the timestamp.

The real problem is this:
`tcp 127.0.0.1:8081: getsockopt: connection refused`
`tcp 127.0.0.1:8081: connect: connection refused`

I'm getting this locally as well (on master) so I'm thinking that the error text changes between Golang versions. We would need to go through our tests and audit/update them all for 1.10.3.

I think this is a worthwhile effort, and should be part of this PR. Thoughts @maciaszczykm?

shu-mutou:
@jeefy nice catch! I was closely watching only the timestamp. I fixed this test and if new error occurs due to bumping go version I will address them.

maciaszczykm:
Any progress with the tests?

maciaszczykm:
Restarting the build solved the problem :)

shu-mutou:
@maciaszczykm Thanks for restarting build! I could not work in the second half of last week for k8s. I also edited wiki, so please confirm it.

maciaszczykm:
It's okay, thanks!

PR #3349, opened 2018-11-19
Fix inconsistent logs on integration manager test

shu-mutou:
Errors are logged with timestamp on integration manager test,
but actually checked with only message.

This PR fixes inconsistent log on it.
Also, this divides logs for connection error and for difference on messages.


maciaszczykm:
CI is not passing. Make it green before the review.

Issue #3357, opened 2018-11-23
filter job which not belong to other workloads


zouyee:
##### Environment

In the process of using, we found that cronjob will generate a large number of jobs, and in dashboard project, there is no filtering operation for this. For the jobs which has owner, we may filter them out better.
```
Dashboard version: master
Kubernetes version: 1.11
/kind feature
```

maciaszczykm:
We should include this as advanced filtering options. Not by default.

/close

PR #3358, opened 2018-11-23
filter job which has not pure job

zouyee:
https://github.com/kubernetes/dashboard/issues/3357
For the owned jobs, we may filter them out better.
/assgin @maciaszczykm


jeefy:
I'd prefer to not globally set this. If I'm debugging an issue with a cluster, I may want to see all job objects across my whole cluster, not just jobs without an OwnerReference.

I do like the idea of being able to filter things like this for day-to-day use, and a toggle in the UI would be welcome. Thoughts?

maciaszczykm:
Closing this one. Please reopen/create a new pull request if you would like to make it more interactive.

/close

PR #3360, opened 2018-11-24
delete image reference validating info

zhuguihua:
delete image reference validating info.

### DIFF
--- a/src/app/backend/validation/validateimagereference.go
+++ b/src/app/backend/validation/validateimagereference.go
@@ -15,8 +15,6 @@
 package validation
 
 import (
-	"log"
-
 	"github.com/docker/distribution/reference"
 )
 
@@ -36,8 +34,6 @@ type ImageReferenceValidity struct {
 
 // ValidateImageReference validates image reference.
 func ValidateImageReference(spec *ImageReferenceValiditySpec) (*ImageReferenceValidity, error) {
-	log.Printf("Validating %s as an image reference", spec.Reference)
-
 	s := spec.Reference
 	_, err := reference.ParseNamed(s)
 	if err != nil {


zhuguihua:
I signed it

zhuguihua:
I signed it

zhuguihua:
I signed it

jeefy:
I feel like this would make dashboard's logs unnecessarily chatty.

Additionally, if `ValidateImageReference` returns an error it already gets handled.

What value are you trying to add, or what problem are you trying to solve? Thanks in advance!

zhuguihua:
I'd like to know the reason in log when validating image reference failed.

src/app/backend/validation/validateimagereference.go
```diff
@@ -41,7 +41,10 @@ func ValidateImageReference(spec *ImageReferenceValiditySpec) (*ImageReferenceVa
 	s := spec.Reference
 	_, err := reference.ParseNamed(s)
 	if err != nil {
+		log.Printf("Validation result for %s image reference is false, reason: %v", spec.Reference, err)
 		return &ImageReferenceValidity{Valid: false, Reason: err.Error()}, nil
```
shu-mutou: I'm not sure about subsequent handling ( https://github.com/kubernetes/dashboard/blob/664f533db09fc94b158536b5d6cc75fcf99a7283/src/app/backend/handler/apihandler.go#L1054-L1058 ), 
however should the "nil" in L45 and L49 be err object to pass it to caller first?

zhuguihua:
When image name was changed, this api would be called. Maybe it is no necessary to show validating info, just only show validation result?

maciaszczykm:
@zhuguihua Still too much info in the logs IMO.

zhuguihua:
@maciaszczykm I go to try this api again, there is no need to show the validation info indeed. And now I think the validating info can also be deleted.

jeefy:
Your tests are failing because `validateimagereference.go` is still importing the `log` library and it no longer needs to. 

Please remove that import and push your changes. The tests should pass and I can lgtm it. 

Thanks!

zhuguihua:
@jeefy  Thanks, updated.

PR #3399, opened 2018-12-07
Update apiVersion

koooge:


jeefy:
Awesome find.

We should (in our abundant free time) audit the codebase and update these as well. :)

/lgtm

PR #3400, opened 2018-12-07
Security fix (CVE-2018-18264)

### LINKED ISSUE #2672: Remove skip login option

floreks:
Fixes #2672

`Skip` option has been disabled by default and access to the Dashboard SA has been restricted to our backend only. In case user tries to send a request without auth info 401 error will be returned.

If user tries to "manually" skip login page by adding cookie in the frontend, every page that tries to access API server will look like that:
zrzut ekranu z 2018-12-07 11-13-13

Thanks to #3077 most of the work has been already done.


maciaszczykm:
Looks good to me. I will do some more testing tomorrow and if it is also okay with others let's make a release to solve this critical issue.

src/app/backend/client/manager.go
```diff
@@ -279,6 +283,10 @@ func (self *clientManager) extractAuthInfo(req *restful.Request) (*api.AuthInfo,
 		return self.tokenManager.Decrypt(jweToken)
 	}
```
liggitt: not knowing how this method is used, does returning `nil,nil` mean the service account credentials are used instead?

if so, shouldn't the `if req == nil {` case above return an error?

build/serve.js
```diff
@@ -79,6 +79,10 @@ function getBackendArgs(mode) {
     args.push(`--apiserver-host=${conf.backend.envApiServerHost || conf.backend.apiServerHost}`);
   }
 
+  if (!conf.backend.disableSkipButton) {
```
liggitt: since `undefined` is falsy, would a missing config value re-enable the skip button? that seems fragile

src/app/backend/dashboard.go
```diff
@@ -66,7 +66,7 @@ var (
 	argMetricClientCheckPeriod   = pflag.Int("metric-client-check-period", 30, "Time in seconds that defines how often configured metric client health check should be run. Default: 30 seconds.")
 	argAutoGenerateCertificates  = pflag.Bool("auto-generate-certificates", false, "When set to true, Dashboard will automatically generate certificates used to serve HTTPS. Default: false.")
 	argEnableInsecureLogin       = pflag.Bool("enable-insecure-login", false, "When enabled, Dashboard login view will also be shown when Dashboard is not served over HTTPS. Default: false.")
-	argDisableSkip               = pflag.Bool("disable-skip", false, "When enabled, the skip button on the login page will not be shown. Default: false.")
+	argDisableSkip               = pflag.Bool("disable-skip", true, "When enabled, the skip button on the login page will not be shown. Default: false.")
```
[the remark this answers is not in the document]
liggitt: the double negative makes this hard to follow... can the flag be changed to `--enable-skip-login` and default to false?

src/app/backend/client/manager.go
liggitt: the "If request is nil then authentication will be skipped." comment on `Client(req)` sounds fragile (a call that doesn't pass in the request would escalate)

src/app/backend/dashboard.go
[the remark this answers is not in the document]
floreks: Sure we can but it also requires frontend and build pipeline refactoring. As `skip` option should be disabled by default then it is `argDisableSkip=true`.

src/app/backend/client/manager.go
floreks: I actually investigated our code to check if it would be possible that external request comes as nil, and it is not. `restful-go` will always initialize request object for us. The first `if req == nil` is used to initialize `insecureClient` that is used internally by our backend only and `return` at the end is there, not to break dashboard in case someone chooses to actually set `argDisableSkip` to `false`.

build/serve.js
floreks: This is only used by our development pipeline and can't actually affect anyone.

jessfraz:
awesome thanks, agree with @liggitt above but ovverall really great thanks!

src/app/backend/dashboard.go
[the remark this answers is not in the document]
maciaszczykm: I think we should do this anyways.

floreks:
Applied @liggitt 's suggestions and simplified our dev pipeline, not to break anything. To test it on serve/serve:prod use:
`gulp serve:prod --serveHttps --autoGenerateCerts --enableSkipButton=false`

build/conf.js
```diff
@@ -183,6 +183,11 @@ export default {
     systemBannerSeverity: gulpUtil.env.systemBannerSeverity !== undefined ?
         gulpUtil.env.systemBannerSeverity :
         '',
+    /**
+     * Allows to override disable skip button option on the backend.
+     */
+    enableSkipButton: gulpUtil.env.enableSkipButton !== undefined ? gulpUtil.env.enableSkipButton :
```
liggitt: simplify to `gulpUtil.env.enableSkipButton === true`?

src/app/backend/args/builder.go
```diff
@@ -132,15 +132,15 @@ func (self *holderBuilder) SetEnableInsecureLogin(enableInsecureLogin bool) *hol
...
-// SetDisableSettingsAuthorizer 'enable-settings-authorizer' argument of Dashboard binary.
-func (self *holderBuilder) SetDisableSettingsAuthorizer(enableSettingsAuthorizer bool) *holderBuilder {
-	self.holder.disableSettingsAuthorizer = enableSettingsAuthorizer
+// SetDisableSettingsAuthorizer 'disable-settings-authorizer' argument of Dashboard binary.
+func (self *holderBuilder) SetDisableSettingsAuthorizer(disableSettingsAuthorizer bool) *holderBuilder {
+	self.holder.disableSettingsAuthorizer = disableSettingsAuthorizer
 	return self
 }
 
 // SetDisableSkipButton 'disable-skip' argument of Dashboard binary.
```
liggitt: doc needs updating

src/app/backend/client/manager.go
liggitt: > I actually investigated our code to check if it would be possible that external request comes as nil, and it is not.

That may be true at this point in time, but a code path could be introduced that could call this with a nil internally as part of building a response to satisfy an external request.

> The first if req == nil is used to initialize insecureClient (by insecure I mean client that uses dashboard SA privileges) that is used internally by our backend only

I'd like to see the internal client building made more separate from the function that is used to return clients used to handle external requests, for safety

build/conf.js
floreks: The current one is actually correct as we need skip button to be enabled in our dev environments and tests by default. Most of our UI functionality tests rely on dashboard SA privileges.

It only affects our `gulp serve`, `gulp serve:prod` and tasks that rely directly on them, so tests mostly.

So in case someone simply uses `gulp serve` skip option will be visible. To disable it user will have to provide a `enableSkipButton` parameter. `gulp serve --enableSkipButton=false`.

floreks:
PTAL

build/conf.js
liggitt: would `gulp serve` or `gulp serve:prod` ever be used to actually run the dashboard?

floreks: @liggitt I don't think so. They are used only for manual tests before creating a PR or to check the changes. Also, they do not deploy container in the cluster but rather start one outside of the cluster.

src/app/backend/client/manager.go
```diff
@@ -290,6 +291,47 @@ func (self *clientManager) extractTokenFromHeader(authHeader string) string {
 	return ""
 }
 
+func (self *clientManager) isLoginEnabled(req *restful.Request) bool {
+	return req.Request.TLS != nil || args.Holder.GetEnableInsecureLogin()
+}
+
+// Secure mode means that every request to Dashboard has to be authenticated and privileges
+// of Dashboard SA can not be used.
+func (self *clientManager) isSecureModeEnabled(req *restful.Request) bool {
```
liggitt: this logic is extremely difficult to follow. can we do something like this as a first step to make sure this always returns true when skip login is not enabled?

```
if !args.Holder.GetEnableSkipLogin() {
  return true
}
```

floreks: It would have to be a bit different actually.

```go
if self.isLoginEnabled(req) && !args.Holder.GetEnableSkipLogin() {
  return true
}
```

In case login option is completely disabled we can not block user out of Dashboard. First we have to make sure that it is enabled and then skip option has to be disabled. Bascially if we would split that then whole method would become:
```go
func (self *clientManager) isSecureModeEnabled(req *restful.Request) bool {
	if self.isLoginEnabled(req) && !args.Holder.GetEnableSkipLogin() {
		return true
	}

	authInfo, _ := self.extractAuthInfo(req)
	return self.isLoginEnabled(req) && args.Holder.GetEnableSkipLogin() && authInfo != nil
}
```

It might be slightly easier to follow. @liggitt WDYT?

maciaszczykm:
Tested it locally on prod and it works as expected 👍

src/app/backend/client/manager.go
liggitt: that is simpler, but if I've configured the server to use TLS and not skip login, the incoming request shouldn't have any ability to make `isSecureModeEnabled` return `false`. Can we make the initial check only look at how the server was configured?

floreks: > Can we make the initial check only look at how the server was configured, not look at req?

It would make this method even more complicated as we have several arguments responsible for TLS mode. The easiest way is to check if `req.Request.TLS != nil`. If it is then we are sure that the request itself is secure. 

There is no way to alter that from the outside. Only by corrupting `restful-go` used by dashboard someone would be able to somehow influence that.

liggitt: Letting a characteristic of an incoming request control bypassing security seems like a bad idea when the intent to serve securely can be determined from the config options alone. I'd like to see that adjusted in the future.

liggitt:
/lgtm

This improves the current state, though the comments at https://github.com/kubernetes/dashboard/pull/3400/files#discussion_r241135916 are still relevant

PR #3414, opened 2018-12-13
Prepare for 1.10.1 release

maciaszczykm:
Done according to https://github.com/kubernetes/dashboard/wiki/Release-procedures.

Please wait for #3400.

@jeefy What is the current status of metrics container?

@kubernetes/dashboard-maintainers Most probably this is the last release before migration. What do you think about getting `ng-migration` branch merged to master then?

### DIFF
--- a/build/conf.js
+++ b/build/conf.js
@@ -59,7 +59,7 @@ const version = {
   /**
    * Current release version of the project.
    */
-  release: 'v1.10.0',
+  release: 'v1.10.1',
   /**
    * Version name of the head release of the project.
    */
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "//-1-//": "Specification of this file can be found at: https://docs.npmjs.com/files/package.json",
   "name": "kubernetes-dashboard",
-  "version": "1.10.0",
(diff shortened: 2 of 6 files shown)


jeefy:
@maciaszczykm Got it working locally. Waiting for a sanity check from @DirectXMan12, but it should not hold this up. 

For now, I suggest we hold off on Metrics server support till the Angular release. That way v1.10.1 can drop sooner.

maciaszczykm:
Tested using HEAD release. Everything seems to work. We can release 🙂 

@bryk @rf232 Can you take over?

floreks:
> Tested using HEAD release. Everything seems to work. We can release :slightly_smiling_face: 

I have also tested and can confirm.

bryk:
Can you perhaps just update only the two first files (build/conf.js and package.json), push this and create a tag? This way travis will automatically build correct image version that we could just push to GCR. 
Then in a next PR you could update yaml files

Would this work?

maciaszczykm:
@bryk Done: https://github.com/kubernetes/dashboard/releases/tag/v1.10.1 🙂

PR #3422, opened 2018-12-14
opt/dataselect filter no necessary continue loop.

epiclen:
if false，no necessary continue loop.


src/app/backend/resource/dataselect/dataselect.go
```diff
@@ -116,7 +116,7 @@ func (self *DataSelector) Filter() *DataSelector {
 			}
 			if !v.Contains(filterBy.Value) {
```
maciaszczykm: You can also mer

PR #3433, opened 2018-12-18
`servicesAndDiscovery` -> `Discovery`

chenrui333:
Relates to #1787 menu update. 

I have encountered some understanding when I help with kubernetes/website documentation i18n process. I think this renaming is necessary and helpful.


chenrui333:
That is just a comment update, not sure why cov-report goes down. 

Also we might need to update some other label-related information.

PR #3434, opened 2018-12-18
Fix FilterPodsForJob in backend

JackWindows:
The current implementation `FilterPodsForJob` in backend assumes the job controller uses automatically generated selector, which doesn't work if `manualSelector: true`.
This commit implements the `MatchLabels` expression in job controller `Selector` and fixes the issue.

### DIFF
--- a/src/app/backend/resource/common/pod.go
+++ b/src/app/backend/resource/common/pod.go
@@ -46,12 +46,21 @@ func FilterPodsByControllerRef(owner metav1.Object, allPods []v1.Pod) []v1.Pod {
 	return matchingPods
 }
 
+// FilterPodsForJob returns a list of pods that matches to a job controller's selector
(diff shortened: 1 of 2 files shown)


jeefy:
This change breaks the current tests. See https://travis-ci.org/kubernetes/dashboard/jobs/469791242#L369 for more info. :)

I also wonder if there's a way to get around putting another for loop into the mix.

Lastly, @JackWindows please sign the CLA. :) Thanks!

JackWindows:
@jeefy 
I updated the commit to fix the test. Turns out to be current test is wrongly written.

For the match logic, I don't think we can simulate the label selector behavior without using the loop. Alternative way is to offload the pods-to-job matching to kubernetes API server, but that requires making an API call for each job, which is very inefficient.

I have signed the CLA and it only got updated by the bot after your comment.

Issue #3066, opened 2018-05-30
HPA list has broken pagination/search


danopia:
##### Environment
I run this container with no alterations: `gcr.io/google_containers/kubernetes-dashboard-amd64:v1.8.3`

```
Dashboard version: v1.8.3
Kubernetes version: v1.9.3 (via kops)
Operating system: N/A
Node.js version: N/A
Go version: N/A
```

I started on dashboard v1.6.0 and have been bumping the version since then, in parallel with cluster upgrades.

##### Steps to reproduce
1. Make more than 10 HPAs
2. Go to /#!/horizontalpodautoscaler
3. Use the pagination arrow to go to the next page

Alternatively, try using the "Search" feature and type in garbage.

##### Observed result
The pagination makes a request and shows a page, but it's still the same 10 HPAs in the list.

Alternatively, the "Search" doesn't actually filter, and there's still the same 10 HPAs in the list.

##### Expected result
The pagination shows the _next_ 10 HPAs in the list.

Alternatively, the "Search" feature actually filters the list down as requested. Searching for garbage should show nothing.

##### Comments
This issue has been present for as long as HPAs have been listable, IIRC. 

I've looked at the network requests and it seems like the backend always returns the full list of HPAs. I see 30+ in the response, and it doesn't change. I compared to the deployment API, and that one returns up to 10, and paginates/searches as intended. So it seems the HPA list backend isn't obeying the page and search parameters.

maciaszczykm:
Closed by #3436.

/close

PR #3436, opened 2018-12-19
Fix HPA pagination

CDillinger:
Fixes issue #3066


jeefy:
Thanks for addressing #3066 

/lgtm

PR #3439, opened 2018-12-20
add rbac resources api

izern:
fix GetRbacRoleBindingList bug
add rbac resources api


izern:
> CLA签名完成

/check-cla

PR #3451, opened 2018-12-20
temporarily fixing issue with deploy directory

jimangel:
After playing with git symlinks (both of directories and of files) it appears none of those options would get us in the clear faster than simply copying the yaml file over.

This is only a patch and we should discuss how to handle this moving forward.


jimangel:
From @floreks on slack:
> We have promoted this way of installing Dashboard and it was our mistake which we have to finally resolve. The only way we can do that is to force people to use installation links that are not referring master branch in any way. It is not about the yaml files. It is about using links that rely directly on the unstable master branch instead of stable tagged branches. Master branch can always be changed in a way that might introduce breaking changes. That is why it is never a good idea to rely on it. We have learned our lesson and updated our documentation. All public documents should be forced to be updated also.

/close

PR #3462, opened 2018-12-23
fix(terminal): syncmutex are used as value

xujihui1985:
this is a potential and obvious problem to use value type when use sync.mutex


xujihui1985:
I signed it

xujihui1985:
I signed it

PR #3469, opened 2018-12-27
 Rework resource Status for Job List Pie Chart

JackWindows:
1. Use individual job status to derive job list status
2. Modify unit tests accordingly


floreks:
The problem here is that we need to take events into consideration upon deciding on job status. For example, there might be a warning saying that image cannot be pulled and Job will "officially" be in a pending state, where it should show failed. The only reason that it will be shown as pending is that pod will be restarted indefinitely until image pull issue will be resolved.

JackWindows:
> The problem here is that we need to take events into consideration upon deciding on job status. For example, there might be a warning saying that image cannot be pulled and Job will "officially" be in a pending state, where it should show failed. The only reason that it will be shown as pending is that pod will be restarted indefinitely until image pull issue will be resolved.

I think the issue with your example is that k8s doesn't have an `ImagePullRetryLimit` argument. Ideally the pod should retry pulling the image until reaching certain threshold (I believe it is 50 in GCP k8s now), and then fail. After enough pods fail the job should trigger `BackoffLimit` and gets transitioned to `Failed` status.
The current implementation could also be misleading, e.g. when pods are spawned by job controller, it can take cluster autoscaler a couple minutes before new nodes are online to accommodate pods, during which the job will be attributed as failed job in the dashboard.

JackWindows:
> This is not entirely true. If there won't be any events marked as errors associated with job pods then pod and job will be correctly seen as pending. I'd keep events condition and make this an `or` statement. If there are any warning events or job is in failed status then mark it as failed.

Before the new nodes are spawned and join the k8s cluster, the pods will be marked with Unscheduleable by the default scheduler, so it is counted towards "Failed" job with current implementation.

I am still not convinced that the job status should be determined by pod event. Another example would be some pods in the job failed due to transitional issue, the job controller re-spawn some pods and the job runs to desired completion count. In that case the job should also be considered "Succeeded", however, with current implementation it will be considered as "Failed" due to the warning pod event.

Also, event is not persistent k8s resource, they get purged after an hour by default. A job resource is persistent unless manually deleted or configured with TTL. I don't think a persistent object's status should be determined by transitional resources.

JackWindows:
@floreks Here is an example screenshot of what happens before cluster autoscaler brings up a new node to host the job pod. The job is considered "Failed" which is undesired. The pod is also considered "Failed" which is not True.
image

The same applies to `ImagePullBackoff`. When such events happen, it means the pod has trouble pulling image at the moment, but that doesn't necessarily mean the pod/job has failed. It could be temporary outage from the docker registry, or maintenance is being conducted for the docker daemon on node, etc. The pod should only be considered "Failed" when some kind of threshold has been exceeded, which is what kubernetes currently lacks of. The decision of whether a job has failed due to too many image pull failures shouldn't fall on to the dashboard, but rather should be handled by kubernetes itself.

floreks:
> The same applies to ImagePullBackoff. When such events happen, it means the pod has trouble pulling image at the moment, but that doesn't necessarily mean the pod/job has failed. It could be temporary outage from the docker registry, or maintenance is being conducted for the docker daemon on node, etc. The pod should only be considered "Failed" when some kind of threshold has been exceeded, which is what kubernetes currently lacks of. The decision of whether a job has failed due to too many image pull failures shouldn't fall on to the dashboard, but rather should be handled by kubernetes itself.

Basically, it confirms that we should take events into consideration. If some resource has issues at any given moment we should show errors in real-time, not silence them and hope that they will be fixed in 5/10/15 min. Once everything is ok and image has been pulled or nodes have been brought up and are ready to start new pods errors will be gone.

JackWindows:
> Basically, it confirms that we should take events into consideration. If some resource has issues at any given moment we should show errors in real-time, not silence them and hope that they will be fixed in 5/10/15 min. Once everything is ok and image has been pulled or nodes have been brought up and are ready to start new pods errors will be gone.

I think you may have mixed the status "running but with warning/transit error" with "failed".
"failed" status should indicate the job's life cycle has ended and the job failed. For example, a job should be marked as "failed" if the pod fails to pull docker image for `x` (e.g. 50) times.
While "running but with warning/transit error" status means the job is still running, just with some temporary issues, the job controller is still going to retry.

floreks:
> I think you may have mixed the status "running but with warning/transit error" with "failed".

Running means that the pod is actually running and the container of the pod has to also be running. We can not show `Running` or `Pending` if it is stuck in a loop or crashing. We can never be sure that any error such as `backOffLimit` or `insufficient cpu/gpu` is just a temporary one. We always have to assume the worst case scenario, that image just might got deleted from repository and pod might be restarted indefinitely or that user has exceeded his node quotas and there are no more nodes coming any time soon.

We might make related events less restrictive or introduce new status (Running/Completed with warnings), but in any case we can not just get rid of events. They have to be taken into consideration somehow.

JackWindows:
> Running means that the pod is actually running and the container of the pod has to also be running. 

I want to clarify that I'm modifying how job status is calculated, not pod status. It seems you have mixed them up?
`Failed` and `Succeeded` are considered to be terminal status of a job. A `Failed` job should stay `Failed`(i.e. not become `Running`) without manual intervention (e.g. restart).
Let me describe a few scenarios:
1. A newly submitted job creates a few pods, however, currently there is no room in the cluster to accommodate these pods, and autoscaler is in the process of scaling up the cluster.
_What status should the job be? Under current implementation the job is considered to be `Failed` because there is warning message (cannot schedule) on its pods. Personally I think the job should be in `Pending` status._
2. A job has finished and failed, but its pods have been purged due to resource being reclaimed.
_What status should the job be? Under current implementation the job is considered to be `Succeeded` because there is no pod with warning message associated with it. Personally I think the job should be in `Failed` status._

Anyway this is my last effect trying to reason with this issue. If the kubernetes UI team still thinks the current implementation is "Work As Intended", then I guess I will be stuck on my fork until more statuses are introduced and this misbehavior has been fixed.

floreks:
> A newly submitted job creates a few pods, however, currently there is no room in the cluster to accommodate these pods, and autoscaler is in the process of scaling up the cluster.
What status should the job be? Under current implementation the job is considered to be Failed because there is warning message (cannot schedule) on its pods. Personally I think the job should be in Pending status.

If you could adapt the frontend also to show `Pending` but also show the warnings messages it would be a good solution for me.

> A job has finished and failed, but its pods have been purged due to resource being reclaimed.
What status should the job be? Under current implementation the job is considered to be Succeeded because there is no pod with warning message associated with it. Personally I think the job should be in Failed status.

In this case I agree. If job has failed and did not finish its job then it should show `Failed`.

maciaszczykm:
I have a question that may be out of topic, but do you know how does `kubectl` handle it? `get` and `describe`?

JackWindows:
> I have a question that may be out of topic, but do you know how does `kubectl` handle it? `get` and `describe`?

`kubectl get` only shows `NAME`, `DESIRED`, `SUCCESSFUL`, `AGE` of a job.
`kubectl describe` doesn't show inferred job running/succeeded/failed status, it has `Pods Statuses: 0 Running / 1 Succeeded / 0 Failed`.
I believe there is no reason to involve pod events for inferring job status.
The job status should be determined by:
1. If `job.status.conditions` is `Failed`, job is failed.
2. Else if `job.status.conditions` is `Complete`, job is succeeded.
3. Else job is running.

The above is inline with how individual job status is determined in kubernetes-dashboard backend. https://github.com/kubernetes/dashboard/blame/master/src/app/backend/resource/job/list.go#L164
However, for some reason, if resource overview (the pie chart), job status is calculated differently and leads to inconsistently behavior (which is what I'm trying to fix in this PR).

JackWindows:
> If you could adapt the frontend also to show `Pending` but also show the warnings messages it would be a good solution for me.

I think you are talking about something else. My PR does not affect the individual job entry in the dashboard. My PR only tries to fix the job overview (the pie chart). So I'm not sure what you mean by "but also show the warnings messages", as my PR doesn't touch [that part](https://github.com/kubernetes/dashboard/blame/master/src/app/backend/resource/job/list.go#L150). The warning events showing at each job entry (in the job list, not pie chart) are kept intact in this PR.

Also, the job overview pie chart already has `Pending` status, and if you look closely to my PR, the logic is that "if a job is neither `Complete` nor `Failed`, and does not have running pod, it's in pending state".

I both scenarios described, there is no need for pod events to determine job status (in the pie chart overview).

maciaszczykm:
I have no strong opinion on this. @floreks WDYT about last comments from @JackWindows?

floreks:
> My PR does not affect the individual job entry in the dashboard. My PR only tries to fix the job overview (the pie chart). So I'm not sure what you mean by "but also show the warnings messages", as my PR doesn't touch [that part](https://github.com/kubernetes/dashboard/blame/master/src/app/backend/resource/job/list.go#L150) and it's not possible to show warning messages in the pie chart. The warning events showing at each job entry (in the job list, not pie chart) are kept intact in this PR.

It is true that you are not directly touching the job entries on the list view, but they have to also reflect state of the piechart. I think (haven't checked for sure) that after this change some Jobs might be showed as valid/pending on the chart whereas on the list they will be actually marked as failed with some warnings displayed. Those 2 parts have to be in sync.

JackWindows:
> Those 2 parts have to be in sync.

I agree with this, but the fact is these two don't match at the moment, and the discrepancy is huge.
My PR brings the gap closer but not necessarily eliminate it.
The issue is, the job list only has 3 states (Succeeded/Failed/Running), while the pie chart has 4 states (with an additional Pending state). This leads to the discrepancy between these two.
My PR only tries to address the issue in pie chart at this time. There can be a subsequent PR to fix the issue with entries in job list, and you are welcome to help working on it.

PR #3490, opened 2019-01-09
Update component.spec.ts

mohdshakir:
Bump up copyright year


floreks:
@maciaszczykm agree with that. This does not change anything.

PR #3503, opened 2019-01-10
Fix breadcrumbs font weight

### LINKED ISSUE #3498: Breadcrumbs font is too heavy
zrzut ekranu 2019-01-10 o 15 00 22

It seems like it is the impact of moving to new Angular Material.

maciaszczykm:
Fixes #3498.

/cc @floreks


maciaszczykm:
Most probably the issue exists only on macOS. I think we can ignore it since it may be some font issue.

/close

PR #3505, opened 2019-01-10
Disable job scaling as it is not supported by scale client

### LINKED ISSUE #3500: Disable job scaling
We are using our custom code in the backend to provide scaling functionality for Job resources. `kubectl` decided against it an I think we should follow the same path. Let's use scaling client for resources that can be scaled by it and ignore other resources. It might get confusing for some people that by scaling we change parallelism of the Job. WDYT?

maciaszczykm:
Fixes #3500.


floreks:
lgtm

PR #3513, opened 2019-01-11
Update README.md

julioxavierr:


floreks:
This is not needed. Angular configures default path.

Issue #3440, opened 2018-12-20
Complete Angular migration


maciaszczykm:
Things left to do:

- [x] merge `ng-migration` to `master` branch
- [ ] ~~deployment list and details (85% done; missing: new replica set list, hpa list in details @daminisatya )~~
- [x] cron job list and details
- [x] align actionbar actions with context actions in list menu for every resource
- [ ] ~~graphs for cluster, nodes, overview, workloads, daemon sets, deployments, pods, rc, rs, and stateful sets (@eloyekunle)~~
- [ ] ~~error handling (@floreks)~~
- [x] remove skip button
- [x] update `README.md`
- [ ] ~~update wikis~~
- [x] workload statuses for workload and overview page (@ajatprabha)
- [ ] ~~tests~~
- [ ] ~~check `frontend_old` for any leftovers~~
- [ ] ~~solving all TODOs from the code~~
- [ ] ~~manual tests and final code review~~
- [ ] ~~release and annoucements~~

Related to #3152.

junsionzhang:
is Angular migration done now and what is the plan ?

floreks:
As you can see, there are still some pending tasks left to be implemented/fixed. Now, that the migration has been merged to master it will be easier to maintain and move forward with it.

maciaszczykm:
@jeefy Do you have any updates about metrics container?

jeefy:
@maciaszczykm It works! I'm still waiting on @DirectXMan12 to give it another look over (https://github.com/jeefy/metrics-sidecar) but after that I can PR in support for it (See https://github.com/kubernetes/dashboard/compare/master...jeefy:sidecar?expand=1) 

I still have to pull in the angular migration changes and apply them to that branch though.

ajatprabha:
I'd like to work on `align actionbar actions with context actions in list menu for every resource` task.
cc: @maciaszczykm

maciaszczykm:
@daminisatya You can take a look at one of following:

>- [ ] deployment list and details (85% done; missing: hpa list in details)
>  - [ ] cron job list and details (80% done; missing: inactive job list, trigger option in list, pause action)

We can discuss details here or at #sig-ui Slack.

drazul:
Why you removed skip button on loging screen?
Are there a new way to avoid authentication?

All my team work on our local cluster before go to real cluster and we don't want to need to set up a new credential each time we need to use the dashboard.

maciaszczykm:
[...]

PR #3514, opened 2019-01-14
align actionbar actions with context actions

ajatprabha:
This PR is to align action-bar actions with context actions in list menu for every resource. WIP towards #3440

### DIFF
--- a/src/app/frontend/common/components/actionbar/detailactions/component.ts
+++ b/src/app/frontend/common/components/actionbar/detailactions/component.ts
@@ -14,6 +14,7 @@
 
 import {Component, Input} from '@angular/core';
 import {ObjectMeta, TypeMeta} from '@api/backendapi';
+import {ResourceMeta} from '../../../services/global/actionbar';
 
 @Component({
   selector: 'kd-actionbar-detail-actions',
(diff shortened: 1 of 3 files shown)


maciaszczykm:
We have a misunderstanding here :) I was not so clear in the issue, so let's rephrase:

To align items from context-menu (list view) and action bars (details view) you need to make sure that in both views we have same actions for every resource.

For example for Pods I'd like to use logs, exec, edit and delete from list view and details view.

It is not about CSS alignment, but if it works you can leave it and we will review. The real issue is a little bit more time consuming :)

ajatprabha:
Okay, got it. I'll see if I can make this work and update the PR.

ajatprabha:
Both menus should have same actions in detail and list views, right?

maciaszczykm:
Thank you for your patience, I am reviewing it 🙂

maciaszczykm:
/lgtm
/approve

Great piece of work. What I like is that it is clean and code is in one place. Thanks a lot! We can iterate on it to make these actions more visible, but it is already good.

ajatprabha:
Thank you @maciazsczykm 🙂

ajatprabha:
Thank you @maciaszczykm 🙂

PR #3518, opened 2019-01-15
removed skip button

phoenixking25:
Removed skip button from login component. Related to #3440


floreks:
The description there is not clear. It's not about removing it completely. It has to be hidden based on certain conditions. The logic has to be the same as for `v1.10.1` [here](https://github.com/kubernetes/dashboard/tree/v1.10.1/src/app/frontend/login).

phoenixking25:
Sorry its my misunderstanding

PR #3522, opened 2019-01-17
Fix in-cluster config

### LINKED ISSUE #3008: Handle Kubernetes API server failover
##### Environment

```
Dashboard version: v1.8.3
Kubernetes version: v1.10.2
Operating system: CentOS Linux release 7.4.1708 (Core)
```

##### Steps to reproduce
1. Have a multi-master Kubernetes cluster
2. Run dashboard with in-cluster config
3. Stop one of the API servers

##### Observed result
* Dashboard hangs while trying to load cluster resources, until Linux eventually timeouts the TCP connection. 
* The dashboard pod is not killed and restarted automatically either, because the liveness probe does not exercise the Kubernetes API connection
* Even after the TCP timeout and eventual reconnect, the log still repeatedly shows `Synchronizer kubernetes-dashboard-key-holder-kube-system exited with error: kubernetes-dashboard-key-holder-kube-system watch ended with timeout` (see https://github.com/kubernetes/dashboard/issues/2723#issuecomment-382378523) 

##### Expected result
* Dashboard has a shorter timeout on the API server connection and reconnects (might depend on https://github.com/kubernetes/client-go/issues/374)
* The dashboard has a liveness probe that actually tests core functionality of the dashboard (`/api/v1/settings/global` might do, but ideally a designated health URI)

floreks:
Fixes current head image to work with in-cluster config.

Also fixes:
 - Closes #3008 - I have added max count of synchronizer restarts before Dashboard pod will be restarted. After restart it should correctly reconnect to the API server.

TODO:
 - [x] Add some tests


src/app/backend/client/api/common.go
```diff
@@ -29,3 +33,14 @@ func ToSelfSubjectAccessReview(namespace, name, resource, verb string) *v1.SelfS
 		},
 	}
 }
+
+// GenerateCSRFKey generates random csrf key
+func GenerateCSRFKey() string {
+	bytes := make([]byte, 256)
+	_, err := rand.Read(bytes)
+	if err != nil {
+		panic("Fatal error. Could not generate csrf key")
```
maciaszczykm: It is common to start error messages with a small letter in Go, so perhaps `panic("could not generate csrf key")` would be better?

src/app/backend/sync/overwatch.go
```diff
@@ -81,6 +87,10 @@ func (self *overwatch) monitorRestartEvents() {
 	go wait.Forever(func() {
 		select {
 		case name := <-self.restartSignal:
+			if self.restartCount[name] > MaxRestartCount {
+				panic(fmt.Sprintf("Synchronizer %s restart limit execeeded. Restarting pod.", name))
```
maciaszczykm: As above.

PR #3536, opened 2019-01-24
Allow deep-linking of specific container logs

### LINKED ISSUE #2760: Deep-linking to logs of a specific container within a pod has become broken
##### Environment
```
Dashboard version: 1.8.0 (git commit 054916b8fbadebd32dc16d32108fdd3c9ccbf400)
Kubernetes version: 1.8.0 (via minikube 0.24.1)
Operating system: Windows 10 x64 host with VirtualBox 5.1.30 and Minikube guest OS on Buildroot 2017.02 with kernel Linux minikube 4.9.13
```

##### Steps to reproduce

1. Open the logs page for a Pod with two or more containers.
2. Note the URL in the browser address bar.
3. Select a different container.
4. Note the URL in the browser address bar doesn't change.
5. Open the URL in a new tab and note the new tab shows the logs for a different container in the pod.

##### Expected result
The container selected on the Log page should be represented in the URL so that bookmarking or sharing the URL will still refer to the correct container. In a Dashboard version prior to 1.8.0 (sorry not sure which) the container was part of the URL.

Previously the URL took the form:
`/#!/log/<namespace>/<pod>/<container>?namespace=<namespace>`

But now in Dashboard 1.8.0 the URL to the log page has replaced the container name with the constant literal "pod":
`/#!/log/<namespace>/<pod>/pod?namespace=<namespace>`

maciaszczykm:
Example URL: http://localhost:8080/#/log/default/hello-1548341040-n8gdp/pod?namespace=default&container=hello

Closes #2760.

### DIFF
--- a/src/app/frontend/logs/component.ts
+++ b/src/app/frontend/logs/component.ts
@@ -14,7 +14,7 @@
 
 import {HttpParams} from '@angular/common/http';
 import {Component, OnDestroy} from '@angular/core';
-import {MatDialog} from '@angular/material';
+import {MatDialog, MatOptionSelectionChange, MatSelectChange} from '@angular/material';
 import {LogDetails, LogLine, LogSelection, LogSources} from '@api/backendapi';
 import {StateService} from '@uirouter/core';
 import {GlobalSettingsService} from 'common/services/global/globalsettings';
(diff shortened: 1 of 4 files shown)


src/app/frontend/logs/component.ts
```diff
@@ -124,6 +127,15 @@ export class LogsComponent implements OnDestroy {
     return showTimestamp ? `${line.timestamp}  ${line.content}` : line.content;
   }
 
+  appendContainerParam() {
+    this.state_.go('.', {container: this.container}, {notify: false});
```
maciaszczykm: Appends params w/o reload.

PR #3540, opened 2019-01-28
Change e2e tests to use kind

### LINKED ISSUE #3539: Use KinD instead of Minikube for e2e tests
We should move our e2e tests to use kind (https://github.com/kubernetes-sigs/kind) 

##### Why is this needed
Lighter weight testing, removes the need for minikube, and it's the general direction testing is going.

##### Comments
This is a placeholder issue, I'm gonna look to tackle this in the next couple weeks.

/assign

jeefy:
Fixes #3539 

Several things are moving to use [Kind](https://github.com/kubernetes-sigs/kind) in place of Minikube. This is luckily a pretty easy swap for us as our e2e test is the only thing reliant on Minikube. 

Performance comparisons:
Minikube: https://travis-ci.org/kubernetes/dashboard/builds/484938232
Kind: https://travis-ci.com/jeefy/dashboard/builds/98754147

While this might seem unnecessary now, it will definitely benefit us later.

Comment, discuss, and iterate. :)


aio/scripts/start-cluster.sh
```diff
@@ -15,38 +15,32 @@
 
 # Import config.
 ROOT_DIR="$(cd $(dirname "${BASH_SOURCE}")/../.. && pwd -P)"
+echo "We're in ${ROOT_DIR}"
```
maciaszczykm: Should be removed after teting.

aio/scripts/start-cluster.sh
```diff
@@ -67,24 +61,17 @@ function start-ci-heapster {
...
 
-function start-local-minikube {
-  say "\nStarting local cluster"
-  sudo -E ${MINIKUBE_BIN} start --kubernetes-version=${MINIKUBE_K8S_VERSION}
-}
-
-function start-minikube {
+function start-kind {
+  ${KIND_BIN} create cluster --name="k8s-cluster-ci"
+  sleep 5
```
maciaszczykm: Why `sleep` is required?

aio/scripts/start-cluster.sh
jeefy: Thanks removed.

aio/scripts/start-cluster.sh
jeefy: It's not, I was testing something else and accidentally left it in. Thanks for the find! Removed.

package.json
```diff
@@ -30,8 +30,9 @@
     "test:backend": "go test github.com/kubernetes/dashboard/src/app/backend/...",
     "test:backend:coverage": "./aio/scripts/coverage-backend.sh",
     "test:coverage": "npm run test:frontend:coverage && npm run test:backend:coverage",
-    "test:e2e": "ng e2e",
-    "cluster": "./aio/scripts/start-cluster.sh",
+    "test:e2e": "npm run cluster:start && ng e2e && npm run cluster:stop",
```
maciaszczykm: Not sure if we should start the test cluster by default. What if I would like to test against some other cluster? @floreks WDYT?

floreks: Yes, this should stay as it was. Usually, we already have a cluster started while developing.

.travis.yml
```diff
@@ -33,13 +33,7 @@ jobs:
...
         - rm -rf $TRAVIS_BUILD_DIR/.tmp
         - bash <(curl -s https://codecov.io/bash)
-    - addons:
-        sauce_connect: true
-      script:
-        - npm run cluster
-        - npm run test:e2e
-      env:
-        - KUBE_DASHBOARD_KUBECONFIG=$HOME/.kube/config
+    - script: npm run test:e2e
```
floreks: Why dropping `sauce_connect`?

package.json
jeefy: Sounds good!

.travis.yml
jeefy: We aren't using sauce_labs for our tests anymore.

If we plan to use them again, I'll revert, but they were also the cause of a lot of flakes so I'd opt against. 

Let me know your thoughts!

floreks: This is quite weird because in the protractor conf we do have config for sauce labs.

https://github.com/kubernetes/dashboard/blob/1bbe49858d461d8f11eb4fbed126c0cf6480b5f0/aio/protractor.conf.js#L38-L63

Those variables are also defined in the travis settings.

.travis.yml
```diff
@@ -33,13 +33,7 @@ jobs:
...
         - rm -rf $TRAVIS_BUILD_DIR/.tmp
         - bash <(curl -s https://codecov.io/bash)
-    - addons:
-        sauce_connect: true
-      script:
-        - npm run cluster
-        - npm run test:e2e
-      env:
-        - KUBE_DASHBOARD_KUBECONFIG=$HOME/.kube/config
+    - script: npm run cluster:start && npm run test:e2e && npm run cluster:stop
```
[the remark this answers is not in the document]
jeefy: Hmm. Probably not. I'll change that.

.travis.yml
jeefy: I had checked several job logs for any sign of Sauce Labs running and found none. It looks like no actual tests are running. 

Example: https://api.travis-ci.org/v3/job/486470328/log.txt (this is off of #3542) 

```
[15:24:50] I/update - chromedriver: chromedriver_2.45 up to date
[15:24:51] I/launcher - Running 1 instances of WebDriver
[15:24:51] I/local - Starting selenium standalone server...
[15:24:53] I/local - Selenium standalone server started at http://10.20.4.242:47825/wd/hub
Jasmine started

Executed 0 of 0 specs[32m SUCCESS[39m in 0.005 sec.
[15:24:58] I/local - Shutting down selenium standalone server.
[15:24:58] I/launcher - 0 instance(s) of WebDriver still running
[15:24:58] I/launcher - chrome #01 passed
```

If we're still intending to use Sauce Labs I'll revert that change and keep moving forward. :)

maciaszczykm:
/lgtm

Leaving it to @floreks to approve.

maciaszczykm:
Too late... 😅

Issue #3501, opened 2019-01-10
Stop using deprecated-dynamic package


maciaszczykm:
We should stop using `deprecated-dynamic` package ASAP:
https://github.com/kubernetes/dashboard/blob/b30e01f8a7fe85ae8f7cfad48fb72ecff4ce2a5e/src/app/backend/resource/deployment/deploy.go#L33

PR #3549, opened 2019-02-07
Use client-go/dynamic instead client-go/deprecated-dynamic

shu-mutou:
Fixes: kubernetes/dashboard#3501


floreks:
Cool!

@maciaszczykm should we also remove `deprecated-dynamic` package from the `Gopkg.lock` by applying `dep ensure`?

maciaszczykm:
Yep, we should, but we can do it in the following PR.

shu-mutou:
Removed deprecated-dynamic from Gopkg.lock and vendor. :wink:

PR #3552, opened 2019-02-07
Fix breadcrumbs and refactor context actions code

maciaszczykm:
- Fixed the issue with broken breadcrumbs when logs were opened for resource different than a Pod.
- Refactored logic conneted to our context menus, applied some minor fixes.
- Added components for logs, scale, exec, edit and delete context actions. It was the result of half a day of work on trying to make menu look a bit more similar to Gmail. Unfortunately it is not done yet, but components might be useful for future development. At the moment code would be too hacky or the actions would take too much space, that's why I opted to staying with the menu instead of mutiple action buttons.

### DIFF
--- a/src/app/frontend/common/components/list/column/menu/component.ts
+++ b/src/app/frontend/common/components/list/column/menu/component.ts
@@ -16,7 +16,7 @@ import {Component, Input, OnDestroy} from '@angular/core';
 import {ObjectMeta, TypeMeta} from '@api/backendapi';
 import {ActionColumn} from '@api/frontendapi';
 import {StateService} from '@uirouter/core';
-import {Subscription} from 'rxjs/Subscription';
+import {first} from 'rxjs/operators';
 
 import {logsState} from '../../../../../logs/state';
 import {LogsStateParams} from '../../../../params/params';
@@ -33,6 +33,8 @@ const scalableResources: string[] = [
   Resource.deployment, Resource.replicaSet, Resource.replicationController, Resource.statefulSet
 ];
 
+const executableResources: string[] = [Resource.pod];
+
 @Component({
   selector: 'kd-resource-context-menu',
   templateUrl: './template.html',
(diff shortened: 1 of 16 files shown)


src/app/frontend/common/components/list/column/menu/template.html
```diff
@@ -5,7 +5,7 @@
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
```
jeefy: Indent this please. No indentation causes the `check:license` test to fail.

src/app/frontend/common/components/list/column/menu/component.ts
```diff
@@ -57,16 +55,8 @@ export class MenuComponent implements ActionColumn, OnDestroy {
     this.typeMeta = typeMeta;
   }
 
-  ngOnDestroy(): void {
```
jeefy: This Angular component requires an `ngOnDestroy()` method. 

Running `npm start` on this branch causes the following error:
```
[1] ERROR in src/app/frontend/common/components/list/column/menu/component.ts(42,14): error TS2420: Class 'MenuComponent' incorrectly implements interface 'OnDestroy'.
[1]   Property 'ngOnDestroy' is missing in type 'MenuComponent'.
[1]
[1] ℹ ｢wdm｣: Failed to compile.
```

floreks: It's mainly because this class implements `OnDestroy` which requires this function. It should be removed from class definition.

maciaszczykm: Yep, didn't notice as it was changing it multiple times. Thanks for checking that.

maciaszczykm:
PTAL

Issue #1083, opened 2016-07-30
Support `tail`ing and `follow`ing logs


SeMeKh:
`kubectl logs`'s `--tail` and `--follow` options are pretty handy and missing in dashboard. Enabling a sane tail default (e.g. 1000 lines) could also relieve the issue with the browser becoming unresponsive due to huge amount of logs.

bryk:
This is a good feature request. Thanks @SeMeKh for reporting it. We'll take a deeper look at it in our next iteration.

steelbrain:
A friendly bump, this feature would be wonderful to have

bryk:
A general revamp of logs viewing experience is high on our priority list. So we cluster all logs-related issues like this into one big project. We'll let know when this is being taken. 

And yes, please ping the issue if you feel this is taken too slowly.

innovia:
Last comment 10 months ago is this is still in progress?

floreks:
Not until kubernetes API changes. It is very limited when it comes to logs. Right now we could maybe provide proxy link to apiserver endpoint and open raw logs page (pure text) with streaming enabled. We don't really want to do complex logs processing. At least not until other higher priority tasks are finished.

cheld:
--tail option is now implemented internally to avoid out of memory. It is not exposed to the user as option. I don't think it makes sense.

--follow is not implemented. I think a follow should be exposed to the user.

maciaszczykm:
It is already tracked from #1828.

PR #3561, opened 2019-02-11
Logs Auto Scroll

### LINKED ISSUE #2610: On logs refresh logs window should be scrolled to the bottom
Window stays at the same position and scrollbar moves up.

zrzut ekranu z 2017-11-22 14-27-23

### LINKED ISSUE #3509: Support following logs
Picked the last issue from #1828.

eloyekunle:
Update the Logs view to support autoscroll when Logs refresh.

Closes #2610

### DIFF
--- a/src/app/frontend/common/services/global/logs.ts
+++ b/src/app/frontend/common/services/global/logs.ts
@@ -22,7 +22,8 @@ export class LogService {
   inverted_ = true;
   compact_ = false;
   showTimestamp_ = false;
-  following_ = false;
+  following_ = true;
+  autoRefresh_ = false;
 
   constructor(private readonly http_: HttpClient) {}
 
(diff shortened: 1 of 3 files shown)


maciaszczykm:
It does not solve #3509. The original issue with more context is https://github.com/kubernetes/dashboard/issues/1083. I will take a look tomorrow.

maciaszczykm:
Works fine for page refresh, but not when switching pages. It is quote uncomfortable at the moment, but newest logs are at the end. Can you scroll also on page changes?

And remove #3509 from fixed issues please.

eloyekunle:
I've removed #3509, I'll have to thrash that on another PR.

The current behavior is it scrolls only when the user is already at the bottom of the logs viewer. This is to avoid a situation where you're trying to read something higher up in the page but the logs keep scrolling down during auto-refresh.

Currently, when you switch pages, you arrive at the top of the target page so it doesn't scroll because you're not at the bottom.

What do you think?

floreks:
I'd say that on page change to next one we should stay on top, but on page change to the previous one, we should go to the bottom to make it feel like you can continue reading without scrolling manually.

eloyekunle:
Okay, so:

- Navigation to previous pages should go to the bottom.
- Add new option to follow logs.

Sounds good.

jeefy:
Also it looks like you have a conflict in `src/app/frontend/logs/template.html` 

When you make your additional changes, please resolve that before pushing. :) 

Thanks!

eloyekunle:
Implemented:

- Option to follow logs. This adds a new icon.
- Scroll to logs bottom when navigating to previous page.
- Scroll to logs top when navigation to next page.

jeefy:
@eloyekunle Your latest changes broke some HTML formatting. 

See https://travis-ci.org/kubernetes/dashboard/jobs/496191334#L297

You can test these CI checks locally by running `npm run check` (and others, see `package.json` for the other options)

After that quick formatting fix I'll give it one more look and lgtm it. :) Thanks!

eloyekunle:
@jeefy Thanks, I've fixed the HTML formatting.

PR #3593, opened 2019-02-21
Support $KUBECONFIG

jeefy:
As we start using Kind more, we're reliant on the way they manage their kubecfg files: Through setting $KUBECONFIG`. Locally I keep having to change `package.json` or `.npmrc` to use it. We can do better.

This wraps `gulp serve` in a script that checks to see if `$KUBECONFIG` is set. If it is, it uses it. Otherwise, it defaults to the `kubernetes-dashboard:kubeconfig` values in `.npmrc`

I would have avoided _yet another wrapper script_ but I can't put the necessary logic in `.npmrc`, and it'd be super ugly in `package.json`

Feedback, if any, is appreciated. :)


floreks:
You could actually override it using builtin mechanism of npm, i.e. you can pass `--userconfig` flag to specify custom `.npmrc` file stored outside of the project. More information can be found here `https://docs.npmjs.com/files/npmrc`

jeefy:
I agree re: wrappers scripts 100%

But we should be able to support `$KUBECONFIG` out-of-the-box since it is a pretty standard Kubernetes convention. Adding another step for someone that uses `$KUBECONFIG` over the typical `~/.kube/config~` isn't a good experience and I believe should be addressed (one way or another)

I'm happy to let this hang (or close it and create an issue) as we ponder the more unified approach.

floreks:
> But we should be able to support `$KUBECONFIG` out-of-the-box since it is a pretty standard Kubernetes convention. Adding another step for someone that uses `$KUBECONFIG` over the typical `~/.kube/config~` isn't a good experience and I believe should be addressed (one way or another)

This is a bit misleading assumption as we can't really compare `out-of-the-box` experience to running a dev build of application for development/testing. It would be great to have more ways of overriding parameters passed to dev scripts, but it should be a unified solution, not a wrapper for current logic.

maciaszczykm:
+1 to what @floreks said.

maciaszczykm:
@jeefy @floreks So what's the plan with it? Do we really need it?

jeefy:
I was super dumb. There's a much simpler way to do this. PTAL. Tested it locally without any issue.

package.json
```diff
@@ -11,8 +11,8 @@
     "start:https": "concurrently \"npm run start:backend:https\" \"npm run start:frontend:https\"",
     "start:frontend": "npm run postversion && ng serve --aot --progress=false --proxy-config aio/proxy.conf.json --host=$npm_package_config_bind_address",
     "start:frontend:https": "node aio/scripts/version.js && ng serve --progress=false --aot --proxy-config aio/https-proxy.conf.json --ssl --host=$npm_package_config_bind_address",
-    "start:backend": "gulp serve --kubeconfig $npm_package_config_kubeconfig --sidecarServerHost $npm_package_config_sidecar_host",
-    "start:backend:https": "gulp serve --kubeconfig $npm_package_config_kubeconfig --autoGenerateCerts true --sidecarServerHost $npm_package_config_sidecar_host",
+    "start:backend": "KUBECONFIG=${KUBECONFIG:-$npm_package_config_kubeconfig}; echo $KUBECONFIG; gulp serve --kubeconfig $KUBECONFIG --sidecarServerHost $npm_package_config_sidecar_host",
+    "start:backend:https": "KUBECONFIG=${KUBECONFIG:-$npm_package_config_kubeconfig}; echo $KUBECONFIG; gulp serve --kubeconfig $KUBECONFIG --autoGenerateCerts true --sidecarServerHost $npm_package_config_sidecar_host",
```
maciaszczykm: Remove the echo.

jeefy: lol thanks

maciaszczykm:
Thanks :)

/lgtm
/approve

PR #3605, opened 2019-02-26
Bump golangci-lint to 1.15.0 and add go linter for pre-commit

shu-mutou:
Newly added auto-fix feature for goimports into golangci-lint 1.15.0.
So we can use this for fixing go format in our tasks, i.e. pre-commit, check:backend and fix:backend.

Also, added config file for golangci-lint as `.golangci.yml`.

Standalone goimports installation was removed in favor of this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "check:frontend:html": "./aio/scripts/format.sh --html --check",
     "check:license": "gulp check-license-headers",
     "fix": "concurrently \"npm run fix:backend\" \"npm run fix:frontend\" \"npm run fix:license\"",
-    "fix:backend": "goimports -w src/app/backend",
+    "fix:backend": "golangci-lint run -c .golangci.yml --fix src/app/backend/...",
     "fix:frontend": "concurrently \"npm run fix:frontend:ts\" \"npm run fix:frontend:scss\" \"npm run fix:frontend:html\"",
     "fix:frontend:ts": "gts fix",
     "fix:frontend:scss": "scssfmt -r 'src/**/*.scss'",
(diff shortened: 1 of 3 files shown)


shu-mutou:
/hold
need to confirm behavior in travis ci.

shu-mutou:
following should be error.
https://github.com/kubernetes/dashboard/commit/954b11462feb20154ae0441d62ae47583d43eaac#diff-e9d0bac66c36b1a4df65b4a6f8dedc1bR37

and got error as expected!!
https://travis-ci.org/kubernetes/dashboard/jobs/498641046#L288

shu-mutou:
/hold cancel
returned to correct commit.

package.json
```diff
@@ -41,15 +41,15 @@
     "check:frontend:html": "./aio/scripts/format.sh --html --check",
     "check:license": "gulp check-license-headers",
     "fix": "concurrently \"npm run fix:backend\" \"npm run fix:frontend\" \"npm run fix:license\"",
-    "fix:backend": "goimports -w src/app/backend",
+    "fix:backend": ".cached_tools/golangci-lint run -c .golangci.yml --fix src/app/backend/...",
```
maciaszczykm: This one looks wrong. Can be like this in the CI, but not there.

package.json
```diff
@@ -64,6 +64,10 @@
     "src/**/*.scss": [
       "scssfmt",
       "git add"
+    ],
+    "src/**/*.go": [
```
maciaszczykm: I like this one a lot :)

package.json
maciaszczykm: Can it be used like `goimports` was?

shu-mutou: > This one looks wrong. Can be like this in the CI, but not there.

It's my mistake. Thanks!

> Can it be used like `goimports` was?

Yes. We can format our codes using not only goimports, but also other linters included in golangci-lint. Let's add other linters in subsequent PRs. ;)

PR #3610, opened 2019-02-27
toggle selected mode

phoenixking25:
Toggle file format on switching json/yaml button in edit resource dialog.

### DIFF
--- a/src/app/frontend/common/dialogs/editresource/template.html
+++ b/src/app/frontend/common/dialogs/editresource/template.html
@@ -30,7 +30,7 @@ limitations under the License.
     <mat-icon>info</mat-icon>
     <div>
       <span>This action is equivalent to:</span>
-      <code>kubectl apply -f &lt;spec.json&gt;</code>
+      <code>kubectl apply -f &lt;spec.{{selectedMode_}}&gt;</code>
     </div>
   </div>
 </mat-dialog-content>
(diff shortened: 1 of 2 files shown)


src/app/frontend/common/dialogs/editresource/template.html
```diff
@@ -30,7 +30,7 @@ <h2 mat-dialog-title>Edit a resource</h2>
     <mat-icon>info</mat-icon>
     <div>
       <span>This action is equivalent to:</span>
-      <code>kubectl apply -f &lt;spec.json&gt;</code>
+      <code>kubectl apply -f &lt;spec.{{selectedMode_}}&gt;</code>
```
jeefy: When trying to run this locally I get:

```
[1] ERROR in src/app/frontend/common/dialogs/editresource/template.html(33,13): : Property 'selectedMode_' is private and only accessible within class 'EditResourceDialog'.
[1] ℹ ｢wdm｣: Failed to compile.
```

Which is true, `selectedMode_` is [declared as private](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/dialogs/editresource/dialog.ts#L33)

phoenixking25: so should i make another public variable for this?

maciaszczykm: You should change existing one to public and rename it to remove `_` from name.

phoenixking25: ok thanks

phoenixking25:
@maciaszczykm done

maciaszczykm:
@phoenixking25 Have you tested it locally?

phoenixking25:
yes

PR #3614, opened 2019-02-28
[WIP] Move lint backend script to package.json and remove linting staged Go files as it fails

maciaszczykm:
Moves lint backend script to package.json. Having dependency install in `postinstall` should be better because previously it was not possible to run `npm run fix` as the `golangci-lint` would be unavailable.


package.json
```diff
@@ -64,10 +64,6 @@
     "src/**/*.scss": [
       "scssfmt",
       "git add"
-    ],
-    "src/**/*.go": [
-      "golangci-lint run -c .golangci.yml --fix",
-      "git add"
```
shu-mutou: `check:backend` only reports error. Since `--fix` option fixes actually our go codes in pre-commit by husky, so I think we need these lines.

package.json
```diff
@@ -49,7 +49,7 @@
     "fix:license": "gulp update-license-headers",
     "clean": "rm -rf .go_workspace .tmp coverage dist npm-debug.log",
     "postversion": "node aio/scripts/version.js",
-    "postinstall": "node aio/scripts/version.js && ./node_modules/protractor/bin/webdriver-manager update --gecko=false"
+    "postinstall": "node aio/scripts/version.js && ./node_modules/protractor/bin/webdriver-manager update --gecko=false && go get -u github.com/golangci/golangci-lint/cmd/golangci-lint"
```
shu-mutou: `${GOPATH}/bin` may be not included in `${PATH}` in Travis CI.

package.json
maciaszczykm: Did you try to modify Go files and push the commit? It fails for me :/

shu-mutou: Yes, of course. I confirm its behavior in several commits on local and on travis at previous PR. If your problem means this [ https://travis-ci.org/kubernetes/dashboard/jobs/499890953#L320 ], could you check my comment [ https://github.com/kubernetes/dashboard/pull/3617#discussion_r262768339 ] ?

package.json
shu-mutou: Because https://github.com/kubernetes/dashboard/blob/43c9568a08e844a70518f5c958e6ca689a96f950/.travis.yml#L19 changes `${GOPATH}`, I guess.

jeefy:
I believe we're going to have to be redundant and duplicate `go get -u github.com/golangci/golangci-lint/cmd/golangci-lint` in the `before_install` section of `.travis-ci.yml` 

Travis runs `npm ci` instead of `npm i` if it detects `package-lock.json`. That won't trigger a `post-install` hook and `golangci-lint` won't get installed. This is likely why the current Travis test fails. :(

maciaszczykm:
Don't bother this PR at the moment. BTW could you verify if lint-staged for Go files works for you? Could modify a few of them and commit to some branch? Please make sure that entry is in package.json too.

maciaszczykm:
Can someone check it?

shu-mutou:
I found `postinstall` is running in Travis CI as follow [1].
[1]: https://travis-ci.org/kubernetes/dashboard/jobs/505700981#L246

Also, confirmed `${PATH}` [2] does not include `${GOPATH}/bin` [3].
[2]: https://travis-ci.org/kubernetes/dashboard/jobs/505700981#L256
[3]: https://travis-ci.org/kubernetes/dashboard/jobs/505700981#L257

So I propose adding `export PATH=${GOPATH}/bin:${PATH}` after GOPATH setting [4].
[4] https://github.com/kubernetes/dashboard/blob/43c9568a08e844a70518f5c958e6ca689a96f950/.travis.yml#L19

Also, I found when `npm ci` shows failure like "cannot run in wd", `--unsafe-perm` helped for running `postinstall` in my environment that means using `aio/develop/run-npm-on-container.sh` [5].
[5] https://github.com/kubernetes/dashboard/wiki/Getting-started#building-dashboard-inside-a-container

maciaszczykm:
I'll close this one and create an issue to observe if lint-staged for Go files work.

/close

@shu-mutou Feel free to pick up these changes to your PR.

PR #3615, opened 2019-02-28
Introduce angular router and add error handling

floreks:
This PR is far from finished yet :slightly_smiling_face: 

Current changes:
- Refactor error handling on the backend to a single package. All errors should be created using our internal package.
- Start replacement process of `ui-router` to `angular-router`. 

What's missing:
- [x] Proper query params handling. By default, it should at least have `namespace` query set to `default` namespace.
- [x] Proper path params handling for detail views. Only `Cluster Role` details view is working right now.
- [x] Remove usage of old `ui-router` classes i.e. `StateService`, `Transition`, etc. in all places...
- [x] ...and replace them with angular router equivalents.
- [x] Breadcrumbs.
- [x] Exec, logs, search etc.
- [x] Final review. Check every state.

Closes https://github.com/kubernetes/dashboard/issues/3869.

PR #3627, opened 2019-03-02
add conform dialog WIP

DanielUche:
@eloyekunle check this out https://ui-router.github.io/sample-app-angular/#/home


floreks:
We are actually in the middle of transition to angular router and we will drop the ui-router completely from the project.

DanielUche:
@floreks this makes sense @angular/router canActivate would work well for this

DanielUche:
I would also like to contribute to the transitioning

jeefy:
I don't think this PR is valid anymore and should be closed. The work migrating from ui-router is being done in another PR (#3615) already. @DanielUche If you're still looking to contribute to the migration effort, look in that PR and work with @floreks to coordinate efforts. :)

/close

PR #3637, opened 2019-03-08
Workloads status component for resource list components

ajatprabha:
This PR adds workloads status section charts to the resource list components.
Works towards #3440


jeefy:
Have you rebased to master recently? There are some files/changes in there I know have been removed as of the Angular migration in this PR. Totally get it's a WIP, just looking to help out. 😄 

Thanks!

ajatprabha:
Yes, I rebased with master before creating the PR. The changes are working fine locally. Can you point out the files that need attention, I'll make the necessary changes after I add the statuses for workloads.

src/app/backend/handler/apihandler.go
```diff
@@ -555,6 +556,16 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
 			To(apiHandler.handleLogFile).
 			Writes(logs.LogDetails{}))
 
+	apiV1Ws.Route(
+		apiV1Ws.GET("/overview/").
+			To(apiHandler.handleOverview).
+			Writes(overview.Overview{}))
+
+	apiV1Ws.Route(
```
maciaszczykm: It was removed. Why do you add this? You might solve merge conflicts wrongly, otherwise please explain.

src/app/backend/handler/apihandler.go
```diff
@@ -1035,6 +1046,25 @@ func (apiHandler *APIHandler) handleGetReplicationControllerList(request *restfu
 	response.WriteHeaderAndEntity(http.StatusOK, result)
 }
 
+func (apiHandler *APIHandler) handleOverview(request *restful.Request, response *restful.Response) {
```
maciaszczykm: It was removed. Why do you add this? You might solve merge conflicts wrongly, otherwise please explain.

src/app/backend/resource/cronjob/detail.go
```diff
@@ -22,12 +22,12 @@ import (
...
 // CronJobDetail contains Cron Job details.
 type CronJobDetail struct {
-	// Extends list item structure.
-	CronJob `json:",inline"`
-
 	ConcurrencyPolicy       string `json:"concurrencyPolicy"`
 	StartingDeadLineSeconds *int64 `json:"startingDeadlineSeconds"`
 
+	// Extends list item structure.
+	CronJob `json:",inline"`
```
maciaszczykm: Why do you move this?

src/app/backend/resource/overview/overview.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/config"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/discovery"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/workload"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Overview is a list of objects present in a given namespace.
+type Overview struct {
```
maciaszczykm: It was removed. Why do you add this? You might solve merge conflicts wrongly, otherwise please explain.

src/app/backend/resource/workload/workload.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/kubernetes/dashboard/src/app/backend/resource/job"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/pod"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/replicaset"
+	rc "github.com/kubernetes/dashboard/src/app/backend/resource/replicationcontroller"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/statefulset"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Workloads structure contains all resource lists grouped into the workloads category.
+type Workloads struct {
```
maciaszczykm: It was removed. Why do you add this? You might solve merge conflicts wrongly, otherwise please explain.

maciaszczykm:
Pointed a few places, please review all of your changes or cherry pick your commits into the master.

ajatprabha:
As discussed on Slack, I have added the status bar on each list component. This allowed using existing data in the component as suggested by @maciaszczykm. cc: @jeefy

jeefy:
I tested this and things look good.

@ajatprabha Is this still considered a WIP? Or is this PR ready for review?

ajatprabha:
I was thinking to also add `PodList.cumulativeMetrics` to this PR as well. I'll open a new PR for that and remove WIP from this one.

floreks:
IMHO adding a status bar on top of every list makes everything a bit overwhelming. Lists should be clean and contain only key information about the objects. Adding such bars might distract the user. I'd stick to the original charts at the top of the page to show statuses. List itself already contains status for every object. We should not need any more indicators to keep it as simple and clean as possible.

maciaszczykm:
I generally like the idea of having additional information next to the lists but I think we could use other design here. At the moment it doesn't match the styling of the rest of the component, is a bit too bold and colors are a bit overwhelming. We could use the help of someone who knows more than us about design, @danielromlein do you have some time for it?

I think we will still need a component to display statuses of all workloads in one place. This might be some feature enabled in the settings once we will style it. Both have their advantages, one is easier to reach, the second gives a better overview. 

Additionally like @floreks noticed we have a lot of things in the card components, we have to be careful here.

ajatprabha:
I had the same thought as @floreks while implementing this.

How do I pull data for a central component from backend then? Suggestions would be helpful. A new endpoint in the backend or maybe get it from state?

maciaszczykm:
Is calling all workload endpoint an option? Then we can send multiple calls asynchronously and display graphs when the responses come.

ajatprabha:
We'll replicating calls that way. Maybe we can use `ViewChild` to get data from list component or use `Output`?

Is there a Service which manages this data from backend? I think component calls the endpoint itself right now.
I know about the EndpointManager but that only provides endpoints right?

maciaszczykm:
I know, output was the thing that I was mentioning before on Slack. The problem with this is coupling components. With separate calls it is independent. @floreks WDYT?

floreks:
How about adding emitter to the list and emit data on list update. This way we could avoid hard coupling and other components that require similar data could easily subscribe to it. I don't remember exactly how backend looked like before. I'd have to check it.

maciaszczykm:
@floreks We had group resources, i.e. workloads, overview, config etc.

ajatprabha:
Closing in favor of #3667

Issue #3631, opened 2019-03-04
Show tooltip timestamp on time hover across pages


eloyekunle:
##### What would you like to be added
We should show tooltip timestamps on hover of absolute and relative time displayed across the components to give specific time information to the users.

For example, dates as shown below are not very helpful, because I might want to know exactly when the events occured.

![screenshot from 2019-03-04 18-18-49](https://user-images.githubusercontent.com/11166933/53750446-150c4980-3eaa-11e9-8c6b-87706616fced.png)

##### Why is this needed
To improve the user experience of users by giving them more information about dates and times across the dashboard.

##### Comments
In general, tooltips can look like below, with the dotted underline added as a visual cue for users to know they can hover:

![screenshot from 2019-03-04 18-15-43](https://user-images.githubusercontent.com/11166933/53750338-ca8acd00-3ea9-11e9-9c9e-dec79ba32794.png)
(credit: [this article](https://uxmovement.com/content/absolute-vs-relative-timestamps-when-to-use-which/))

To proceed, you might need to check places where times/dates are shown, e.g. events, conditions etc, and then investigate the use of [Material Tooltips](https://material.angular.io/components/tooltip/overview) for the tooltips.

/good-first-issue

anuragprafulla:
@eloyekunle I would like to give this a try. I have thought of a solution and am working on it.

anuragprafulla:
Is this issue still open?

PR #3643, opened 2019-03-09
added tooltips to event timestamps

DanielUche:
This PR adds 
1. Timestamp tooltips to the events table.

### DIFF
--- a/src/app/frontend/common/components/resourcelist/event/template.html
+++ b/src/app/frontend/common/components/resourcelist/event/template.html
@@ -71,8 +71,7 @@ limitations under the License.
       <ng-container [matColumnDef]="getDisplayColumns()[6]">
         <mat-header-cell *matHeaderCellDef>Last Seen</mat-header-cell>
         <mat-cell *matCellDef="let event"
-          matTooltip="{{event.lastSeen | date:'medium'}}"
-          matTooltipPosition="above">
+          matTooltip="{{event.lastSeen | date:'medium'}}">
           {{event.lastSeen | date}}
         </mat-cell>
       </ng-container>
--- a/src/app/frontend/common/components/resourcelist/event/template.html
+++ b/src/app/frontend/common/components/resourcelist/event/template.html
@@ -62,14 +62,20 @@ limitations under the License.
 
       <ng-container [matColumnDef]="getDisplayColumns()[5]">
         <mat-header-cell *matHeaderCellDef>First Seen</mat-header-cell>
-        <mat-cell *matCellDef="let event">
+        <mat-cell *matCellDef="let event"
+          matTooltip="{{event.firstSeen | date:'medium'}}"
+          [matTooltipPosition]="top"
+        >
           {{event.firstSeen | date}}
         </mat-cell>
       </ng-container>
 
       <ng-container [matColumnDef]="getDisplayColumns()[6]">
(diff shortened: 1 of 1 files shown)


src/app/frontend/common/components/resourcelist/event/template.html
```diff
@@ -62,14 +62,20 @@
 
       <ng-container [matColumnDef]="getDisplayColumns()[5]">
         <mat-header-cell *matHeaderCellDef>First Seen</mat-header-cell>
-        <mat-cell *matCellDef="let event">
+        <mat-cell *matCellDef="let event"
+          matTooltip="{{event.firstSeen | date:'medium'}}"
+          [matTooltipPosition]="top"
```
jeefy: This should be `matTooltipPosition="above"`

Enclosing the attribute in [] tries binding the value within the angular component and that's not needed.

"top" is not a valid value for this either, it's "above" -- See https://material.angular.io/components/tooltip/overview

src/app/frontend/common/components/resourcelist/event/template.html
```diff
@@ -62,14 +62,20 @@
...
         </mat-cell>
       </ng-container>
 
       <ng-container [matColumnDef]="getDisplayColumns()[6]">
         <mat-header-cell *matHeaderCellDef>Last Seen</mat-header-cell>
-        <mat-cell *matCellDef="let event">
+        <mat-cell *matCellDef="let event"
+          matTooltip="{{event.lastSeen | date:'medium'}}"
+          [matTooltipPosition]="top"
+        >
```
maciaszczykm: Formatting is broken here.

DanielUche: will fix and revert

src/app/frontend/common/components/resourcelist/event/template.html
DanielUche: will fix and reverse thanks

src/app/frontend/common/components/resourcelist/event/template.html
```diff
@@ -62,14 +62,18 @@
...
           {{event.firstSeen | date}}
         </mat-cell>
       </ng-container>
 
       <ng-container [matColumnDef]="getDisplayColumns()[6]">
         <mat-header-cell *matHeaderCellDef>Last Seen</mat-header-cell>
-        <mat-cell *matCellDef="let event">
+        <mat-cell *matCellDef="let event"
+          matTooltip="{{event.lastSeen | date:'medium'}}"
+          matTooltipPosition="above">
```
maciaszczykm: We keep the default position in other places. Remove it and we can merge.

jeefy: @DanielUche Ping! Make that small change and we can merge this in. :) Thanks!

maciaszczykm:
Applied the changes.

/lgtm
/approve

eloyekunle:
I don't think the issue completely fixes #3631, although it is a step in the right direction.
The goal is to add these tooltips everywhere a time is displayed on the dashboard.

maciaszczykm:
There is a formatting issue. I will close it as @DanielUche is not responding. Please reopen if you are willing to apply required changes, then we can merge.

/close

PR #3667, opened 2019-03-17
update WorkloadStatusComponent

ajatprabha:
Work towards #3440

Preview:

### DIFF
--- a/aio/closure/webpack.config.js
+++ b/aio/closure/webpack.config.js
@@ -69,6 +69,7 @@ module.exports = {
     'styles': [
       '../node_modules/material-design-icons/iconfont/material-icons.css',
       '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css',
+      '../node_modules/nvd3/build/nv.d3.css',
       '../src/app/frontend/styles.scss'
     ],
   },
@@ -93,6 +94,8 @@ module.exports = {
               process.cwd(), '../node_modules/material-design-icons/iconfont/material-icons.css'),
           path.join(
               process.cwd(), '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css'),
+          path.join(
+              process.cwd(), '../node_modules/nvd3/build/nv.d3.css'),
           path.join(process.cwd(), '../src/app/frontend/styles.scss')
         ],
         'test': /\.css$/,
@@ -110,6 +113,8 @@ module.exports = {
               process.cwd(), '../node_modules/material-design-icons/iconfont/material-icons.css'),
           path.join(
               process.cwd(), '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css'),
+          path.join(
+              process.cwd(), '../node_modules/nvd3/build/nv.d3.css'),
           path.join(process.cwd(), '../src/app/frontend/styles.scss')
         ],
         'test': /\.css$/,
@@ -127,6 +132,8 @@ module.exports = {
               process.cwd(), '../node_modules/material-design-icons/iconfont/material-icons.css'),
           path.join(
               process.cwd(), '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css'),
+          path.join(
+              process.cwd(), '../node_modules/nvd3/build/nv.d3.css'),
           path.join(process.cwd(), '../src/app/frontend/styles.scss')
         ],
         'test': /\.scss$|\.sass$/,
@@ -148,6 +155,8 @@ module.exports = {
               process.cwd(), '../node_modules/material-design-icons/iconfont/material-icons.css'),
           path.join(
               process.cwd(), '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css'),
+          path.join(
+              process.cwd(), '../node_modules/nvd3/build/nv.d3.css'),
           path.join(process.cwd(), '../src/app/frontend/styles.scss')
         ],
         'test': /\.scss$|\.sass$/,
--- a/aio/closure/webpack.config.js
+++ b/aio/closure/webpack.config.js
@@ -69,7 +69,6 @@ module.exports = {
     'styles': [
       '../node_modules/material-design-icons/iconfont/material-icons.css',
       '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css',
-      '../node_modules/nvd3/build/nv.d3.css',
       '../src/app/frontend/styles.scss'
     ],
   },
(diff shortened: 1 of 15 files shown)


src/app/frontend/overview/workloadstatus/template.html
```diff
@@ -14,132 +14,124 @@
...
-<!--size="140">-->
-<!--</kd-allocated-resources-chart>-->
-<!--<div class="kd-graph-title">-->
-<!--[[Cron Jobs|Label 'Cron Jobs' for the resource status chart on overview page.]]-->
-<!--</div>-->
-<!--</div>-->
-<!--</div>-->
+<kd-card>
+  <div title
+       fxLayout="row">Workload Statuses</div>
```
jeefy: I might leave this as a singular "Workload Status"

ajatprabha:
@jeefy I'm thinking to address the font and hover behavior in another PR since it is more related to `AllocationChartComponent`. Is it okay?

src/app/frontend/overview/workloadstatus/template.html
```diff
@@ -14,132 +14,124 @@
...
-<!--[[Stateful Sets|Label 'Stateful Sets' for the resource status chart on overview page.]]-->
-<!--</div>-->
-<!--</div>-->
-<!--</div>-->
-<!--</div>-->
+    <div class="kd-graph-container"
+         *ngIf="resourcesRatio.replicaSetRatio.length">
+      <div fxLayout="column"
+           fxLayoutAlign="center center">
+        <kd-allocation-chart [id]="'replicaSets'"
```
floreks: This could be just `id="replicaSets"` if you are not using binding anyway. Same applies to other resources.

src/app/frontend/overview/workloadstatus/component.ts
```diff
@@ -12,96 +12,29 @@
...
+  replicationControllerRatio: [],
+  statefulSetRatio: []
+};
+
+@Component(
+    {selector: 'kd-workload-statuses', templateUrl: './template.html', styleUrls: ['./style.scss']})
 export class WorkloadStatusComponent {
-  // @Input() ratio: any;
+  @Input() resourcesRatio: ResourcesRatio;
   colors: string[] = ['#00c752', '#f00', '#ffad20', '#006028'];
```
floreks: I see on the preview that colors are very bright and with hight contrast. Can you use same colors as before?

floreks:
Just a small comments. Nice work!

src/app/frontend/overview/workloadstatus/component.ts
ajatprabha: The hex codes are the same as before, I believe the problem is with `AllocationChartComponent`. In v1.10.1, it changes color on hover, which is not the behavior in the current implementation of the component. It needs some more fixes too as @jeefy pointed out in his review. I'm thinking to address this in another PR.

jeefy: @ajatprabha I think we should take the time to get it right on the first try. This isn't time critical, and this is fairly narrow-scoped. :) 

I could be convinced otherwise, but I'd say let's keep it in a single PR.

ajatprabha: Updated and pushed the changes.

src/app/frontend/common/components/allocationchart/component.ts
```diff
@@ -78,7 +74,11 @@ export class AllocationChartComponent implements OnInit {
                       .labelType(labelFunc);
 
     chart.tooltip.enabled(this.enableTooltips);
-    chart.tooltip.contentGenerator((obj) => {
+    chart.tooltip.contentGenerator(obj => {
+      if (obj.data.key.includes(':')) {
```
floreks: What is this logic?

aio/closure/webpack.config.js
```diff
@@ -69,6 +69,7 @@ module.exports = {
     'styles': [
       '../node_modules/material-design-icons/iconfont/material-icons.css',
       '../node_modules/roboto-fontface/css/roboto/roboto-fontface.css',
+      '../node_modules/nvd3/build/nv.d3.css',
```
floreks: What is this change? Files in the `closure` dir are not actually used anywhere. They are leftovers from old builds kept until we finish everything in case they are needed.

src/app/frontend/overview/workloadstatus/style.scss
```diff
@@ -0,0 +1,58 @@
...
+  padding-bottom: $baseline-grid;
+  text-align: center;
+}
+
+.kd-graph-legend-entry {
+  display: flex;
+  margin: 0 auto;
+  padding: $baseline-grid;
+  text-align: left;
+  width: 240px;
```
floreks: Should be baseline grid multiplication.

aio/closure/webpack.config.js
ajatprabha: This is the CSS I was talking about on Slack, it's not included by default. I was not aware of how are CSS files included from `node_modules` in kube-dashboard. I looked for similar includes and found this in webpack config, added the statement and it worked. If you can suggest any other method to include this CSS, I'll try that. The goal is just to include this CSS somehow in the bundle.

floreks: I mean that this should only be added to the `angular.json` file. This `webpack.config.js` file is basically not used at all.

src/app/frontend/common/components/allocationchart/component.ts
ajatprabha: > I'd prefer to see the hover behavior seen on [nvd3's site](http://nvd3.org/examples/pie.html)

@jeefy asked for a similar look in the link above. The `key` is of the type `label: count` eg: `Success: 10` and the `value` contains percentage eg: 100, so I could not use value directly. I used the key and split it to achieve the look.
If `:` is present in the key, I'm assuming it is of the form `label: count` and it'll split it, otherwise the old behaviour will be followed.

src/app/frontend/overview/workloadstatus/style.scss
ajatprabha: I'll update this.

aio/closure/webpack.config.js
ajatprabha: Okay, my bad. It works with changes to `angular.json` only.

ajatprabha:
Result so far.
Screenshot from 2019-03-22 21-55-26

src/app/frontend/common/components/allocationchart/component.ts
floreks: Could you add short comment as this is not obvious what it does?

floreks:
Looks great! I'll take a closer look later.

src/app/frontend/common/components/allocationchart/component.ts
ajatprabha: Added.

Issue #3656, opened 2019-03-14
Observe if lint-staged works with Go files


maciaszczykm:
Tracker mentioned in https://github.com/kubernetes/dashboard/pull/3614.

shu-mutou:
Following comments and observations were summerized from #3653:

https://github.com/kubernetes/dashboard/blob/1e9a6b1927da456cf6b2aeb07e4c0d317207bd2b/package.json#L68-L70

According to lint-staged documentation [1], file pathes would be passed as list.
[1] https://github.com/okonet/lint-staged#lintstagedrc-example

like: `golangci-lint run -c .golangci.yml --fix src/app/backend/handler/apihandler.go src/app/backend/integration/metric/heapster/client_test.go`

But if multiple file paths was passed to golangci-lint, it causes error like follow:
`ERRO Running error: context loading failed: no go files to analyze`

If single file path was passed to golangci-lint, the check runs successfully.

In my observations:
* It seems to run once even if we change 20 go files actually.
* Adding `src/app/backend/...` at the end, golangci-lint seems to run as we expected.

shu-mutou:
I found that due to [bug on golangci-lint][https://github.com/golangci/golangci-lint/issues/391], multiple files check failes sometimes.
We need to check each files independently.

PR #3676, opened 2019-03-20
Remove `golangci-lint` temporarily from `lint-staged`

shu-mutou:
`golangci-lint` in `lint-staged` does not work as we expected
due to its strage behavior (maybe bug in golangci-lint).
Not to block development, this removes `golangci-lint` temporarily
from `lint-staged` until we understand its behavior perfectly.

Observations will be tracked by #3656.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,10 +65,6 @@
     "src/**/*.scss": [
       "scssfmt",
       "git add"
-    ],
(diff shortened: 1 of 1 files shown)


maciaszczykm:
/lgtm
/approve

Thanks a lot!

PR #3679, opened 2019-03-20
Cleanup travis environment

### LINKED ISSUE #3677: Observe if $GOPATH should be set as `.tmp` in Travis CI
Tracker mentioned in #3653.

Found that golangci-lint does not check go files outside $GOPATH. So we seem to have no way rather than copying .golangci.yml to and changing directory under $GOPAH, i.e. under `.tmp`, before running golangci-lint, for now.

shu-mutou:
* remove lint-backend.sh
* set GOPATH to `~/gopath` on travis

Also, fixes goimorts error.

Fixes #3677


maciaszczykm:
Great :) 

/lgtm
/approve

PR #3721, opened 2019-04-05
Use settingsApi.SettingsManager in CreateHTTPAPIHandler

bouk:
Currently the CreateHTTPAPIHandler requires a settings.SettingsManager instead of an api.SettingsManager, which means you can't provide a different implementation if you want to embed the Handler with a different SettingsManager. The SettingsManager interface wasn't actually used anywhere and was out of sync with the implementation, so I've updated the interface to match.

The SettingsManager took in a clientManager, which was only used by the SettingsHandler. This PR moves the clientManager to the SettingsHandler.

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -108,7 +108,7 @@ func main() {
 	authManager := initAuthManager(clientManager)
 
 	// Init settings manager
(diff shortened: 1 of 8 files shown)


maciaszczykm:
Thanks, looks good!

/lgtm
/approve

PR #3729, opened 2019-04-12
Update CONTRIBUTING.md

wangxy518:
Updated an url


maciaszczykm:
The current one works as expected. Why do you want to change?

/close

PR #3767, opened 2019-04-27
Show tooltip timestamp on time hover across pages

### LINKED ISSUE #3631: Show tooltip timestamp on time hover across pages

feloy:
Fixes #3631


jeefy:
This looks :100: 

Small suggestion: Could you possibly the formatted times have some style that lets you know you _can_ hover? Even as simple as a dotted underline would be great. 

/assign

jeefy:
Changes look good.

/approve

maciaszczykm:
Please rebase/merge master to fix the issues with the CI job. Sorry for the inconvenience.

src/app/frontend/common/components/date/component.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+@Component({selector: 'kd-date', templateUrl: './template.html', styleUrls: ['./style.scss']})
+export class DateComponent {
+  @Input() date: string;
+  @Input() format = 'medium';
+
+  _relative: boolean;
+  @Input('relative')
+  set relative(v: boolean) {
+    this._relative = v !== undefined;
```
maciaszczykm: What if somebody would use `[relative]="false"`?

feloy: You are right, I'll use this case

jeefy:
Rebase looks good. @maciaszczykm PTAL

PR #3770, opened 2019-04-28
French translation

### LINKED ISSUE #3574: Support i18n
It is not supported since Angular migration and we should keep that in our roadmap.

feloy:
Fixes #3574

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "start:frontend:https": "node aio/scripts/version.js && ng serve --progress=false --aot --proxy-config aio/https-proxy.conf.json --ssl --host=0.0.0.0",
     "start:backend": "gulp serve --kubeconfig $npm_package_config_kubeconfig",
     "start:backend:https": "gulp serve --kubeconfig $npm_package_config_kubeconfig --autoGenerateCerts true --defaultCertDir $npm_package_config_kubeconfig_dir",
-    "start:prod": "npm run build && ./$npm_package_config_dashboard_binary_path --kubeconfig $npm_package_config_kubeconfig",
+    "start:prod": "npm run build && (cd $(dirname ./$npm_package_config_dashboard_binary_path) && ./$(basename $npm_package_config_dashboard_binary_path) --kubeconfig $npm_package_config_kubeconfig)",
     "build": "./aio/scripts/build.sh",
     "build:cross": "./aio/scripts/build.sh -c",
     "build:frontend": "npm run clean && ng build --aot --prod --outputPath=$npm_package_config_frontend_build_dir",
--- a/src/app/frontend/chrome/notifications/template.html
+++ b/src/app/frontend/chrome/notifications/template.html
@@ -15,7 +15,7 @@ limitations under the License.
 -->
 
 <div class="kd-notifications-container">
-  <span style="display: none"
+  <span hidden
         i18n
         #translatedTooltip>{isOpen(), select, 1 {Close notifications panel} other {Open notifications panel}}</span>
   <button mat-icon-button
--- a/src/app/frontend/chrome/notifications/template.html
+++ b/src/app/frontend/chrome/notifications/template.html
@@ -15,7 +15,9 @@ limitations under the License.
 -->
 
 <div class="kd-notifications-container">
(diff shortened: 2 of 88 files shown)


maciaszczykm:
Let us know when you would like to receive some feedback/review.

feloy:
Feedbacks are welcome at this point. Thanks


> Let us know when you would like to receive some feedback/review.

package.json
```diff
@@ -13,7 +13,7 @@
     "start:frontend:https": "node aio/scripts/version.js && ng serve --progress=false --aot --proxy-config aio/https-proxy.conf.json --ssl --host=0.0.0.0",
     "start:backend": "gulp serve --kubeconfig $npm_package_config_kubeconfig",
     "start:backend:https": "gulp serve --kubeconfig $npm_package_config_kubeconfig --autoGenerateCerts true --defaultCertDir $npm_package_config_kubeconfig_dir",
-    "start:prod": "npm run build && ./$npm_package_config_dashboard_binary_path --kubeconfig $npm_package_config_kubeconfig",
+    "start:prod": "npm run build && (cd $(dirname ./$npm_package_config_dashboard_binary_path) && ./$(basename $npm_package_config_dashboard_binary_path) --kubeconfig $npm_package_config_kubeconfig)",
```
maciaszczykm: `cd` looks ugly here. Can we avoid using it from `npm` scripts?

src/app/frontend/chrome/notifications/template.html
```diff
@@ -15,11 +15,13 @@
 -->
 
 <div class="kd-notifications-container">
+  <span style="display: none"
```
maciaszczykm: We are trying to not use styles from templates. Why do you need it? What are these 3 lines doing?

src/app/frontend/common/components/card/template.html
```diff
@@ -35,10 +35,11 @@
          class="kd-card-actions kd-muted">
       <ng-content select="[actions]"></ng-content>
     </div>
-
+    <span style="display: none"
+          #translatedTooltip
```
maciaszczykm: As above.

src/app/frontend/chrome/notifications/template.html
feloy: The content is used some lines below, for the matTooltip translation. Unfortunately, there is a current limitation in the i18n angular code and it is not possible to use ICU ({VAR, select, ...}) in a i18n-matTooltip (or any i18n-*), so we need to create an invisible element for that.

It seems it is not possible either to use an ng-template element.

@ocombe do you have some advice on this problem?

ocombe: So most likely the content of i18n-matTooltip is used by the directive directly and is not analyzed as template code, which is why the ICU doesn't work?
`ng-template` should work though

you can try `ng-container` to create an invisible element

feloy: Here is the exact description of the problem: https://github.com/angular/angular/issues/21615

And an explanation of the hack I'm using: http://chrisgriffing.com/coding/angular/2017/07/22/angular-i18n-icu-messages-attributes/

Unfortunately, that does not work with ng-container (the content is not invisible) or ng-template (the textContent is empty)

feloy:
I have finished the translation of the templates.

I will have to look at the MSG_ and figure out how to translate from the TS code (probably in another pull request).

maciaszczykm:
Please rebase/merge master to fix the issues with the CI job. Sorry for the inconvenience.

package.json
floreks: It is defined in [.npmrc](https://github.com/kubernetes/dashboard/blob/master/.npmrc).

floreks:
@jeefy @feloy @maciaszczykm  

I don't want to revert it because overall it is really good work, but I'd address 2 things.

1. Remove usage of the hidden element as a workaround for tooltip message logic. Export the logic to a function and when angular will support translations from the ts/js code also we will use that.
2. Definitely remove `cd ...` logic from the `package.json`. We have to keep it simple. Building translated versions was already supported by our build scripts for prod.

feloy:
I will work on this asap

feloy:
@maciaszczykm I have proposed a patch on #3812

PR #3794, opened 2019-05-08
Update webdriver-manager to 12.1.4

jeefy:
There were still flaky e2e tests. This seems to fix it across all platforms I can test.


jeefy:
Merging this to more sanely work on the other PRs. Failing test was due to poorly-linted code that got in. Will address soon.

PR #3795, opened 2019-05-09
Fix ts errors

shu-mutou:
Fix ts errors in favor of gts 1.0.0

- [x] fix format errors by `prettier`
- [x] fix `deprecation` errors by `gts`

`ban-ts-ignore` and `interface-over-type-literal` rules for `gts` are remained as ignore.

### DIFF
--- a/aio/closure/main.aot.ts
+++ b/aio/closure/main.aot.ts
@@ -1,7 +1,7 @@
-import {enableProdMode} from '@angular/core';
-import {platformBrowser} from '@angular/platform-browser';
+import { enableProdMode } from '@angular/core';
+import { platformBrowser } from '@angular/platform-browser';
 // noinspection TypeScriptCheckImport
-import {RootModuleNgFactory} from './index_module.ngfactory';
+import { RootModuleNgFactory } from './index_module.ngfactory';
 
 enableProdMode();
 
--- a/aio/scripts/version.js
+++ b/aio/scripts/version.js
@@ -39,8 +39,9 @@ writeFileSync(
 
 // clang-format off
 
-import {VersionInfo} from '@api/frontendapi';
+import { VersionInfo } from '@api/frontendapi';
 
+// prettier-ignore
 export const version: VersionInfo = ${JSON.stringify(gitInfo, null, 2).replace(/\"/g, '\'')};
 `,
     {encoding: 'utf-8'});
--- a/src/app/frontend/about/actionbar/component.ts
+++ b/src/app/frontend/about/actionbar/component.ts
@@ -12,11 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component} from '@angular/core';
-import {VersionInfo} from '@api/frontendapi';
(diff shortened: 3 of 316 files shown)


maciaszczykm:
This changes a lot. Why do we need this? Previous code format was pretty simple, now it seems a bit custom to me.

cc @floreks WDYT?

maciaszczykm:
Okay, I see. Master is failing since the: https://travis-ci.org/kubernetes/dashboard/builds/525741804 (https://github.com/kubernetes/dashboard/pull/3761)

cc @jeefy

shu-mutou:
> This changes a lot. Why do we need this? Previous code format was pretty simple, now it seems a bit custom to me.
> 
> cc @floreks WDYT?

As you said, simple is the best. :) And I was also seeking the way to fix them little by little, but not found as now, especially the way to avoid `prettier`'s error. This code formatting is default settings of gts, so this might be acceptable for many people, I think. I'm not picky about formatting, so any way is fine for me. Uhm.... but I don't like to use operators without spaces.

shu-mutou:
This PR contains three commits. The 1st one is only formatting typescripts. So could you guys review 2nd and 3rd commits?
[1] Fix typescript formatting in favor of gts 1.0.0 https://github.com/kubernetes/dashboard/pull/3795/commits/fbe1c37dad72a863b5a9ad230babf85dad143de8
[2] Fix deprecated errors in favor of gts 1.0.0 https://github.com/kubernetes/dashboard/pull/3795/commits/5b71890371cb43dcaac0811aa23fd3d1e584bdd9
[3] Ignore version.ts in gts and prettier https://github.com/kubernetes/dashboard/pull/3795/commits/9e4d84e2e593c298a3de4c5155f7ccbefd65297d

maciaszczykm:
@shu-mutou Totally makes sense. We have to use it. Perhaps in the future we will adjust it.

Thanks for fixing the issue.

/lgtm
/approve

eloyekunle:
@shu-mutou Should we consider integrating `prettier` into our workflow?

maciaszczykm:
@eloyekunle I think it woule be better to restore gts < v1.0. It was more readable for me.

cc @floreks

shu-mutou:
@eloyekunle @maciaszczykm @floreks If we keep to use gts and to update in the future, we can use `prettier` that gts integrates and its settings. To update gts to 1.0.0, we need to change so many files and it becomes difficult to merge other PRs. But it would be a good chance to update gts for us, because we would be in `feature freeze` period before new release. Also, I think the formatting of gts 1.0.0 would be acceptable for many people and it looks good for me.

jeefy:
My two cents -- the [Release Notes](https://github.com/kubernetes-sigs/release-notes) Site currently uses `prettier` and I have zero complaints. And I'm always a fan of updating libraries when it's warranted, and this might be warranted.

shu-mutou:
I will try to investigate again how to use gts 1.0.0 with custom `prettier` setting. 😉

shu-mutou:
I'm trying at #4034, please see it. 😉

PR #3812, opened 2019-05-14
French translation fixes

feloy:
Revert some hacks introduced in #3770 
- Revert the usage of hidden  hidden element as a workaround for tooltip message logic
- Add a locale-config parameter to the dashboard binary


maciaszczykm:
Ups, `/hold` didn't work...

floreks:
It's fine. We did not have support for i18n testing on prod before. LGTM. Thanks!

PR #3829, opened 2019-05-27
Add dynamic client for ClusterRole resource

zreigz:
Dynamic client from runtime-controller project reads from a cache and write to the API server. It will speed up reading operations when this client replace go-client for k8s


src/app/backend/resource/clusterrole/detail.go
```diff
@@ -32,9 +34,9 @@ type ClusterRoleDetail struct {
 }
 
 // GetClusterRoleDetail gets Cluster Role details.
-func GetClusterRoleDetail(client k8sClient.Interface, name string) (*ClusterRoleDetail, error) {
-	rawObject, err := client.RbacV1().ClusterRoles().Get(name, metaV1.GetOptions{})
-	if err != nil {
+func GetClusterRoleDetail(client ctrlruntimeclient.Client, name string) (*ClusterRoleDetail, error) {
+	rawObject := &rbac.ClusterRole{}
+	if err := client.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: metav1.NamespaceAll, Name: name}, rawObject); err != nil {
```
maciaszczykm: @zreigz One thing I would like to mention here is context creation. Should we do it at this point? Can we use something that already exists in the request for example? Also, we should propagate it once it will be needed in more places.

maciaszczykm:
Please reopen if you are willing to finish the implementation.

PR #3877, opened 2019-06-06
Remove redundant namespace filters

eloyekunle:
In `resourcechannels.go`, REST queries already filtered by namespace should no longer be filtered by the filtering loop.


maciaszczykm:
Not sure about this anymore. Indeed `ToRequestParam` will filter with namespaces but it can be also set to all if there are more than 2 namespaces. `Matches` works in a different way, filtering using specific namespaces from `NamespaceQuery`.

@floreks Was this intended? Can we pick let's say 2 or 3 namespaces somewhere (not all of them)?

eloyekunle:
You're right @maciaszczykm, `ToRequestParam` sets to `ALL` if there are more than 1 namespace.
So the filters are indeed necessary.

maciaszczykm:
Okay then, let's close it as it has some function.

/close

PR #3878, opened 2019-06-06
Migrate from dep to go mod

eloyekunle:
Since Go 1.11 there is a new way to structure packages, which don't require the project to live under `$GOPATH`, and provides version management.

`go mod` could superseed `dep`, given it is builtin on the core tooling of the language and requires less dependencies to contribute.

This commit follows the [migration instructions](https://github.com/golang/go/wiki/Modules#migrating-to-modules) to adopt `go mod`.


maciaszczykm:
Great! Looking forward to merging this. Could you take a look at the errors in the CI?

maciaszczykm:
Master is broken, not this PR. You can take a look on it if you have some time.

cc @floreks @jeefy

shu-mutou:
At least, before tests or building image that use vendor directory, we need to set `GO111MODULE` and run `go mod`  like follow.
```
export GO111MODULE=on
go mod download
go mod vendor
```
We can run almost tests on local. See https://github.com/kubernetes/dashboard/wiki/Getting-started#to-run-npm-commands-as-you-want . It's almost same as on travis. Please run test locally at first.

eloyekunle:
The backend tests pass currently when I run it locally. This current broken build is from master, not the go modules migration.

floreks:
Rebase please to fix travis.

eloyekunle:
I've rebased @floreks

shu-mutou:
For now, I couldn't find why vendor directory [1] that were used by tests and build becomes unnecessary. But we need to clean up the use of vendor directory, if vendor is really unused. Could you clean up them?
[1] https://github.com/kubernetes/dashboard/search?q=vendor&unscoped_q=vendor

maciaszczykm:
/approve

Please solve conflicts and answer to @shu-mutou comment.

eloyekunle:
@shu-mutou The `vendor` directory is still used during build and is still necessary.
By running `go mod vendor` locally, the appropriate packages are added to the `vendor` directory.
In the future with go modules, we might not need to vendor dependencies but instead, as you've said, run `go mod download` and `go mod vendor` during builds. With this approach, the issue the community is currently trying to figure out is how to ensure that builds are not affected by repository changes (e.g. deletion) of required packages.

shu-mutou:
/lgtm
Great! And thank you for your answer. I think, we need to memo somewhere how to build container image, but I could confirm that we can launch the dashboard, the CI checks is passing and we can fix concerns later.

PR #3890, opened 2019-06-11
[WIP] Fix git-describe in version.js

shu-mutou:
Fix error in `aio/scripts/version.js` .


shu-mutou:
`git-describe` does not get tags in git repository.

shu-mutou:
travis does not seems to fetch tags information, therefore git-describe fail to get tag info and semver.

shu-mutou:
On travis, run `git tag` but `git describe --tags` does not return  with tag information.
@maciaszczykm Could you advice about this?

floreks:
I have already fixed that. :slightly_smiling_face: We just needed to increase cloning depth of the git in the travis. It was set to `250` and we are over `300` commits in front of master. It was not able to get version.

shu-mutou:
@floreks @maciaszczykm Thanks a lot!!

Issue #3927, opened 2019-06-14
Error view is broken


ajatprabha:
##### Environment

```
Installation method: Resository checkout
Kubernetes version: 1.14.3
Dashboard version: master(89f56f0)
Operating system: MacOS v10.14.5
Node.js version ('node --version' output): v12.1.0
Go version ('go version' output): go1.12.5 darwin/amd64
```

##### Observed result
Navigating to `/error` throws the following error:
```js
TypeError: Cannot read property 'toggle' of undefined
    at NavService.push../src/app/frontend/common/services/nav/service.ts.NavService.setVisibility (service.ts:27)
    at ErrorComponent.push../src/app/frontend/error/component.ts.ErrorComponent.ngOnInit (component.ts:33)
```

##### Expected result

Error view should work as expected.

maciaszczykm:
Can you try again with the latest master?

ajatprabha:
It is there on the latest master. Can you put some light on why `NavService` is required here?

maciaszczykm:
@ajatprabha It is indeed some issue, but I think it does not exist on master. One bug was causing another. We can add checks for null here, but I think this was not normal use case.

maciaszczykm:
Seems to be fixed now anyways.

/close

PR #3928, opened 2019-06-14
fix error view

ajatprabha:
Ref: #3927

### DIFF
--- a/src/app/frontend/error/component.ts
+++ b/src/app/frontend/error/component.ts
@@ -18,7 +18,6 @@ import {StateError} from '@api/frontendapi';
 import {map} from 'rxjs/operators';
 
 import {KdError} from '../common/errors/errors';
-import {NavService} from '../common/services/nav/service';
 
 @Component({
   selector: 'kd-error',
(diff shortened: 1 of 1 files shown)


src/app/frontend/error/component.ts
```diff
@@ -27,11 +26,10 @@ import {NavService} from '../common/services/nav/service';
 })
 export class ErrorComponent implements OnInit {
   private error_: KdError;
-  constructor(private readonly nav_: NavService, private readonly route_: ActivatedRoute) {}
 
-  ngOnInit(): void {
-    this.nav_.setVisibility(false);
+  constructor(private readonly route_: ActivatedRoute) {}
```
floreks: This way when entering error view, sidebar is still visible right?

ajatprabha: Yes, I can hide it if needed.

floreks: It would be best to hide it in case we are entering this view. You could use injector to get this service.

src/app/frontend/chrome/template.html
```diff
@@ -54,7 +54,8 @@
   <mat-toolbar class="kd-second-toolbar kd-bg-primary">
     <div class="kd-toolbar-tools"
          fxFlex>
-      <kd-nav-hamburger fxFlex="none"></kd-nav-hamburger>
+      <kd-nav-hamburger *ngIf="showNav()"
```
floreks: We should use nav service for that. This feels more like a workaround. If nav service does not work, then just leave it visible for now (restore previous fix).

PR #3934, opened 2019-06-17
Fix golanci-lint for pre-commit

shu-mutou:
Previously, we add `src/app/backend/...` at the end of `golangci-lint` command to run it for pre-commit.
But we do not need to take such tricky way for now.

Fixes #3918

PR #3935, opened 2019-06-17
Disable `go mod` once in travis

shu-mutou:
We need to investigate to use `go mod` in travis.
For now, change `GO111MODULE` to off in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ cache:
 
 before_install:
   - export NG_CLI_ANALYTICS=ci
-  - export GO111MODULE=on
+  - export GO111MODULE=off
   - export PATH=$PATH:$GOPATH/bin
   - eval "$(gimme 1.12.6)"
   - cd $HOME


shu-mutou:
`npm ci` seems to update go.mod file to use latest. Also, prometheus/client-golang was bumped latest to 1.0.0 few days ago, then prometheus.Handler() removed.

maciaszczykm:
Hmm, @eloyekunle can you help us there?

shu-mutou:
I updated this PR as disabling `go mod` in travis.
Also, I posted Issue #3936. @eloyekunle Could you investigate this issue?

maciaszczykm:
cc @floreks AFAIK you enabled Go modules. We need them, arent we?

shu-mutou:
Basically, I agree with using `go mod`, but we need to investigate it more.  😉

eloyekunle:
I will look into it.

eloyekunle:
`go get` for `golangci-lint` in the `postinstall` script was updating some packages. In #3934, we download `golangci-lint` via `curl` instead.

shu-mutou:
@eloyekunle Thank you for your help! I understand we should not use `go get` to install tools.

PR #3944, opened 2019-06-19
Use minimist instead of gulp-util

shu-mutou:
Due to gulp-util was deprecated, use minimist instead.

### DIFF
--- a/aio/gulp/conf.js
+++ b/aio/gulp/conf.js
@@ -15,7 +15,7 @@
 /**
  * @fileoverview Common configuration constants used in other build/test files.
  */
-import gulpUtil from 'gulp-util';
+import minimist from 'minimist';
 import path from 'path';
 
 /**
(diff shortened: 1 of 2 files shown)


aio/gulp/conf.js
```diff
@@ -69,6 +69,11 @@ const version = {
  */
 const imageNameBase = 'kubernetes-dashboard';
 
+/**
+ * Arguments
+ */
+const argv = minimist(process.argv.slice(2));
```
maciaszczykm: What does this exactly mean? Why there is `.slice(2)`?

shu-mutou: `process.argv` has [0] as `node`, [1] as `gulp` and [2] as `serve`. So we need after them.

shu-mutou:
I found `gulp-license-check` also uses `gulp-util`, and `minimist` used by lots of other packages. So, `package-lock.json` was not updated by `npm i`. 😨

aio/gulp/conf.js
maciaszczykm: Makes sense, thank you.

maciaszczykm:
Ah, that might be the case.

/lgtm
/approve

PR #3971, opened 2019-06-26
Add `i18n` tasks

shu-mutou:
Add npm tasks `check:i18n` and `fix:i18n` that execute `ng xi18n`.
Also, add `pre-commit` task for html files that check updates in `i18n/messages.xlf` file.


shu-mutou:
Specify versions for gulp and angular/cli

PR #3975, opened 2019-06-26
Japanese translation

shu-mutou:
Add Japanese translation file `i18n/messages.ja.xlf`.

### DIFF
--- a/i18n/locale_conf.json
+++ b/i18n/locale_conf.json
@@ -1,1 +1,1 @@
-{"translations": [ "en", "fr" ]}
+{"translations": [ "en", "fr", "ja" ]}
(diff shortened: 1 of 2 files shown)


shu-mutou:
/hold
until #3971 merged.

aio/develop/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN curl -sL https://deb.nodesource.com/setup_11.x | bash - \
 	&& apt-get clean
 
 # Install dependencies. This will take a while.
-RUN npm install -g npm@latest gulp
+RUN npm install -g npm@latest gulp @angular/cli@latest
```
maciaszczykm: `latest` can be problematic with the new releases. Let's keep that in mind for the future.

shu-mutou: Thanks! Versions for gulp and angular/cli were specified same as written in package.json in #3971

maciaszczykm:
Thank you for this huge effort, looks great.

Issue #3863, opened 2019-06-06
Manually test the HEAD image and release the beta


maciaszczykm:

maciaszczykm:
Hey, I think we are ready for a pre-release. What do you think?

How we will handle the metrics sidecar, should we add it to the recommended setup?

jeefy:
At this point the scraper has 1.8k pulls and only one outstanding issue (that is more of a feature request) so I'm comfortable moving it into the recommended deployment.

I'll cut a PR now cloning it over. Besides that PR, I think we cut an RC release soon. I tested it a bunch just now and things looked solid. :)

discostur:
just deployed the 2.0-beta1 - it looks really well! great job;)

PR #3987, opened 2019-06-29
Add scraper to recommended.

jeefy:
Ref #3863 

Also removed amd64 from the `-head` script because we have manifests in place on Docker Hub.

When we cut a release for 2.0.0-alpha, we can actually remove the `amd64` from the recommended YAML as well as remove the arm64-specific files since this release will also push a manifest up to gcr now. :)


maciaszczykm:
We should do the release from the same PR, so there will be no situation like users want to use it but it doesn't work.

maciaszczykm:
I will update versions on top of that PR, so we can make the release from it.

maciaszczykm:
Closed by https://github.com/kubernetes/dashboard/pull/4019.

PR #3995, opened 2019-07-01
Add go modules verification and update scripts

### LINKED ISSUE #3993: Verify vendored modules in pull requests
Add scripts to _update vendor_ and _verify vendor_. Vendor verification should run in Travis for all pull requests to ensure the `go.mod` file includes the correct dependencies.
The _update vendor_ script will run `go mod tidy` to prune all extraneous requirements.

##### Why is this needed
To ensure that the contents of the `go.mod` file and the `/vendor` directory are always correct and verified.
##### Comments

This is based on the current implementation in the [Kubernetes repository](https://github.com/kubernetes/kubernetes):

1. The[ `update-vendor.sh`](https://github.com/kubernetes/kubernetes/blob/master/hack/update-vendor.sh) script.
2. The [`verify-vendor.sh`](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-vendor.sh) script.

eloyekunle:
Closes #3993 

Adds two new `package.json` scripts:

1. `check:backend:vendor`
2. `fix:backend:vendor`


floreks:
@eloyekunle we do not need vendor directory in the dashboard repo anymore as go mod allows us to drop it. During default `go build` it is not used anyway. We would have to explicitly call `go build -mod=vendor` to use it. Required versions are defined in `go.mod` file. I have deleted this directory. Dashboard will be much smaller now thanks to that.

eloyekunle:
@floreks I have deleted the `vendor/` directory from this PR and updated the scripts.

shu-mutou:
Cleaning up `go.mod` file using `go mod tidy` looks good for me. But could you let me know why we should keep vendor directory even if on local environment or travis. Can `go mod` not ensure library dependencies without running` go mod vendor`?

floreks:
@shu-mutou  Go docs only describes one reason to keep vendor directory. That is if user would use older Go version or have disabled go mod feature. Otherwise there is no benefit of keeping it inside the repo.

shu-mutou:
@floreks You mean we do not need to add `go mod vendor` into CI process, don't you? Our CI process does not seem to face any problem actually, so I agree with not to add `go mod vendor`.

floreks:
@shu-mutou `go mod vendor` only creates a vendor directory and is not used by the `go build` by default. We can drop it everywhere.

aio/scripts/update-vendor.sh
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+TMP_DIR="${TMP_DIR:-$(mktemp -d /tmp/update-vendor.XXXX)}"
+
+echo "vendor: running 'go mod tidy'"
+go mod tidy -v
+
+echo "vendor: running 'go mod vendor'"
+go mod vendor
```
floreks: This is not needed as it would create a local vendor dir.

floreks:
@eloyekunle those vendor scripts are more `kubernetes` related. They won't really work for us since we will not be using vendor directory. All dependencies will be downloaded to your `$GOPATH` anyway. Also bumping and tidying Go deps is rather manual task for us. We do not need to do this on every `npm run check`. It should not be done silently in background as it might introduce some breaking changes.

@maciaszczykm WDYT?

maciaszczykm:
I agree. For me `go mod` is something like `npm` right now. It is a tool to be used directly. We don't have any scripts to update Node deps.

eloyekunle:
I agree, the scripts are unnecessary since we're not keeping the `vendor` directory anymore.

/close

Issue #3996, opened 2019-07-01
Add ability to register a custom plugin through a CRD


ajatprabha:
##### What would you like to be added
A dashboard user should be able to register a new custom plugin via a `kubectl` command. For eg: `kubectl create plugin plugin-name --source-from-file=<path-to-bundle.js>`

##### Why is this needed
This will enable us to register plugins with Kubernetes dashboard and extend its capabilities without too much hassle for the end user.

PR #4017, opened 2019-07-04
Plugin controller to manage plugin source from dashboard backend

ajatprabha:
Ref: #3996
- [x] Add CRD definition for custom plugins
- [x] Add go types for `Plugin` kind
- [x] Read plugin source from ConfigMapRef
- [x] Serve plugin sources via backend whose CRD exists


aio/test-resources/plugin-test.yml
```diff
@@ -0,0 +1,24 @@
...
+# A manifest that creates a plugin resource.
+
+apiVersion: dashboard.k8s.io/v1alpha1
+kind: Plugin
+metadata:
+  name: plugin-test
+spec:
+  source:
+    configMapRef:
+      name: plugin-test-src
```
maciaszczykm: You should also add the definition of CRD here I think. Some people will try to deploy it not finding the definition in the same directory.

aio/scripts/update-codegen.sh
```diff
@@ -0,0 +1,33 @@
...
+# generate the code with:
+# --output-base    because this script should also be able to run inside the vendor dir of
+#                  kl8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
+#                  instead of the $GOPATH directly. For normal projects this can be dropped.
+"${CODEGEN_PKG}"/generate-groups.sh "deepcopy,client,informer,lister" \
+  github.com/kubernetes/dashboard/src/app/backend/plugin/client github.com/kubernetes/dashboard/src/app/backend/plugin \
+  apis:v1alpha1 \
+  --output-base "$(dirname "${BASH_SOURCE[0]}")/../../../../.."
+
+# To use your own boilerplate text append:
```
maciaszczykm: Do we need this?

aio/scripts/update-codegen.sh
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/../..
+CODEGEN_PKG=${GOPATH}/src/k8s.io/code-generator
+# generate the code with:
```
maciaszczykm: Add a new line.

src/app/backend/handler/apihandler.go
```diff
@@ -584,6 +586,16 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
 			To(apiHandler.handleLogFile).
 			Writes(logs.LogDetails{}))
 
+	// Plugin routes
```
maciaszczykm: Would be nice to have these routers in a separate package. We moved, for example, some routes to the `integration` package.

src/app/backend/handler/apihandler.go
```diff
@@ -2264,6 +2276,44 @@ func (apiHandler *APIHandler) handleLogFile(request *restful.Request, response *
...
+	}
+	namespace := request.PathParameter("namespace")
+	pluginName := request.PathParameter("pluginName")
+
+	result, err := plugin.GetPluginSource(pluginClient, k8sClient, namespace, strings.Split(pluginName, ".")[0])
+	if err != nil {
+		errors.HandleInternalError(response, err)
+		return
+	}
+	_, _ = response.Write(result)
```
maciaszczykm: Ignoring everything here? What if an error occurs, will it be silenced?

maciaszczykm:
Looks good, added some comments.

src/app/backend/handler/apihandler.go
ajatprabha: Yes, it'll be silenced. I can follow the original signature if that's suggested.

aio/scripts/update-codegen.sh
ajatprabha: Not really right now but I thought we might forget that we have this feature if I remove the comment.

src/app/backend/handler/apihandler.go
ajatprabha: Moved to `plugin/handler.go`

src/app/backend/handler/apihandler.go
ajatprabha: Left the error unhandled.

src/app/backend/plugin/handler.go
```diff
@@ -0,0 +1,90 @@
...
+    return
+  }
+  k8sClient, err := h.cManager.Client(request)
+  if err != nil {
+    errors.HandleInternalError(response, err)
+    return
+  }
+  namespace := request.PathParameter("namespace")
+  // Removes .js extension if it's present
+  pluginName := strings.Split(request.PathParameter("pluginName"), ".")[0]
```
maciaszczykm: Wouldn't it be a bit better to remove only the last part after dot? `plugin.abc.js => plugin.abc`.

src/app/backend/handler/apihandler.go
floreks: I don't see `plugin/handler.go` file.

src/app/backend/client/manager.go
```diff
@@ -110,6 +114,21 @@ func (self *clientManager) APIExtensionsClient(req *restful.Request) (apiextensi
 	return self.InsecureAPIExtensionsClient(), nil
 }
 
+// PluginClient returns a plugin client. In case dashboard login is enabled and
+// option to skip login page is disabled only secure client will be returned, otherwise insecure
+// client will be used.
+func (self *clientManager) PluginClient(req *restful.Request) (pluginclientset.Interface, error) {
```
floreks: I am wondering how RBAC management for CRDs will work. Will logged in user require privileges to get/list Plugin CRD in order to see the plugins? Should allow plugins to be listed always and only enforce privileges if they will i.e. ask for a list of pods but the user will not have privileges to get them?

aio/test-resources/plugin-crd.yml
```diff
@@ -0,0 +1,41 @@
...
+  name: plugins.dashboard.k8s.io
+spec:
+  group: dashboard.k8s.io
+  scope: Namespaced
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+  names:
+    kind: Plugin
```
kfox1111: Not sure how conflicts are handled. If someone else registered a plugin, how would you then list plugins?

Should this be dashboardplugin?

ajatprabha: Because of the group name, I think there won't be any issues/clashes.

src/app/backend/client/manager.go
ajatprabha: I think we can move forward with the current changes for now. It may be desirable to not show the available plugins to all users. WDYT?

aio/test-resources/plugin-crd.yml
kfox1111: A quote from leggitt on a similar issue: "shortnames can be used within different groups, like kubectl get svc.mygroup.io" So, yeah. I think your right. its a non issue. Thanks. :)

src/app/backend/client/manager.go
floreks: We'll see how it impacts users. We can move forward for now.

src/app/backend/plugin/handler.go
ajatprabha: Makes sense, I'll update this.

src/app/backend/plugin/detail_test.go
```diff
@@ -0,0 +1,72 @@
...
+        },
+        Filename: filename}},
+  })
+
+  _, err = GetPluginSource(pcs, cs, ns, pluginName)
+  if err == nil {
+    t.Errorf("error 'configmaps \"%s\" not found' did not occur", cfgMapName)
+  }
+
+  _, _ = cs.CoreV1().ConfigMaps(ns).Create(&coreV1.ConfigMap{
```
maciaszczykm: Ignoring error?

ajatprabha: Yes, this is a fake client so I thought of leaving the error unchecked. I don't think it will return an error here.

maciaszczykm: Ahh, sorry. Didn't notice that this is a test 🤦‍♂

floreks:
Awesome!

/lgtm

maciaszczykm:
/approve

@floreks will you LGTM?

PR #4019, opened 2019-07-04
Update deploy files

floreks:
- Removed `create` privileges for `secrets` and `configmap` in favor of creating required resources from yaml.
- Removed architecture based files. Now that we are deploying multiarch image it is not required to split it this way.
- Split resources into different files and dirs for easier management.
- Added single file deployments to top `aio/deploy` dir. In the future we could add some script to either deploy using `curl ...` or to generate them.
- Moved kubernetes to a seperate namespace `kubernetes-dashboard` and `kubernetes-dashboard-head` for head.


bdonnahue:
The removal of kubernetes-dashboard.yaml has broken the installation documentation here: https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/

What yaml files need to be deployed to fix the documentation?

floreks:
@bdonnahue docs will be fixed by https://github.com/kubernetes/website/pull/15290

bdonnahue:
thanks @floreks !

trezm:
This seems also to have broken arm (raspberry pi) installation. Running with the `recommended.yaml` doesn't seem to work.

Looks like the last one that's working is here: https://raw.githubusercontent.com/kubernetes/dashboard/523a6fb16e164d70489ff80f47ec128403e8e9ab/aio/deploy/alternative/kubernetes-dashboard-arm.yaml

PR #4034, opened 2019-07-09
Bump gts to 1.0.0

shu-mutou:
Also, drop `clang-format` in favor of `prettier`.


shu-mutou:
`gts` does not seems to have way to disable `prettier`. To ignore all ts files, `.prettierignore` file was not effective. Therefore if we keep to using and updating `gts`, we need to accept `prettier`'s formatting.

293 files are changed by default settings.
235 files changed by [custom settings](https://github.com/kubernetes/dashboard/commit/50eaf665af77cf3edff65d2804c92d74c4d32a5a#diff-28c2b0e39a7056f1d1357e6efb94fd3b).

src/app/frontend/common/components/resourcelist/deployment/component.ts
```diff
@@ -12,41 +12,61 @@
...
+    this.registerBinding(
+      this.icon.checkCircle,
+      'kd-success',
+      this.isInSuccessState
+    );
+    this.registerBinding(
+      this.icon.timelapse,
+      'kd-muted',
+      this.isInPendingState
+    );
```
maciaszczykm: Styling is not so bad but there are a few things I don't like. For me `prettier` is using too much of new lines like in the example above. Can we limit it somehow?

shu-mutou: I think so too. We can set maximum length for each line to suppress adding new line, but `pettier` does not have option such as allowing to enumerate arguments on the second and subsequent lines.
I tried to set options for `prettier` to format closer as current code in subsequent commit. Please check it. 😉

maciaszczykm:
Looks better already.

/approve

@floreks WDYT?

shu-mutou:
/hold
This PR tends to conflict due to other mergings and would cause lots of conflict in other PRs. Also, in order to create this PR, it is necessary to recreate manually instead of rebasing. Please let me know when it is time to merge this PR.
If the time will come, I will recreate this PR ASAP.

maciaszczykm:
I am fine with moving it forward, @floreks @jeefy WDYT?

floreks:
I think we can do this after we release `beta2`. Maybe today.

maciaszczykm:
@shu-mutou It is the right time to rebase. I can LGTM then.

shu-mutou:
If conflicts occur in your PR due to this PR, please handle your PR before rebasing as follow:
1. change following files as same as this PR.
  - `.prettierrc.yaml`
  - `aio/scripts/version.js`
  - `package.json`
  - `tslint.json`
2. run `npm i`
3. run `npm fix:frontend`
4. then rebase your PR to master

PR #4046, opened 2019-07-11
Support dashboard-metrics-scraper on development container

shu-mutou:
Deploy `dashboard-metrics-scraper` on `kind` cluster.
Then pass `--sidecar-host` to dashboard via `npm` or `gulp`.
Also, cleanup handlings of `kubeconfig` for `kind`.


.gitignore
```diff
@@ -27,5 +27,3 @@ coverage/
 # npm debug log:
 npm-debug.log
 
-# kubeconfig for cluster made by kind:
-kind.kubeconfig*
```
maciaszczykm: Why was it removed?

.npmrc
```diff
@@ -9,3 +9,4 @@ kubernetes-dashboard:frontend_build_dir = dist/frontend/en
 kubernetes-dashboard:dashboard_binary_path = dist/amd64/dashboard
 kubernetes-dashboard:dashboard_locale_config = dist/amd64/locale_conf.json
 kubernetes-dashboard:bind_address = 127.0.0.1
+kubernetes-dashboard:sidecar_host = http://localhost:8000
```
maciaszczykm: Missing newline at the end of the file.

.gitignore
shu-mutou: Because moved its destination directory to `/tmp` on container.

.npmrc
shu-mutou: done.

.gitignore
maciaszczykm: Will not it be a problem when developing locally without a container?

package.json
```diff
@@ -7,13 +7,13 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "start": "concurrently \"npm run start:backend\" \"npm run start:frontend --kubernetes-dashboard:bind_address=$npm_package_config_bind_address\"",
+    "start": "concurrently \"npm run start:backend --kubernetes-dashboard:sidecar_host=$npm_package_config_sidecar_host\" \"npm run start:frontend --kubernetes-dashboard:bind_address=$npm_package_config_bind_address\"",
```
floreks: Isn't adding `--kubernetes-dashboard:sidecar_host=$npm_package_config_sidecar_host` obsolete if it's already added to the `start:backend` task?

.gitignore
shu-mutou: kubeconfig for `kind` was originally created into `${HOME}/.kube/config` by `aio/scripts/start-cluster.sh` https://github.com/kubernetes/dashboard/blob/master/aio/scripts/conf.sh#L81
`./kind.kubeconfig` was created for only development container.

package.json
shu-mutou: Sometimes, I run `start:backend` and `start:frontend` independently. So the new argument is need for both.

.gitignore
maciaszczykm: Okay, thank you.

package.json
floreks: But `start:backend` already has `--sidecarServerHost $npm_package_config_sidecar_host`. Why is `--kubernetes-dashboard:sidecar_host=$npm_package_config_sidecar_host` needed then for `npm start` if it already executes `npm start:backend`?

My understanding is that wheter we run `npm start` or `npm start:backend` they both should use `sidecarServerHost` parameter defined for `npm start:backend`.

shu-mutou: I also have thought same as you. But actually `$npm_package_config_sidecar_host` value was not passed to `npm start:backend` when I did not set it from `npm start`. So I concluded we need to set it to both.

floreks: That's quite confusing. Thank you for checking that.

PR #4087, opened 2019-07-18
add tmp-volume to metrics-scraper

MarkDeckert:
Add an emptyDir /tmp volume to the metrics-scraper so it doesn't have to run as root.  Matches setup with the dashboard itself.


MarkDeckert:
sorry about multiple commits.  Didn't realize at first there was slightly different indenting between all the yamls

floreks:
Thanks!

/lgtm

Issue #4069, opened 2019-07-17
Add plugin support on the frontend


ajatprabha:
##### What would you like to be added
#4017 adds support for serving plugins from the backend. The dashboard should be able to consume them and mount them in the UI.

##### Why is this needed
See Ref: #1832 

##### TODOs
- [x] Add plugin discovery service to dashboard
- [x] Mount plugins in the UI
- [x] Add routing for plugins
- [x] Error Handling
- [ ] Plugin support feature should be toggleable during production build until a stable version is released

ajatprabha:
Closing since #4094 is merged.

PR #4094, opened 2019-07-19
Add plugin support on the frontend

ajatprabha:
Ref: #4069 
- [x] List registered plugins in a ResourceList based component
- [x] Add a link to all plugins in the Sidebar
- [x] Add a route for plugin UI to render
- [x] Render the AOT compiled plugin at runtime.
- [x] Error Handling
- [x] Fetch plugins based on the namespace


ajatprabha:
Current progress:
1. List of installed plugins  

Screenshot from 2019-07-22 17-04-19

2. A sample plugin displaying available pods in the cluster  

Screenshot from 2019-07-22 17-03-43

maciaszczykm:
Testing it :)

maciaszczykm:
The plugin works, that's great.

Few remarks from my side:
- If there is no `Plugin` CRD registered in the cluster I'd avoid listing `Plugins` in the menu. Thanks to that users can use dashboard like they do now if the have no plugins. Otherwise, when they enter plugin page they will see errors:

Zrzut ekranu 2019-07-25 o 11 34 45

Another thing to consider here might be adding CRD from our backend and listing `Plugins` in the menu only if there are any plugins in the cluster.

- I'd like to know more how the plugins can be created. Which parts of the backend I can use, which parts of the frontend I can use. How to build the plugins. It will require some detailed documentation.

- Can you add plugin that uses material design card and styling is similar to the rest of the app? It would be valuable to show it to the people.

- I'd really appreciate possibility to pin some of the plugins to the menu like we advised for CRDs. I mean in this demo `plugin1` could be pinned to the menu and then appear just under `Plugins`. Like `Pods` appear under `Workloads`.

At the moment that's it. I really appreciate all the effort and I think most crucial part is already done. Keep it up!

floreks:
I have found couple of issues while testing:
- Not working with secure Dashboard (login enabled)
Zrzut ekranu z 2019-07-25 12-19-29
- Delete and edit is available on the plugin list but it is not working
Zrzut ekranu z 2019-07-25 12-24-36
- Sorting not applied on the plugin list when first entering the page, only after manually hitting sort or clicking again on `Plugins` entry.

Questions:
1. How dependencies are calculated?
2. Are we able to inject plugin view into any view?

Besides that, really nice progress. It looks better and better.

src/app/frontend/common/services/global/plugin.config.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017 The Kubernetes Authors.
```
floreks: You can rename it to simply `plugin.ts`.

src/app/frontend/common/services/global/plugin.config.ts
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+import {HttpClient} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {AppConfig} from '@api/backendapi';
+import {VersionInfo} from '@api/frontendapi';
+import {Observable} from 'rxjs/Observable';
+import {version} from '../../../environments/version';
+
+interface PluginMetadata {
```
floreks: Can we move those interfaces to `frontendapi.ts` typings?

PR #4101, opened 2019-07-21
Help access dashboard with v2

feloy:


README.md
```diff
@@ -31,6 +31,12 @@ Now access Dashboard at:
 [`http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/`](
 http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/).
 
+Or, if you are using the v2:
+
+[`http://localhost:8001/api/v1/namespaces/kubernetes-dashboard/services/https:kubernetes-dashboard:/proxy/`](
```
maciaszczykm: There is no v2 mention anywhere else in the menu. I'd rephrase:

```
**NOTE**: If you are using one of v2.0.0 beta releases then use: ...
```

Adding some link to the latest beta release might be good idea too.

@floreks @jeefy WDYT?

floreks: We'd have to add it also for latest v1.

floreks: Actually, I'd rather mention v2 only in the README and move v1 access part to the docs. We could only mention that if you want to use `v1` follow this guide.

PR #4143, opened 2019-07-30
add metrics-provider's no check metrics option

xieyanker:
I add a metrics-provider's option, it can cancel metrics' check, otherwise it will log a plenty of loggs if my k8s cluster no sidecar or heapster exist.

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -56,7 +56,7 @@ var (
 		"to connect to in the format of protocol://address:port, e.g., "+
 		"http://localhost:8080. If not specified, the assumption is that the binary runs inside a "+
 		"Kubernetes cluster and local discovery is attempted.")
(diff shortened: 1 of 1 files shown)


src/app/backend/dashboard.go
```diff
@@ -130,8 +130,10 @@ func main() {
 	case "heapster":
 		integrationManager.Metric().ConfigureHeapster(args.Holder.GetHeapsterHost()).
 			EnableWithRetry(integrationapi.HeapsterIntegrationID, time.Duration(args.Holder.GetMetricClientCheckPeriod()))
+	case "noprovider":
```
floreks: I'd change that to `none`.

@maciaszczykm WDYT?

maciaszczykm: I agree.

xieyanker: Thank you for your reply, done!

PR #4151, opened 2019-08-01
Use binary search in getLineIndex method

### LINKED ISSUE #3900: Use binary search in getLineIndex method
Timestamps should be compared.

eloyekunle:
Closes #3900

### DIFF
--- a/src/app/backend/resource/container/logs_test.go
+++ b/src/app/backend/resource/container/logs_test.go
@@ -15,6 +15,7 @@
 package container
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


src/app/backend/resource/container/logs_test.go
```diff
@@ -47,6 +48,40 @@ var log5 = logs.LogLine{
 	Content:   "log5",
 }
 
+var details *logs.LogDetails
+
+func benchmarkGetLogDetails(lines int, timestamp string, b *testing.B) {
```
floreks: IIRC benchmarks are not executed as part of the test spec by default (they need `-bench` arg). It's a nice addition though. Have you compared old implementation performance of `getLineIndex` with the new one?

eloyekunle: @floreks Yes I compared the performance. The summary is here: https://github.com/kubernetes/dashboard/issues/3900#issuecomment-517395365.

Overall, there's a negligible speedup.

PR #4156, opened 2019-08-01
Show time tooltips for events FirstSeen and LastSeen

eloyekunle:
Also refactored the `kd-date` component to also show tooltips for non-relative timestamp displays.

### DIFF
--- a/src/app/frontend/common/components/date/template.html
+++ b/src/app/frontend/common/components/date/template.html
@@ -14,11 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<span *ngIf="_relative"
-      class="kd-date"
+<span class="kd-date"
       [matTooltip]="date | date:format">
-  {{date | kdRelativeTime}}
-</span>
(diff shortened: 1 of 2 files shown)


src/app/frontend/common/components/date/template.html
```diff
@@ -14,11 +14,7 @@
...
 
-<span *ngIf="_relative"
-      class="kd-date"
+<span class="kd-date"
       [matTooltip]="date | date:format">
-  {{date | kdRelativeTime}}
-</span>
-<span *ngIf="!_relative">
-  {{date | date:format}}
+  {{_relative ? (date | kdRelativeTime) : (date | date)}}
```
maciaszczykm: Shouldn't it be `date:format` at the end?

eloyekunle: `date:format` gives the longer version of the timestamp which is sometimes longer than column displays. Since the longer version is already shown in the tooltip, I think the main display can be the shorter one with `date`.

Issue #4153, opened 2019-08-01
[Parent Issue]: Add e2e Tests


eloyekunle:
##### What would you like to be added
With the addition of the [Cypress](https://cypress.io) test framework in #4002, we can add e2e tests for some of the features that we currently have.

##### Why is this needed
To improve the quality and stability of the project.

smohtadi:
Hi, is this issue still open? I would like to work on this.

eloyekunle:
@smohtadi Yes it is still open. We want to have e2e tests for more components of the dashboard, so this is a parent issue to track this feature.

PR #4161, opened 2019-08-03
CRD - e2e Tests

eloyekunle:
This PR adds e2e tests for CRDs using Cypress.

- [x] CRD List
- [x] CRD Details Page
- [x] CRD Object Details Page
- [x] Update `apiextensions` from `v1beta1` to `v1` 

Ref: #4153

### DIFF
--- a/cypress/integration/crd.spec.js
+++ b/cypress/integration/crd.spec.js
@@ -142,8 +142,13 @@ describe('Custom Resource Definitions', () => {
 
     cy.wait('@deleteObject');
 
-    // browser goes to the crd list page on deletion
-    cy.url().should('include', '/#/customresourcedefinition?namespace=default');
+    cy.go('back'); // to crd detail
+
+    cy.get('kd-crd-object-list').within(() => {
+      cy.get('#zero-state').should('be.visible');
+    });
+
+    cy.go('back'); // to crd list
   });
 
   it('delete crd', () => {
--- a/cypress/integration/crd.spec.js
+++ b/cypress/integration/crd.spec.js
@@ -1,0 +1,167 @@
+describe('Custom Resource Definitions', () => {
+  before(() => {
+    cy.visit('/');
+  });
+
+  it('crd list is initially empty', () => {
+    cy.get('#sidebar-crd').click();
+    cy.url().should('include', '/#/customresourcedefinition');
+
+    cy.get('#zero-state').should('be.visible');
+  });
+
+  it('create a crd', () => {
+    cy.fixture('crd.yaml').then(crd => {
+      cy.createResource({content: crd});
+      cy.reload();
+    });
+  });
+
+  it('crd list should contain the new crd', () => {
+    cy.get('#zero-state').should('not.be.visible');
+    cy.get('kd-crd-list').within(() => {
+      cy.get('mat-row').should('have.length', 1);
+    });
+  });
+
+  it('go to crd details page', () => {
+    cy.get('kd-crd-list').within(() => {
+      cy.get('a')
+        .first()
+        .click();
+    });
+    cy.url().should(
+      'include',
+      '/#/customresourcedefinition/foos.samplecontroller.k8s.io?namespace=default',
+    );
+
+    // loads correctly
+    cy.get('kd-object-meta').within(() => {
+      cy.get('kd-property')
+        .first()
+        .contains('foos.samplecontroller.k8s.io');
+    });
+
+    // has resource information
+    cy.get('#resource-information').within(() => {
+      cy.get('kd-property').should('have.length', 3);
+
+      cy.get('kd-property')
+        .eq(0)
+        .contains('v1alpha1');
+      cy.get('kd-property')
+        .eq(1)
+        .contains('Namespaced');
+      cy.get('kd-property')
+        .eq(2)
+        .contains('samplecontroller.k8s.io');
+    });
+
+    // has accepted names
+    cy.get('#accepted-names').within(() => {
+      cy.get('kd-property').should('have.length', 4);
+
+      cy.get('kd-property')
+        .eq(0)
+        .contains('foos');
+      cy.get('kd-property')
+        .eq(1)
+        .contains('foo');
+      cy.get('kd-property')
+        .eq(2)
+        .contains('Foo');
+      cy.get('kd-property')
+        .eq(3)
+        .contains('FooList');
+    });
+
+    // has empty object section
+    cy.get('kd-crd-object-list').within(() => {
+      cy.get('#zero-state').should('be.visible');
+    });
+
+    // has one version
+    cy.get('kd-crd-versions-list').within(() => {
+      cy.get('mat-row').should('have.length', 1);
+
+      cy.get('mat-cell')
+        .first()
+        .contains('v1alpha1');
+    });
+  });
+
+  it('create a crd object', () => {
+    cy.fixture('example-foo.yaml').then(object => {
+      cy.createResource({content: object, namespace: 'default'});
+      cy.reload();
+    });
+  });
+
+  it('crd objects list should contain the new object', () => {
+    cy.get('kd-crd-object-list').within(() => {
+      cy.get('#zero-state').should('not.be.visible');
+
+      cy.get('mat-row').should('have.length', 1);
+      cy.get('mat-cell')
+        .first()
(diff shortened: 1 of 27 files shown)


cypress/support/commands.js
```diff
@@ -1,3 +1,31 @@
-// ***********************************************
-// Create various custom commands and overwrite existing commands.
-// ***********************************************
+Cypress.Commands.add(
```
floreks: This feels like an API tests, not UI. Isn't it possible to click over the frontend to do that?

cypress/integration/crd.spec.js
```diff
@@ -0,0 +1,177 @@
...
+    });
+
+    cy.fixture('example-foo.yaml').then(object => {
+      cy.createResource({content: object, namespace: 'default'});
+    });
+  });
+
+  it('shows objects if they exist', () => {
+    cy.visit('/#/crd/foos.samplecontroller.k8s.io?namespace=default');
+    cy.wait(500);
```
floreks: Try to avoid this type of `cy.wait` at all cost. It's not reliable. You can use `cy.route` and `cy.wait(alias)` to wait for actual request from the server.

floreks:
We should configure cypress to use typescript and write PO objects to encapsulate some of the page logic into the class. 

Example page object
```typescript
export class LoginPage {
  static getLoginBtn(): Cypress.Chainable<any> {
    return cy.get('#login-button');
  }

  static getLogoutBtn(): Cypress.Chainable<any> {
    return cy.get('#navbar-logout-btn');
  }

  static getLogoutMenuBtn(): Cypress.Chainable<any> {
    return cy.get('#navbar-user-menu');
  }

  static visit(): void {
    cy.visit('/');
  }
}
```

cypress/support/commands.js
eloyekunle: It's not a test. I added it as a utility to be used in other tests where we're not explicitly testing for resource creation.
To test app deployment/creation from the UI, we can add a test for that particular aspect, where we click over the UI to do it, but other cases might be able to use this utility method to test only features that they're concerned with.

floreks: Still, UI e2e are supposed to do their tests by clicking over the UI. If page objects will be created it will be much easier to write this. Also, it's best to write e2e in the form of stories. 

1. Going to deploy from file
2. Deploying crd
3. Going to crd list view
4. Should contain x
5. Delete crd
6. Should show empty list
etc.

This way the test will simulate user actions otherwise it's not much different than unit tests with mocked API responses.

cypress/integration/crd.spec.js
maciaszczykm: Can you change that?

eloyekunle: Yes, I'll fix this soon.

maciaszczykm: Any progress?

eloyekunle: Yes. This is done.

cypress/integration/crd.spec.js
```diff
@@ -211,3 +179,29 @@ describe('crd objects page - valid object', () => {
     cy.url().should('include', '/#/customresourcedefinition?namespace=default');
   });
 });
+
+describe('deletes crd', () => {
+  it('deletes crd from menu dropdown', () => {
+    cy.visit('/#/customresourcedefinition');
```
maciaszczykm: Please use menu entries for navigation as it will cover navigation better. Current test doesn't check if the navigation works.

eloyekunle: Okay, I'll do that.

eloyekunle: Done.

cypress/support/commands.js
eloyekunle: I've tried to rewrite the tests in form of stories.

I also tried to create resources via the input form at `/#/create`, but I couldn't find a good approach.

Using `cypress.type` inputs wrongly-formatted YAML because pressing `enter` on Ace Editor preserves indent, and the YAML file has its own indent so we get something like:
image

Setting the value automatically via the Ace module, (e.g. via below snippet) does not trigger Angular's change detection so the input is not formatted, and the `Upload` button is greyed out.

Snippet:
```js
import * as ace from 'brace';
...
cy.get('ace-editor').then(ele => {
  const editor = ace.edit(ele[0]);
  editor.setValue(content);
  editor.clearSelection();
});
```
Display:
image

cypress/integration/crd.spec.js
```diff
@@ -0,0 +1,167 @@
...
+        .first()
+        .contains('foos.samplecontroller.k8s.io');
+    });
+
+    // has resource information
+    cy.get('[data-cy=resource-information]').within(() => {
+      cy.get('kd-property').should('have.length', 3);
+
+      cy.get('kd-property')
+        .eq(0)
```
maciaszczykm: We cannot assume order here, please use text matching or add IDs to the template.

cypress/support/commands.js
floreks: How about using `setValue` + typing new line to make it run the validation code?

cypress/integration/crd.spec.js
```diff
@@ -0,0 +1,167 @@
...
+      // Namespace
+      cy.get('kd-property')
+        .eq(1)
+        .contains('default');
+    });
+  });
+
+  it('delete crd object', () => {
+    cy.server();
+    cy.route('DELETE', '/api/v1/_raw/foos.samplecontroller.k8s.io/**').as('deleteObject');
```
floreks: It's an API test, not UI. UI should offer functionality to delete it.

src/app/frontend/common/components/list/column/menu/template.html
```diff
@@ -38,15 +38,18 @@
           (click)="isPinned() ? onUnpin() : onPin()"
           i18n>{{ isPinned() ? "Unpin" : "Pin" }}</button>
   <button mat-menu-item
+          data-cy="edit"
```
floreks: Adding framework specific attributes seems a bit weird to me. You should use `id` in such cases.

@maciaszczykm WDYT?

maciaszczykm: I agree.

cypress/support/commands.js
eloyekunle: I tried that, didn't work. The script sees the entire input value as a newline (`\n`) (or whatever else I typed).

src/app/frontend/common/components/list/column/menu/template.html
eloyekunle: Cypress has some best practices here regarding element selection: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements.

That's what I tried to follow.

cypress/integration/crd.spec.js
eloyekunle: It is deleted via the UI. This is only telling Cypress to properly wait for the XHR to complete. The actual deletion occurs in following lines.

This follows your suggestion to use `route` and `alias` for waiting.

src/app/frontend/common/components/list/column/menu/template.html
floreks: Yes, but it is very cypress specific. If we'd like to move to another framework we'll have all this `data-cy` attributes that we do not need. We should always use something generic.

eloyekunle: Okay, so should we rename `data-cy` to something else? Like `kd-test`...

maciaszczykm: No, just use IDs, like `<div id="...">...</div>`.

eloyekunle: Okay, will update.

eloyekunle: Done.

cypress/integration/crd.spec.js
eloyekunle: Done

eloyekunle:
2 more updates:

1. Bumped kubernetes version used in `kind` to `1.16.2`.
2.  Update apiextensions from `v1beta1` to `v1`.

maciaszczykm:
Great, we can stick with it for now. In the future, it would be really nice to use UI for all of the actions.

/lgtm
/approve

PR #4170, opened 2019-08-06
Add `go mod download` in `postinstall`

shu-mutou:
`go mod` often fails downloading modules when `go run` or etc.
`go mod download` is good to download modules surely than `go run`.
To ensure downloading modules, run `go mod download` in `postinstall`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "fix:i18n": "ng xi18n --outFile ../i18n/messages.xlf && xliffmerge",
     "clean": "rm -rf .go_workspace .tmp coverage dist npm-debug.log",
     "postversion": "node aio/scripts/version.js",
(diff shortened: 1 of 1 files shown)


floreks:
Is it actually needed? Modules are downloaded during the build.

shu-mutou:
Yes. `go mod` tries to download modules, but actually it sometimes fails to download.

shu-mutou:
`go mod download` tends to download modules successfully comparing with downloading in `go run` or etc.. 😉

floreks:
Most often when it fails on the CI, it's just an issue on their side. Travis tends to have networking issues from time to time. We can add it and see if it improves anything. From my experience, if the network fails on Travis then it fails for good. Retry usually does not help.

shu-mutou:
OK, I see. Thank you for your sharing experience. I'd like to remove script for retrying. But I would leave `go mod download`, because it seems better for downloading modules than `go run` etc..

PR #4171, opened 2019-08-06
Fix typo: requestes -> requests

kirillDanshin:

PR #4172, opened 2019-08-07
Migrate DOCKER_CLI_EXPERIMENTAL and GOPROXY env from travis

shu-mutou:
To clarify, log and test automatically changes for environment variables in travis, migrate these definitions from repository settings in travis UI to `.travis.yml` under repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,9 @@ cache:
 
 before_install:
   - export NG_CLI_ANALYTICS=ci
+  - export DOCKER_CLI_EXPERIMENTAL=enabled
   - export GO111MODULE=on
+  - export GOPROXY=https://proxy.golang.org
   - export PATH=$PATH:$GOPATH/bin
   - eval "$(gimme 1.12.6)"
 


.travis.yml
```diff
@@ -19,6 +19,8 @@ cache:
 before_install:
   - export NG_CLI_ANALYTICS=ci
   - export GO111MODULE=on
+  - echo $GOPROXY
+  - export $GOPROXY=https://proxy.golang.org
```
jeefy: Remove the `$` when exporting a variable.

Also all examples I've seen setting GOPROXY don't include the protocol (see https://twitter.com/JetstackHQ/status/1158698823910395905 for example)

shu-mutou: I found following log in travis job.
```
Setting environment variables from repository settings
$ export DOCKER_CLI_EXPERIMENTAL=enabled
$ export GOPROXY=proxy.golang.org
```
That may the reason to fail.

shu-mutou: I added "http://", jobs seem to run successfully.

shu-mutou: Failures in other PR are actually as follow:
```
invalid $GOPROXY setting: malformed URL or invalid scheme (must be http, https, file)
```

Schema would be needed, I think.  😉

jeefy: Updated GOPROXY to be https:// and testing against #4171

shu-mutou: Do changes about `repository settings` be tested? If not, it's better to migrate their settings to `.travis.yaml`, I think.

jeefy: Environment settings (currently) we manage within Travis itself (for secrets and whatnot)

This kind of change would be nice to be something we could keep in the `.travis.yml` file BUT I was in the middle of testing the GOPROXY variable change (in Travis) almost at the same time you were submitting the PR. (Great minds think alike!)

I personally think (for now) we should keep the change in Travis (so we can close this PR) and (soon) migrate any non-secret config info out of Travis (itself) and into .travis.yml

Thoughts @shu-mutou ?

shu-mutou: I updated this PR into migration of environment variables for travis. 
We can consider in other PR or issue whether we should set GOPROXY or not. :wink:

floreks:
Any explanation of why `GOPROXY` would be needed?

shu-mutou:
> Any explanation of why `GOPROXY` would be needed?

https://groups.google.com/forum/#!topic/kubernetes-dev/NBjEBHCeBA8
I guess it is effective for downloading modules.

jeefy:
Just for follow up, I've removed those two env vars from the Travis Settings. :) Cheers!

Issue #3945, opened 2019-06-19
Endpoint for health checks


astorath:
##### What would you like to be added
1. Either serve both secure and insecure endpoints simultaneously
2. Or create a separate endpoint for health checks (on separate port)

##### Why is this needed
In order to protect Dashboard from unauthorised access I want to use a sidecar auth proxy. But with current options I have to either bind to 127.0.0.1 address (liveness probe can't be invoked in this case) or expose unauthorised endpoint on 0.0.0.0 (making it possible to bypass proxy).

##### Comments
Maybe there's another way to make port reachable for health checks only?

floreks:
Kubernetes uses default API port as liveness probe. See https://github.com/kubernetes/dashboard/blob/master/aio/deploy/recommended.yaml#L209

I don't really understand your use case. Simple check if API is alive should be enough.

astorath:
@floreks 
Use case is as follows:

1. I have dashboard serving on http port without any authentication
2. Dashboard is bound to 127.0.0.1 (so that noone except for sidecar can connect)
3. Auth proxy (as a sidecar) performs auth and proxies user requests

This way I'm unable to use liveness probe:
- Dashboard do not accept connections from kube
- I have no option to enable second interface (with authentication) to support liveness probe
- Container is built from scratch - no option to run some `exec curl` there also

floreks:
I think you are using some custom Dashboard deployment with your own sidecar container. If you choose not to expose Dashboard on `0.0.0.0` inside the pod then you have to make sure that your auth proxy will do necessary health checks as it is the only part that can access Dashboard. We don't want to run another API inside the container just for health checks.

astorath:
I see, thanks for reply

PR #4185, opened 2019-08-09
Add healthz endpoint

distol:
If you use dashboard in insecure mode and reverse-proxy to secure it – you need to have some endpoint in the dashboard, which can be exposed without any authentication to be used in readiness and liveness probes.

### DIFF
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -175,6 +175,9 @@ func main() {
 	http.Handle("/config", handler.AppHandler(handler.ConfigHandler))
 	http.Handle("/api/sockjs/", handler.CreateAttachHandler("/api/sockjs"))
 	http.Handle("/metrics", prometheus.Handler())
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "OK")
+	})
 
 	// Listen for http or https
 	if servingCerts != nil {


floreks:
Kubernetes already can handle health check and it does not require explicit healthz endpoint to do that. I don't see why we'd need that.

https://github.com/kubernetes/dashboard/blob/master/aio/deploy/alternative.yaml#L205

distol:
@floreks, let me explain:
1. We have some kind of authentication (actually external auth, but that's not important) on ingress controller. So we know the user on ingress and make the access decision there.
2. We use the dashboard in insecure mode, so it works with kubernetes under its own service account (pod service account).
3. But we need to secure connection between ingress and dashboard pod. And for that purpose, we configure the dashboard to listen on 127.0.0.1 and add nginx proxy to the dashboard pod. And this nginx checks client certificate (our ingress makes requests with the certificate, that is signed by known CA).

So. We can't do readiness/liveness on /, because kubelet has no certificate. We need some extra location, which is safe to be exposed without any auth.

From the other hand, this change should not hurt anyone.

floreks:
I have answered to the same issue in here: https://github.com/kubernetes/dashboard/issues/3945

> I think you are using some custom Dashboard deployment with your own sidecar container. If you choose not to expose Dashboard on 0.0.0.0 inside the pod then you have to make sure that your auth proxy will do necessary health checks as it is the only part that can access Dashboard. We don't want to run another API inside the container just for health checks.
>
> Personally, I'd prefer to use network policies to allow only certain services access to Dashboard and expose it normally on 0.0.0.0.

distol:
Oh. Sorry for not checking issues, my fault. Yes – the issue you've mentioned is exactly the same case.

> I think you are using some custom Dashboard deployment with your own sidecar container. If you choose not to expose Dashboard on 0.0.0.0 inside the pod then you have to make sure that your auth proxy will do necessary health checks as it is the only part that can access Dashboard. We don't want to run another API inside the container just for health checks.

The problem is that it's a little bit hard to implement on the proxy side. At least, being quite experienced in Nginx, I don't know the clean and easy way of doing this.

And from the other hand, the `/healthz` endpoint pattern is quite popular in cloud-native apps.
* kube-scheduler and kube-controller-manager both have /healthz endpoint, and special command line option `--authorization-always-allow-paths` (Default: [/healthz]), specifically for the same case.
* Prometheus: https://github.com/prometheus/prometheus/blob/be67b8d460cb30b94de771df24753816443993a0/web/web.go#L366-L372
* Ingress Nginx: https://github.com/kubernetes/ingress-nginx/blob/771fc9fc2ab707be0fe7d3eaf7163b5fcca3a57f/cmd/nginx/flags.go#L96-L99
* Dex: https://github.com/dexidp/dex/blob/master/server/server.go#L301
* Example app from Kelsey Hightower: https://github.com/kelseyhightower/app-healthz

Why not just add this endpoint? It's a really common pattern.

floreks:
Is there a reason why i.e. `/config` endpoint could not be used as health indicator? It basically only returns server time and does not require any authentication. `healthz` endpoint was just never needed because we could easily do a health check using any other exposed endpoint. I understand that some apps might need such endpoint if their internal API needs time to start and they are exposing it while starting. This way `healthz` would sometimes return something like `not ready`. In our case, we do not need anything to start, that's why I don't see a real reason why it would be needed if other exposed endpoint could just as easily be used as health indicators.

distol:
Yes, if `/config` endpoint has no secret information – it can be used.

But two questions arrises immediately:
1. If, at any point, some kind of secret information will be required in /config (for example for a new feature) – what are the chances that the new endpoint (for ex. /secret_config) be created?
2. How will new users find the right solution if they need distinct endpoint for healthz? Shall using of /config be documented as a recommended way?

If we will have `/healthz` endpoint – it's known pattern, but using `/config` IMHO it's just using undocumented capabilities.

> I understand that some apps might need such endpoint if their internal API needs time to start and they are exposing it while starting

That's a little bit different story. Kubernetes has liveness and readiness probes, and they work in a completely different way. I would say that for the dashboard case only one endpoint is enough (/healthz), and it can be used both in readiness and liveness probes.

maciaszczykm:
It would be better if this health endpoint would return the actual status of the app, not only `OK` if the backend is reachable. What if the rest of the app will be in the crashed state? It will be misleading.

maciaszczykm:
Can you improve it?

floreks:
> It would be better if this health endpoint would return the actual status of the app, not only OK if the backend is reachable. What if the rest of the app will be in the crashed state? It will be misleading.

Agree. Returning `200 - OK` at all times without any check feels pointless IMHO. It doesn't improve anything as you can take any other endpoint and simply check if it returns anything at all. The result would be pretty much the same. It returns something, therefore it works. If it can be improved and actually check if "state" of our API, in fact, is `healthy/non-healthy` it would be ok to have.

distol:
Yeah! Totally agree! If there is a way of checking backend internal health – it would be awesome to have in this endpoint. But if we go this way, we may go a bit deeper – the internal state should be distinguished from the overall state:
- If the dashboard has no connection to Kubernetes API (or any other external dependencies) – it's healthy, but not ready. So the ready endpoint should be used in Kubernetes readiness probe, and the pod will be removed from the list of endpoints if it's not ready.
- If the dashboard has some internal problems (some crash, hang, etc.), that could be resolved only by a restart – it's unhealthy. And this kind of endpoint should be used in Kubernetes liveness probe.

The first case (readiness) is especially useful for long starting apps and for environments where there is a high chance of network partitioning. But I'm not sure whether we need it in the dashboard. So IMHO we can go on with one endpoint checking only real internal health (and not the external dependencies).

Speaking about improving current implementation – unfortunately, this requires a much better understanding of backend internals and I've no idea where to start. It's only you guys who can improve this.

Maybe we can use the simplest possible implementation provided, as a starting point?

floreks:
Like I mentioned, I am hesitant to add something that does not improve the application in any way and only returns a hardcoded response. In case of our API, internal issues would be issues with the K8S API as our API is only a proxy here. If there is no connection to the K8S API then it fails to start. I don't imagine any other reason it could fail. I'd even say that in our case liveness probe == readiness probe.

IMHO apps like nginx, etc. should be configurable to accept any endpoint as a health indicator and user should be able to configure what kind of response would mean for him that it is up and running. It should not be required by Dashboard or any other app to add some endpoint (`/healthz`) and always return `OK` solely for the purpose of some other app to be able to read health status.

Kubernetes itself does not have any issues with that. We only need to define `livenessProbe`, because in case of our API when it lives, then it is also ready to be used.
```
livenessProbe:
httpGet:
  scheme: HTTPS
  path: /
  port: 8443
initialDelaySeconds: 30
timeoutSeconds: 30
```

maciaszczykm:
We will track from #4543 since there is no progress here.

PR #4195, opened 2019-08-13
Support 'Scale' subresource for CRDs

### LINKED ISSUE #4167: Support 'Scale' subresource for CRDs
The `Scale` option should be supported CRDs that have the `Scale` subresource enabled.

eloyekunle:
Backend implementation via `k8s.io/client-go/scale` currently unable to work with cluster-scoped resources. Tracking issue: https://github.com/kubernetes/kubernetes/issues/81380

Closes #4167

### DIFF
--- a/src/app/backend/resource/customresourcedefinition/objects.go
+++ b/src/app/backend/resource/customresourcedefinition/objects.go
@@ -29,8 +29,8 @@ import (
 
 // CustomResourceObject represents a custom resource object.
 type CustomResourceObject struct {
-	TypeMeta   metav1.TypeMeta   `json:"typeMeta"`
-	ObjectMeta metav1.ObjectMeta `json:"objectMeta"`
+	TypeMeta   api.TypeMeta   `json:"typeMeta"`
+	ObjectMeta api.ObjectMeta `json:"objectMeta"`
 }
 
 func (r *CustomResourceObject) UnmarshalJSON(data []byte) error {
@@ -44,8 +44,8 @@ func (r *CustomResourceObject) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	r.TypeMeta = tempStruct.TypeMeta
-	r.ObjectMeta = tempStruct.ObjectMeta
+	r.TypeMeta = api.NewTypeMeta(api.ResourceKind(tempStruct.TypeMeta.Kind))
+	r.ObjectMeta = api.NewObjectMeta(tempStruct.ObjectMeta)
 	return nil
 }
 
@@ -125,7 +125,7 @@ func GetCustomResourceObjectList(client apiextensionsclientset.Interface, config
 	list.ListMeta = api.ListMeta{TotalItems: filteredTotal}
 
 	for i := range list.Items {
-		replaceCRDObjectKind(&list.Items[i], customResourceDefinition.Name)
+		toCRDObject(&list.Items[i], customResourceDefinition)
 	}
 
 	return list, nil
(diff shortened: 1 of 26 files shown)


eloyekunle:
This is ready for review, PTAL.

/assign @jeefy @maciaszczykm @floreks

maciaszczykm:
Please rebase.

eloyekunle:
Rebased.

src/app/backend/resource/customresourcedefinition/objects.go
```diff
@@ -31,6 +31,9 @@ import (
 type CustomResourceObject struct {
 	TypeMeta   metav1.TypeMeta   `json:"typeMeta"`
 	ObjectMeta metav1.ObjectMeta `json:"objectMeta"`
+
+	// Represents whether or not the CRD's 'scale' subresource is enabled.
+	IsScalable bool `json:"isScalable"`
```
floreks: s/IsScalable/Scalable

I'd change that everywhere. `scalable` itself already suggests that it is a boolean value.

Also maybe instead of adding a new `scalable` information only to CRD, we could maybe extend `TypeMeta` information with `scalable` variable. This would simplify frontend part a lot.

@maciaszczykm WDYT?

src/app/frontend/common/components/list/column/menu/component.ts
```diff
@@ -63,6 +64,10 @@ export class MenuComponent implements ActionColumn {
     this.typeMeta = typeMeta;
   }
 
+  setIsScalable(isScalable: boolean): void {
```
floreks: s/IsScalable/scalable as mentioned above.

We can use setter/getter methods here.

```typescript
set scalable(...)

get scalable(): boolean
```

Usage will be more straightforward then.

src/app/backend/resource/customresourcedefinition/objects.go
maciaszczykm: Fine for me, extending metadata is always good when there is opportunity (all resources have the same property).

eloyekunle: I agree that extending `TypeMeta` simplifies things. I didn't do that originally because I thought the fields in `TypeMeta` should always correspond to the main k8s representation.

floreks: They do not have to be the same. This is our internal API used by the frontend. If it is useful for us, then it can be added.

eloyekunle: I'll work on extending the field today.

src/app/frontend/common/components/list/column/menu/component.ts
eloyekunle: We don't need this anymore, because we're embedding the value in `TypeMeta`.

src/app/backend/resource/customresourcedefinition/objects.go
eloyekunle: Done. Should I move [this implementation](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/list/column/menu/component.ts#L34) of `scalableResources` for other resources to the backend now?

src/app/frontend/common/components/actionbars/scaledefault/component.ts
```diff
@@ -49,4 +49,11 @@ export class ScaleDefaultActionbar implements OnInit, OnDestroy {
     this._unsubscribe.next();
     this._unsubscribe.complete();
   }
+
+  scalable(): boolean {
+    if (this.resourceMeta.typeMeta.scalable !== undefined) {
```
floreks: Why return true if scalable is undefined?

eloyekunle: It has to do with [this](https://github.com/kubernetes/dashboard/pull/4195#discussion_r317648007). For deployments and other scalable resources, we're not yet setting this field from the backend.

floreks: Add comment with some todo here that backend should be adapted and this should be changed.

I'd also change it to `return this.resourceMeta.typeMeta.scalable === undefined || this.resourceMeta.typeMeta.scalable`.

eloyekunle: Done :+1:

eloyekunle: I've moved the `scalableResources` implementation to the backend for other resources.

maciaszczykm:
Tested locally.

/lgtm
/approve

Issue #4128, opened 2019-07-26
sh: ./dist/amd64/dashboard: not execute binary file


smartkuk:
Hi again...

When I execute ```npm run start:prod``` command , and the standard output log showed that the message ```sh: ./dist/amd64/dashboard: not execute binary file``` seemed to be a hint, but the binary file had the right to execute.

I want to contribute to translation work, but I can't see the result.
I cloned kubernetes/dashboard today again for resolving this issue.
The process I did is as the follows.
I dont know why... ***Please help***

### ENV

Minikube

```shell
$ minikube version
minikube version: v1.2.0
```

Kubernetes CLI

```shell
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.4", GitCommit:"c27b913fddd1a6c480c229191a087698aa92f0b1", GitTreeState:"clean", BuildDate:"2019-03-01T23:34:27Z", GoVersion:"go1.12", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:32:14Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

OS

```shell
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.5
BuildVersion:	18F132
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 10.14.5 (18F132)
      Kernel Version: Darwin 18.6.0
      Boot Volume: 무제
      Boot Mode: Normal
      Computer Name: KimJeongkuk의 MacBook Pro
      User Name: KimJeongkuk (smartkuk)
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 2:51
```

Npm & Node

```shell
$ npm version
{
  'kubernetes-dashboard': '2.0.0-beta2',
  npm: '6.9.0',
  ares: '1.15.0',
  brotli: '1.0.7',
  cldr: '35.1',
  http_parser: '2.8.0',
  icu: '64.2',
  llhttp: '1.1.4',
  modules: '72',
  napi: '4',
  nghttp2: '1.38.0',
  node: '12.6.0',
  openssl: '1.1.1c',
[...]

PR #4235, opened 2019-08-29
Fix prod build for macOS

maciaszczykm:
`GOOS` is set automatically, no need to block it only for `linux`.


shu-mutou:
closes #4128

PR #4239, opened 2019-08-29
Remove redundant settings in deploy configs

mapshen:
* ClusterRoleBinding is cluster scoped, not bound to a namespace
* Secret kubernetes-dashboard-certs is not used when --auto-generate-certificates not set

### DIFF
--- a/aio/deploy/alternative.yaml
+++ b/aio/deploy/alternative.yaml
@@ -50,17 +50,6 @@ kind: Secret
 metadata:
   labels:
     k8s-app: kubernetes-dashboard
-  name: kubernetes-dashboard-certs
-  namespace: kubernetes-dashboard
-type: Opaque
-
----
-
-apiVersion: v1
-kind: Secret
-metadata:
-  labels:
-    k8s-app: kubernetes-dashboard
   name: kubernetes-dashboard-csrf
   namespace: kubernetes-dashboard
 type: Opaque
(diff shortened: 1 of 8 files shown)


floreks:
I'd remove the cert, but keep permissions. This way user won't have to remember about granting more privileges if he decides to use his own certs and creates the secret.

mapshen:
> I'd remove the cert, but keep permissions. This way user won't have to remember about granting more privileges if he decides to use his own certs and creates the secret.

Thanks for reviewing my PR @floreks !

My understanding is that if the user would like to use the cert, they would use the `recommended` one. If they choose to use the one in `alternative` or `head`, that means they are aware of the cert won't be used by default unless they make some changes on top. Typically, this kind of users knows what they are doing and can figure that out.

floreks:
I am saying that we can remove a secret resource, but keep the privileges of Dashboard's SA.

mapshen:
> I am saying that we can remove a secret resource, but keep the privileges of Dashboard's SA.

Yes, I get it. What I do here is to follow the Principle of Least Privilege. If a user uses his own certs and creates the secret, he would have to either recreate the role, granting access to the secret or create a new role and assign it to the SA, which is cleaner, in my opinion. 

That said, I think this is not a common use case, and for that, the user should go for `recommended`. But let me know if you still want to keep it, I could make the change.

floreks:
I wouldn't want to do that if it would be a security issue, but in this case, we grant explicit privilege for secret with that name. If it doesn't exist then there is no issue. If someone decides to create it and use their own certs, then there is 1 less thing to remember about.

I don't see any downside here of keeping this one. If you see any potential issues then share it, please.

mapshen:
I can't think of any either so far, so I added that back :)

Let me know if there are any other issues.

PR #4258, opened 2019-09-05
Fix build script

shu-mutou:
In MacOS, `date` command does not have `%N` option.
We do not need elapsed time in nanosecond.

Related: #4128


maciaszczykm:
/lgtm
/approve

Thanks!

Issue #4008, opened 2019-07-03
Move wiki documents under @docs/` and update them


shu-mutou:
##### Reasons

1. Due to instruction by k8s org, only maintainer can edit wiki pages since this year.
2. Therefore, contributor can not propose documentation updates by PR.
3. Wiki contents are out of dates. Partially fairly.
4. Before release 2.0.0, we need to update docs.

##### TODOs

- [x] for `2.0.0 or later`
    - [x] Reorganize `docs/`, create indexes and blank pages. #4009
    - [x] Fill blank pages with new descriptions or add new pages if needed. (divided work)
- [x] for `1.10.1 or before`
    After migration to `docs/`, revert wiki for <=1.10.1.
    - [x] Revert wiki revision to 1.10.1 (to the end of last year). (maintainer only)
    - [x] Add notice for `<=1.10.1` on wiki pages. (maintainer only)
    - [x] Add link to `docs/` for `>=2.0.0`" on wiki pages. (maintainer only)
    - Let's consider migration of docs for `<=1.10.1` later.
    Old documents are all removed.

shu-mutou:
Wiki pages to be migrated and assignee:
- [ ] Common
  - [ ] FAQ
  - [ ] Roadmap
  - [ ] Dashboard arguments
- [ ] User Guide
  - [ ] Installation
  - [ ] Certificate management
  - [ ] Accessing Dashboard
    - [ ] 1.7.X and above
    - [ ] 1.6.X and below
  - [ ] Access control
    - [ ] Creating sample user
  - [ ] Integrations
  - [ ] Labels
- [ ] Development Guide
  - [x] Getting started @shu-mutou 
  - [ ] Release procedures
  - [ ] Dependency management
  - [ ] Architecture
  - [ ] Code conventions
  - [ ] Text conventions
  - [x] Internationalization @shu-mutou

maciaszczykm:
Once the docs are moved we should also update docs button on `About` page in the app.

maciaszczykm:
We should close the Wikis once this will be done.

maciaszczykm:
Arch image should be updated to replace Heapster.

maciaszczykm:
Please also add https://github.com/kubernetes/dashboard/pull/3504#issuecomment-501353220 info to the docs. Can be placed in FAQ or some dev doc.

maciaszczykm:
@jeefy Would be nice if you would describe dev process with the metrics. Some manual, tips etc.

[...]

PR #4276, opened 2019-09-11
Add docs for plugin development

ajatprabha:


docs/plugins/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Registering Plugin
+
+Once the custom CRD is registered we can now create [instances](../../aio/test-resources/plugin-test.yml) of the CRD which will hold the spec for plugin.
+
+```
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/aio/test-resources/plugin-test.yml
+```
+
+> Note: The backend reads the built plugin from a ConfigMap and we new to create that also.
```
feloy: The end of the phrase is not clear

ajatprabha: Thank you @feloy for reviewing, fixed it.

shu-mutou:
Could you add link for this into https://github.com/kubernetes/dashboard/blob/master/docs/README.md ?

shu-mutou:
Related to #4008.

ajatprabha:
@shu-mutou Should this be mentioned under a separate category or under Developer Guide?

shu-mutou:
@ajatprabha I think this should be described for third party that includes both of user and developer.

docs/README.md
```diff
@@ -28,5 +28,7 @@
 * [Text conventions](developer/text-conventions.md)
 * [Internationalization](developer/internationalization.md)
 
+## [Plugin Guide](plugins/README.md)
```
maciaszczykm: Put it in the dev guide.

ajatprabha: Done

maciaszczykm:
/lgtm
/approve

It's good as a first step. I'd like to have more infor about how to write plugins, what we can use,  even some samples with screenshots showing the active plugin in the UI.

docs/plugins/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Once the custom CRD is registered we can now create [instances](../../aio/test-resources/plugin-test.yml) of the CRD which will hold the spec for plugin.
+
+```
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/aio/test-resources/plugin-test.yml
+```
+
+> Note: The backend reads the compiled plugin source from a ConfigMap and we need to create that also.
+
+### Creating ConfigMap
```
xm2: kubernetes configmap has size limitation, (<=1MB), so the compiled plugin javascript file should not exceed 1MB. That is the limitation of this plugin solution, right?

ajatprabha: Yes, you're right. We can look into other ways to obtain the compiled plugin file instead of just configmap. We wanted to keep dashboard stateless and using configmaps was the easiest way to tackle this for now.

PR #4138, opened 2019-07-29
Specified Linux as OS in nodeSelector, to make compatible with AKS cluster with Windows nodes

greggbjensen:
Updated dashboard to specify Linux as OS in nodeSelector, to make it compatible with AKS cluster that has Windows nodes.

```yaml
    spec:
      nodeSelector:
        "beta.kubernetes.io/os": linux
```

PR #4288, opened 2019-09-13
Apply Security hardening

### LINKED ISSUE #4287: Apply Docker/Kubernetes Security Best Practices
Reduce technical debt by applying  security best practices:

Low hanging fruit:

- [ ] Run as non-root (& disallow privilege escalation)
- [ ] Run with the default seccomp profile (i.e.  `runtime/default`)
- [ ] ReadOnlyRootFilesystem
- [ ] Avoid unnecessary HostPath volumes
- [ ] Don't mount service account token

More advanced:

- [ ] Reduce dependencies in the base image (https://github.com/kubernetes/kubernetes/issues/40248 https://github.com/kubernetes/kubernetes/issues/70249)
- [ ] Custom seccomp profile
- [ ] Custom AppArmor profile
- [ ] Drop unneeded capabilities

This is a copy of the issue create on the Kubernetes upstream project:
https://github.com/kubernetes/kubernetes/issues/82647
Which was inspired on @tallclair `s:
https://github.com/kubernetes/kubernetes/issues/38541

/cc @tallclair 
/sig auth

pjbgf:
Hardens the dashboard add-on by implementing some security best practices, such as:
- Run as non-root & disallow privilege escalation
- ReadOnlyRootFilesystem
- Replaces deprecated seccomp
- Reduces dependencies in the base image (https://github.com/kubernetes/kubernetes/issues/40248 https://github.com/kubernetes/kubernetes/issues/70249)

Fixes #4287

/assign @tallclair
/sig auth

### DIFF
--- a/aio/Dockerfile
+++ b/aio/Dockerfile
@@ -17,7 +17,7 @@
 
 # Scratch can be used as the base image because the backend is compiled to include all
 # its dependencies.
-FROM gcr.io/distroless/static
+FROM scratch
 
 # Add all files from current working directory to the root of the image, i.e., copy dist directory
 # layout to the root directory.
--- a/aio/Dockerfile
+++ b/aio/Dockerfile
@@ -17,7 +17,7 @@
 
 # Scratch can be used as the base image because the backend is compiled to include all
 # its dependencies.
-FROM scratch
+FROM gcr.io/distroless/static
 
 # Add all files from current working directory to the root of the image, i.e., copy dist directory
 # layout to the root directory.
--- a/aio/deploy/alternative.yaml
+++ b/aio/deploy/alternative.yaml
@@ -257,6 +257,8 @@ spec:
     metadata:
       labels:
         k8s-app: dashboard-metrics-scraper
+      annotations:
+        seccomp.security.alpha.kubernetes.io/pod: 'runtime/default'
     spec:
       containers:
         - name: dashboard-metrics-scraper
--- a/aio/deploy/alternative.yaml
+++ b/aio/deploy/alternative.yaml
@@ -208,6 +208,11 @@ spec:
               port: 9090
             initialDelaySeconds: 30
             timeoutSeconds: 30
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            runAsUser: 1001
+            runAsGroup: 2001
       volumes:
         - name: tmp-volume
           emptyDir: {}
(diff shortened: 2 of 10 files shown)


maciaszczykm:
LGTM

/assign @jeefy @floreks

aio/Dockerfile
```diff
@@ -17,7 +17,7 @@
 
 # Scratch can be used as the base image because the backend is compiled to include all
 # its dependencies.
-FROM scratch
+FROM gcr.io/distroless/static
```
floreks: Why is it better than a `scratch` image? We don't really need extra stuff that is available inside this image.

pjbgf: @floreks we are aligning all k8s "system containers" with the same base image/version, to make it easier to monitor and handle CVEs across the Estate. More info: https://github.com/kubernetes/kubernetes/issues/40248
https://github.com/kubernetes/kubernetes/issues/70249

@tallclair anything else you may want to add?

floreks: @pjbgf if there is no logic behind changing the base image other than the desire to align all images, I don't really want to change it. Considering the security I don't see an actual advantage of changing it.

As mentioned in kubernetes/kubernetes#40248:
> Dynamically loaded libraries: some images (e.g. kubeproxy, fluentd) rely on more than a statically built binary, so basing off something like scratch is non-trivial.

In our case, Dockerfile only copies statically built binary and runs it. There are no additional steps, so we actually rely only on a single binary and compiled frontend code that is served by the binary. This does not apply to our case.

jeefy: Second @floreks -- Our image footprint is already as minimal as possible. Inheriting from distroless isn't required.

maciaszczykm:
@pjbgf Would you be willing also to do change like in https://github.com/kubernetes/dashboard/pull/4138?

aio/Dockerfile
pjbgf: Hey @floreks @jeefy, thank you for assessing my changes. I have now removed the rebasing of the dashboard. 

Given that this was a [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md) decision, is it worth creating a issue to discuss this further?

pjbgf:
@maciaszczykm sure, sending a new commit in a few minutes.

aio/Dockerfile
tallclair: To add a bit of context on this:
distroless/static doesn't add any executable dependencies to the image. It only adds configuration files (and ca-certs) that are required for certain go functions, even for static binaries. For example, see this issue: https://github.com/kubernetes/kubernetes/issues/69195

It's possible that these networking functions aren't required for the dashboard, but IMO there's basically no reason not to include them, and doing so might avoid some confusing bugs if those functions are called in the future.

jeefy: Thanks for the context! With that, background I'd totally accept changing the base image to `distroless/static`

pjbgf: @tallclair thank you for adding the context.

@jeefy thank you for reviewing this. I have just submitted another PR to make that change.

Issue #4255, opened 2019-09-04
UI Feature: Create or Update Secret From Env File


ryanelian:
##### What would you like to be added

![image](https://user-images.githubusercontent.com/8892187/64242916-98c5f200-cf30-11e9-8535-9fc225528faf.png)

Another tab should be added next to the "Create From Form" tab labeled "Upsert Secret From Env File" or whatever.

This tab has two forms:

1. Text Box labeled "Secret Name"

2. Dropdown labeled "Namespace" similar to the one in "Create From Form"

3. A multi-line source code text area for pasting secret in env file format:

```env
NAME=martin
GENDER=male
```

-----

That is the basic form, but a more sophisticated form may also be created:

1. The name text box instead is a dropdown, which allows selecting existing secret or creating a new one. (Similar to the one in "Create From Form")

2. When the dropdown item is selected, the text area should be populated with decoded secret from the actual base64 encoded secret.

##### Why is this needed

Enhances and solves the non-user-friendly text-based secret editing in Kubernetes Dashboard.

##### Comments

Related issues:

https://github.com/kubernetes/dashboard/issues/2782

https://github.com/kubernetes/dashboard/issues/1593

https://stackoverflow.com/questions/51419102/is-it-possible-to-source-a-env-file-to-create-kubernetes-secrets

Issue #1600, opened 2017-01-27
Full Support for Object Creation


ianlewis:
This issue is for tracking full support for object creation without having to upload YAML files. Individual object types may be supported via other issues.

- [x] Deployment
- [ ] Add service to existing deployment (#1569)
- [ ] Headless service
- [ ] Replica Set
- [ ] Stateful Set (#1527)
- [ ] Daemon Set
- [ ] ConfigMap
- [ ] Secret (#1593)
- [ ] Namespace
- [ ] PersistentVolume (#1598)
- [ ] PersistentVolumeClaim (#1598)

floreks:
I'd also add ThirdPartyResource to the list. I can start initial work on that and introduce a way to easily add new object creation pages.

bryk:
Are we talking about custom workflows for individual objects or handling this with YAML-generic editors?

floreks:
Custom forms for different objects. YAML-like editor may be hard to use for beginners.

floreks:
@IanLewis @bryk @maciaszczykm @rf232 

I'm thinking where to start here. What do you think about introducing wizard-like style of object creation? 

Flow could be as follow:
1. Choose whether to create object from yaml or use custom wizard.
2. Assuming that user have chosen our custom object creator he is presented with list of supported resources + ThirdPartyResources + Object instances for TPRs dynamically loaded if there are any TPR templates already in the cluster.
3. For most resources we just create wizard step with custom fields. The tricky part is TPR object creation. We have to provide some kind of structure creator with types. User can provide multiple custom fields with different types (int, string, array, object) that will be part of TPR object instance.

I will work on some sketches and think about how to code it but we can discuss do you think about this approach.

floreks:
@kubernetes/dashboard-maintainers 

I've prepared initial mockups for object creator wizard. It's very raw so I expect some changes, but we have to start somewhere.

Here is a [PDF](https://drive.google.com/open?id=0B4DGVEQHPRToMm5ydWU4Uk8xNlk) (there are some clickable fields but you have to download it) and if someone wants to adjust some things there is also my [balsamiq project file](https://drive.google.com/open?id=0B4DGVEQHPRToell0d0ItdjAxT1E).

[...]

PR #4290, opened 2019-09-15
Upsert secret

anshulahuja98:
Basic form for issue #4255 

image

### DIFF
--- a/i18n/fr/messages.fr.xlf
+++ b/i18n/fr/messages.fr.xlf
@@ -1147,7 +1147,7 @@
         </context-group>
       </trans-unit>
       <trans-unit id="14a5136cbdeef6ed97abfd58112f979f20f619f1" datatype="html">
-        <source><x id="INTERPOLATION" equiv-text="{{ isPinned() ? &quot;Détacher&quot; : &quot;Épingler&quot; }}"/></source>
+        <source><x id="INTERPOLATION" equiv-text="{{ isPinned() ? &quot;Unpin&quot; : &quot;Pin&quot; }}"/></source>
         <target><x id="INTERPOLATION" equiv-text="{{ isPinned() ? &quot;Détacher&quot; : &quot;Épingler&quot; }}"/></target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/common/components/list/column/menu/template.html</context>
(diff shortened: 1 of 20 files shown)


floreks:
Few things:
1. Upsert is a very technical word. I wouldn't use that for the UI.
2. We don't really want to end up with tab per every resource. I'd rather want to see some kind of combobox to switch the resource type on already existing creation form and then show proper inputs to create that resource.

anshulahuja98:
@floreks 
We can switch from using `Upsert`, I actually initially followed the terminology as mentioned in Issue #4255. Please suggest a suitable naming scheme for the same, I'll make the required changes.

This tab isn't for every resource, this is just an independent form to create secrets in a given namespace.
I probably didn't understand what exactly you meant by the second point.

floreks:
I mean that we should use a `create from form` tab and add there some kind of selector to create resource this way. This is basically `create a secret from form` option. We can have a single tab + selector to create: App, Secret, Pod, Role, etc. I wouldn't add separate tab just to create new resource type.

anshulahuja98:
Okay, I think I got it.
So you basically want to make changes to the current `create from form` option and add the method to create secrets using the method of just pasting the secret in plain text apart from the existing version which takes base64 encoded secrets.

This would just require changes in the CreateSecret dialog right?

anshulahuja98:
Can we perhaps update the issue #4255 to reflect on what exactly is required.

floreks:
> Okay, I think I got it.
So you basically want to make changes to the current create from form option and add the method to create secrets using the method of just pasting the secret in plain text apart from the existing version which takes base64 encoded secrets.
>
> This would just require changes in the CreateSecret dialog right?

No. When you open a `create from form` tab, there should be additional selector/combobox with options i.e. `Application|Secret|Role`. When you select a `Secret` it basically shows fields from your new tab `Upsert Secrets`. When you switch back to `Application` it shows the current status, etc.

maciaszczykm:
We cannot accept it in the current form. It would have to be moved to the "create from form" tab as @floreks pointed out. Please reopen if you are willing to make such changes.  We will track from https://github.com/kubernetes/dashboard/issues/1600.

anshulahuja98:
Okay!
I'll have a try at it again and let you know.

PR #4292, opened 2019-09-16
Add node selectors for linux

pjbgf:
Add node selectors to support clustes with windows nodes.

A copy of #4138 given that that PR does not have a CLA signed @maciaszczykm asked me to make the change as a new PR.


jeefy:
Travis job flaked because of jasmine, restarting.

PR #4301, opened 2019-09-17
Change base image to distroless/static

### LINKED ISSUE #4287: Apply Docker/Kubernetes Security Best Practices

pjbgf:
Replaces the base image with distroless/static. This is a continuation of the PR #4288 

Context around this change can be found on its [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md) and on upstream issue: kubernetes/kubernetes#70249

Fixes #4287

### DIFF
--- a/aio/Dockerfile
+++ b/aio/Dockerfile
@@ -17,7 +17,7 @@
 
 # Scratch can be used as the base image because the backend is compiled to include all
 # its dependencies.
-FROM scratch
+FROM gcr.io/distroless/static
 
 # Add all files from current working directory to the root of the image, i.e., copy dist directory
 # layout to the root directory.


pjbgf:
@jeefy here's the PR that implements the base image as we agreed on #4288

jeefy:
/approve
/lgtm
/hold
(For discussion)

@maciaszczykm @floreks See https://github.com/kubernetes/dashboard/pull/4288#discussion_r325288164 for context along with the KEP linked above. With all that, it does make sense to bring Dashboard in line with the rest of the project and it's low impact for us. 

If one/both agree, just remove the hold. :)

maciaszczykm:
+0.5 from my side, we can use configs but I still feel it is not required atm.

floreks:
To be honest I don't feel comfortable changing the base image just for the sake of changing it. If it ain't broken don't try to fix it. We don't need any additional dependencies + scratch image does not add any new layers and in my opinion, fewer dependencies equal better security. Even if it's only certificate/networking stuff that for the time being is considered secure.

If we'll ever need it, we can change it then. For now, I'd leave it as is.

maciaszczykm:
We'll reopen when/if we'll adding some dependency to our image.

jchauncey:
We are having security scans flagged because the image is created using root as the default executing user. Moving to either distroless/static:nonroot or copy /etc/passwd from a builder image like the following:

```Dockerfile
FROM ubuntu:latest
RUN useradd -u 10001 scratchuser
FROM scratch
COPY dosomething /dosomething
COPY --from=0 /etc/passwd /etc/passwd
USER scratchuser
```

PR #4308, opened 2019-09-19
Install `code-generator` into development container

shu-mutou:
To enable `npm run check:codegen`, download `code-generator`.


maciaszczykm:
I think we can also make you an owner of this Dockerfile. Send us a PR and I will approve.

Issue #4327, opened 2019-09-24
Add support for Kubernetes v1.16


LQBing:
##### Environment

```
Installation method: kubectl 
Kubernetes version: v1.16.0
Dashboard version: kubernetesui/dashboard:v2.0.0-beta4
Operating system: ubuntu 18.04.1
Node.js version ('node --version' output): n/a
Go version ('go version' output): n/a
```

##### Steps to reproduce

Choose Workloads -> Deployments -> a deployment, click 'edit'. 
![image](https://user-images.githubusercontent.com/6857508/65488550-fc15c500-dedb-11e9-9af7-1c9ac3b4f883.png)

##### Observed result

The deployment yaml content not load, and I find there is a 404 request in the browser dev panel.
##### Expected result

It need display deployment resource yaml content.

anshulahuja98:
Can you perhaps also share the console output as well?

RnkeZ:
Hi, I have the same problem here is the console output.
![image](https://user-images.githubusercontent.com/1742984/65542443-7ad82580-df0f-11e9-949a-cb727e1f2d4b.png)

anshulahuja98:
I replicated your dev environment. 
The issue seems to be with using Kubernetes version: v1.16.0
That's a fairly recent release, not sure if we should worry about that version as of now. 
The current dashboard release v2.0.0-beta4 only claims to support till 1.15 k8s.
And this feature works fine on 1.15

/awaiting-more-evidence

RnkeZ:
Tkanks for the info.
Do we have ETA on v1.16.0 support?

anshulahuja98:
I cannot comment on that
@maciaszczykm @floreks can perhaps give better info on this.

floreks:
It all depends on client-go updating their code to support v1.16.

zerkms:
Client go **DOES** support v1.16

floreks:
@zerkms I don't see any official release that would support `v1.16` on their github page: https://github.com/kubernetes/client-go#compatibility-matrix

LQBing:
![image](https://user-images.githubusercontent.com/6857508/65571613-e9f95c80-df97-11e9-814a-745df0983be2.png)

![image](https://user-images.githubusercontent.com/6857508/65573234-e2888200-df9c-11e9-9f5a-4b90a2991edd.png)

@anshulahuja98 here is the console output and kubernetes-dashboard pod log.

floreks:
[...]

PR #4346, opened 2019-09-26
Update Kubernetes deps to v1.16

maciaszczykm:
Closes https://github.com/kubernetes/dashboard/issues/4327.

Using https://github.com/kubernetes/client-go/releases/tag/kubernetes-1.16.0.

Changes impacting our code:
- https://github.com/kubernetes/kubernetes/commit/49f50484b898f55c1db1e25b2a9c2bb76fb45a5d#diff-081b6d43e1a4dd91481a666cdb8f82b5
- https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#deprecations-and-removals


floreks:
Edit for resources that the user has access to has to work, as it works with v1.15 now. Last issue exists on the current master I think, so we can create an issue for that and fix later.

shu-mutou:
> Edit for resources that the user has access to has to work, as it works with v1.15 now. Last issue exists on the current master I think, so we can create an issue for that and fix later.

The last one is described in Issue #4370.

PR #4353, opened 2019-09-27
Add labels key to i18n/ja/OWNERS

atoato88:
This PR adds `labels` item to `i18n/ja/OWNERS` for automatic `language/ja` labeling.

### DIFF
--- a/i18n/ja/OWNERS
+++ b/i18n/ja/OWNERS
@@ -2,3 +2,6 @@ approvers:
   - shu-mutou
   - atoato88
 
+labels:
+- language/ja
+


i18n/ja/OWNERS
```diff
@@ -2,3 +2,6 @@ approvers:
   - shu-mutou
   - atoato88
 
+labels:
+- language/ja
```
floreks: Can you add proper labels to OWNERS files in other i18n directories?

atoato88: Thank you for comment.
I will do it, after confirm that "language/ja" is attached automatically by other PR.

PR #4354, opened 2019-09-27
Fix typo in i18n/ja/messages.ja.xlf

atoato88:
This commit fixes typo in `i18n/ja/messages.ja.xlf`.

/assign @shu-mutou


shu-mutou:
/lgtm
Thank you so much!! We could confirm that `language/xx` label is useful for watching updates of `messages.xx.xlf` files, so we can improve i18n workflow much more. At last, let's spread this method into other i18n teams and describe about it in documentation.

Issue #3160, opened 2018-07-20
Allow the user to search for labels


Vargsyn:
Is it possible to search for labels in the dashboard? When i tried it last time it didn't seem to work, so if not it would be nice to search for elements containing labels, similiar to kubectl (see [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#api) for reference).

Dashboard version: 1.8.3
Kubernetes version: 1.10.3

pmpfr:
There's also demand for this on [stackoverflow](https://stackoverflow.com/questions/49700983/how-to-filter-pods-in-kubernetes-dashboard-using-labels).

maciaszczykm:
This should also cover application support (https://github.com/kubernetes/dashboard/issues/1599) as apps are connected with labels. The application CRDs can be viewed already, user can even pin them to the menu.

mumrau:
Has anything been started regarding this topic?

maciaszczykm:
As far as I know, no. We do not have enough resources and this is not the highest priority. If you are willing to work on this we will help you.

nfrmtkr:
I also vote for this feature.

nishafnaeem:
+1

avanish23:
I have opened PR #9572 where we can filter pods based on labels, annotations, and status. 
I will expand its horizon with other workload types like Deployments, Daemon sets, etc. 
However, I believe this issue is about supporting it from the dashboard search bar. 
I will take this up as I have more bandwidth. 
/assign

pmpfr:
Thanks @avanish23 !  FWIW, when I commented on this in 2019 it's because I wanted to be able to filter programatically rather than through the UI, so your PR is exactly the sort of thing I hoped for 👍

PR #4365, opened 2019-10-01
Support label based filtering to search for resources

anshulahuja98:
Work towards #3160 
Disscussions>slack

### DIFF
--- a/src/app/backend/resource/dataselect/dataselect.go
+++ b/src/app/backend/resource/dataselect/dataselect.go
@@ -109,7 +109,7 @@ func (self *DataSelector) Filter() *DataSelector {
 	for _, c := range self.GenericDataList {
 		for _, filterBy := range self.DataSelectQuery.FilterQuery.FilterByList {
 			v := c.GetProperty(filterBy.Property)
-			if v == nil || v.Contains(filterBy.Value) {
+			if v.Contains(filterBy.Value){
         filteredList = append(filteredList, c)
       }
 		}
--- a/src/app/backend/resource/dataselect/dataselect.go
+++ b/src/app/backend/resource/dataselect/dataselect.go
@@ -107,16 +107,11 @@ func (self *DataSelector) Filter() *DataSelector {
 	filteredList := []DataCell{}
 
 	for _, c := range self.GenericDataList {
-		matches := true
 		for _, filterBy := range self.DataSelectQuery.FilterQuery.FilterByList {
 			v := c.GetProperty(filterBy.Property)
-			if v == nil || !v.Contains(filterBy.Value) {
-				matches = false
-				break
-			}
-		}
-		if matches {
-			filteredList = append(filteredList, c)
+			if v == nil || v.Contains(filterBy.Value) {
+        filteredList = append(filteredList, c)
+      }
 		}
 	}
 
--- a/src/app/backend/resource/persistentvolumeclaim/common.go
+++ b/src/app/backend/resource/persistentvolumeclaim/common.go
@@ -15,7 +15,8 @@
 package persistentvolumeclaim
 
 import (
-	"log"
+  "encoding/json"
+  "log"
 	"strings"
 
 	"github.com/kubernetes/dashboard/src/app/backend/errors"
@@ -94,7 +95,10 @@ func (self PersistentVolumeClaimCell) GetProperty(name dataselect.PropertyName) 
 		return dataselect.StdComparableTime(self.ObjectMeta.CreationTimestamp.Time)
 	case dataselect.NamespaceProperty:
 		return dataselect.StdComparableString(self.ObjectMeta.Namespace)
-	default:
+  case dataselect.LabelProperty:
+    s, _ := json.MarshalIndent(self.ObjectMeta.Labels, "", "\t")
+    return dataselect.StdComparableString(s)
+  default:
 		// if name is not supported then just return a constant dummy value, sort will have no effect.
 		return nil
 	}
--- a/src/app/backend/resource/clusterrole/common.go
+++ b/src/app/backend/resource/clusterrole/common.go
@@ -15,7 +15,8 @@
 package clusterrole
 
 import (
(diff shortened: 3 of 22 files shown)


src/app/backend/resource/dataselect/dataselect.go
```diff
@@ -107,16 +107,11 @@ func (self *DataSelector) Filter() *DataSelector {
...
 		for _, filterBy := range self.DataSelectQuery.FilterQuery.FilterByList {
 			v := c.GetProperty(filterBy.Property)
-			if v == nil || !v.Contains(filterBy.Value) {
-				matches = false
-				break
-			}
-		}
-		if matches {
-			filteredList = append(filteredList, c)
+			if v == nil || v.Contains(filterBy.Value) {
```
ajatprabha: Can you explain this change?

src/app/backend/resource/persistentvolumeclaim/common.go
```diff
@@ -15,7 +15,8 @@
 package persistentvolumeclaim
 
 import (
-	"log"
+  "encoding/json"
+  "log"
```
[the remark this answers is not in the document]
anshulahuja98: `encoding/json` has` json.MarshallIndent` which is being used below in line 99

src/app/backend/resource/dataselect/dataselect.go
anshulahuja98: I just realized a mistake in this, should've removed the `v==nil`

anshulahuja98: So overall
Currently this function Filter(), basically takes the intersection of various `filterBy` conditions passed to it

I instead changed it to work in a union fashion

Because I wanted to get results filtered by either name or labels whichever is found

I even mentioned this decision on slack.

src/app/backend/resource/persistentvolumeclaim/common.go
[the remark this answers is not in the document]
ajatprabha: Ah, I was talking about "log" import statement, I thought it's added now. Missed the line changed above.

anshulahuja98:
@maciaszczykm @floreks  can you please review this PR?

anshulahuja98:
> Can you give some examples how API calls will change? How will request/response to get resource look like:
> 
> * by label only
> * by name
> * by name and/or label?

* by label only : ```/api/v1/service?itemsPerPage=10&page=1&sortBy=d,creationTimestamp&filterBy=labels,label_data```
* by name only(same as before) : ```/api/v1/service?itemsPerPage=10&page=1&sortBy=d,creationTimestamp&filterBy=name,name_data```
* by name **or** label: ```/api/v1/service?itemsPerPage=10&page=1&sortBy=d,creationTimestamp&filterBy=name,name_data,labels,label_data```

Response will be same as before
only the support to provide labels based filtering and name **or** label filtering is supported

floreks:
I am especially interested in how `label_data` looks like. I'm worried that marshalling/unmarshalling content of the whole label will make it harder for us to write user friendly UI.

anshulahuja98:
> I am especially interested in how `label_data` looks like. I'm worried that marshalling/unmarshalling content of the whole label will make it harder for us to write user friendly UI.

To make it clear
label_data can contain any part of the label, it's just a string

The actual labels are marshalled and then just a string check to see if the searched keyword is in the labels json. The label_data just stands for the searched keyword.

What my solution overall does is
using the current search bar field, it sends the searched keyword as the searched label and name
```/api/v1/service?itemsPerPage=10&page=1&sortBy=d,creationTimestamp&filterBy=name,searched_keyword,labels,searched_keyword```

Not sure what UI change you are talking about
The labels data is internally marshalled just for searching amongst all labels easily.

floreks:
Let's say that I have 3 labels:
`key1: value1`
`key2: value2:
`key3: key1`

I want to:
- filter by key1
- filter by value1
- filter by exact match `key3: key1`

How would the url look like for this examples? Can we actually do that? How to ensure that it only filters by keys if values may be the same as some keys?

anshulahuja98:
> Let's say that I have 3 labels:
> `key1: value1`
> `key2: value2`
> `key3: key1`
> 
> I want to:
> 
> * filter by key1
> * filter by value1
> * filter by exact match `key3: key1`
> 
> How would the url look like for this examples? Can we actually do that? How to ensure that it only filters by keys if values may be the same as some keys?


My approach doesn't give a method to match only by key or value only specifically.

We must decide if we want to give the user option to filter only by either only key or value or both. We would need the user to give search inputs a specific manner then. Do we need this complexity
Results as per my approach will be 
* filter by key1
Will return 
`key1: value1` and
`key3: key1`

* filter by value1
will return 
`key1: value1`

floreks:
In the UI we are going to need more options. I wouldn't want to filter by labels in the same field as the name currently. It would be confusing. I'd expect some advanced search area.

In example we would have things like:
- `name`
- `label`
- `date`
- `state`

Label section would offer different types of search:
- by key
- by value
- by exact match

If multiple fields were provided then we require all of them (search for resources that match all search criteria).

anshulahuja98:
Umm, that would be an elaborate feature to work on. 
If that's something we can finalize the specifications on, I can perhaps work on it.

zehuaiWANG:
Maybe we can add something such like `--field-selector` for filtering? I think it will be difficult to handler different kind of label selector, and it will make it harder for us to write user friendly UI.

floreks:
What do you mean?

zehuaiWANG:
@floreks I think we can add [--field-selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) feature for filtering. I think it's hard for us to use both label and name for filtering by adding label selector which will makes the UI interface difficult to design. And if we use` --field-selector`, we can keep the search conditions consistent with kubectl.

floreks:
I mean how do you think it would work and look in the UI? I can't image such feature.

zehuaiWANG:
Such like this? Maybe we need more UI/UX design. @floreks

PR #4377, opened 2019-10-03
Update API versions

maciaszczykm:


maciaszczykm:
/cc @shu-mutou 

Can you verify it?

shu-mutou:
@maciaszczykm I do not have time this week. If no need to rush, I'll check this next week.

src/app/backend/scaling/scale.go
```diff
@@ -17,6 +17,7 @@ package scaling
 import (
 	"strconv"
 
+	"k8s.io/api/apps/v1"
```
shu-mutou: Please add "v1" at the first of this line as package alias.
We need to name for packages that are just version number like "v1"  explicitly.
Please see https://github.com/golang/go/issues/29041#issuecomment-443285973

floreks:
Quickly tested overall functionalities and haven't found issues. Could you confirm @shu-mutou?

shu-mutou:
/lgtm
I also do not face problems for now.

PR #4384, opened 2019-10-04
Redesign errors on login page

### LINKED ISSUE #3904: Redesign errors displayed on login page
At the moment we are using simple component to display the error messages. It can be improved (see: `src/app/frontend/login/template.html:19`).

anshulahuja98:
Fix #3904 

Inspired from Error template in frontend/error

image

Suggestions welcome!

### DIFF
--- a/i18n/fr/messages.fr.xlf
+++ b/i18n/fr/messages.fr.xlf
@@ -2949,7 +2949,7 @@
         <target>Kubernetes Dashboard</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/login/template.html</context>
-          <context context-type="linenumber">27</context>
+          <context context-type="linenumber">23</context>
         </context-group>
       </trans-unit>
       <trans-unit id="cf9072061094a1ad33866a311152c495c05890a0" datatype="html">
@@ -2957,7 +2957,7 @@
         <target>Kubeconfig</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/login/template.html</context>
-          <context context-type="linenumber">38</context>
+          <context context-type="linenumber">34</context>
         </context-group>
       </trans-unit>
       <trans-unit id="380ab580741bec31346978e7cab8062d6970408d" datatype="html">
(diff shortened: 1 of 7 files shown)


maciaszczykm:
If it is inside card, don't use another one. Some warning message would be better.

You can use card outside, but still I have no idea how to make it look good.

The current version doesn't look good.

anshulahuja98:
> If it is inside card, don't use another one. Some warning message would be better.
> 
> You can use card outside, but still I have no idea how to make it look good.
> 
> The current version doesn't look good.

Okay, I'll try out few more designs

anshulahuja98:
1 more design idea with different placements

image

image

maciaszczykm:
I think the nice idea would be using `<mat-error>` below input field.

anshulahuja98:
I'll try that

anshulahuja98:
image

Did it with `<mat-error>`
Pushing the code

anshulahuja98:
Let me know your opinion on this.

maciaszczykm:
Looks better, can we make it red?

Any opinions from you guys, @kubernetes/dashboard-maintainers?

anshulahuja98:
image


Red probably looks better, I'll push that

floreks:
It does look better. I think we can go with it for now.

maciaszczykm:
I've restarted tests but it doesn't look like flake. Please check this if they will fail again.

floreks:
One important thing. What about the scenario where user token expires and he is forced logged out. Error shown then will be definitely misleading if it will be under the input.

anshulahuja98:
> I've restarted tests but it doesn't look like flake. Please check this if they will fail again.

Fixed the build

anshulahuja98:
> One important thing. What about the scenario where user token expires and he is forced logged out. Error shown then will be definitely misleading if it will be under the input.

Umm, but it does point to filling the fields again, and if the error is descriptive enough already, I hope this won't be much of a problem, we can maybe even change the current error message.
Otherwise, I am open to suggestions as to how you would consider handling this specific form of error separately.

anshulahuja98:
> Ye, this is the error message. The issue is that it's not really related in any way to the input fields. That's what might be confusing.

 I am open to suggestions as to how you would consider handling this specific form of error separately.

floreks:
Not sure right now how we could improve on that. Possibly we need some designer to figure out a common way to display errors nicely.

/lgtm
/approve

PR #4432, opened 2019-10-18
Use golint

zehuaiWANG:
I have used golint and find some typo error and comment missing，and this is some of them.  It is only in backend/cert, backend/client and backend/resource/deployment. And I will do the others next time if it is necessary. @floreks @maciaszczykm


src/app/backend/cert/manager.go
```diff
@@ -22,13 +22,13 @@ import (
 	certapi "github.com/kubernetes/dashboard/src/app/backend/cert/api"
 )
 
-// Implements cert/api/types.Manager interface. See Manager for more information.
+// Manager is used to implements cert/api/types.Manager interface. See Manager for more information.
```
maciaszczykm: s/implements/implement

zehuaiWANG: Do you mean to remove the 's' form "implements" ? I fix it now.

zehuaiWANG: @floreks

zehuaiWANG: @maciaszczykm

PR #4439, opened 2019-10-21
fix golint failures

FayerZhang:


zehuaiWANG:
I think there is no problem with this. It looks good to me.

PR #4453, opened 2019-10-24
fix golint failures

FayerZhang:


zehuaiWANG:
I don't think it's necessary to change “self” to “cell”, it's easy to know the know the meaning of "self" (it's also very common in other programming language, such like python),  conversely，difficult kind of object use the same wrold "cell" will be confused.

floreks:
I know that golint does not like `self`, but we are ok with using that. Warnings are fine in this case.

zehuaiWANG:
I think it should have a more specific specific name instead of just using "cell"， “cell" is also a generic name in this solution although it can pass golint，I think it also not well.

FayerZhang:
> I agree with @floreks. It seems that golint does not like "self", but it just a warning, like a suggestion.

It's no need to fix the golint warnings.

PR #4458, opened 2019-10-24
fix golint failures

FayerZhang:


zehuaiWANG:
It seems the ci is checks were not successful. And I think `var parallelism int32 = 0` is better than `var parallelism int32 `, although we know int32 will be init as 0,  just in my opinion.

FayerZhang:
> It seems the ci is checks were not successful. And I think `var parallelism int32 = 0` is better than `var parallelism int32 `, although we know int32 will be init as 0, just in my opinion.

I  think the ci is not stable, and I  have no permission to rebuild it.  But I ran ok in my own ci.

zehuaiWANG:
emm.. maybe, it seems that the ci is time out, and I don't know why. Could you use the command `/rebuild`  ? @FayerZhang Do we have any idea to deal with the CI timeout ? @floreks

maciaszczykm:
Restarted the build.

FayerZhang:
> Restarted the build.



> > emm.. maybe, it seems that the ci is time out, and I don't know why. Could you use the command `/rebuild` ? @FayerZhang Do we have any idea to deal with the CI timeout ? @floreks
> 
> I used the command /rebuild , but it does not work.

CI is ok now.

FayerZhang:
> Restarted the build.

CI is ok now

PR #4469, opened 2019-10-28
remove the duplicated code in sidercar/common.go and heapster/common.go

zehuaiWANG:


src/app/backend/integration/metric/common/common.go
```diff
@@ -0,0 +1,83 @@
...
+// limitations under the License.
+
+package common
+
+import (
+  "github.com/kubernetes/dashboard/src/app/backend/api"
+  metricapi "github.com/kubernetes/dashboard/src/app/backend/integration/metric/api"
+)
+
+type CommonSelector struct {
```
floreks: Heapster and sidecar selectors are similar only because sidecar was built to mimic heapster API. This will not always be true, that is why logic was split. Additionally, heapster integration will be soon removed completely from our code.

floreks:
Once we delete heapster this will no longer be duplication and it will once more belong only to the sidecar. Basically we will have to move it again as it will not be common anymore.

zehuaiWANG:
okay, I agree with you. close this pr.

PR #4472, opened 2019-10-29
remove another useless code in apihandler.go

zehuaiWANG:
remove another useless code in apihandler.go


zehuaiWANG:
@maciaszczykm  sorry but i find some other useless codes in apihandler.go which is the same as before. I had checked all the file this time. Please have a look, thanks.

maciaszczykm:
Please create 1 PR with things like this.

zehuaiWANG:
@maciaszczykm what's the meaning of "create 1 PR with things like this.". I am no sure what's the problem was.

PR #4477, opened 2019-10-30
Modifying the namespace in creating-sample-user.md to match the namespace in README.

darshanprasad:
Hello,

I hit an error when following the instructions in create-sample-user.md: Error from server (NotFound): error when creating "dashboard-adminuser.yaml": namespaces "kubernetes-dashboard" not found.  I realized this is because when I used the `kubectl apply` command on the [README](https://github.com/kubernetes/dashboard#getting-started), the namespace used is `kube-system` and not `kubernetes-dashboard`.  I'm not fully sure if this change is appropriate, but I am submitting it in case it will help other users like me who are setting up the dashboard for the first time.  Thank you for a great project!

-Darshan


maciaszczykm:
https://github.com/kubernetes/dashboard/blob/master/docs/user/installation.md already describes installation for v2, so it is fine to use the same namespace as in the v2. The official release is close, I'd to avoid changing that.

law-lee:
great！It helps me.

Issue #4522, opened 2019-11-12
Enable to switch context when kubeconfig used for login


shu-mutou:
##### What would you like to be added

- [x] Add contexts in kubeconfig into user menu.
- [x] Store kubeconfig in browser side.
- [x] Re-login when current context is switched.
- [ ] Store `server` for `cluster` in kubeconfig into JWE token.
- [ ] Create client for `server` in each request.
- [ ] Get `dashboard-metrics-scraper` endpoint from `server` and cache them.
- [ ] Create client for metrics plugin in each request.

##### Why is this needed
To enable to switch context from browser, rather than switching by `kubectl proxy`.

##### Comments

![switch_context](https://user-images.githubusercontent.com/12838129/68647548-791afd80-0561-11ea-9829-f1c3c08f4e08.png)

floreks:
That looks really interesting. How about adding an option to local settings on settings view to add new API server config. We could ping it before adding to see if it's up. Then on settings view we could switch contexts and trigger re-login.

Also, it would be good to show available contexts on the login view.

shu-mutou:
Nice idea! Added config editor with validation to this spec.

ivelichkovich:
Is it possible we could have an all-clusters option, kinda like all namespaces that shows an aggregated view as well?

shu-mutou:
@ivelichkovich `All Clusters` option added into this issue. It will be impremented in the future, but not 2.0.0. Thanks! 😉

PR #4534, opened 2019-11-13
Enable to switch context in kubeconfig

shu-mutou:
When user is logging in with kubeconfig that have multiple contexts, add buttons for the contexts into user menu and enable to switch contexts.

**This solution assumes that dashboard-metrics-scraper is deployed on each cluster.**
To get metrics from dashboard-metrics-scraper in each cluster, add `sidecarHost: [endpoint URL for scraper]` into `cluster` in kubeconfig like follow:

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxxxxxxxxxxxx
    server: http://192.168.100.13:8001
    sidecar-host: http://192.168.100.13:30080
  name: minikube
- cluster:
    certificate-authority-data: xxxxxxxxxxxxx
    server: http://192.168.100.14:8001
    sidecar-host: http://192.168.100.14:30080
  name: minikube2
contexts:
- context:
    cluster: minikube2
    user: minikube2-admin
  name: minikube2-admin@minikube2
- context:
    cluster: minikube
    user: minikube-admin
  name: minikube-admin@minikube
current-context: minikube-admin@minikube
kind: Config
preferences: {}
users:
- name: minikube-admin
  user:
    client-certificate-data: xxxxxxxxxxxxx
    client-key-data: xxxxxxxxxxxxx
    token: xxxxxxxxxxxxx
- name: minikube2-admin
  user:
    client-certificate-data: xxxxxxxxxxxxx
    client-key-data: xxxxxxxxxxxxx
    token: xxxxxxxxxxxxx
```

Endpoints for `server` and `sidecarHost` would be passed as cookies and clients for them would be created dynamically in server side.

As first step, this implements basic architecture. Config editor and etc. would be implemented in future PR.

Note: To test this with `npm run start*`, enlarge maximum http header size with `export NODE_OPTIONS=--max-http-header-size=102400`.

Partial implements #4522


maciaszczykm:
We will go back to this after the release, please ping if I will forget.

shu-mutou:
/hold
Due to String Freeze (from 2020-2-28), PR for new feature that needs translation will be freezed until new release (2020-3-30).

shu-mutou:
/hold cancel
because 2.0.0 released.

shu-mutou:
Hi @maciaszczykm, could you review this?

src/app/frontend/chrome/userpanel/template.html
```diff
@@ -29,6 +29,12 @@
 
     </ng-container>
   </div>
+  <mat-divider *ngIf="contexts.length"></mat-divider>
```
maciaszczykm, line 32: Before init funtion it is undefined. Please add check for that.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+
+type Cluster = {
+  name: string;
+  cluster: {
+    server: string;
+    'certificate-authority-data': string;
+    'sidecar-host': string;
+  };
+};
+type User = {
```
maciaszczykm, line 37: Add new lines between types.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+    private readonly state_: Router,
+    @Inject(CONFIG_DI_TOKEN) private readonly CONFIG: Config,
+  ) {}
+
+  setKubeconfig(configYaml: string): void {
+    if (!configYaml.length) {
+      return null;
+    }
+    // Store kubeconfig from session storage
+    this.sessionStorage_.store(this.CONFIG.kubeconfigSessionStorageName, configYaml);
```
maciaszczykm, line 81: Could you explain why you have picked session storage to keep the data?

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  getKubeconfig(): string {
+    // Load kubeconfig from session storage
+    return this.sessionStorage_.retrieve(this.CONFIG.kubeconfigSessionStorageName) as string;
+  }
+
+  getContexts(): string[] {
+    const kubeconfig = this.sessionStorage_.retrieve(this.CONFIG.kubeconfigSessionStorageName) as string;
+    if (kubeconfig === null) {
```
maciaszczykm, line 110: Use `!kubeconfig` which also checks for `undefined` and empty string.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+    const conf: Kubeconfig = loadYaml(kubeconfig);
+    return conf.contexts.map(ctx => {
+      return ctx.name;
+    });
+  }
+
+  getCurrentContext(): string {
+    const kubeconfig = this.sessionStorage_.retrieve(this.CONFIG.kubeconfigSessionStorageName) as string;
+    if (kubeconfig === null) {
```
maciaszczykm, line 121: Same as before.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+      return '';
+    }
+    const conf: Kubeconfig = loadYaml(kubeconfig);
+    return conf['current-context'];
```
maciaszczykm, line 125: Since `conf` has the `Kubeconfig` type you can use `current-context` property using dot accessor.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+    const kubeconfig = this.sessionStorage_.retrieve(this.CONFIG.kubeconfigSessionStorageName) as string;
+    if (kubeconfig === null) {
+      return null;
+    }
+    const conf: Kubeconfig = loadYaml(kubeconfig);
+    const contexts = conf.contexts.filter(ctx => {
+      return ctx.name === contextName;
+    });
+    if (!contexts.length) {
+      return null;
```
maciaszczykm, line 138: If null can be returned change function type to `Context|null`.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+    return clusters[0];
+  }
+
+  getCurrentServer(): string {
+    const contextName = this.getCurrentContext();
+    return this.getServer(contextName);
+  }
+
+  getServer(contextName: string): string {
+    // Get Context
```
maciaszczykm, line 164: These comments are not required, it's easy to see what is this code is doing by looking at it. Remove.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+  getCluster(name: string): Cluster {
+    const kubeconfig = this.sessionStorage_.retrieve(this.CONFIG.kubeconfigSessionStorageName) as string;
+    if (kubeconfig === null) {
+      return null;
+    }
+    const conf: Kubeconfig = loadYaml(kubeconfig);
+    const clusters = conf.clusters.filter(cluster => {
+      return cluster.name === name;
+    });
+    if (!clusters.length) {
```
maciaszczykm, line 152: Please use `!_.isEmpty()` to check for clusters list emptiness. This is lodash utility function that can be imported with following line `import * as _ from 'lodash';`.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+  switchContext(context: string) {
+    const kubeconfig = this.sessionStorage_.retrieve(this.CONFIG.kubeconfigSessionStorageName) as string;
+    const conf: Kubeconfig = loadYaml(kubeconfig);
+
+    if (
+      conf.contexts.find(ctx => {
+        return ctx.name === context;
+      })
+    ) {
+      conf['current-context'] = context;
```
maciaszczykm, line 213: As before, on known types please use dot to access object params.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+  context: {
+    cluster: string;
+    user: string;
+  };
+};
+type Kubeconfig = {
+  apiVersion: string;
+  clusters: Cluster[];
+  contexts: Context[];
+  'current-context': string;
```
maciaszczykm, line 56: There is no need to escape it. Can we change it to camelCase?

maciaszczykm:
@floreks What do you think about the usability and approach here? I have done initial code review so far but we should also consider if that will not impact our current setup.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+import {load as loadYaml, dump as dumpYaml} from 'js-yaml';
+
+import {AsKdError, K8SError} from '../../errors/errors';
+import {AuthService} from './authentication';
+import {PluginsConfigService} from './plugin';
+import {LoginSpec} from '@api/backendapi';
+import {KdError} from '@api/frontendapi';
+import {Config, CONFIG_DI_TOKEN} from '../../../index.config';
+
+type Cluster = {
```
floreks, line 29: Those are backend api types. Move it to the `backendapi.ts`.

src/app/frontend/common/services/global/kubeconfig.ts
floreks: We have to be very careful with storing the kubeconfig. First, we have to make sure that we can support used kubeconfig file as a login option, not to store invalid kubeconfig files that might contain sensitive certificate information. Also, session storage might not be secure enough to store such sensitive data.

src/app/frontend/common/services/global/kubeconfig.ts
```diff
@@ -0,0 +1,246 @@
...
+  cluster: {
+    server: string;
+    'certificate-authority-data': string;
+    'sidecar-host': string;
+  };
+};
+type User = {
+  name: string;
+  user: {
+    token: string;
```
floreks, line 40: We can't store token here as it is already part of our login mechanism and is stored in an encrypted cookie. This would be extremely insecure. Also, we do not support `client-certificate-data` as a way of logging in anyway, so this is obsolete.

maciaszczykm:
I have tested it locally using `npm run start:prod` command and the context are visible after logging with kubeconfig. However, after I try to switch context I see an error in the console (`POST https://localhost:8080/api/v1/login 500`) and I am redirected to the login view.

src/app/frontend/common/services/global/kubeconfig.ts
shu-mutou: Actually, I didn't want to store such a sensitive data as kubeconfig. But to enable to reload page, i.e. to be able to reload kubeconfig by angular service , we need to store somewhere. Also, kubeconfig is not so small to store in cookie. So I think session storage is better place to store kubeconfig while the session than the other.

shu-mutou:
Thank you very much for your reviews! I will continue considering pointed things, check and modify this next week.

shu-mutou:
> I have tested it locally using `npm run start:prod` command and the context are visible after logging with kubeconfig. However, after I try to switch context I see an error in the console (`POST https://localhost:8080/api/v1/login 500`) and I am redirected to the login view.

I checked `npm run start:prod` and it works in my environment. Could you check your kubeconfig?

src/app/frontend/common/services/global/kubeconfig.ts
shu-mutou: I found that `js-yaml` can not convert keys to camelcase. Instead I found a library [`camelcase-keys`](https://github.com/sindresorhus/camelcase-keys) but it need to use `require` in `import`. Do you know better library or is it acceptable, i.e. using `require`?

maciaszczykm:
I will try to verify again tomorrow but it seemed okay as I was able to login with it. Is some additional info required?

src/app/frontend/common/services/global/kubeconfig.ts
shu-mutou: If we store all the tokens contained in kubeconfig in cookies, I'm worried that all HTTP requests will get too big and impact performance.
I'm considering to encrypt tokens in jwe and store them into cookie before login sequence. But do you have any other ideas?

shu-mutou:
@maciaszczykm I improved my codes, but I'm considering about encrypting tokens with jwe and store them into cookie yet. But I'm happy if you try this UX.

src/app/frontend/common/services/global/kubeconfig.ts
maciaszczykm: You can use require.

shu-mutou:
/hold
need more investigations.

shu-mutou:
/hold cancel
All tokens are stored in cookie as jweTokens now.
Instead, please note that set env `NODE_OPTIONS=--max-http-header-size=1024000` when run `npm run start*`.

floreks:
The solution that requires us to raise limits for HTTP header size is not really acceptable. We should try to limit data passed between browser and server as much as possible. Most proxies set their limits to 8KB and many people are using that. This change would break many configurations. 

Processing all tokens at all times is also not good. Only one should be used at the time and others should naturally expire over time.

Another issue is encrypting all auth information to all servers with a single encryption key that is stored in the main cluster. We should not do that. It decreases security and makes all servers vulnerable in case of a single cluster breach.

There are a couple of more issues, but I think the general approach should be reconsidered. First, we should have some architectural concepts prepared. Possibly something like that should be handled in the user settings and we should have an option to add additional servers that we want to access. We could store some data in secrets/config maps and use multiple encryption tokens for actual auth info.

shu-mutou:
As I expected, the HTTP header size exceeded the limit.
Let's redesign token handling.
If we store sensitive informations into secrets, do you have any idea how to clear expired secrets?

floreks:
Closing as stale. New PR can be opened as it requires major changes.

PR #4535, opened 2019-11-13
add RollbackDeployment feature

zehuaiWANG:
In my opinion, deployment rollback is a very important feature. In the client-go and rancher issues have mentioned the relevant issues (https://github.com/kubernetes/client-go/issues/398), I simply implemented this function in the backend‘s code.

### DIFF
--- a/src/app/backend/resource/deployment/rollout.go
+++ b/src/app/backend/resource/deployment/rollout.go
@@ -97,7 +97,7 @@ func GetReplicateSetFromDeployment(client client.Interface, namespace, deploymen
 	}
 	var result []v1.ReplicaSet
 	for _, rs := range allRS.Items {
-		if metaV1.IsControlledBy(&rs, deployment){
+		if metaV1.IsControlledBy(&rs, deployment) {
 			result = append(result, rs)
 		}
 	}
--- a/src/app/backend/resource/deployment/rollout.go
+++ b/src/app/backend/resource/deployment/rollout.go
@@ -34,12 +34,15 @@ func RollbackDeployment(client client.Interface, namespace string, deploymentNam
 		return errors.New("No revision for rolling back ")
 	}
 	matchRS, err := GetReplicateSetFromDeployment(client, namespace, deploymentName)
+	if err != nil {
+		return err
+	}
 	for _, rs := range matchRS {
 		if rs.Annotations["deployment.kubernetes.io/revision"] == revisionNumber {
 			updateDeployment := deployment.DeepCopy()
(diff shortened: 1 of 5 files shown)


src/app/backend/resource/deployment/rollout.go
```diff
@@ -0,0 +1,45 @@
...
+func RollbackDeployment(client client.Interface, namespace string, deploymentName, replicaSetName string) error {
+	deployment, err := client.AppsV1().Deployments(namespace).Get(deploymentName, metaV1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	currRevision := deployment.Annotations["deployment.kubernetes.io/revision"]
+	if currRevision == "1" {
+		return errors.New("No revision for rolling back ")
+	}
+	replicaSet, err := client.AppsV1().ReplicaSets(namespace).Get(replicaSetName, metaV1.GetOptions{})
```
floreks: Wouldn't it be better to provide a revision number that we want to roll back to and then search for specific replica set with provided revision number?

zehuaiWANG: Yes, thank you for your suggestion, it might be better, is there any other problem? I will modify it in the code that follows. This is just a beta version.

src/app/backend/resource/deployment/rollout.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil
+	}
+	updateDeployment := deployment.DeepCopy()
+	updateDeployment.Spec.Template.Spec = replicaSet.Spec.Template.Spec
+	_, err = client.AppsV1().Deployments(namespace).Update(updateDeployment)
+	if err != nil {
+		return nil
```
floreks: Should be `return err` instead of `return nil`.

zehuaiWANG: yes， I make a mistake, thank you, anything else？

floreks:
Besides minor comments looks good. It's a very useful feature. You also need to register a new endpoint for that in the API.

zehuaiWANG:
thank you, I will modified it later. @floreks

zehuaiWANG:
CI fail：
```
Firefox 63.0.0 (Linux 0.0.0) AboutComponent should print current year FAILED
	Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 6350)
```
Could you help me to find out what the ci problem is, or if it have any other problem in my pr? @floreks @maciaszczykm

PR #4554, opened 2019-11-16
Implement the front end of clusterrole

zehuaiWANG:
Implement the front end of clusterrole


floreks:
Zrzut ekranu z 2019-11-22 09-57-15

Mostly looks good. Few things:
- Add `-` when there is no namespace.
- Move `name` in `RoleRef` to the first place.
- Make `name` of the related role clickable so we can go directly to it.

I'd also change `RoleRef` to simply `Role` or even `Cluster Role`? Can we bind ClusterRoleBinding to a regular Role or only ClusterRole?

zehuaiWANG:
Accoring to `https://kubernetes.io/docs/reference/access-authn-authz/rbac/`, `roleRef`  is the field defined by kubernetes, so I didn't modify him.

floreks:
> Accoring to https://kubernetes.io/docs/reference/access-authn-authz/rbac/, roleRef is the field defined by kubernetes, so I didn't modify him.

I know, but as card title we want to use names that are understandable for regular users, not strictly for developers. Simply `Role` or `Cluster Role` would be sufficient. Depending on if we can bind a role to a cluster role binding.

zehuaiWANG:
Okay, I understand what you mean now. I can not find the answer in the kubernetes official documentation。So I do a example in my cluster.
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  namespace: default
  name: pod-reader
rules:
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["get", "watch", "list"]
```
And
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: read-secrets-global
subjects:
- kind: Group
  name: manager
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: Role
  name: pod-reader
  apiGroup: rbac.authorization.k8s.io
```
And it will show `The ClusterRoleBinding "read-secrets-global" is invalid: roleRef.kind: Unsupported value: "Role": supported values: "ClusterRole"`, so it jsut suppot `clusterrole`

floreks:
We can name it `Cluster Role` then.

zehuaiWANG:
ahh, okay~ I will modify it later.

zehuaiWANG:
@floreks hi~floreks, I have modified it.

PR #4558, opened 2019-11-16
Skip metrics that do not exist, e.g. when pod is in pending or succeeded phase.

wjywbs:


src/app/backend/integration/metric/api/types.go
```diff
@@ -238,7 +237,6 @@ func (self MetricPromises) GetMetrics() ([]Metric, error) {
 		metric, err := metricPromise.GetMetric()
 		if err != nil {
 			// Do not fail when cannot resolve one of the metrics promises and return what can be resolved.
-			log.Printf(`Cannot get metrics: %v`, err)
```
floreks: Can you just restore this? We'd like to still have some log of error even if it's not critical.

floreks:
This should help with graphs/sparklines not being displayed when there are pods in a different state than running.

src/app/backend/integration/metric/api/types.go
wjywbs: In my use case, there are many pods in the not running status, so there are many log lines per request. I added the "received %d resources from sidecar instead of %d" log below to only print one line.

Issue #4261, opened 2019-09-05
Workload statuses cause memory leak


floreks:
I've noticed that leaving `Overview` page open for a couple of minutes causes high memory spike. I've been investigating this issue and narrowed it down to a `kd-workload-statuses` component. It has to be investigated further to see why exactly this happens and how to fix it.

PR #4582, opened 2019-11-21
Fix workload statuses

maciaszczykm:
Fixes https://github.com/kubernetes/dashboard/issues/4261.

Adds https://github.com/swimlane/ngx-charts library.


package.json
```diff
@@ -104,6 +104,7 @@
     "@angular/platform-browser": "8.0.3",
     "@angular/platform-browser-dynamic": "8.0.3",
     "@angular/router": "8.0.3",
+    "@swimlane/ngx-charts": "^12.1.0",
```
floreks: Remove "^"

src/app/frontend/common/components/workloadstatus/style.scss
```diff
@@ -15,7 +15,9 @@
 @import '../../../variables';
 
 .kd-graph-container {
-  flex: auto;
```
dudo: Why drop flex?

maciaszczykm: It's not ready yet as you can see in the title. I still have to fix the styling.

maciaszczykm:
Zrzut ekranu 2019-11-28 o 11 27 35

PTAL

PR #4587, opened 2019-11-22
Dashboard update zh support

zehuaiWANG:
/assign @hwdef

### DIFF
--- a/i18n/zh/OWNERS
+++ b/i18n/zh/OWNERS
@@ -1,7 +1,6 @@
 approvers:
   - hwdef
   - tanjunchen
-  - zehuaiWANG
 
 labels:
 - language/zh
--- a/i18n/zh/OWNERS
+++ b/i18n/zh/OWNERS
@@ -1,6 +1,7 @@
 approvers:
   - hwdef
   - tanjunchen
+  - zehuaiWANG
 
 labels:
 - language/zh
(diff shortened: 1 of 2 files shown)


hwdef:
/lgtm
/approve
Thanks for your contributions.

floreks:
I am wondering if we should actually translate Kubernetes specific resource names. It seems that official documentation chose not to translate them also (mostly). https://kubernetes.io/zh/docs/concepts/workloads/controllers/cron-jobs/

IMHO it would be good to keep them in original forms. Considering i.e. polish translation if I'd ever add one, trying to translate those names would not make much sense.

zehuaiWANG:
Yes, for the k8s developers, they know the native language and English of these kubernetes resources, so there is no big difference between English version and the translated version. But if only a part of it is translated, it will make the whole page very strange. For example, you can imagine that the page is generally in English and polish, so I decided to translate it after discussion with @hwdef , which is also the result after referring to the Japanese version of dashboard.

floreks:
I haven't really reviewed translated versions of Dashboard in detail before, so other versions may have translated names that are specific to Kubernetes. I am saying that we should discuss which approach to follow and make it consistent across all translations. For me translating names like `Pod` would be similar to translating company name like `Google`. Those are specific names that usually are not translated and should be kept in original form.

@maciaszczykm wdyt?

zehuaiWANG:
Yes, I agree with your opinion, it should be discussed and find out which approach to follow and make it consistent across all translations.

maciaszczykm:
I agree I'd avoid translating proper names like pod or daemon set.

hwdef:
I agree with @maciaszczykm and @floreks.
We better not translate them.

zehuaiWANG:
okay, I wonder if we should define what should be translated next and consistent across all translations. It would be better if it could be written in something like a translation contribution guide.

Issue #4589, opened 2019-11-22
Support passing CA certificate for validating apiserver_host


jr0d:
##### What would you like to be added
I would like a CLI argument to be added to the dashboard that would capture the path to a CA certificate. This path would be used to populate `ConfigOverrides.ClusterInfo`. 

##### Why is this needed
This feature is needed to verify connections to an `apiserver_host` that is protected by TLS

##### Comments
I will have a PR up shortly which provides this feature.

jr0d:
Closing this issue, this feature turned out not to be necessary.

PR #4590, opened 2019-11-22
Add apiserver-host-ca-file flag to dashboard

jr0d:
This flag allows users to provide a CA certificate for the purpose
of validating connections to the API server. This is mostly helpful
when using an `apiserver-host` that is protected by TLS.

Related to https://github.com/kubernetes/dashboard/issues/4589

### DIFF
--- a/src/app/backend/client/manager.go
+++ b/src/app/backend/client/manager.go
@@ -285,7 +285,7 @@ func (self *clientManager) initConfig(cfg *rest.Config) {
 // empty then in-cluster config will be used and if it is nil the error is returned.
 func (self *clientManager) buildConfigFromFlags(apiserverHost, apiserverHostCAFile, kubeConfigPath string) (
 	*rest.Config, error) {
-	if len(kubeConfigPath) > 0 || len(apiserverHost) > 0 || len(apiserverHostCAFile) > 0 {
+	if len(kubeConfigPath) > 0 || len(apiserverHost) > 0 {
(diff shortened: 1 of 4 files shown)


src/app/backend/client/manager.go
```diff
@@ -281,12 +283,16 @@ func (self *clientManager) initConfig(cfg *rest.Config) {
 
 // Returns rest Config based on provided apiserverHost and kubeConfigPath flags. If both are
 // empty then in-cluster config will be used and if it is nil the error is returned.
-func (self *clientManager) buildConfigFromFlags(apiserverHost, kubeConfigPath string) (
+func (self *clientManager) buildConfigFromFlags(apiserverHost, apiserverHostCAFile, kubeConfigPath string) (
 	*rest.Config, error) {
-	if len(kubeConfigPath) > 0 || len(apiserverHost) > 0 {
+	if len(kubeConfigPath) > 0 || len(apiserverHost) > 0 || len(apiserverHostCAFile) > 0 {
```
floreks: I guess we can ignore checking that and simply pass it to client config. Default value is empty string, so it will be ignored anyway. If someone will pass only `apiserverHostCAFile` without `apiserverHost` it will ignore it.

jr0d: :+1:

jr0d:
This PR turned out not to be needed. We will rely on the existing ICC with impersonation headers.

PR #4619, opened 2019-11-28
Rewrite allocation charts

maciaszczykm:
Zrzut ekranu 2019-11-28 o 14 57 48


hwdef:
/lgtm
Thanks for your contribution.

floreks:
It would be nice to use core and memory filters that we have to show bytes/cores values with proper suffixes. billions of bytes is not very user friendly.

maciaszczykm:
@floreks It is not possible to apply them there, unfortunately. I can change them to cores and megabytes though.

src/app/frontend/resource/cluster/node/detail/template.html
```diff
@@ -129,91 +129,39 @@
...
-          </div>
-          <div class="kd-graph-legend-entry">
-            <div class="kd-graph-legend-placeholder"></div>
-            <span i18n>Capacity</span>
-            <span class="kd-graph-legend-value">
-              {{node?.allocatedResources.cpuCapacity | kdCores}}
-            </span>
-          </div>
-        </div>
+        <div class="kd-h1">CPU</div>
```
shu-mutou: Could you add `i18n` attribute for some elements that should be translated?

PR #4622, opened 2019-11-29
Implement the front end of clusterrole

zehuaiWANG:
Implement the front end of clusterrole，base on #4554 and fix the merge error


src/app/frontend/common/components/roleref/component.ts
```diff
@@ -0,0 +1,42 @@
...
+import {RoleRef} from '@api/backendapi';
+
+@Component({
+  selector: 'kd-role-ref-meta',
+  templateUrl: './template.html',
+})
+export class RoleRefComponent {
+  @Input() initialized = false;
+
+  private roleRef_: RoleRef;
```
floreks: I don't think we need get/set here. It can simply be `@Input() roleRef: RoleRef = {} as RoleRef;` and then in `ngOnInit` method you can do something like `this.roleRef = this.roleRef === undefined ? {} as RoleRef : this.roleRef;`.

src/app/frontend/common/components/roleref/component.ts
```diff
@@ -0,0 +1,42 @@
...
+  @Input()
+  set roleRef(val: RoleRef) {
+    if (val === undefined) {
+      this.roleRef_ = {} as RoleRef;
+    } else {
+      this.roleRef_ = val;
+    }
+  }
+
+  getClusterRef(name: string): string {
```
floreks: s/getClusterRef/getClusterRoleHref

Do not hardcode the namespace. It should never be changed this way. The namespace should stay as is when changing views.

src/app/frontend/common/components/subject/component.ts
```diff
@@ -0,0 +1,49 @@
...
+  @Input() initialized: boolean;
+  @Input() subjects: Subject[];
+  private columns = {
+    0: 'kind',
+    1: 'apiGroup',
+    2: 'name',
+    3: 'namespace',
+  };
+
+  ngOnInit(): void {}
```
floreks: Do not implement `OnInit` if you are leaving this method empty.

src/app/frontend/common/components/roleref/component.ts
zehuaiWANG: Yes, I also feel strange beacause when I visit `/clusterrole/xxx?namespace=default`, it's ok but when I visit `/clusterrole/prometheus?namespace=` it will occur some problem, it will show blank, but clusterrole is nothing to do with `namespace`, I wonder if it's caused by where we set the `namespace` to be non empty in someway.

src/app/frontend/common/components/roleref/component.ts
zehuaiWANG: okay, I will modifiy it later.

src/app/frontend/common/components/subject/component.ts
zehuaiWANG: Okay, I will remove it.

src/app/frontend/common/components/roleref/component.ts
floreks: Check how other views are doing that.

shu-mutou:
/hold
Due to String Freeze (from 2020-2-28), PR for new feature that needs translation will be freezed until new release (2020-3-30).

shu-mutou:
/hold cancel
because 2.0.0 released.

PR #4667, opened 2019-12-16
Modified the PodRequestAndLimits function

zehuaiWANG:
Modified the PodRequestAndLimits function and Add overhead for running a pod to the sum of requests and to non-zero limits.
do as kubectl [PodRequestsAndLimits in kubectl](https://github.com/kubernetes/kubernetes/blob/3d17fd5c4f6b1ed22499c4db192e1ed6a867d579/staging/src/k8s.io/kubectl/pkg/util/resource/resource.go)


zehuaiWANG:
Thanks for reviewing it. @floreks

Issue #4696, opened 2019-12-21
The Pods status in ReplicaSet is not correct


zehuaiWANG:
hi~ everyone, I found that when replicaset gets its podInfo by following code, it only uses the label selector. I think we should add the owner reference judgment to ensure that the pod belongs to replicaset.
```
func getReplicaSetPodInfo(client k8sClient.Interface, replicaSet *apps.ReplicaSet) (*common.PodInfo, error) {
	labelSelector := labels.SelectorFromSet(replicaSet.Spec.Selector.MatchLabels)
	channels := &common.ResourceChannels{
		PodList: common.GetPodListChannelWithOptions(client, common.NewSameNamespaceQuery(replicaSet.Namespace),
			metaV1.ListOptions{
				LabelSelector: labelSelector.String(),
				FieldSelector: fields.Everything().String(),
			}, 1),
	}

	pods := <-channels.PodList.List
	if err := <-channels.PodList.Error; err != nil {
		return nil, err
	}

	podInfo := common.GetPodInfo(replicaSet.Status.Replicas, replicaSet.Spec.Replicas, pods.Items)
	return &podInfo, nil
}
```
What more, I test it by create two replcicaSet with same label selectors：
```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: test0
  labels:
    app: guestbook
    tier: frontend
spec:
  replicas: 1
  selector:
    matchLabels:
      tier: frontend
  template:
    metadata:
      labels:
        tier: frontend
    spec:
      containers:
      - name: php-redis
        image: gcr.io/google_samples/gb-frontend:v3
```
```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: test1
  labels:
    app: guestbook
    tier: frontend
spec:
  replicas: 1
  selector:
    matchLabels:
      tier: frontend
  template:
    metadata:
      labels:
        tier: frontend
    spec:
      containers:
      - name: php-redis
        image: gcr.io/google_samples/gb-frontend:v3
```
and the view will be:
![image](https://user-images.githubusercontent.com/22339920/71308352-6fc16280-2436-11ea-9c41-435e9608b25f.png)

I wonder if it is a bug, if so, I am willing to submit a PR to modify it.
/assign @maciaszczykm @floreks thanks

Additional, I wonder if the front-end this part have something wrong either in the Resource Infomation:
[...]

PR #4698, opened 2019-12-22
fix bug for pod status in replicaSet

zehuaiWANG:
bug fix: refer to issue #4696 

In addition, I don't think we need to judge all the pods under the namespace, just judge the pods selected by the label selector


floreks:
> In addition, I don't think we need to judge all the pods under the namespace, just judge the pods selected by the label selector

ReplicaSet can't control pods across namespaces, but different namespace can have pod with the same name, so it could cause conflicts.

/lgtm

Issue #4259, opened 2019-09-05
Update translation files before releasing 2.0.0


shu-mutou:
##### What would you like to be added
Before releasing 2.0.0, update translation files, i.e. `i18n/[locale]/messages.[locale].xlf`, for new releases.

##### Why is this needed
Since the last translation, new commits for frontend have been made and translation files have updated.

##### Comments
To add or update translation file, see [this document](https://github.com/kubernetes/dashboard/blob/master/docs/developer/internationalization.md).

##### Translation files to be updated (2020-1-22)
- [x] fr: done by #4828
- [x] ja: done by #4784
- [x] ko: done by #4819, #4859
- [x] zh (Simplified Chinese): ~#4806~ #4856
- zh-cn (Simplified Chinese): Linked to `zh` #4814
- zh-sg (Simplified Chinese): Linked to `zh` #4856
- [x] zh-tw (Traditional Chinese): #4815
- [x] zh-hk (Traditional Chinese): #4815

##### PRs that require translation
- ~Implement the front end of clusterrole #4622~ **_hold_**
- ~Enable to switch context in kubeconfig #4534~ **_hold, >2.0.0_**

##### String fix (2020-2-28)
- [x] In principle, additions or changes of strings that require further translation will not occur. #4911
  Above PRs that need translation would be putted off to next release.

##### Translation files to be merged (should be merged until 2020-3-6)
- [x] de: #4905 

##### Angular 9 support (starts at 2020-3-9, until 2020-3-13)
- [x] Use new locale id: #4969
  zh-cn -> zh-Hans, zh-sg -> remove, zh-tw -> zh-Hant, zh-hk -> zh-Hant-HK, zh -> remove

##### Translation files to be checked (starts at 2020-3-16, until 2020-3-20)
- [x] Final check of translation files by approver of each language.

##### Translation files to be checked (starts at 2020-3-23, until 2020-3-27)
- [x] Final check of translation files. (cleanup `status="new"` and update `linenumber`s) @shu-mutou #4969

shu-mutou:
@feloy @shu-mutou @fuergaosi233 @seokho-son please follow this issue.
related #3812 #4173 #4063 #4174

shu-mutou:
@feloy @fuergaosi233 @seokho-son BTW, if you would get permission to approve translation files, i.e. if directories for each languages with OWNERS file that includes you would be added, would you guys be happy?

feloy:
Yes, that's ok for me. Thanks

[...]

PR #4710, opened 2019-12-24
Update translation files for Chinese

hwdef:
Ref: #4259


hwdef:
@shu-mutou 
This is the last PR for update translation files , I think almost everything that should be translated has been translated.

PR #4726, opened 2020-01-06
fix: node pie graphs to use consistent powers in divisor

### LINKED ISSUE #4725: node pie graphs sometimes apply wrong power in divisor
##### Steps to reproduce

* Run dashboard on a cluster with a node that has a lot of memory headroom, such that the capacity would be rendered in GiB, while the requests and limits would (all other things being equal) be rendered in MiB.
* Navigate to that node in the tree (#/node/my-node?namespace=default or whatever)

##### Observed result

* Observe that the memory allocation % is off (by a factor of 1024):

image

##### Expected result

Instead of 7,3xx% requests for memory allocation, I expect something closer to 7%. Same for limits, e.g.:

image

##### Comments

The code tries to normalize the units, and that means multiplying and dividing by powers of 1024 and 1000 for CPU and memory, respectively. Two overlapping bugs in that code are resulting in an incorrect divisor.

I have a fix that I will PR momentarily.

laurelnaiad:
1. The code determining whether or not the units of requests, limits and
capacity  required normalization was leaving out capacity in its
determination. As such, if the capacity was of different units than
requests and limits, they'd be left  inconsistent and the percentage
would be off by at least a factor of 1000 or 1024.
2. The code normalizing units was erroneously reversing a value in one
conditional, and thus erroneously triggering a second conditional which
was reversing the work of the first one (multiplying by the same factor
by which it had just divided). Rather than fix the bug, it's easier and
better to just eliminate this code. The conditionals aren't even necessary,
because negative exponentiation "just works".

### DIFF
--- a/src/app/frontend/common/components/graph/helper.ts
+++ b/src/app/frontend/common/components/graph/helper.ts
@@ -57,21 +57,10 @@ export class FormattedValue {
     if (expectedPower < 0) {
       throw new Error(`Suffix '${suffix}' not recognized.`);
     }
-
-    let powerDiff = expectedPower - currentPower;
-
-    if (powerDiff < 0) {
-      powerDiff = -powerDiff;
-      this.value_ = this.value_ * Math.pow(this.base_, powerDiff);
-      this.suffix_ = suffix;
-    }
-
-    if (powerDiff > 0) {
-      this.value_ = this.value_ / Math.pow(this.base_, powerDiff);
-      this.suffix_ = suffix;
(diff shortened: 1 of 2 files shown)


laurelnaiad:
This is a fix for https://github.com/kubernetes/dashboard/issues/4725 . I marked it as "Closes #...", but the robot didn't like that. :)

floreks:
Closes #4725 

Thanks!

/lgtm

Issue #4740, opened 2020-01-09
the pod status is not true


zehuaiWANG:
hi, everyone,
I found that the dashboard can not handle the pod `Terminating` status,
BTW, when the pod phase is `Running ` and the `DeletionTimestamp` is not nil, dashboard will show the pod is Running, but it's `Terminating`  if I use `kubectl describe pod`.
```
func isTerminating(pod *corev1.Pod) bool {
	if pod.Spec.ActiveDeadlineSeconds != nil && *pod.Spec.ActiveDeadlineSeconds >= int64(0) {
		return true
	}
	return false
}
```
I found some code in kubernetes/kubernetes and I think we need more information about the pod status because it's a important part, the kubernetes offical doc seems not clear to describe it also.

mukundmckinsey:
try this one https://grafana.com/grafana/dashboards/11802

PR #4742, opened 2020-01-09
Dashboard fix pod status

zehuaiWANG:
refer to issue #4740 

I found that  dashboard can not handle the pod terminating status,
BTW, when the pod phase is Running and the DeletionTimestamp is not nil, dashboard will show the pod is Running, but it's Terminating if I use kubectl describe pod

cc @maciaszczykm @floreks @shu-mutou 
Please don't merge it now. I think it should be discussed. Maybe the front-end should be changed either.

Above it's some reference about pod status:
[How to use the kubernetes go-client to get the same Pod status info that kubectl gives](https://stackoverflow.com/questions/57297817/how-to-use-the-kubernetes-go-client-to-get-the-same-pod-status-info-that-kubectl)
[kubernetes/kubernetes printpod](https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L558)

### DIFF
--- a/src/app/backend/resource/pod/common.go
+++ b/src/app/backend/resource/pod/common.go
@@ -46,7 +46,7 @@ func getPodStatus(pod v1.Pod, warnings []common.Event) PodStatus {
 	}
 }
 
-// getPodStatusPhase returns one of four pod status phases (Pending, Running, Succeeded, Failed, Unknow, Terminating)
+// getPodStatusPhase returns one of four pod status phases (Pending, Running, Succeeded, Failed, Unknown, Terminating)
 func getPodStatusPhase(pod v1.Pod, warnings []common.Event) v1.PodPhase {
 	// For terminated pods that failed
 	if pod.Status.Phase == v1.PodFailed {
--- a/src/app/backend/resource/pod/common.go
+++ b/src/app/backend/resource/pod/common.go
@@ -46,7 +46,7 @@ func getPodStatus(pod v1.Pod, warnings []common.Event) PodStatus {
 	}
 }
 
-// getPodStatusPhase returns one of four pod status phases (Pending, Running, Succeeded, Failed)
+// getPodStatusPhase returns one of four pod status phases (Pending, Running, Succeeded, Failed, Unknow, Terminating)
 func getPodStatusPhase(pod v1.Pod, warnings []common.Event) v1.PodPhase {
 	// For terminated pods that failed
 	if pod.Status.Phase == v1.PodFailed {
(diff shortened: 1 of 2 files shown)


floreks:
I'd leave base statuses as a list of:
- Running
- Pending
- Succeded
- Failed

We could display additional information on the details view, but on the list view I'd stick to the above 4.

zehuaiWANG:
@floreks okay, I agree with you that display the base statuses as a list of: Running、Pending、Succeded、Failed. But I think we should add some code to handle the case if the pod is terminating because when the pod status is terminating, the pod is not use anymore, but in dashboard, we still display it as running.

zehuaiWANG:
some code such like:
```
if pod.DeletionTimestamp != nil {
		return v1.PodFailed			reason = "Terminating"
	}		
}
```
we should take `pod.DeletionTimestamp` into consideration.

floreks:
So maybe we could add another status like `Terminating`. What is the status of pod in kubectl when it is being deleted?

zehuaiWANG:
when a pod is being deleted:

- The object is still visible via the REST API

- The object’s deletionTimestamp is set

And the pod status display in `kubectl get pod xxx` will be `Terminating`. But in general, the pod will be deleted quickly, which makes it difficult for us to see the `Termination` state. However, if the pod cannot be deleted for some reasons (for example, the resource of the pod is occupied by other processes), it will be in the `Termination` state.

More than that, I think we may need to redetermine the status of pods, because if a node loses contact, then all the pods on the node will be in the `Unknown` state. There are also some cases, such as image pull failure or container startup failure. Do we set the pod status to `Failed`? Because the `Failed` state in the kubectl command actually means that the pod error has exited and will not be restarted, and the container pull failure will be restarted.

This PR is written according to `kubectl get pod` command in kubernetes, but I think it needs to be discussed and decided, so I hold him.

@floreks

zehuaiWANG:
crash
Terminating
such like this. @floreks

floreks:
We can add `Terminating` as additional status. For containers that are constantly restarted due to some error we have `Pending` status + error icon and we can click on the pod to see related error. If we can determine that node has lost connection and Pods are in an `Unknown` state then I'd add that also to the list of states. We would present some icon with the question mark then as a status icon.

zehuaiWANG:
okay, I agree with you. And I'd modify this pr and add `Unknow` and `Terminating`  to the list of states later. Thanks.

src/app/backend/resource/pod/common.go
```diff
@@ -46,7 +46,7 @@ func getPodStatus(pod v1.Pod, warnings []common.Event) PodStatus {
 	}
 }
 
-// getPodStatusPhase returns one of four pod status phases (Pending, Running, Succeeded, Failed)
+// getPodStatusPhase returns one of four pod status phases (Pending, Running, Succeeded, Failed, Unknow, Terminating)
```
floreks: s/Unknow/Unknown

maciaszczykm:
@zehuaiWANG Will you update it?

zehuaiWANG:
Yes. This is an important PR. Sorry I forgot about it, I think I need to look at it again to make sure there are no new problems. Maybe about next week I will update it.

zehuaiWANG:
@maciaszczykm I've already tested it and found no problem. Thank you for fixing the typo.

maciaszczykm:
It's problem with Travis.

PR #4775, opened 2020-01-14
golangci-lint need to run against whole go files

shu-mutou:
`golangci-lint` can not check files on different directory.
It would reports "named files must all be in one directory" error.
So we need to check whole files in once without filepaths from `lint-staged`.

Also, bump golangci-lint from 1.17.1 to 1.22.2.

Closes: #3656

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
       "git add"
     ],
     "src/**/*.go": [
-      "golangci-lint run -c .golangci.yml --fix",
+      "./aio/scripts/pre-commit-golangci-lint.sh",
       "git add"
     ],
     "src/**/*.html": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -71,7 +71,7 @@
       "git add"
     ],
     "src/**/*.go": [
-      "golangci-lint run -c .golangci.yml --fix",
+      "./aio/scripts/pre-commit-golangci-lint.sh",
```
maciaszczykm: If I change 10 files, will it run 10 times the same script?

shu-mutou: lint-staged passes all staged filepaths at once.
So this script will be called only once.

maciaszczykm: Ah, sorry. Didn't notice.

PR #4792, opened 2020-01-16
Add additional locale directories for zh

floreks:
I have configured directories for additional Chinese translations. They are ready for translation.

@hwdef @tanjunchen @zehuaiWANG 

Feel free to update the OWNERS files.

### DIFF
--- a/docs/developer/internationalization.md
+++ b/docs/developer/internationalization.md
@@ -9,6 +9,10 @@ Based on current browser locale the Dashboard can be displayed in one of the sup
 | Japanese           | ja   |
 | Korean             | ko   |
 | Simplified Chinese | zh   |
+| Chinese (PRC)      | zh-cn|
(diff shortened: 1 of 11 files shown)


hwdef:
Thank you very much.
Maybe it is better to copy the zh folder to zh-cn and zh-sg, because their contents are the same.
Only zh-tw and zh-hk need translation

floreks:
@hwdef
Unfortunately, it would not be picked up by the backend this way. We have to use translation per directory right now.

hwdef:
@floreks 
I got it, thank you for creating these folders.I will translate zh-cn and zh-sg later.
And keep zh-tw and zh-hk as they are. How about this?

floreks:
Right now they are all templates in English ready to be translated. You can reuse `zh` where it is possible. Just copy translated parts over.

PR #4796, opened 2020-01-16
update documentation for building on darwin

ivelichkovich:
otherwise you can't build docker when working in a darwin environment, you get incompatible architectures from go build.

### DIFF
--- a/docs/developer/getting-started.md
+++ b/docs/developer/getting-started.md
@@ -80,6 +80,12 @@ Open a browser and access the UI under `localhost:9090`. The following processes
 
 `Dashboard backend (9090) ---> Kubernetes API server (8080)`
 
+To build the docker image on darwin OS you will need to set environment variable for go to build as linux:
+
+```
+export GOOS=linux
+```
+
 In order to package everything into a ready-to-run Docker image, use the following task:
 
 ```


floreks:
Add that before `npm run build` block.

ivelichkovich:
Moved it, I can move it around more or change the text if needed

docs/developer/getting-started.md
```diff
@@ -82,6 +82,12 @@ Open a browser and access the UI under `localhost:9090`. The following processes
 
 In order to package everything into a ready-to-run Docker image, use the following task:
```
floreks: This part should be after `export GOOS=linux`

ivelichkovich: ah yeah that makes more sense, updated

ivelichkovich:
while we're at it, is there a faster way to build this docker image? If I'm only changing the go for example, the frontend takes ages to build, is there a way I can just manually run go build and build a docker container from that?

ivelichkovich:
Also what should I do about the test coverage?

floreks:
> while we're at it, is there a faster way to build this docker image? If I'm only changing the go for example, the frontend takes ages to build, is there a way I can just manually run go build and build a docker container from that?

We do not have separate task for that, but you could manually compile backend binary, replace it in the dist directory and run only `gulp docker-image:head:cross`. I don't see why you would build the image on your own frequently. You develop on `serve` and then do the final test on `prod`. That should be enough most of the time.

/lgtm

Issue #4763, opened 2020-01-13
Having trouble adding languages to i18n


hwdef:
##### What should be cleaned up or changed

`docs/developer/internationalization.md`

##### Provide any links for context

I am doing the work of adding languages follow `internationalization.md` , and I find two troubles:

1. I cannot find  `xfillmergeOptions` array , only `xliffmergeOptions`

![image](https://user-images.githubusercontent.com/13084946/72236231-ed7d2600-3610-11ea-828a-ddcdfa3d1802.png)

![image](https://user-images.githubusercontent.com/13084946/72236318-4cdb3600-3611-11ea-9b71-87f061ed2e6c.png)

2.I add `zh-cn` in xliffmergeOptions array ,and I run 

```shell
npm run fix:i18n
```
the result is 

![image](https://user-images.githubusercontent.com/13084946/72236371-89a72d00-3611-11ea-86eb-099846a9f3bf.png)

Is this a bug or am I running the wrong process?

floreks:
Those steps are not really valid.

After `messages.zh-cn.xlf` is generated, create directory for your translation i.e. `zh-cn` and move the translation file there. Then add `zh-cn` to the `locale_conf.json` file. That should be all. Then you can translate the file.

hwdef:
@floreks 
Thanks for your answer.
How should problem 1 be solved?Is it a spelling mistakes? Modify the array name in the `internationalization.md`?
I can perfect this document

maciaszczykm:
@hwdef You can fix the mistake and add the description from @floreks to it.

hwdef:
@maciaszczykm 
OK, I will do these.

floreks:
Updated Documentation in #4792 

/close

Issue #4734, opened 2020-01-08
Add language setting in local setting.


hwdef:
##### What would you like to be added

language setting in local setting

##### Why is this needed

In most browsers, when the region is China, the default language is zh-CN, it can not match zh, so dashboard will display English.

Chinese has four branch: zh-CN , zh-TW, zh-HK, zh-SG, correspond to China, Taiwan, Hong Kong and Singapore.

I think adding an option for language settings is an easy and intuitive way to solve this.

##### Comments

![image](https://user-images.githubusercontent.com/13084946/71961081-90106000-3231-11ea-8c61-723a2a7648d6.png)

tanjunchen:
![image](https://user-images.githubusercontent.com/19339970/71983783-82270300-3262-11ea-8fd2-020db4997cef.png)

floreks:
> In most browsers, when the region is China, the default language is zh-CN, it can not match zh, so dashboard will display English.

That is not really an issue. We can simply add translation for `zh-CN`. If base is the same as `zh` then it's as simple as creating new directory named `zh-CN` and copying the translation file.

It would be harder to let user choose language as backend auto-discovers and handles serving translations on a single endpoint and we would need to dynamically add endpoint per translation to handle it similar to i.e. Kubernetes docs.

hwdef:
@floreks 
Thanks for your suggestion , it is good.
I can add a new directory named `zh-CN`.
But it creates a new problem, Chinese includes Simplified Chinese and Traditional Chinese, zh-cn and zh-sg is Simplified Chinese, zh-TW and zh-HK is Traditional Chinese. the `zh` folder need to make a choice.I think the `zh` folder is best kept as it is now, Simplified Chinese. Because Simplified Chinese has the most users. What do you think about it.

floreks:
@hwdef I think we should have all directories `zh`, `zh.CN`, `zh-TW`, etc. Depending on what locale browser will present, the exact directory will be chosen for translation. For example it looks like chrome supports `zh_CN` and `zh_TW` as locale information.

https://developer.chrome.com/webstore/i18n#localeTable

We should check other browsers and decide how to name our i18n directories.

hwdef:
@floreks 
I can create all the directories, but I cannot write Traditional Chinese.So they will keep Simplified Chinese until someone has translated Traditional Chinese.

In firefox ,there are more settings,I should create `zh-CN`, `zh-TW` and 'zh-HK'.
[...]

PR #4814, opened 2020-01-21
Support multi locales in one translation file

shu-mutou:
`zh` is same as `zh-cn`, so maintain `zh` and change `zh-cn` as link to `zh`.

Also, added support for `i18n/locales_not_for_build_local` that keeps locales not for build as local setting for developer. Because building dashboard takes much time.

Ref: #4734 [comment](https://github.com/kubernetes/dashboard/issues/4734#issuecomment-574036444)
Ref: #4763


floreks:
I am not sure if creating a symbolic link is a correct way of doing this. Every translation file in the header has information about `target-language`

https://github.com/kubernetes/dashboard/blob/14ae7c1f159f218a98aaf6c9e994f0c18da57a40/i18n/zh-cn/messages.zh-cn.xlf#L3

If we create a symbolic link to a different language code, then `target-language` will not be correct.

shu-mutou:
We can use `zh-Hans` representing Simplified Chinese and and `zh-Hant` for Traditional Chinese.
http://www.iana.org/assignments/lang-tags/zh-Hans
http://www.iana.org/assignments/lang-tags/zh-Hant

Firefox and Chrome don't seem to have `zh-Hans` and `zh-Hant`, but we can create them for translation work and link to them from other `zh*`s.

It has inconsistent in `target-language` but reasonable, I think. npm builds dashboard for each languages with symbolic linked `messages.xx.xlf` file and the binary works perfectly as expected.

I'll investigate how not to build `zh-hans` and `zh-hant` that won't be use and update document.

/hold

hwdef:
This may not be the right way.
Although both mainland China and Singapore use simplified Chinese, some words are different.
Although Taiwan and Hong Kong both use traditional Chinese, they are also very different.
So we should use different files for each region

shu-mutou:
/hold
@hwdef I see situations for zh locales. I'll revert zh* reconstructions. Thanks.

shu-mutou:
@hwdef, how about `zh` and `zh-cn` ? Are there deference between them?

hwdef:
@shu-mutou  
zh represents all kinds of Chinese, it is better to make zh consistent with zh-cn, because zh-cn is the most used.

shu-mutou:
@hwdef I got it. I'll make `zh-cn` as link to `zh`.

shu-mutou:
@floreks Changes for this PR are coming back to first situation and inconsistency in `target-language` is back. But `zh` means "all kinds of Chinese" and its contents is same as `zh-cn`, as hwdef said, so it's reasonable doing so, I think.

floreks:
I am not fully sure if keeping target locale that is different than configured locale code will not cause some issues during compiling the translated versions. Having multiple overlapping target languages might cause issues.

shu-mutou:
@floreks No worry, I think. Dashboard is built with `ng build --i18nLocale = [locale]` and the Angular documentation says that the locale specified by this option will be set to the `LOCALE_ID` of the compiled file.
https://github.com/angular/angular/blob/master/aio/content/guide/i18n.md#i18n-pipes

In fact, `LOCALE_ID` in compiled JavaScript files for `zh-cn` that created as a symbolic link to`zh` in this PR is `zh-cn` instead of `zh`.
As same as `zh-cn`, compiled files for `ja` or` fr` that do not have `target-language` in xlf files have each `LOCALE_ID` specified with `--i18nLocale` option.

floreks:
@shu-mutou makes sense.

/lgtm

Issue #2947, opened 2018-04-08
Scrambled Credentials - Accessing dashboard with chrome, on windows


oustn:
##### Environment

```
Dashboard version: k8s.gcr.io/kubernetes-dashboard-amd64:v1.8.3
Kubernetes version: v1.10.0
Host: Windows10 1709  VirtualBox 5.2.8

Operating system: CentOS-7.2  Linux version 3.10.0-693.21.1.el7.x86_64 (builder@kbuilder.dev.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-16) (GCC) )
```

##### Steps to reproduce
1. Master Node IP: 192.168.56.3
2. Deploying  Dashboard 
```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/src/deploy/recommended/kubernetes-dashboard.yaml
```
3. Can accessing dashboard with kubectl proxy

```
kubectl proxy

curl http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/

<!doctype html> <html ng-app="kubernetesDashboard"> <head> <meta charset="utf-8"> <title ng-controller="kdTitle as $ctrl" ng-bind="$ctrl.title()"></title> <link rel="icon" type="image/png" href="assets/images/kubernetes-logo.png"> <meta name="viewport" content="width=device-width"> <link rel="stylesheet" href="static/vendor.93db0a0d.css"> <link rel="stylesheet" href="static/app.93e259f7.css"> </head> <body ng-controller="kdMain as $ctrl"> <!--[if lt IE 10]>
      <p class="browsehappy">You are using an <strong>outdated</strong> browser.
      Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your
      experience.</p>
    <![endif]--> <kd-login layout="column" layout-fill="" ng-if="$ctrl.isLoginState()"> </kd-login> <kd-chrome layout="column" layout-fill="" ng-if="!$ctrl.isLoginState()"> </kd-chrome> <script src="static/vendor.bd425c26.js"></script> <script src="api/appConfig.json"></script> <script src="static/app.b5ad51ac.js"></script> </body> </html>
```

It's seems work correct.

4.  Edit kubernetes-dashboard service,  Change type: ClusterIP to type: NodePort

```
NAME                   TYPE       CLUSTER-IP       EXTERNAL-IP   PORT(S)         AGE
[...]

PR #4860, opened 2020-02-04
Fix browser prompting ERR_CERT_INVALID for direct HTTPS access

oneoneonepig:
The `notBefore` and `notAfter` x509 field is missing when creating the certificate, causing the certificate to have no valid date information. The notBefore and notAfter decodes to "00010101000000Z", which seems like a "GeneralizedTime" according to [RFC5280](https://tools.ietf.org/html/rfc5280#section-4.1.2.5.2) and decodes to something like 0001/1/1, eventually causing the browser to prompt certificate error. 

If the self-signed certificate is invalid because of the duration, subject name, insecure algorithm or EKU, user should be able to skip the certificate invalid prompt since most of the self-signed certificate are for developing. 

Also I've modified the deployment YAMLs so that the self-signed certificates can show proper subject name instead of leaving it empty.

### DIFF
--- a/aio/deploy/recommended.yaml
+++ b/aio/deploy/recommended.yaml
@@ -199,6 +199,19 @@ spec:
             # If not specified, Dashboard will attempt to auto discover the API server and connect
             # to it. Uncomment only if the default does not work.
             # - --apiserver-host=http://my-address:port
+          env:
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
           volumeMounts:
             - name: kubernetes-dashboard-certs
               mountPath: /certs
--- a/aio/deploy/recommended.yaml
+++ b/aio/deploy/recommended.yaml
@@ -199,19 +199,6 @@ spec:
             # If not specified, Dashboard will attempt to auto discover the API server and connect
             # to it. Uncomment only if the default does not work.
             # - --apiserver-host=http://my-address:port
-          env:
-            - name: POD_NAME
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.name
-            - name: POD_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
-            - name: POD_IP
-              valueFrom:
-                fieldRef:
-                  fieldPath: status.podIP
           volumeMounts:
             - name: kubernetes-dashboard-certs
               mountPath: /certs
--- a/src/app/backend/cert/ecdsa/creator.go
+++ b/src/app/backend/cert/ecdsa/creator.go
@@ -26,6 +26,7 @@ import (
 	"math/big"
 	"net"
 	"os"
+	"time"
 
 	certapi "github.com/kubernetes/dashboard/src/app/backend/cert/api"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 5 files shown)


oneoneonepig:
Fixes https://github.com/kubernetes/dashboard/issues/2947

aio/deploy/recommended.yaml
```diff
@@ -199,6 +199,19 @@ spec:
             # If not specified, Dashboard will attempt to auto discover the API server and connect
             # to it. Uncomment only if the default does not work.
             # - --apiserver-host=http://my-address:port
+          env:
```
floreks: We can add that to head deployments, but I'd expect that for the recommended one people will use their own certificates. It would be even better for them to see some error and force them to use proper custom certificates instead of self-signed one.

oneoneonepig: I removed the environment variables in recommended deployment YAMLs. 

I understand your concern on forcing user to create their own certificates, though I still believe most of them will create another self-signed certificate just to skip the "unskippable" auto-generated self-signed certificate, or even finding other ways to skip the error prompt by using CentOS, Firefox or modifying flags in Chrome.

We all know using self-signed certificate is not a best practice, but I don't think this should block people from deploying a quickstart environment. If forcing them to generate a proper custom certificate is a must, then I would recommend to not generate a self-signed certificate at all and let the program exit with error.

WDYT

floreks: To be honest I haven't encountered any issues with an autogenerated certificate that we use right now. Anyway, even forcing someone to generate a custom self-signed certificate requires the user to be aware of what he is doing. It requires some knowledge, force you to read the documentation and all security issues that will come with it will be on the user.

oneoneonepig: ## Spec

- Operating System: Ubuntu 18.04.3 LTS
- Kubernetes version: v1.17.0
- Dashboard version: v2.0.0-rc3
- Browser: Chrome 79.0.3945.130 (Official Build) (64-bit) on Windows 10

## How to reproduce:

1. Install Kubernetes v1.17.0 with kubeadm and install network plugin
1. Install dashboard with command `kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/v2.0.0-rc3/aio/deploy/recommended.yaml`
1. Change the service type for `kubernetes-dashboard/kubernetes-dashboard` to `NodePort`
1. Connect to the dashboard through `https://NodeIP:NodePort`
dashboard-cert-error

src/app/backend/cert/ecdsa/creator.go
```diff
@@ -54,8 +55,15 @@ func (self *ecdsaCreator) GenerateCertificate(key interface{}) []byte {
...
 
+	notBefore := time.Now()
+	validFor, _ := time.ParseDuration("8760h")
+	notAfter := notBefore.Add(validFor)
+
+
 	template := x509.Certificate{
 		SerialNumber: self.generateSerialNumber(),
+		NotAfter: notAfter,
+		NotBefore: notBefore,
```
shu-mutou: To pass check on travis, add spaces before `notAfter|notBefore` and align its position with `self.generateSerialNumber(),` on the previous line:
```
		SerialNumber: self.generateSerialNumber(),
		NotAfter:     notAfter,
		NotBefore:    notBefore,
```

oneoneonepig: Executing the `gofmt -w src/app/backend/cert/ecdsa/creator.go` will do the trick

shu-mutou:
/lgtm
Thanks!! I could confirm that `NET::ERR_CERT_INVALID` was changed to `NET::ERR_CERT_AUTHORITY_INVALID` and we can proceed anyway, i.e. using link `Proceed to localhost (unsafe)`.

Screenshot from 2020-02-06 15-52-32

PR #4875, opened 2020-02-06
Fix development container

shu-mutou:
* Ensure installing dependencies.
* Enable to pass K8S_DASHBOARD_NPM_CMD as env to run-npm-on-container.sh.
* Fix typo.

### DIFF
--- a/aio/develop/npm-command.sh
+++ b/aio/develop/npm-command.sh
@@ -13,8 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Install dependencies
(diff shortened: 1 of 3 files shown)


floreks:
Actually instead of install Dashboard it should say install dependencies.

aio/develop/npm-command.sh
```diff
@@ -13,8 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Install dashboard.
```
maciaszczykm: `Install dependencies.`

PR #4878, opened 2020-02-06
Fix display of long URLs inside chips

### LINKED ISSUE #4809: Annotations with longer url getting truncated and not treated as a url
##### Environment

```
Installation method: helm chart
Kubernetes version: 1.15.8
Dashboard version: Dashboard v2.0.0-rc2
Operating system: Ubuntu/Debian
Node.js version ('node --version' output): NA
Go version ('go version' output): NA
```

##### Steps to reproduce

This is same with annotations in other resources. Just giving an example in configmap here.

Created a configmap with following config

```
{
    "kind": "ConfigMap",
    "apiVersion": "v1",
    "metadata": {
        "name": "gsree-test-configmap",
        "namespace": "test",
        "labels": {
            "app": "gsree-test-app",
            "chart": "gsree-test-chart"
        },
        "annotations": {
            "build.mydomain.io/url": "https://buildkite.com/domainname/infrastructure-deploy/builds/123456"
        }
    },
    "data": {
        "kubernetes-config-map.properties": "app = test"
    }
}
```

##### Observed result

annotations_example

The annotation looks like a link and if I click on that another empty text box appears which cannot be closed.

annotation_example_2

During this I checked the console log and I am seeing the below error.

```
polyfills.40201eb41e540c91bd50.js:1 Uncaught SyntaxError: Unexpected token h in JSON at position 0
    at JSON.parse (<anonymous>)
    at t.prettify_ (main.f12a4c3c443c78899ff2.js:1)
    at t.ngOnInit (main.f12a4c3c443c78899ff2.js:1)
    at main.f12a4c3c443c78899ff2.js:1
    at main.f12a4c3c443c78899ff2.js:1
    at Sb (main.f12a4c3c443c78899ff2.js:1)
    at Qb (main.f12a4c3c443c78899ff2.js:1)
    at Object.updateDirectives (main.f12a4c3c443c78899ff2.js:1)
    at Object.updateDirectives (main.f12a4c3c443c78899ff2.js:1)
    at Ab (main.f12a4c3c443c78899ff2.js:1)
```

Need to refresh the page later to come back to the normal display.

##### Expected result

The annotation with long url should be displayed as a url which can be access by clicking on the url

expected_annotation

##### Comments

NA

maciaszczykm:
Fixes #4809.

Zrzut ekranu 2020-02-6 o 13 09 19

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
     "rxjs-compat": "6.5.4",
     "sockjs-client": "1.4.0",
     "systemjs": "6.2.0",
+    "truncate-url": "^1.0.0",
     "tslib": "1.10.0",
     "web-animations-js": "2.3.2",
     "xterm": "3.14.5",
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "rxjs-compat": "6.5.4",
     "sockjs-client": "1.4.0",
     "systemjs": "6.2.0",
-    "truncate-url": "^1.0.0",
+    "truncate-url": "1.0.0",
     "tslib": "1.10.0",
     "web-animations-js": "2.3.2",
     "xterm": "3.14.5",
--- a/src/app/frontend/common/components/chips/component.ts
+++ b/src/app/frontend/common/components/chips/component.ts
@@ -18,6 +18,8 @@ import {StringMap} from '@api/backendapi';
 
 import {ChipDialog} from './chipdialog/dialog';
 
+const truncateUrl = require('truncate-url');
+
 export interface Chip {
   key: string;
   value: string;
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -130,6 +130,7 @@
     "rxjs-compat": "6.5.4",
     "sockjs-client": "1.4.0",
     "systemjs": "6.2.0",
+    "truncate-url": "^1.0.0",
```
floreks: Remove `^`

src/app/frontend/common/components/chips/component.ts
```diff
@@ -18,6 +18,8 @@ import {StringMap} from '@api/backendapi';
 
 import {ChipDialog} from './chipdialog/dialog';
 
+const truncateUrl = require('truncate-url');
```
floreks: isn't it possible to use import here? 

In example
`import * as truncateUrl from 'truncate-url'`

maciaszczykm: Unfortunately not. The typings are missing and the build fails.

package.json
maciaszczykm: Done

PR #4879, opened 2020-02-06
Increase version

maciaszczykm:


kharouny:
The image referenced in alternative.yaml is not accessible.

PR #4899, opened 2020-02-12
Bump node.js to 12.x LTS

shu-mutou:
Now dashboard supports nodejs 11.6, but it had been end-of-life already.
So bump nodejs to 12.x LTS.


shu-mutou:
/hold
travis seems to be able to install v12.16.0, so I'll specify it strictly.

PR #4921, opened 2020-02-19
Set default login mode to first authentication-mode option

shu-mutou:
authentication-mode options specified in backend, i.e. `token` (default), `basic` or both, are retrieved from frontend, and login mode selections are set in order of the options specified.

But `Kubeconfig` is set as default in frontend for now.

This commit enables to set default selection by operator's specification of authentication-mode.

Ref: #2641


src/app/frontend/login/component.ts
```diff
@@ -65,6 +65,11 @@ export class LoginComponent implements OnInit {
       .get<EnabledAuthenticationModes>('api/v1/login/modes')
       .subscribe((enabledModes: EnabledAuthenticationModes) => {
         this.enabledAuthenticationModes_ = enabledModes.modes;
+        Object.entries(LoginModes).forEach(([_, mode]) => {
+          if (mode === enabledModes.modes[0]) {
```
jeefy: Is there ever a condition or possibility of `enabledModes.modes` to have no entries? Always worry about hard coding an array index.

floreks: How about simply switching order in the template, so that kubeconfig option is always last? This way enabled mode will be first anyway.

shu-mutou: > How about simply switching order in the template, so that kubeconfig option is always last? This way enabled mode will be first anyway.

We need to set selectedAuthenticationMode to one of enabeldModes explicitly. Otherwise, if hard coded default selectedAuthenticationMode is not included in enabled mode, i.e. enabled mode was set as only `basic` but hard coded default is `token`, selection for token does not exist but input field is shown as `token`.

shu-mutou:
/hold
firefox on travis is too old (v64) and latest firefox (v74) in my development environment pass the test.
so bump it at first with #4927

shu-mutou:
/hold cancel
rebased master includes #4927

floreks:
@shu-mutou I have simplified the logic and updated argument description.

src/app/frontend/login/component.ts
```diff
@@ -65,6 +65,8 @@ export class LoginComponent implements OnInit {
       .get<EnabledAuthenticationModes>('api/v1/login/modes')
       .subscribe((enabledModes: EnabledAuthenticationModes) => {
         this.enabledAuthenticationModes_ = enabledModes.modes;
+        this.enabledAuthenticationModes_.push(LoginModes.Kubeconfig);
```
floreks: I think we can simplify the logic a bit here. Even if enabled modes will be empty, then we push kubeconfig at the end and select first option from the list as active one.

shu-mutou: Excellent! Thanks!!

Issue #4922, opened 2020-02-19
Firefox can not login


shu-mutou:
AFAIK, firefox had worked fine in Jan, but I can't login now.

Router does not work with firefox.
`navigate(['overview'])` at [here](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/login/component.ts#L107) was called but not jump to overview page.

floreks:
Will investigate that tomorrow.

shu-mutou:
Bumping angular modules from 8.0.3 to 8.2.14 was not effective for this. But chrome works fine.

Wilm0r:
Same here, I have to switch to chromium to get past the login screen..

shu-mutou:
I also see this. Angular 9 did not seem to fix this.

floreks:
Not sure why it does not work for you, but I have latest FF and on dev/prod it works just fine.

Wilm0r:
I could reproduce it even in a Private Window, using Firefox 68 (latest ESR packaged in for example Debian).

I'd be surprised if it mattered, but the Dashboard instance is running in k8s from the v2.0.0-rc5 recommended.yaml.

floreks:
This has been fixed on master so no release with the fix yet.

shu-mutou:
I could confirm that my latest firefix v73 on ubuntu and windows works fine today. 🎉 
Yesterday, dependencies might not be installed perfectly in my dev environment. 🤔

shu-mutou:
This fix will be released in rc6 or later.

PR #4929, opened 2020-02-20
Bump angular modules to 8.x latest

shu-mutou:
Almost angular modules are updated to 8.2.14.
This works fine in Chrome 80.0.3987.100, but does not fix login issue in Firefox 73 (#4922).


floreks:
Actually I wanted to do migration to angular 9. It enables ivy by default, has lots of improvements and should also speed up our build.

shu-mutou:
Agree. But it's time to finish work today. I'll tackle to version 9 tomorrow afternoon. 😉

floreks:
> Agree. But it's time to finish work today. I'll tackle to version 9 tomorrow afternoon.

I am working on this one.

PR #4930, opened 2020-02-20
Migrate angular to v9

### LINKED ISSUE #4893: ConfigMap view breaks after changing
##### Environment

```
Installation method: via YAML file
Kubernetes version: 1.17.2
Dashboard version: 2.0.0-rc5
Operating system: Windows 10
Browser: Firefox 72, Chrome 79
Node.js version ('node --version' output):
Go version ('go version' output):
```

##### Steps to reproduce

1. Select Config Maps
2. Select any configmap from list
3. Make sure you see pretty formatted configmap
изображение
4. Edit configmap
5. Save it
##### Observed result

изображение

##### Expected result

You can see edited configmap in the pretty view

floreks:
Fixes #4893

- Updated CRD data to also use `ace-editor` component instead of custom implementation.
- Removed `ng2-ace-editor` dependency and reused vanilla implementation.
- Tested Firefox 73.0 on prod and I was able to log in correctly (related to #4922). Will test once again after release and close if resolved.

## I18n pipeline update
Angular 9 introduced a couple of changes to the localization framework.

### Locale codes
New locale codes have to be used as described in: https://angular.io/guide/i18n#setting-up-the-locale-of-your-app

Example changes:
`zh-cn` -> `zh-Hans`
`zh-tw` -> `zh-Hant`

To be able to properly use `zh-Hant-HK` (codes with 3 parts) we have to wait for the issue https://github.com/angular/angular-cli/issues/17032 to be fixed. I have already renamed directories, removed symlinks as new i18n pipeline do not seem to like them. I have had to also slightly modify the locale handler in the backend.

 ### Frontend build
We do not need to build the frontend for every locale separately now. Localization configuration is defined in `angular.json` file and when `--localize` flag is passed to `ng`, localized frontend will be built.

@shu-mutou @zehuaiWANG please check if I have properly assigned new locale codes to languages.


floreks:
There seems to be an issue with prod build.

/hold

shu-mutou:
Locales except Chinese seem good. Since `zh-Hans` and `zh-Hant` aren't implemented in firefox and chrome, so only `zh` is actually available. I'm investigating the available way to use actual locale IDs.

shu-mutou:
I could not find the way to use `zh-*` locales that are sent via HTTP `Accept-Language` header today.... 😞

floreks:
@shu-mutou The new angular only supports new locale codes that are based on BCP47 regulations. Browsers should adapt them. I was actually using [language switcher](https://chrome.google.com/webstore/detail/advanced-page-language-sw/mhgjlbolkmcookadjpolimbddngjgbol) plugin to test if they work and it is also using new locale codes. For backward compatibility, we could add a map to our backend that will try to resolve `zh-cn` as `zh-Hans` if an exact translation is not found. It seems like an easy solution and should work for the time being. In the frontend we can use new language codes.

floreks:
I have fixed locale code mapping to also support old locales for now. I have also found a bug with default language package in Go and fixed locale matching logic. 

It is now ready for tests. Once angular releases patch version for the mentioned issue, we can bump it and restore support for missing languages.

maciaszczykm:
I will test it later.

hwdef:
I just noticed this PR, I think it's confusing to change the file directory like this. Can we simplify `zh-Hans-SG` `zh-Hans` ` zh-Hant-HK` `zh-Hant` and `zh`?

shu-mutou:
I'm tackling it at #4969. Please check it.

hwdef:
image

This is their structure. The ones in parentheses are `Accept-Language`.
Now we are missing tw and cn.
And many folders are duplicated, such as `zh-hans-sg` and `zh-hans` and `zh`.

If we want to distinguish by the smallest unit, we can keep the lowest directory. Which is reserved `zh-hans-CN`,`zh-hans-SG`,`zh-hant-HK`,`zh-hant-TW`. or `zh-CN`,`zh-SG`,`zh-HK`,`zh-TW`. 

Or we can keep the middle level. `zh-hans` and `zh-hant`.Then we link `CN`,`SG`,`HK`,`TW` to them.This way we only need to translate two files. But I don't recommend this way. Because although HK and TW belong to zh-hant, they also have some differences.

shu-mutou:
I'm considering to keep 'zh-Hans',  'zh-Hant' and 'zh-Hant-HK' in #4969, not to duplicate large translation files. Could you review it?

hwdef:
@shu-mutou 
I am sorry , I just noticed it. You discussed a lot and I need a little time to read.
Thanks for your PR.

PR #4962, opened 2020-02-28
Fix name of ClusterRole

dhoppe:
The ```ClusterRole``` should be ```cluster-admin```, because ```admin-Role``` does not exist.


floreks:
Thanks

/lgtm

PR #4963, opened 2020-02-28
Added support for ANSI color codes to logged terminal output.

greggbjensen:
.NET Core outputs ANSI color codes for it's standard logging, and these are not currently translated.  This leads to output that contains unprintable characters.

image

Adding support for ANSI color codes will allow for more readable logs, and consistent support for .NET Core as well as other applications.

image


package.json
```diff
@@ -117,6 +117,7 @@
     "@types/file-saver": "2.0.1",
     "@types/highlight.js": "9.12.3",
     "ace-builds": "1.4.8",
+    "ansi-to-html": "^0.6.14",
```
floreks: Remove `^`

src/app/frontend/common/pipes/safehtml.ts
```diff
@@ -15,17 +15,31 @@
...
 /**
  * Formats the given value as raw HTML to display to the user.
  */
 @Pipe({name: 'kdSafeHtml'})
 export class SafeHtmlFormatter {
   constructor(private readonly sanitizer: DomSanitizer) {}
 
-  transform(value: string): SafeHtml {
-    return this.sanitizer.sanitize(
+  transform(value: string, translate: string): SafeHtml {
```
floreks: s/translate/mode

I'd also make it an enum.

```typescript
enum TextMode {
    Default,
    Colored
}
```

Then to make `mode` really optional assign it a default value of `TextMode.Default` and use `switch(mode)` to handle the logic.

package.json
greggbjensen: The `^` is now removed.

src/app/frontend/common/pipes/safehtml.ts
greggbjensen: The enum and switch is now in place, thanks for taking a look at this so quickly.

src/app/frontend/common/pipes/safehtml.ts
```diff
@@ -15,17 +15,41 @@
...
+    let content = this.sanitizer.sanitize(
       SecurityContext.HTML,
       value.replace('<', '&lt;').replace('>', '&gt;'),
     );
+
+    // Handle conversion of ANSI color codes.
+    switch (mode) {
+      case TextMode.Colored:
+        content = ansiColor.toHtml(content.replace(/&#27;/g, '\u001b'));
+        result = this.sanitizer.bypassSecurityTrustHtml(content);
```
floreks: I am worried about using `bypassSecurityTrustHtml`. This might lead to security issues and allow execution of 3rd party code through logs.

greggbjensen: All of the HTML was sanitized in the prior step, the only new HTML would be that added by the ANSI color.  I believe this to be fully safe at and that point.

floreks:
It should be safe, given that we trust ansi library to be fully safe.

PR #4969, opened 2020-03-03
Fix locale mapping

### LINKED ISSUE #5027: Redandant large translation files in `zh*` locales

There are redandant large translation files in `zh*` locales.
i.e. `i18n/zh/messages.zh.xlf`, `i18n/zh-Hans/messages.zh-Hans.xlf` and `i18n/zh-Hans-SG/messages.zh-Hans-SG.xlf` are same.

We can specify translation file for each locale easily, so we should remove redandant files and specify it for locales.
e.g.
* leave `i18n/zh-Hans`
* specify `i18n/zh-Hans/messages.zh-Hans.xlf` for `zh` and `zh-Hans-SG` in angular.json to build.

#4969

shu-mutou:
Now, Angular>=9.0.5 supports `zh-Hant-HK` and `zh-Hans-SG`, so we can build for them.
However due to following reason, remove `zh-Hant-SG` and `zh` support for now.
* Chinese is not official in Singapore
* `zh` is not used often
* Translation files for `zh` and `zh-Hans-SG` is same as `zh-Hant` for now.

Also, `Match` of `golang.org/x/text/language` returns `zh-u-rg-cnzzzz` for `zh-CN` and
`zh-u-rg-txzzzz` for `zh-TW`, so we need to handle them properly before.

Since `Match` seems flaky and it's not easy to align specs between `Match`, Angular 9 and browsers, we decided not to use `Match`.

In conclusion, dashboard supports strictly matched locale IDs as followings:
* en (default)
* de
* fr
* ja
* ko
* zh-CN (zh-Hans)
* zh-TW (zh-Hant)
* zh-HK (zh-Hant-HK)

Also, support locales in lowercase and clean up `state="new"` in some translation files.
Closes #5027, #4259


angular.json
```diff
@@ -37,9 +37,17 @@
             "translation": "i18n/zh-Hans/messages.zh-Hans.xlf",
             "baseHref": "/"
           },
+          "zh-SG": {
```
floreks: Remove them as they are not supported by angular anyway and language codes will change. Like I have mentioned in https://github.com/kubernetes/dashboard/pull/4930 we will update them once angular releases patch for this. They have already fixed issue that I have created, so it should not take long.

go.sum
```diff
@@ -294,6 +294,7 @@ go.etcd.io/bbolt v1.3.3/go.mod h1:IbVyRI1SCnLcuJnV2u8VeU0CEYM7e686BmAb1XKL+uU=
 go.etcd.io/etcd v0.0.0-20191023171146-3cf2f69b5738/go.mod h1:dnLIgRNXwCJa5e+c6mIZCrds/GIG4ncV9HhK5PX7jPg=
 go.mongodb.org/mongo-driver v1.0.3/go.mod h1:u7ryQJ+DOzQmeO7zB6MHyr8jkEQvC8vH7qLUO4lqsUM=
 go.mongodb.org/mongo-driver v1.1.1/go.mod h1:u7ryQJ+DOzQmeO7zB6MHyr8jkEQvC8vH7qLUO4lqsUM=
+go.mongodb.org/mongo-driver v1.1.2 h1:jxcFYjlkl8xaERsgLo+RNquI0epW6zuy/ZRQs6jnrFA=
```
floreks: Did you do any go dep update?

src/app/backend/handler/localehandler.go
```diff
@@ -29,12 +30,14 @@ import (
 
 // TODO(floreks): Remove that once new locale codes are supported by the browsers.
 // For backward compatibility only.
-var localeMap = map[string]string{
-	"zh-cn": "zh-Hans",
-	"zh-sg": "zh-Hans-SG",
-	"zh-tw": "zh-Hant",
-	"zh-hk": "zh-Hant-HK",
-}
+// zh-SG, zh-Hans-SG, zh-HK and zh-Hant-HK are not supported by Angular 9 for now.
```
floreks: Leave the todo about it being a temporary solution and that it should be removed once browsers adopt new language codes.

floreks:
> Also, issue related golang/go#24211 is not exist as of now,

This is still not closed on the Go side. I have been testing it last week and it still happened to me. I'd leave the additional code that handles this case for now, until we can 100% confirm that this is fixed for everyone.

go.sum
shu-mutou: No. I did only `npm ci`. It may be contamination on my development environment.

angular.json
shu-mutou: I need some more investigation(#4973) on Angular 9 side. Please wait.

shu-mutou:
/hold
I investigated the issue golang/go#24211 at #4973. I'll fix my implements as comment. But I need some more investigation on Angular side.

angular.json
```diff
@@ -34,7 +34,7 @@
             "baseHref": "/"
           },
           "zh-Hans": {
-            "translation": "i18n/zh-Hans/messages.zh-Hans.xlf",
+            "translation": "i18n/zh/messages.zh.xlf",
```
floreks: I guess it should work the other way around. `zh` is generic language code without region, so it should be mapped to a specific region. `zh-Hans` should stay as `zh-Hans`.

package.json
```diff
@@ -89,8 +89,6 @@
       "ja",
       "ko",
       "zh",
-      "zh-Hans",
```
floreks: I'd leave all supported translations here, even if we use some translations for multiple locales. It's easier to handle that in the backend and find out which locales are actually supported. If we have to handle mappings both here and in the backend it becomes harder to manage.

src/app/backend/handler/localehandler.go
```diff
@@ -131,46 +122,40 @@ func (handler *LocaleHandler) determineLocalizedDir(locale string) string {
...
-	// If locale match is exact, then we have to manually look for proper locale code as language
-	// library contains a bug that returns invalid locale string.
-	// Related issue: https://github.com/golang/go/issues/24211
-	if confidence == language.Exact {
-		matchedLocale = ""
-		for _, l := range locales {
-			base, _ := tag.Base()
-			if l.String() == base.String() {
-				matchedLocale = l.String()
+	localeDir := ""
```
floreks: Why did you change the original logic here and removed string replacer? Previous version should work just fine. 

1. Replace 
2. Try to match the language
3. If an exact match, look for proper locale dir (can be omitted once go/text bug is fixed)
4. Check if dir exists and return

angular.json
shu-mutou: I think, top level language code should have translation file, and using parent's translation or own is children's choice. Also, translation `OWNERS` should be set by who can translate realy.

package.json
shu-mutou: From point of translator, maintaining same large translation files is messy. If translation for each region is realy needed, we can create it at the time. Also, mapping here should be set by translation `OWNERS`, I think.

About mapping on backend, we do not need it in new strategy. See my next comment.

src/app/backend/handler/localehandler.go
shu-mutou: I found the other solution that fix our issue without removing temporary fix for the bug and replacing string.
According to the spec of `language` module, we can use not only `Base` of the language but also `Parent`.
Using `Parent`, we can map all languages that we support now, without string replaces.
More details, see https://github.com/kubernetes/dashboard/issues/4973

Also, I added comments on my codes.

src/app/backend/handler/localehandler.go
```diff
@@ -131,46 +122,45 @@ func (handler *LocaleHandler) determineLocalizedDir(locale string) string {
...
+	case language.Exact:
+		// If locale match is exact, then we have to manually look for proper locale code as language
+		// library contains a bug that returns invalid locale string.
+		// Related issue: https://github.com/golang/go/issues/24211
+		// TODO(shu-mutou): Handling with `Extensions` is hack for the bug.
+		// So possively we need to fix again this.
+		// More details, see https://github.com/kubernetes/dashboard/issues/4973
+		if len(tag.Extensions()) > 0 {
+			// e.g. map `fr-FR` to `fr`, `zh-CN` to `zh`, `zh-TW` to `zh-Hant`
+			localeDir = getSupportedLocaleDir(tag.Parent().String(), locales)
```
floreks: I wouldn't want to rely on their internal implementation here and map specific regional locale to more generic ones. We do know already that they have an issue with returning proper locales. Relying too much on their internal logic might result in further issues. Let's stick to the basic check and rely only on confidence. If it's exact, then look for proper locale on our own as it was done before.

package.json
floreks: Translators do not have to manage other translation files as their only step would be to copy/paste it to another directory. We could even prepare a simple script that would keep those files in sync for us. It is also much easier to understand the final structure as every locale would be mapped to the same corresponding directory on the server. No more secret mapping different locales to the same directory.

From developer point of view, it was really confusing to understand whole i18n configuration before as some files contained all supported locales, some only the "actual ones" that we used (directories). I'd prefer to keep all files in sync and have the same list of locales/directories. It introduces small redundancy but is easier to understand.

angular.json
floreks: Those regional codes are there for a reason, even though we have tried to simplify it before, they can contain small differences for some words. That is why we shouldn't just take zh-XX and map it to the generic one (zh). 

It's like "mapping" car and scooter to a vehicle. You will lose features that are specific to each one of them. We should do this the other way around and say that for most people vehicle of choice (zh) will be a car (zh-Hans).

shu-mutou:
@floreks I found angular/angular-cli#17033 that fixes your bug report (angular/angular-cli#17032) was introduced in angular/cli 9.0.5 . And angular/cli 9.0.5 was introduced in dashboard.
I'll check it and consider the best way again with your comments.

src/app/backend/handler/localehandler.go
shu-mutou: I updated my investigation #4973 after update Angular to 9.0.5, implementations are getting better. 😉

src/app/backend/handler/localehandler.go
shu-mutou: I found the issue is not simply bug in `golang.org/x/text/language`. If we can set supported locales properly, it matches locales properly. Matcher maps `zh-SG` and `zh-HK` to `zh-Hans-SG` and `zh-Hant-HK`. And now, in Angular>=9.0.5 supports `zh-Hans-SG` and `zh-Hant-HK`, so we can build with them and we do not need to replace them. However, we need to replace `zh-CN` and `zh-TW` yet.

package.json
shu-mutou: It's not difficult to understand these settings, i.e. for create/update translation files in package.json and set locales for build in angular.json and locale_conf.json. I updated docs also, please see it.

src/app/backend/handler/localehandler.go
floreks: It is a bug if it returns weird `xx-yy-zzzzzzz-n` strings.

shu-mutou: After all, it seems weird but does not seem to be a bug. See more,  https://godoc.org/golang.org/x/text/language#NewMatcher

src/app/backend/handler/localehandler.go
```diff
@@ -116,6 +108,17 @@ func (handler *LocaleHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	if acceptLanguage == "" {
 		acceptLanguage = r.Header.Get("Accept-Language")
 	}
+
+	// TODO(floreks): Remove that once new locale codes are supported by the browsers,
+	// or golang.org/x/text/language map properly.
+	// For backward compatibility only.
+	localeMap := strings.NewReplacer(
+		"zh-CN", "zh-Hans",
```
floreks: Why `zh-sg` and `zh-hk` mappings are not here?

shu-mutou: I found that Match returns tag as `zh-Hant-HK` for `zh-HK`, and `zh-Hans-SG` for `zh-SG`, so we don't need to replace them. However we need to replace `zh-CN` and `zh-TW`.

floreks:
I have performed detailed tests to see how complex it is to determine what will be the actual directory used for translation based on input locale.

Zrzut ekranu z 2020-03-12 15-00-47

We can see that:
- Some mappings are very complex and it is hard to determine which directory will be used as a result.
- There is a bug with the different locale being returned for the same input locale.
- There are many indirect mappings that are not easily understandable unless the developer investigates the logic behind it.
- Worst case scenario we have an indirect mapping to base language in the backend and then again to regional one in the frontend. This was very confusing.

Because complexity level here is quite high and I'd like to have really straightforward mapping even if we will have to introduce some redundancy, this should be changed.

maciaszczykm:
> There is a bug with the different locale being returned for the same input locale.

> Because complexity level here is quite high and I'd like to have really straightforward mapping even if we will have to introduce some redundancy, this should be changed.

This sounds good to me. The main thing is that we should be consistent even if it requires some redundant configuration IMO.

shu-mutou:
@floreks I have not seen browsers that send locale ID in lowercases only, e.g. `zh-cn` or `zh-tw`. If such browser exists, I need to add them into `NewReplacer`. Do such browser exists?

@maciaszczykm Do you mean it's better to keep configs like follows rather than to keep same large translation files? If so, that's how I'd like to do.
* `xliffmergeOptions` in `package.json`: to merge extracted strings into each translation files.
* `angular.json`: building dashboard for each locales
* `aio/locale_conf.json`: supported locales for dashboard binary

floreks:
> I have not seen browsers that send locale ID in lowercases only, e.g. zh-cn or zh-tw. If such browser exists, I need to add them into NewReplacer. Do such browser exists?

There are dozens of different browsers. We have to consider lowercase as a potential option also. Especially, that Go text library returns different results for uppercase/lowercase options.

> Do you mean it's better to keep configs like follows rather than to keep same large translation files?

I think it means that it is better to introduce small redundancy over complexity to simplify the logic.

shu-mutou:
Added support for locales in lowercase

shu-mutou:
Also, improve tests

floreks:
I see that mine and @maciaszczykm comments' are not applied. It still uses this confusing way of indirect mapping.

SataQiu:
Expect this PR to solve the language problem 👍 
Ref: #5021

shu-mutou:
> I see that mine and @maciaszczykm comments' are not applied. It still uses this confusing way of indirect mapping.

It's just specifying translation files to build for locales in angular.json. It's really easy. But there is not much time until release date, so leave redandant large translation files and defer discussion of non-functional requirements for now.

@zehuaiWANG @hwdef @tanjunchen @chenrui333 Please participate discussion about this in #5027.

shu-mutou:
Reflected current conclusion in #4969.

hwdef:
I agree with these,
@shu-mutou Thanks for your contributions!

src/app/backend/handler/localehandler.go
```diff
@@ -57,7 +50,7 @@ func CreateLocaleHandler() *LocaleHandler {
 		glog.Warningf("Error when loading the localization configuration. Dashboard will not be localized. %s", err)
 		locales = []language.Tag{}
 	}
-	return &LocaleHandler{SupportedLocales: locales}
+	return &LocaleHandler{SupportedLocales: locales, LocaleDirectories: getLocaleDirs()}
```
floreks: I don't think we need `LocaleDirectories` at all. Old way of looking for matched locale was fine. Exact match + mapping and fallback to english if not found.

shu-mutou: To provide language service as much as we can, as you said too, this change provides base locale for Exact and High confidence that matched not strictly. Requirements makes logic complicated. It is unavoidable. Please see comments in source code. If we provide only strictly matched locales, we don't need to use confusing text/language module.

floreks: If we no longer have `zh` support then high level match and falling back to base is obsolete. Mapping + exact match and fallback to english is enough right now.

shu-mutou: Agree about `zh`. I mean we can support following mappings.
* mapping `en-US` to `en`, `fr-FR` to `fr`, `ko-KR` to `ko`, etc in Exact match.
* mapping `en-GB` to `en`, `fr-CA` to `fr`, etc in High match.

Do you mean we don't need support them?

floreks: We need to apply the same logic to other languages as to `zh`. The same principle applies to all of them.

shu-mutou: I think there are gaps between the principle, implementations in Angular, go module and browsers and user requirements.
As of now, carry over conclusions, prioritize the new release and wait for user feedbacks.

floreks: In example, our `en` is actually `en-US`. There is no guideline as to how to handle base language and we usually have to decide on the region. Some countries have general locale, like i.e. Germany (`de`), but it has its' regional code also (`de-DE`).

In general, I'd like to avoid multiple indirect mappings due to all those subtle differences. It all makes handling localization more complex. We should decide and follow one strategy.

My idea would be to:
1. Map base language code `zh`, `de`, `fr` to the most common regional code. This is the only code that I'd be okay with to map in `angular.json` to a different directory. We'd have `zh-Hans`, `de-DE`, `fr-FR`. This would have to be noted clearly inside our documentation.
2. Always try to find an exact match first.
3. If exact match not found, return base.

This way even if base is returned and we do not support it, it will automatically fallback to English.

shu-mutou: Your idea 1 did not go well.
* Angular (https://github.com/angular/angular/tree/master/packages/common/locales) does not support `de-DE`, `fr-FR`, `ja-JP`, `ko-KR`. So warnings are shown while building locale for them. Angular seems to use `de` instead `de-DE`, etc. Only `zh*` locales seems odd.
* Also, in Angular docs https://angular.io/guide/i18n , new locales do not have region code for the most common region. e.g. `zh-Hans-CN` -> `zh-Hans`. `Hans` is not region.
* Many links take lots of memory while building binary. In my environment, it had taken over 16GB, so I could not finish the build.

floreks: > Angular (https://github.com/angular/angular/tree/master/packages/common/locales) does not support de-DE, fr-FR, ja-JP, ko-KR. So warnings are shown while building locale for them. Angular seems to use de instead de-DE, etc. Only zh* locales seems odd.

I'll check. It should follow BCP47 standard and it allows regional codes.

> Also, in Angular docs https://angular.io/guide/i18n , new locales do not have region code for the most common region. e.g. zh-Hans-CN -> zh-Hans. Hans is not region.

True, `Hans` is actually a script code, not a region identifier. I guess Chinese is a bit more complicated than other languages. We still have to decide what to do here for the base locale.

> Many links take lots of memory while building binary. In my environment, it had taken over 16GB, so I could not finish the build.

With Angular 9 I don't believe it's true. Right now build is very fast and during the build, it did not use more than 3GB of RAM for me. It seems to be more CPU heavy.

floreks: It seems that not supporting `en-US` and `de-DE` might be yet again a bug on angular side.
https://stackoverflow.com/questions/59514703/angular-9-locale-data-for-en-us-cannot-be-found-no-locale-data-will-be-includ

I'd really like to avoid so many workarounds, but it seems that there are issues on every step (angular, golang text library). :slightly_frowning_face:

shu-mutou: Current changes follow current Angular 9 spec. If we wait for consistency of all steps, we won't release new dashboard. For now, let's merge this and release new dashboard. 😉

src/app/backend/handler/localehandler.go
```diff
@@ -131,46 +147,32 @@ func (handler *LocaleHandler) determineLocalizedDir(locale string) string {
...
-	if mappedLocale, ok := localeMap[locale]; ok {
-		locale = mappedLocale
-		tags, _, err := language.ParseAcceptLanguage(locale)
-		if (err != nil) || (len(tags) == 0) {
-			return language.Tag{}, language.No, err
+func (handler *LocaleHandler) getSupportedLocaleDir(locale string) string {
+	assetsDir := getAssetsDir()
+	locales := handler.SupportedLocales
+	localeDirs := getLocaleDirs()
+	for _, l := range locales {
```
floreks: Use map here for clarity and to reduce complexity from O(n^2) to O(n)

floreks:
I have simplified the logic. We don't need to do exact match since we anyway map to our supported locale dir afterwards. We can directly parse header and search for supported locale.

shu-mutou:
Matcher in text/language module seems flaky and confuses us. Agree with not using it.

shu-mutou:
@floreks @maciaszczykm Thank you so much for lots of discussion on this. 😄

Issue #4972, opened 2020-03-04
Line graphs are not getting narrower


shu-mutou:
##### Environment

```
Installation method: development container
Kubernetes version: - 
Dashboard version: master (2020-3-3)
Operating system: -
Node.js version ('node --version' output): -
Go version ('go version' output): -
```

##### Steps to reproduce
1. Show `Nodes` or `Overview` page.
2. Expand window until the width of line graphs increase as following image.
3. Getting narrower window.
![before](https://user-images.githubusercontent.com/12838129/75835072-37a8a980-5e01-11ea-8f5a-e8269551abe1.png)

##### Observed result
Line graphs are not getting narrower.
![screenshot-192 168 1 120_20443-2020 03 04-10_14_37](https://user-images.githubusercontent.com/12838129/75835051-30819b80-5e01-11ea-8be7-affa2b39b80c.png)

##### Expected result
Line graphs narrow responsive with window width.

PR #4980, opened 2020-03-05
Fix line graph width

shu-mutou:
`ngx-charts` need `width` param to adopt it to graph width, so we need to set `width` explicitly.
Especially, firefox does not work `width` with `flex`, so change `flex` to `width`.

Closes: #4972


floreks:
This can be used purely by using flex layout and we actually do not need additional styles here. I'll update it.

PR #5000, opened 2020-03-12
Hide area under zero in graph

### LINKED ISSUE #4995: Setting a minimal "Max Y axis" for (CPU) graphs.
I propose that the metrics graph gets some constraints on the min/max of the Y axis.

For the CPU graph I propose to have a 'smallest' range from 0 to 0.1

##### Why is this needed
In a namespace with very low load the graphs are misleading and give the impression of huge fluctuations when almost nothing happens.

Using v2.0.0-rc5 I see this in a namespace with no load.
image

When there is very little load I see this:
image

and a few minutes later
image

image

##### Comments
I think the Y axis minimum must be fixed on 0 and the lowest possible value for the top of the Y axis on 0.1 (or 0.5).

shu-mutou:
Set `yScaleMax` not to show area under zero, if the maximum metrics data is zero.

Closes #4995

### DIFF
--- a/src/app/frontend/common/components/graph/component.ts
+++ b/src/app/frontend/common/components/graph/component.ts
@@ -34,7 +34,7 @@ export class GraphComponent implements OnInit, OnChanges {
   customColors = {};
   yAxisLabel = '';
   yAxisTickFormatting = (value: number) => `${value} ${this.yAxisSuffix_}`;
-  yScaleMax = '';
+  yScaleMax = 0;
 
   private suffixMap_: Map<number, string> = new Map<number, string>();
   private yAxisSuffix_ = '';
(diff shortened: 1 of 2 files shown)


src/app/frontend/common/components/graph/template.html
```diff
@@ -20,6 +20,7 @@
                        [showYAxisLabel]="true"
                        [yAxisLabel]="yAxisLabel"
                        [yAxisTickFormatting]="yAxisTickFormatting"
+                       [yScaleMax]="yScaleMax"
```
floreks: This is not a string, but a number.

src/app/frontend/common/components/graph/component.ts
```diff
@@ -92,6 +94,15 @@ export class GraphComponent implements OnInit, OnChanges {
...
     this.yAxisSuffix_ = highestSuffix;
 
+    // Find out max value
+    this.metric.dataPoints.forEach(value => {
+      if (maxValue < value.y) {
+        maxValue = value.y;
+      }
+    });
+    // Set yScaleMax not to show area under zero
+    this.yScaleMax = maxValue ? '' : '1';
```
floreks: It will basically always be `this.yScaleMax = '1'` as `maxValue` will always have some value if there are any points. If there are no points graphs are not displayed anyway.

floreks:
I'll fix that. It's better to simply set `yScaleMin` to `0`.

Issue #3058, opened 2018-05-26
need a TLS MinVersion arg 


wu105:
##### Environment

```
Dashboard version: v1.7.1
Kubernetes version: v1.9.1
Operating system: Oracle Linux
Node.js version: N/A
Go version: N/A
```
This is a feature request: when serving over https, dashboard offers TLSv1.0 thus fails our security scan.  Would like to have an argument for higher TLS MinVersion.
The TLS ciphers offerred is not currently a problem for us, but that could change, thus an argument to control TLS cipher set might be useful. 
##### Expected result

When TLS MinVersion is set to TLSv1.2, for instance, the dashboard would offer TLSv1.2 as the lowest version during SSL hand shake, thus pass security scan which requires TLSv1.2 or above.

wu105:
/remove-lifecycle stale
/remove-lifecycle rotten

This is minor but when we need to pass the security scan we had to resort to creating our own version.
We would have the same issue when upgrade.

maciaszczykm:
/lifecycle frozen

@wu105 Perhaps you can prepare some pull request for it?

wu105:
@maciaszczykm What I did is quite messy, did not do anything on TLS ciphers nor adding arguments, thus it may be better to share what I did to get it to pass our security scan.  We started with cloning version v1.7.1 source.

- [ ] hardwired MinVersion 
diff --git a/src/app/backend/dashboard.go b/src/app/backend/dashboard.go
index 5d7b993..f3fde4f 100644
--- a/src/app/backend/dashboard.go
+++ b/src/app/backend/dashboard.go
@@ -15,6 +15,8 @@
 package main
 
 import (
+	"crypto/tls"
+
 	"flag"
 	"fmt"
 	"log"
@@ -109,7 +111,19 @@ func main() {
 	if *argCertFile != "" && *argKeyFile != "" {
 		log.Printf("Serving securely on HTTPS port: %d", *argPort)
 		secureAddr := fmt.Sprintf("%s:%d", *argBindAddress, *argPort)
-		go func() { log.Fatal(http.ListenAndServeTLS(secureAddr, *argCertFile, *argKeyFile, nil)) }()
+                servingCert, err := tls.LoadX509KeyPair(*argCertFile, *argKeyFile)
+                if err != nil {
+		       handleFatalInitError(err)
+	        }
+                var servingCerts []tls.Certificate;
+                servingCerts = []tls.Certificate{servingCert}
+		server := &http.Server{
[...]

PR #5013, opened 2020-03-19
Set MinVersion for TLS

shu-mutou:
Limits TLS <= v1.1 due to security reason.

Closes: #3058


floreks:
I think we can go with that. v1.2 is there for a while, and seem to be supported by most browsers.

https://caniuse.com/#feat=tls1-2

Issue #4755, opened 2020-01-10
How to display Chinese on my dashboard 2.0-beta8?


wanghaiqing2015:
How do I set parameters in yaml

floreks:
You do not. It's based on your browser's locale. We might have an issue with choosing proper locale right now as discussed in #4734. We will prepare the correct directories soon. Right now you should be able to switch language by using locale switcher extensions for your browser.

hwdef:
Change your browser language to `zh`, neither `zh-CN` nor `zh-TW`.

wanghaiqing2015:
tks

PR #5021, opened 2020-03-21
Add zh code alias to make dashboard compatible with more browsers

SataQiu:
In fact, many browsers don't set the `Accept-Language` to lowercase.
For instance, Chrome's `Accpt-Language` is `zh-CN` rather than `zh-cn` for simplified Chinese by default, as well as Firefox.

With this in mind, we'd better make dashboard compatible with this common language format.

This PR adds some zh code aliases to make dashboard compatible with more browsers.

图片

Ref: #4755


floreks:
Check open PRs. There is already one that aims to solve this. It's more complex than that.

/close

PR #5024, opened 2020-03-23
update translation

hwdef:
Ref: #4259


hwdef:
/hold 
for zh-hant.

hwdef:
@tanjunchen @zehuaiWANG 
Thanks for your review.

Issue #4282, opened 2019-09-12
Dashboard does not remember namespace after adding resource.


resuls:
### Issue details
Dashboard does not remember last selected namespace after adding a resource.

### Environment
Dashboard version: v2.0.0-beta3+0.g677aff93
Kubernetes version: v1.15.0
Operating system: Debian
Node.js version: n/a
Go version: n/a

### Steps to reproduce
Select a namespace other than the default.
Build a simple YAML of any resource (example a secret), upload it through the YAML upload dialog.

### Observed result
After the successful creation of a resource, the namespace will be set as `default` again.

### Expected result
It should remember the last selected namespace after creation.

PR #5033, opened 2020-03-26
Fix deply redirect and error messages

maciaszczykm:
Solves the issue mentioned in https://github.com/kubernetes/dashboard/pull/4328#issuecomment-604432517.


floreks:
Fixes #4282

PR #5039, opened 2020-03-28
Added quick edit icon to Secrets page with dialog to decode and update a data value

greggbjensen:
I added a quick edit next to the eye icon of the Secret page, that allows you to open a modal for one entry in the secret and save. This is direct text unencoded, and should provide a good user experience.

image

image

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -33,6 +33,10 @@
           <context context-type="sourcefile">../src/app/frontend/common/dialogs/editresource/template.html</context>
           <context context-type="linenumber">43</context>
         </context-group>
+        <context-group purpose="location">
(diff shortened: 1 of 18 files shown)


floreks:
This is not user friendly. In the current state, edit dialog for secret should be modified. The best way would be to simply be able to show input/text area instead of read-only secret value and be able to edit it. Something like an edit mode for a single value.

greggbjensen:
I had thought about the inline edit, but I felt it would be inconsistent with the edit experience everywhere else on the site.  It is very quick to click the pencil and then click update.  Are you sure this is not the experience you want?

greggbjensen:
I can work on the other experience, I would probably cancel this pull request and make a new one.  Just let me know.

greggbjensen:
I re-implemented it as requested on a separate pull request:

**Implemented inline editing for data entries on Secrets page.** #5040

PR #5040, opened 2020-03-29
Implemented inline editing for data entries on Secrets page.

greggbjensen:
I updated the read only Data entries on the Decrets page to make them a text box with the ability to quickly update a single unencoded entry.

image


floreks:
Seems better. I'll review when I'll have some time.

shu-mutou:
/lgtm
This  works fine. Great!

floreks:
Why was it merged? I did not review it yet and it is not done as I have said. Entries should not be editable by default. There should be view/edit mode available. Eye icon should only show read-only inline value. Text box is too big and does not scale to single-line values.

floreks:
@greggbjensen will you work on improving that or should I revert for now?

maciaszczykm:
Let's revert if there will no answer today.

greggbjensen:
@floreks I can get those changes in.  I'll submit another pull request.  Sorry I'm in a GMT-7.

greggbjensen:
Can you re-merge if I update this pull request? 

@floreks you would like the eye and the pencil icons, leaving the eye with the same behavior and the pencil with the new behavior?

floreks:
Yes, edit icon instead of edit by default. Also, you can check if entry is a single line or multi-line value and use input/text area accordingly. I'd also try to scale the text area, because it takes too much space for small files.

greggbjensen:
Okay, will do.  I'll update this pull request.

maciaszczykm:
> Can you re-merge if I update this pull request?

I am not sure if this is possible. Opening the pull request is on your side so you can check and if not do it from the new branch.

greggbjensen:
@maciaszczykm no problem, I will do a new branch and pull request as an enhancement to this.  Then I will link it back here.

greggbjensen:
@floreks and @maciaszczykm here are all the requested changes:

**Inline edit secret enhancements** #5046

Issue #4205, opened 2019-08-17
Have a Warning Icon on a StatefulSet if the Pods can't be created


kimwnasptd:
##### What would you like to be added
Have a warning icon instead of the green circle check for `StatefulSets` if the Pods cannot be created.

For example, if I create a `StatefulSet` with a non existent `ServiceAccount`, then the `StatefulSet` controller will not be able to create the Pod to the API Server. The problem is that the `StatefulSet`'s Status doesn't reflect that error and only shows `replicas: 0`. But, we can get this info by looking at the StatefulSet's `Events`.

##### Why is this needed
If the StatefulSet controller can't create the Pods, then the UI will show a green check which implies that the STS is OK. 

Althought, on the card it shows that there are 0/1 Pods available and I can see what is wrong if I click on the STS and see its Events. But I think it would be useful if the icon would be a warning to indicate that something is wrong.

##### Comments

You can check this behavior by creating this StatefulSet
```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  ports:
  - port: 80
    name: web
  clusterIP: None
  selector:
    app: nginx
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  selector:
    matchLabels:
      app: nginx # has to match .spec.template.metadata.labels
  replicas: 1
  serviceName: "nginx"
  template:
    metadata:
      labels:
        app: nginx # has to match .spec.selector.matchLabels
    spec:
      terminationGracePeriodSeconds: 10
      serviceAccountName: nonexistent-service-account-test
      containers:
      - name: nginx
        image: k8s.gcr.io/nginx-slim:0.8
        ports:
        - containerPort: 80
          name: web
        volumeMounts:
        - name: www
          mountPath: /usr/share/nginx/html
```

PR #5055, opened 2020-04-08
Adds support for showing warning icon in StatefulSet

Harkishen-Singh:
Signed-off-by: Harkishen Singh <harkishensingh@hotmail.com>

Fixes: #4205 

**Explanatory Image:**

Screenshot from 2020-04-09 12-41-25

### DIFF
--- a/src/app/frontend/common/components/resourcelist/statefulset/component.ts
+++ b/src/app/frontend/common/components/resourcelist/statefulset/component.ts
@@ -78,25 +78,15 @@ export class StatefulSetListComponent extends ResourceListWithStatuses<
   }
 
   isInPendingState(resource: StatefulSet): boolean {
-    return (resource.podInfo.warnings.length === 0 && resource.podInfo.pending > 0) || (resource.podInfo.running !== resource.podInfo.desired);
+    return resource.podInfo.warnings.length === 0 && (resource.podInfo.pending > 0 || resource.podInfo.running !== resource.podInfo.desired);
   }
 
   isInSuccessState(resource: StatefulSet): boolean {
-    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending === 0 && (resource.podInfo.running === resource.podInfo.desired);
-  }
-
-  getDisplayStatus(podInfo: PodInfo): string {
-    let msgState = 'Running';
-
-    if (podInfo.pending > 1 || (podInfo.running !== podInfo.desired)) {
-      msgState = 'Pending';
-    }
-
-    return msgState;
+    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending === 0 && resource.podInfo.running === resource.podInfo.desired;
   }
 
   getDisplayColumns(): string[] {
-    return ['statusicon', 'name', 'labels', 'status', 'pods', 'age', 'images'];
+    return ['statusicon', 'name', 'labels', 'pods', 'age', 'images'];
   }
 
   hasErrors(statefulSet: StatefulSet): boolean {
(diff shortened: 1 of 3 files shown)


floreks:
I'd argue that when desired nr of pods is different than actual number it does not necessarily mean there is something wrong with it. It might be upscaling/downscaling. This case is definitely not a warning.

Harkishen-Singh:
@floreks sure. I actually implemented that way since I did not any other notable things in response from the API (`getData()`). So, I thought that the warning symbol should be kind of at that can signify anything different than the desired value (noticed in gke, I think so). 

What do you suggest should be the recommended way here? @floreks

floreks:
Our error state is actually a warning state. If there are warning/error events associated with any resource, an error icon will appear and proper message will be available. I don't think a warning status is needed here.

Harkishen-Singh:
@floreks does that mean, the _green check_ in `0/1` Pods is acceptable?

floreks:
This is weird because 0/1 should mean that there is a pending pod waiting to be created. You can extend the pending/success checks to check for pending pods and if desired/actual pods match. In such case it should show pending status.

Harkishen-Singh:
Got your point. However, in this case, the pods will forever remain in a pending case. Hence I thought that showing a warning or alert icon will be required, as the issues say.

Harkishen-Singh:
@floreks updated the logic as per discussion.

Screenshot from 2020-04-09 12-41-25

src/app/frontend/common/components/resourcelist/statefulset/component.ts
```diff
@@ -74,24 +73,30 @@ export class StatefulSetListComponent extends ResourceListWithStatuses<
...
-    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending > 0;
+    return (resource.podInfo.warnings.length === 0 && resource.podInfo.pending > 0) || (resource.podInfo.running !== resource.podInfo.desired);
   }
 
   isInSuccessState(resource: StatefulSet): boolean {
-    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending === 0;
+    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending === 0 && (resource.podInfo.running === resource.podInfo.desired);
+  }
+
+  getDisplayStatus(podInfo: PodInfo): string {
```
floreks, line 88: Remove that. Status column is reserved only for pods.

src/app/frontend/common/components/resourcelist/statefulset/component.ts
```diff
@@ -74,24 +73,30 @@ export class StatefulSetListComponent extends ResourceListWithStatuses<
...
-    return resource.podInfo.running !== resource.podInfo.desired;
-  }
-
   isInErrorState(resource: StatefulSet): boolean {
     return resource.podInfo.warnings.length > 0;
   }
 
   isInPendingState(resource: StatefulSet): boolean {
-    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending > 0;
+    return (resource.podInfo.warnings.length === 0 && resource.podInfo.pending > 0) || (resource.podInfo.running !== resource.podInfo.desired);
```
floreks, line 81: Pending and desired/running part should be taken into common parenthesis. It should be

_if there are no warnings and (there are pending pods or running !== desired)_

Adjust success state check accordingly.

src/app/frontend/common/components/workloadstatus/component.ts
```diff
@@ -35,6 +35,11 @@ export class WorkloadStatusComponent {
   @Input() resourcesRatio = emptyResourcesRatio;
   colors: string[] = [];
 
+  constructor() {
```
floreks, line 38: Remove that.

Harkishen-Singh:
@floreks done.

PR #5076, opened 2020-04-17
Use `Created` translation for `Creation time`

shu-mutou:
`Created` is commonly used as item title for `creationTimestamp` now.
So change `Creation time` for metadata creation timestamp to `Created`,
and use its translation for `Created`.


shu-mutou:
@floreks @maciaszczykm Please merge this before new release. Thanks! 😉

Issue #5105, opened 2020-05-06
Auth with Authorization header problem


ruruiye:
We use oauth2_proxy which connect to okta with OIDC method, after the verification, the dashboard will get an  `Authorization` header from the proxy, in the previous version `v1.10.1`, the dashboard will response to logged with Auth header and all the api request will be limited to the default service account role.

after upgrade to `v2.0.0`, I can still login to the UI with the same config, and see the logged with Auth Header in the user icon,  but for all the api request, it will return 401.

I have to provide the default service account role token to the `Authorization` header manually in the ingress of dashboard. 

is this expected? is their anyway to use the default service account role when dashboard is behind an auth proxy? can we hide the login page in this case by default?

Please see my deployment config below:
oauth2_proxy args in deployment
```
            - '--http-address=0.0.0.0:4180'
            - '--cookie-domain=.example.com'
            - '--cookie-secure=false'
            - '--email-domain=*'
            - '--oidc-issuer-url=https://example.okta.com'
            - '--provider=oidc'
            - '--redeem-url=https://example.okta.com/oauth2/v1/authorize'
            - >-
              --redirect-url=https://example.com/oauth2/callback
            - '--set-authorization-header=true'
            - '--silence-ping-logging=true'
            - '--upstream=file:///dev/null'
            - '--whitelist-domain=.example.com'
            - '--config=/etc/oauth2_proxy/oauth2_proxy.cfg'
```
deployment options of kubernetes dashboard:
```
            - --namespace={{ .Release.Namespace }}
            - --insecure-bind-address=0.0.0.0
            - --insecure-port=9000
            - --enable-insecure-login
            - --enable-skip-login
```
ingress for kubernetes dashboard that is not work (return 401 on all api request):
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    # See the bottom of https://oauth2-proxy.github.io/oauth2-proxy/configuration
    nginx.ingress.kubernetes.io/auth-response-headers: Authorization
    nginx.ingress.kubernetes.io/auth-signin: https://$host/oauth2/start?rd=$escaped_request_uri
    nginx.ingress.kubernetes.io/auth-url: https://$host/oauth2/auth
    nginx.ingress.kubernetes.io/configuration-snippet: |
      auth_request_set $name_upstream_1 $upstream_cookie_name_1;

      access_by_lua_block {
        if ngx.var.name_upstream_1 ~= "" then
[...]

PR #5113, opened 2020-05-10
add `force-skip-login` flag to automatically skip login

rokrokss:
We use oauth2-proxy to authorize into kubernetes dashboard at ingress and always skip login, I'm sure there're more groups using the dashboard in this favor. I propose a flag to automatically skip login.


floreks:
If you are using auth proxy, then you are probably passing Authorization header instead of manually logging in. This is what you want then: https://github.com/kubernetes/dashboard/issues/5105#issuecomment-624504263

In this case login view is not shown.

rokrokss:
@floreks oh thanks, I wasn't aware of this.

PR #5115, opened 2020-05-10
Fix list default sort problem

double73:
Default sorting does not work with the "Created" column when loading the page.
When modifying #5076, the matSortActive value change is missing.


floreks:
Nice catch!

/lgtm

PR #5171, opened 2020-05-24
Add update note to README.md

sixcorners:
Is this right? When I run kubectl apply with the new file I see the script file names change on the login page.


floreks:
It doesn't really matter. Deploy does not have to necessarily mean deploy a brand new from scratch, it also can mean deploy a new version. It's fine as it is now.

/close

PR #5174, opened 2020-05-25
Update Japanese translations

shu-mutou:


i18n/ja/messages.ja.xlf
```diff
@@ -2897,11 +2897,11 @@
               here
             <x id="CLOSE_LINK" ctype="x-a" equiv-text="&lt;/a>"/>.
           </source>
-        <target state="new">
-            Insecure access detected. Sign in will not be available. Access Dashboard securely over HTTPS or using localhost. Read more
+        <target>
+            安全でないアクセスを検知しました。サインインは無効になります。HTTPS 経由、あるいは localhost を使用してダッシュボードに安全にアクセスしてください。詳細は、
```
atoato88, line 2901: Is it better "検出", not "検知" ?

atoato88:
Almost LGTM, one nit, please see comment.

i18n/ja/messages.ja.xlf
shu-mutou: I think "検知" is better, because that means machanical detection. "検出" has nuance that Humans actively detect something using machines etc.

atoato88:
/lgtm
/approve
I rethought that "検知" is also OK, because the meaning of "検知"  has "detect" too.

PR #5182, opened 2020-05-26
Fix can-i check for settings configmap

### LINKED ISSUE #5170: `cani` API endpoint returns `false` and cause frontend to refresh content every 5s
##### Environment

```
Installation method: docker image deployed directly to cluster
Kubernetes version: v1.18.1
Dashboard version: v2.0.0-rc5+0.g646b573
Operating system: macOS
Node.js version ('node --version' output): N/A
Go version ('go version' output): N/A
```

##### Steps to reproduce

1. Go the pod list page in any namespace.
2. Click **show all** button to expand the pod labels.
3. Wait for about 5s.
##### Observed result

1. The row will automatically fold again.
2. If I open dev console and switch to "Network" tab, I can see the dashboard frontend is sending some request every 5s to refresh the content of the pod list.
##### Expected result

1. The row **shouldn't** fold again.
2. The frontend shouldn't make requests every 5s to refresh the pod list.

##### Comments

We did some debugging, and identified the issue to be `GET api/v1/settings/global/cani` returns `allow: false` while it should have been `true`.
We use auth head to login, and I verified my logged-in user does have `get` access to the `kubernetes-dashboard-settings` configmap in `kubernetes-dashboard` namespace (I can access the configmap directly on WebUI).

For now, we are using the workaround of `--disable-settings-authorizer` to artificially make the `cani` endpoint return `true`. And this solves the frontend refreshing issue.

floreks:
Fixes #5170

### DIFF
--- a/src/app/backend/client/api/common.go
+++ b/src/app/backend/client/api/common.go
@@ -16,19 +16,23 @@ package api
 
 import (
 	"crypto/rand"
+	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/authorization/v1"
 )
 
 // ToSelfSubjectAccessReview creates kubernetes API object based on provided data.
-func ToSelfSubjectAccessReview(namespace, name, resource, verb string) *v1.SelfSubjectAccessReview {
+func ToSelfSubjectAccessReview(namespace, name, resourceKind, verb string) *v1.SelfSubjectAccessReview {
 	return &v1.SelfSubjectAccessReview{
 		Spec: v1.SelfSubjectAccessReviewSpec{
 			ResourceAttributes: &v1.ResourceAttributes{
 				Namespace: namespace,
 				Name:      name,
-				Resource:  resource,
-				Verb:      verb,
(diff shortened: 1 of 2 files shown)


src/app/backend/client/api/common.go
```diff
@@ -16,19 +16,23 @@ package api
...
-func ToSelfSubjectAccessReview(namespace, name, resource, verb string) *v1.SelfSubjectAccessReview {
+func ToSelfSubjectAccessReview(namespace, name, resourceKind, verb string) *v1.SelfSubjectAccessReview {
 	return &v1.SelfSubjectAccessReview{
 		Spec: v1.SelfSubjectAccessReviewSpec{
 			ResourceAttributes: &v1.ResourceAttributes{
 				Namespace: namespace,
 				Name:      name,
-				Resource:  resource,
-				Verb:      verb,
+				// Name must be in a plural lower-case format,
```
JackWindows, line 32: Probably say "Resource name" instead of "Name" here. It's a bit confusing since "Name" could mean the line above.

PR #5204, opened 2020-05-29
delete 。in chinese translation

hwdef:
图片

Without it, the display will be better


zehuaiWANG:
/LGTM Thank you for your contribution. @hwdef

PR #5214, opened 2020-06-01
Fix wrong job statuses in list view

JackWindows:
The way how the angular frontend calculates the job status is different from how the backend calculates it (https://github.com/kubernetes/dashboard/blob/master/src/app/backend/resource/job/common.go#L80-L88).
This PR fixes the calculation used by frontend and match it with the backend.
Specifically:
1. A job is failed only and if only the job state is `Failed`. This should not be determined by the job's pod states.
2. A job is in success state if the job state is `Complete` or there is at least one pod that is still running. As long as there is still pod(s) running in the job, the job shouldn't be considered failed or pending. The job will later on convert to failed state if the running pod keeps failing and backoff limit is reached.
3. A job is in pending state if the job state is not `Running` or `Failed`, and the number of pending pod is greater than 0.

### DIFF
--- a/src/app/frontend/common/components/resourcelist/job/component.ts
+++ b/src/app/frontend/common/components/resourcelist/job/component.ts
@@ -73,15 +73,19 @@ export class JobListComponent extends ResourceListWithStatuses<JobList, Job> {
   }
 
   isInErrorState(resource: Job): boolean {
-    return resource.podInfo.warnings.length > 0;
+    return resource.jobStatus.status === 'Failed';
   }
 
   isInPendingState(resource: Job): boolean {
-    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending > 0;
+    return (
+      resource.podInfo.running === 0 &&
+      resource.jobStatus.status !== 'Complete' &&
+      resource.jobStatus.status !== 'Failed'
+    );
   }
 
   isInSuccessState(resource: Job): boolean {
-    return resource.podInfo.warnings.length === 0 && resource.podInfo.pending === 0;
+    return resource.podInfo.running > 0 || resource.jobStatus.status === 'Complete';
   }
 
   getDisplayColumns(): string[] {
--- a/src/app/frontend/typings/backendapi.ts
+++ b/src/app/frontend/typings/backendapi.ts
@@ -33,6 +33,11 @@ export interface ObjectMeta {
   uid?: string;
 }
 
+export interface JobStatus {
+  status: string;
+  message: string;
+}
+
 export interface ResourceDetail {
   objectMeta: ObjectMeta;
   typeMeta: TypeMeta;
@@ -285,6 +290,7 @@ export interface Job extends Resource {
   containerImages: string[];
   initContainerImages: string[];
   parallelism: number;
+  jobStatus: JobStatus;
 }
 
 export interface Namespace extends Resource {


floreks:
Our backend is not a good example of how the final status should be calculated. In this backend method, the only available information is the actual Job object. In the frontend, we have so much more information that we can use to calculate the final status. We have Events related to the Job, Pods with their Events and statuses. Everything running as a part of this Job. We shouldn't ignore all of this.

JackWindows:
Hmm, in backend both `podInfo` and `jobStatus` are available (reference https://github.com/kubernetes/dashboard/blob/master/src/app/backend/resource/job/common.go#L77-L78).
In frontend, only `podInfo` is currently in use.
So I would say this PR is at least better than the current frontend code.
It seems you want more info to be considered, what suggestions do you have in mind? We probably can create another PR for further improvement.

floreks:
We should actually check how `kubectl` calculates the final status here. It seems like logic was copied from other resources and might not be adjusted to the job. We should also be consistent with the description at https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#parallel-jobs

Additionally, we can use pod/job events to show potential errors. Proposed logic does not really improve the overall state. It changes the logic but is not really closer to what is expected based on the documentation.

JackWindows:
> We should actually check how `kubectl` calculates the final status here.

`kubectl` doesn't calculate job state.
```
$ kubectl get jobs | head
NAME                                                         COMPLETIONS   DURATION   AGE
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx                       0/2           5d         5d
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx                       0/2           5d1h       5d1h
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx                       1/1           62m        9d
```

> We should also be consistent with the description at https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#parallel-jobs

I took a look of the k8s official job documentation but I didn't find any specifics about how should the job state be defined.

> Additionally, we can use pod/job events to show potential errors.

I think this is already done, and this PR doesn't intend to change that. This PR merely touch the status icon in the job list view. If there are logics that display warning messages, those are not touched in this PR.

> It changes the logic but is not really closer to what is expected based on the documentation.

It's a bit unfair to say this given there is no official documentation on job state. This PR
1. Fixes the issue where a `Failed` job is displayed as successful if the pods of that job are cleaned  out upon job finish. This IMHO is a huge improvement, and the existing logic is a huge confusion to end users. The old angular v1 frontend didn't have such confusion (reference https://github.com/kubernetes/dashboard/blob/v1.10.1/src/app/frontend/job/list/card_component.js#L75-L77).
2. Brings the calculation consistent between frontend and backend.

IMO this is definitely better than existing logic. As for your request on better calculate the job state, I agree on that, but I don't think we have to include that in the scope of this PR. We can use this PR to bring consistent calculation between frontend and backend, and write another PR to change both frontend/backend to a better version of calculation.

floreks:
> kubectl doesn't calculate job state.

Yep, list does not show status of the job, only completions. We should actually add such column to the job list in another PR. Describe however shows a bit more info, including job pods statuses. Status icon shows overall status of the job based on multiple conditions.

> I took a look of the k8s official job documentation but I didn't find any specifics about how should the job state be defined.

There are multiple hints that can help in understanding how status should be calculated here:
- the Job is complete as soon as its Pod terminates successfully
- once at least one Pod has terminated with success and all Pods are terminated, then the Job is completed with success
- the Job represents the overall task, and is complete when there is one successful Pod for each value in the range 1 to `.spec.completions`

In summary, Job should be marked as completed once at least one pod terminates without error. To be honest we should actually have 4 states: `Pending`, `Running`, `Completed`, `Failed`. Based on that I'd propose to use the logic as follows:
- If `podInfo.succeeded > 0` or `jobStatus === 'Complete'` mark Job as `Completed`
- If the above is false and `jobStatus === 'Failed'` or `podInfo.warnings.length > 0` mark Job as `Failed` (potentially there are some errors)
- If it's not completed nor failed and `podInfo.running.length > 0` mark it as `Running`
- In any other case mark it as `Pending`

JackWindows:
Thanks for you detailed suggestion, I did go through a similar thought process and I will address my concerns to some of you proposed items.

> - the Job is complete as soon as its Pod terminates successfully
> - once at least one Pod has terminated with success and all Pods are terminated, then the Job is completed with success
If `podInfo.succeeded > 0` or `jobStatus === 'Complete'` mark Job as `Completed`

I wouldn't consider a job successful unless the number of succeed pods is >= the number of desired completion. So instead of using > 0 here, we should probably use >= `.spec.completions`.

> - the Job represents the overall task, and is complete when there is one successful Pod for each value in the range 1 to `.spec.completions`

Per the k8s doc, this job type is not yet implemented, and it seems we can't tell whether a job falls into this type or not at the moment.

> If the above is false and `jobStatus === 'Failed'` or `podInfo.warnings.length > 0` mark Job as `Failed` (potentially there are some errors).

I agree with the `jobStatus === 'Failed'` part, but not necessary with the `podInfo.warnings.length > 0` part. When a job's pod has warnings, it shouldn't just be treated as `Failed`, for example
1. In a cluster where autoscaler is engaged, there can be a ~2 minutes delay before the autoscaler brings up more machines to join the cluster, during which the pods are unschedulable due to insufficient. With your logic, the job will be displayed as `Failed` as soon as it is submitted due to the unschedulable warning. This creates confusion to the users as actually the job is still running. `Failed` usually is considered a terminal state of a job, it would be super weird for a `Failed` job to return to `Running` state.
2. In the event of `ImagePullBackoff`, the job will also be displayed as `Failed`. However, it could be a temporary docker registry instability that caused the `ImagePullBackoff` and the job will retry and resume running soon. In this case the job state shouldn't be displayed as `Failed` IMO. That said, there lacks a way to determine whether the `ImagePullBackoff` will recover or not, so just displaying the job as `Running` indefinitely isn't really a good idea either, but IMHO it's less confusing than "having a job in `Failed` state and later on returns to `Running` state".

Also note that even tho we choose to display a green checkmark icon for a job with pod warnings, the actual warning text will still be showing on the UI under the job entry, so nothing is really preventing users from noticing a potential error in the job's associated pod(s).

> To be honest we should actually have 4 states: `Pending`, `Running`, `Completed`, `Failed`.

The current implementation isn't too far from this. AFAIK we are just displaying both `Running` and `Completed` jobs with a green checkmark icon, so it should be fine for the time being.

maciaszczykm:
> I wouldn't consider a job successful unless the number of succeed pods is >= the number of desired completion. So instead of using > 0 here, we should probably use >= .spec.completions.

+1

floreks:
> I wouldn't consider a job successful unless the number of succeed pods is >= the number of desired completion. So instead of using > 0 here, we should probably use >= .spec.completions.

I don't think it can ever be bigger than `.spec.competions`, also I am not 100% sure if `podInfo.succeeded === .spec.completions` will always mean that job has completed. According to this:

> once at least one Pod has terminated with success and all Pods are terminated

It might mean that i.e. we have 1 succeeded pod and 3 terminated and Job is still considered as completed by the kubernetes.

JackWindows:
> It might mean that i.e. we have 1 succeeded pod and 3 terminated and Job is still considered as completed by the kubernetes. I'll have to investigate this further to confirm.

It seems that we are complicating things. If a job reaches its completion conditions, k8s will transit this job into the `Complete` state. Therefore, using `jobStatus === 'Complete'` as the sole condition would already be sufficient, and kind of offloading the responsibility to the underlying k8s logic. IMHO `podInfo` isn't really needed at all to determine whether a job is in `Complete` state.

floreks:
I guess for the `Complete` state that's true. It's easier for K8S to determine the status here. Other statuses we can check.

JackWindows:
In light of that, I propose the following conditioning logics.
1. If `jobStatus === 'Complete'`, then mark the job as `Complete`.
2. Else if `jobStatus === 'Failed'`, then mark the job as `Failed`.
Note that both `Complete` and `Failed` are considered terminal state for a job in the underlying k8s logic. That is to say, when a job enters either state, it won't further transition into a different state.
3. Else if `podInfo.running.length > 0`, then mark the job as `Running`.
4. Else mark the job as `Pending`.

And for the UI icon display purpose:
1. Show `icon.checkCircle` for both `Complete` and `Running` jobs.
2. Show `icon.timelapse` for `Pending` jobs.
3. Show `icon.error` for `Failed` jobs.

Note that this is only about the icon of the job state. If the job has associated warning events, the UI will allow you to expand the job object to display these associated warning message. An example of a `Pending` state job with warning is illustrated below:
image
After clicking the "eye" icon the warning message is revealed.
image
Without this PR the above job will be displayed as `Failed` with a red error icon, which is confusing given the job is just in queue and waiting for available resources. `Pending` would be a better description for jobs in such a situation.

Another example would be a `Running` job with failed pod(s). Such a job would be displayed with a red error icon without this PR, and will be displayed with a green checkmark icon with this PR. The logic behind this is that a job with failed pods can still be `Running` as long as there is at least one running pod. In that case, the job is technically still making progress, and has the potential to transition into `Complete` state. In such a situation, it would make more sense to use a green checkmark icon to describe the job state than using a red error icon. Should the job has too many failed pods (hence it would exceed the backoff limit and transition into `Failed` state), or the job takes too long to finish (hence it would exceed the active deadline and also transition into `Failed` state), the icon will become a red error one to reflect the final state of this job.

floreks:
I am still not sure about pending icon with pod event warnings/errors. There is no real graphical indicator here that there might be something wrong with the Job and that it is possible to see errors. We'd need some additional indicator here to hint user about the possibility to see them. Error/Warning icon was doing just that.

JackWindows:
I guess you can think it this way:
The `Pending` state itself is a indicator that the job has something wrong with it. If a cluster is functioning well, the job should probably never be in `Pending` state? (maybe for a split seconds between the job is submitted and before the scheduler allocates resources).
If the job stays in `Pending` state, either there is insufficient resources or there is error pulling the container images, or something else is keeping the containers from starting.

I think in the angular v1 era of this project, we didn't have the additional step of "clicking to show the error message thing", the error messages were directly display if there was any.
What is the motivation to add this additional "clicking to expand" step in angular v2 frontend? I am thinking maybe we can remove that and just go ahead to display error messages if there is any. In that way won't have as much room for confusion.

Another approach is to add a separate icon (e.g. the triangle warning icon) to indicate there are warning messages associated with this job object. The triangle warning icon can serves as a visual indicator for users.

floreks:
Ye, but the `Pending` icon means, well that the job is pending. This is not really an issue. There should be a different icon if it can't start because of pull error or lack of resources. The warning icon would be best here.

JackWindows:
I think what I'm trying to say is the current definition of `Pending` may not be useful.
It seems with that definition, a pod/job in a cluster with autoscaling will never be in `Pending` state. The pod will always be met with a insufficient resource warning message and therefore be displayed as `Error` (a red alert icon). Maybe it's different in an on-premise cluster?

What's our solution to proceed here? Use a 4th icon to suggest a state that job is in a warning state but not yet failed? (TBH this is still a bit confusing in a cluster that utilizes auto-scaling).

floreks:
It definitely can be `Pending` if associated pods are starting but have not started yet.

From K8S docs (`Pending` status of a `Pod`):
> The Pod has been accepted by the Kubernetes system, but one or more of the Container images has not been created. This includes time before being scheduled as well as time spent downloading images over the network, which could take a while.

Sometimes container registry is slow or image takes a lot of space and it takes some time to download and run it. In such cases resource should be in a `Pending` state because it is still starting.

JackWindows:
I guess that makes sense for an on-premise cluster.
However for clusters that use autoscalers, it seems a bit messy.
For example, before autoscaler brings up new nodes, this `Pending` pod is showing the `Failed` icon.
image
And after the autoscaler brings up a new node, but before the pod actually starts. This `Waiting` pod is showing the `Failed` icon again. (Which IMHO, should be showing `Pending` icon).
image

floreks:
The problem here is that events do not expire immediately after the issue is resolved. They only inform about potential problems that have happened. We'd have to make the status logic much more sophisticated and prioritize status over the events to react immediately to such changes. Adding warning icon instead is also an option here if status !== failed and there are related error events.

hagaiovadia:
How come this issue is closed?
This bug still exists.

sneko:
In my case all my jobs completed with success but on the overview I can see (same for `v2.0.3` and `v2.3.1`):
image

sneko:
Ok it didn't solved my issue...

For example today I see 2 failed jobs in the overview
image

But when listing jobs in all namespaces they are all "green"
image

I have to check one by one which one has "X failed" which is really annoying. Hopefully I can use:
```
kubectl get jobs -n bb-dev
```

and check at the column completions for those with `0/1` but it's really far from being intuitive.

I hope this will be fixed in the future.

Thanks,

corford:
We are hitting this issue with cron jobs. With over 300 daily jobs, finding the failed ones via the UI is **extremely** annoying due to this list view bug. Our current workaround is to use K9s or Lens (which both correctly display failed jobs).

Would be great if this issue was re-opened and the PR merged.

aschleifer:
/reopen
/remove-lifecycle-rotten

This issue still exists.

PR #5215, opened 2020-06-01
Add condition to job status return in backend

JackWindows:
Add job condition return in job status in backend.
This can be used in the future by the frontend to render more information about the job such as the job stop time (which is currently not available in frontend).
The PR requires no addition query in the backend, so performance-wise it should stay pretty much the same.

A demo of the revised job detail page:
image

### DIFF
--- a/src/app/frontend/resource/workloads/job/detail/template.html
+++ b/src/app/frontend/resource/workloads/job/detail/template.html
@@ -71,7 +71,7 @@ limitations under the License.
   </div>
 </kd-card>
 
-<kd-condition-list [conditions]="job?.jobStatus.conditions"
+<kd-condition-list [conditions]="job?.jobStatus?.conditions"
                    [initialized]="isInitialized"></kd-condition-list>
 
 <kd-pod-status-card [podInfo]="job?.podInfo"
--- a/src/app/frontend/resource/workloads/job/detail/template.html
+++ b/src/app/frontend/resource/workloads/job/detail/template.html
@@ -71,6 +71,9 @@ limitations under the License.
   </div>
 </kd-card>
 
+<kd-condition-list [conditions]="job?.jobStatus.conditions"
+                   [initialized]="isInitialized"></kd-condition-list>
+
 <kd-pod-status-card [podInfo]="job?.podInfo"
                     [initialized]="isInitialized"></kd-pod-status-card>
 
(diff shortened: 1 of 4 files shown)


maciaszczykm:
Will you take care of the frontend? I'd like to avoid adding it if the frontend will not follow.

JackWindows:
> Will you take care of the frontend? I'd like to avoid adding it if the frontend will not follow.

Unfortunately I don't really have an idea yet on how to change the frontend to display this.
I have a private user script that I use to "augment" the UI and add links to a separate metric dashboard. I want to be able to know the job end time in order to auto-zoom the metrics time series to the duration of the job. This is the motivation of adding this field in the backend.

I understand your idea that the backend shouldn't return anything that the frontend doesn't use in an ideal setup. So I will respect that and probably only put this change in my forked version of the dashboard.

floreks:
It's extremely simple to add in the frontend. Simply use [this component](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/condition/component.ts) and add it to the job details view, similar to how it's done for some other resources.

JackWindows:
@floreks Thanks for the great suggestion. I was under the impression that I need to add a `Job End Time` sort of information to frontend.
If it's just displaying job conditions in the job detail page, I'm happy to also include that in this PR.

JackWindows:
Updated the PR with frontend changes. Also put a screenshot in the description.

src/app/frontend/resource/workloads/job/detail/template.html
```diff
@@ -71,6 +71,9 @@
   </div>
 </kd-card>
 
+<kd-condition-list [conditions]="job?.jobStatus.conditions"
```
maciaszczykm, line 74: It would be safe to put ? also before conditions.

maciaszczykm:
LGTM, only one minor remark.

src/app/frontend/resource/workloads/job/detail/template.html
JackWindows: Good catch.
I adjusted to `job?.jobStatus?.conditions`.

JackWindows:
Hmm I noticed my last commit didn't trigger the travis CI, and now it seems the PR is stuck in waiting for tarvis CI report.
image
What should I do about it?

floreks:
I have manually restarted the job. Travis is quite flaky lately.

PR #5233, opened 2020-06-04
Update dependencies

maciaszczykm:


maciaszczykm:
Test passed but Travis does not return the result:

Zrzut ekranu 2020-06-5 o 08 44 18

PR #5245, opened 2020-06-09
Fixed change detection in resource lists

### LINKED ISSUE #4882: Expanded view of pods will condense after the auto-refresh interval
##### Environment

```
Kubernetes version: 1.17
Dashboard version: 2
Operating system: Centos
```

##### Steps to reproduce

Click the eye or show all buttons that let you expand the view of a pod. 
Wait for the auto-refresh interval, the expanded view will close itself.

##### Observed result
The expanded view will close itself.

##### Expected result
The expanded view should stay open so status changes can be read.

Eddman:
Fixes #4882. 

Fixed change detection on auto-refresh in lists.


src/app/frontend/common/components/date/component.ts
```diff
@@ -36,14 +39,41 @@ import {Component, Input} from '@angular/core';
...
+  changeIteration = 0;
   @Input('relative')
   set relative(v: boolean) {
     this._relative = v !== undefined && v !== false;
   }
+  private _unsubscribe = new Subject<void>();
+
+  constructor(private readonly settings_: GlobalSettingsService, private readonly cdr_: ChangeDetectorRef) {}
+
+  ngOnInit() {
```
floreks, line 58: This feels a bit hacky. Can't we leave it on default change detection?

floreks:
Sorry for the late review. I have been busy. The code looks good. I'll have to find some time to test it properly.

src/app/frontend/common/components/date/component.ts
Eddman: The change detection does not really play the role here. The pipe for relative time is generating each time a different result as time moves on. 

However it is not set as "not pure". So even after change detection cycle you can't see the changed output text. One option is to set the `pure` flag to `false`, second option is to trigger manual update using an additional parameter that is changing over time. 

`pure: false` has a lot of disadvantages. Cause even moving your mouse over the date component triggers update/re-render. That is not really optimal and also causes change detection problems when the pipe result is changed during the change detection loop - which happens a lot if you have, let say, a new pod in the list and the text changes every second.

That's why I switched to the additional parameter that is much more optimal and safer. If it helps we can have a fixed refresh interval cause you need some trigger...

And regarding change detection strategy: `OnPush` has to be set for all sub-components to avoid unneeded change detection loop in components higher in your tree. As far as I remember form my past experiments with change detection: change detection triggered on child component marks your component for view update. So IMHO if you set `OnPush` on the list component it is not really/only partially effective until all column components are `OnPush` as well. 

And in general the speedup with `OnPush` even for a simple component like this one is huge.

floreks: Shouldn't parent component on change event trigger also child component? Also, the date input should change over time and trigger onChange event which in result can mark component for check?

Eddman: Not in case of "pure" pipes. Such pipe is evaluated only if arguments were changed. But the date is the same string, only current time that is not an argument changes -> the pipe is not re-evaluated and text not changed, ever.

We can make current time an argument here, but that would cause an endless change detection loop. So I've chosen  a number that increments on fixed intervals instead.

Eddman: And `pure: true` is default for all pipes.

floreks: I'd then simply define a const interval of 1 minute here to refresh the date.

Eddman: OK, will try to think about something 😉

Eddman: @floreks What about this one:
```typescript
  ngOnChanges() {
    this._unsubscribe.next();
    if (this._relative) {
      // Every 5 seconds the first minute, every minute the first hour, afterwards every hour
      of([5, 5, 11], [60, 60, 59], [3600, 3600, Infinity])
        .pipe(mergeMap(([delay, period, takeFirstN]) =>
          timer(delay * 1000, period * 1000).pipe(take(takeFirstN)))
        )
        .pipe(takeUntil(this._unsubscribe))
        .subscribe(() => {
          this.changeIteration++;
          this.cdr_.markForCheck();
        });
    }
  }
```

Eddman: Verified and tested -  https://github.com/kubernetes/dashboard/pull/5245/commits/6c5dba5c19dfa34d63af9f60e4ca3313aa8b4124

src/app/frontend/common/components/date/component.ts
```diff
@@ -36,14 +38,90 @@ import {Component, Input} from '@angular/core';
...
+  private timeBaseIntervals_ = [
+    60, // Seconds in a minute
+    60, // Minutes in a second
+    24, // Hours in a day
+  ];
+  private unsubscribe_ = new Subject<void>();
+
+  constructor(private readonly cdr_: ChangeDetectorRef) {}
+
+  ngOnChanges() {
```
floreks, line 70: @Eddman I have updated the logic to what I have had in mind. I wanted to calculate the refresh time based on time that has already passed. This way refresh time is adjusted separately for every item on the list.

Eddman: Had a similar thought, but did not want to play around with the `this.date` 🙂 Anyway LGTM 😉 👍

floreks: As a precaution, we could probably transform both dates to a timezone local to the browser and then do the calculation as the server might use a different timezone than the browser.

Eddman: I'm not following now, `new Date(this.date)` parses using ISO format and the `.getTime()` returns timestamp in UTC, or do I miss something?

Eddman: So we're comparing 2 timestamps in UTC I think...

Eddman: Just did a quick test in console to verify and it is UTC timestamp 🙂

floreks: To be extra careful I have updated it to first convert time to the UTC and then calculate the difference.

Eddman: That I guess is overkill :) you can try that in console try setting offset on a date and you’ll see that getTime is always the same ;)

floreks: True, pod `creationTimestamp` according to the documentation is in UTC also. I guess we could assume that `new Date().getTime()` will always return local time in UTC format. Both of them are UTC then.

Eddman:
@floreks Any plan to publish a new release? There were already quite a lot of changes since the last one.

maciaszczykm:
+1, I think we should do it in a week.

floreks:
@Eddman at the latest, we'll do this next Thursday.

floreks:
@Eddman we have just released [v2.0.2](https://github.com/kubernetes/dashboard/releases/tag/v2.0.2)

PR #5246, opened 2020-06-10
Save login mode into cookie

### LINKED ISSUE #5210: ability to set default login-mode?

would like to be able to pass a parameter similar to `--authentication-mode` that will determine the default selected sign-in method, something like `--login-mode=kubeconfig`

`10.x` versions defaulted to "Kubeconfig" which we happen to use most often, and the `2.x` versions now default to "Token" which requires an extra step for our users.

##### Why is this needed

to tailor the ux based on usage patterns

##### Comments

i took a quick glance at the typescript code and i thought `--authentication-mode=kubeconfig` might work, but it didn't. 

> i would be happy to attempt a p/r if the owners think this is a reasonable enhancement.

best,
tony.

shu-mutou:
And use it next time if it exists.

Fixes #5210

### DIFF
--- a/src/app/frontend/login/component.ts
+++ b/src/app/frontend/login/component.ts
@@ -13,12 +13,14 @@
 // limitations under the License.
 
 import {HttpClient, HttpErrorResponse} from '@angular/common/http';
-import {Component, NgZone, OnInit} from '@angular/core';
+import {Component, Inject, NgZone, OnInit} from '@angular/core';
 import {ActivatedRoute, Router} from '@angular/router';
 import {AuthenticationMode, EnabledAuthenticationModes, LoginSkippableResponse, LoginSpec} from '@api/backendapi';
 import {KdError, KdFile, StateError} from '@api/frontendapi';
+import {CookieService} from 'ngx-cookie-service';
 import {map} from 'rxjs/operators';
 
+import {Config, CONFIG_DI_TOKEN} from '../index.config';
 import {AsKdError, K8SError} from '../common/errors/errors';
 import {AuthService} from '../common/services/global/authentication';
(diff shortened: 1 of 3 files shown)


src/app/frontend/login/component.ts
```diff
@@ -36,7 +36,7 @@ enum LoginModes {
 })
 export class LoginComponent implements OnInit {
   loginModes = LoginModes;
-  selectedAuthenticationMode = LoginModes.Kubeconfig;
+  selectedAuthenticationMode = localStorage.selectedAuthenticationMode;
```
floreks, line 39: Use cookie service and save it in a cookie. Try to avoid using things that are not injected through the constructor. It's harder to follow where they're coming from and it's also harder for the angular build to do a proper tree shaking and other optimizations.

shu-mutou: done.

src/app/frontend/login/component.ts
```diff
@@ -36,7 +37,7 @@ enum LoginModes {
 })
 export class LoginComponent implements OnInit {
   loginModes = LoginModes;
-  selectedAuthenticationMode = LoginModes.Kubeconfig;
+  selectedAuthenticationMode = this.cookies_.get('selectedAuthenticationMode') || '';
```
floreks, line 40: Do not hardcode the name. Add it to the [config](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/index.config.ts) and inject into the constructor.
```ts
@Inject(CONFIG_DI_TOKEN) private readonly config_: Config,
```

shu-mutou: done

shu-mutou:
fixed tests

PR #5250, opened 2020-06-10
Move namespace selector

floreks:
As we want to introduce more resources (SA, Roles, etc.) it would cause confusion if we'd add non-namespaced resources under namespace selector or have 2 cluster sections for namespaced and non-namespaced resources. We've decided to move the selector to the top bar and add small indicators to show which sections/resources are namespaced and will be influenced by the namespace change.

Additionally, I have updated `Discovery and Load Balancing` section to `Service` as this is how it's called now in the k8s docs.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#-strong-service-apis-strong-

This is how it looks like
Peek 2020-06-10 16-48

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -2753,7 +2753,7 @@
         <target>Storage Classes</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/chrome/nav/template.html</context>
-          <context context-type="linenumber">50</context>
+          <context context-type="linenumber">56</context>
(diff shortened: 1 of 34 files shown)


maciaszczykm:
What do you think about removing the separator before "Custom Resource Definitions"?

floreks:
Depende, to which section it belongs in K8S API docs?

maciaszczykm:
To the cluster section. However, for us it should be a section on its own as we are displaying the pinned CRDs below.

PR #5256, opened 2020-06-12
[handler] log the real client address

qingwave:
Log the real client address more than proxy ip.

### DIFF
--- a/src/app/backend/handler/filter.go
+++ b/src/app/backend/handler/filter.go
@@ -33,6 +33,12 @@ import (
 	"github.com/kubernetes/dashboard/src/app/backend/errors"
 )
 
+const (
+	originalForwardedForHeader = "X-Original-Forwarded-For"
+	forwardedForHeader         = "X-Forwarded-For"
+	realIPHeader               = "X-Real-Ip"
+)
+
 // InstallFilters installs defined filter for given web service
 func InstallFilters(ws *restful.WebService, manager clientapi.ClientManager) {
 	ws.Filter(requestAndResponseLogger)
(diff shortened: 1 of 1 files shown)


src/app/backend/handler/filter.go
```diff
@@ -177,3 +177,33 @@ func mapUrlToResource(url string) *string {
 	}
 	return &parts[3]
 }
+
+// remoteAddr extracts the remote address of the request, taking into
+// account proxy headers.
+func remoteAddr(r *http.Request) string {
```
maciaszczykm, line 183: Please change to `getRemoteAddr`.

src/app/backend/handler/filter.go
```diff
@@ -177,3 +177,33 @@ func mapUrlToResource(url string) *string {
...
+		if len(proxies) > 0 {
+			remoteAddr := strings.TrimSpace(proxies[0])
+			if remoteAddr != "" {
+				return remoteAddr
+			}
+		}
+	}
+
+	if prior := r.Header.Get("X-Forwarded-For"); prior != "" {
+		proxies := strings.Split(prior, ",")
```
maciaszczykm, line 195: The part inside `if` statement is duplicated, please export it to the new function.

src/app/backend/handler/filter.go
```diff
@@ -177,3 +177,33 @@ func mapUrlToResource(url string) *string {
 	}
 	return &parts[3]
 }
+
+// remoteAddr extracts the remote address of the request, taking into
+// account proxy headers.
+func remoteAddr(r *http.Request) string {
+	if prior := r.Header.Get("X-Original-Forwarded-For"); prior != "" {
```
floreks, line 184: Export those header strings to a custom named string type

qingwave: fixed

src/app/backend/handler/filter.go
qingwave: fixed

PR #5276, opened 2020-06-18
Adding override for ingress paths

jtyr:
This PR is adding the possibility to override the default Ingress paths in the Helm chart with a new block of paths defined as YAML data structure. This allows to configure the Ingress resource to work with the AWS ALB Ingress Controller to automatically [redirect HTTP traffic to HTTPS](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/tasks/ssl_redirect/) on the load balancer.

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.0.3
+version: 2.1.0
 appVersion: 2.0.3
 description: General-purpose web UI for Kubernetes clusters
 keywords:
--- a/aio/deploy/helm-chart/kubernetes-dashboard/README.md
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/README.md
@@ -93,7 +93,7 @@ Parameter                                       | Description                   
 `ingress.labels`                                | Add custom labels                                                                                                                | `[]`
 `ingress.enabled`                               | Enable ingress controller resource                                                                                               | `false`
 `ingress.paths`                                 | Paths to match against incoming requests. Both `/` and `/*` are required to work on gce ingress.                                 | `[/]`
-`ingress.paths_custom`                          | Override for the ingress paths                                                                                                   | `[]`
+`ingress.customPaths`                           | Override the default ingress paths                                                                                               | `[]`
 `ingress.hosts`                                 | Dashboard Hostnames                                                                                                              | `nil`
 `ingress.tls`                                   | Ingress TLS configuration                                                                                                        | `[]`
 `metricsScraper.enabled`                        | Wether to enable dashboard-metrics-scraper                                                                                       | `false`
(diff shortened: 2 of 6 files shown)


aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
```diff
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.0.1
+version: 2.0.1-1
```
floreks, line 17: Not sure why we haven't started chart versioning from the `1.0.0`. Just bump the minor version to `2.0.2`.

desaintmartin: Actually, this was an experiment to keep the versions synchronized and release the chart at the same time as Dashboard. There are cons and pros, I'm open to suggestions.

floreks: IIRC in order to use the new chart features version has to be bumped, otherwise, helm yells at you that chart version didn't change, right? If so, keeping both versions in sync would mean that in order to use those new chart features people would have to wait for official Dashboard release.

desaintmartin: That is the "con" part, indeed. I do not mind having separated versions!

jtyr: OK, I have changed the version to `2.1.0`.

Regarding the mismatch of the app and helm version, I have originally chosen absolutely valid version `2.0.1-1` which maintains the relationship between the two versions but still increments it by specifying the revision portion. I think that's quite a nice compromise.

floreks:
Please rebase your changes.

aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
desaintmartin: After some thoughts according to your inputs, I think un-coupling the versions is the best way to go and the only one to realistically respect semver. 2.0.1-1 is not semver-compliant, and waiting for a new minor release to release this is not realistic either, so 2.1.0 seems the way to go!

aio/deploy/helm-chart/kubernetes-dashboard/README.md
```diff
@@ -93,6 +93,7 @@ Parameter                                       | Description
 `ingress.labels`                                | Add custom labels                                                                                                                | `[]`
 `ingress.enabled`                               | Enable ingress controller resource                                                                                               | `false`
 `ingress.paths`                                 | Paths to match against incoming requests. Both `/` and `/*` are required to work on gce ingress.                                 | `[/]`
+`ingress.paths_custom`                          | Override for the ingress paths                                                                                                   | `[]`
```
desaintmartin, line 96: Convetion for yaml seems to be:
```suggestion
`ingress.pathsCustom`                          | Override for the ingress paths                                                                                                   | `[]`
```

or even customPaths?

aio/deploy/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -119,6 +119,18 @@ ingress:
     - /
   #  - /*
 
+  ## CustomKubernetes Dashboard Ingress paths
```
desaintmartin, line 122:
```suggestion
  ## Custom Kubernetes Dashboard Ingress paths. Will override default paths.
```

floreks:
@desaintmartin https://github.com/kubernetes/dashboard/issues/5280

Currently, helm linting is not yet added as yesterday we have migrated to the github workflows.

desaintmartin:
Thanks!
/lgtm
I'll try to work on monday to automate release of chart versions thanks to the switch to github-actions, so it may take a few days before this change is published.

desaintmartin:
I would like the new CI at https://github.com/kubernetes/dashboard/pull/5286 to get merged before merging new PRs, since it will catch a lot of potential errors. I'll ping you when it's done.

desaintmartin:
Could you rebase again? CI just got merged.

desaintmartin:
Thanks again!
/lgtm

Issue #5086, opened 2020-04-21
github.com/kubernetes/dashboard@v2.0.0 is an invalid go module


bouk:
##### Environment
```
Dashboard version: v2.0.0
Go version ('go version' output): go version go1.14.2 darwin/amd64
```

##### Steps to reproduce

I want to use this project inside my own, but I can't use it as a module because it's an invalid go module right now. To reproduce, run:

```
$ GO111MODULE=on go get github.com/kubernetes/dashboard@v2.0.0
go get github.com/kubernetes/dashboard@v2.0.0: github.com/kubernetes/dashboard@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

##### Observed result

Go says this is an invalid module.

##### Expected result

The module is installed correctly.

##### Comments

I noted this is a problem back in july: https://github.com/kubernetes/dashboard/issues/4110, so I think this does need to get resolved. I can create a PR for it.

maciaszczykm:
If you are willing to work on that we can help you with review and provide some suggestions. It has no big priority for the reasons I have mentioned before but it might be useful for projects like yours. Let us know if you will need anything from us. Also, any early proposal/drafts are welcomed so we can discuss top-level things early.

bouk:
What I think we'll have to do is move the source to `v2/` and then rename all references to that. Doesn't seem to complicated

floreks:
Basically, Dashboard is not intended to be used as a Go module itself. It is a complete application, not only using Go, but also a lot of other things. I don't really want to change our code structure to satisfy this constraint. You can fork it and work on it, but from IMHO it should not be possible to only use it as a pure Go module.

bouk:
Fair enough. I tried to quickly change it but it would involve changing the whole structure of the project which just isn't worth it. I'll fork, thank you!

PR #5294, opened 2020-06-22
go mod: mark as v2

bengadbois:
For go module versions >1, the version should be marked

See https://github.com/google/go-github/blob/master/go.mod as an example


floreks:
I am not really a fan of this solution. We do use go modules but Dashboard is not really intended to be used as a dependency and was never designed to follow go standards for creating modules.

maciaszczykm:
I remember there was similar issue in the past and we had some issues with changing all versions to v2. I agree with @floreks that our focus is not to provide backend that can be reused, we focus to make it work for our app. However, if you make it work and it will not be hacky I think we can merge since it will not harm us.

floreks:
Related to https://github.com/kubernetes/dashboard/issues/5086

floreks:
We might rethink that while designing a new API and splitting UI/API into different containers.

/close

PR #5297, opened 2020-06-23
styles: Turn off hover effects on disabled buttons

### LINKED ISSUE #5293: Turn off hover effect for disabled buttons
Zrzut ekranu 2020-06-22 o 16 40 39

MatonAnthony:
Turns off the applied .mat-icon-button:hover on .mat-icon-button that are disabled.

This change should fix #5293 

This is my first PR to Kubernetes, if anything is missing, I will happily fix it :)

### DIFF
--- a/src/app/frontend/_theming.scss
+++ b/src/app/frontend/_theming.scss
@@ -276,7 +276,9 @@
   }
 
   .mat-icon-button:hover {
-    background-color: $border !important;
+    :not([disabled='true']) and :not(&[disabled='true']) {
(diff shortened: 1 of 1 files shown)


MatonAnthony:
Apparently this changes affect code coverage in backend code and in a component code, this looks flaky to me.

floreks:
@MatonAnthony don't worry about codecov check. Their calculation algorithm is very weird. We've got it only for information purposes when the source code is actually changed. This check does not have to pass.

PR #5307, opened 2020-06-25
Add network policy support

maciaszczykm:
Kubernetes_Dashboard

Before putting ingress and egress rule data in the YAML format I've tried many other solutions involving the tables. Unfortunately, the data can be very different. There are multiple ports that can be mapped to different targets. Each target can also be very different (IP blocks array or pod/namespace selectors). I find it a lot easier to get what's going on when looking at the raw data. It doesn't mean that there is no way, but we should design it first and then we can improve it. The current solution will be okay for now.


maciaszczykm:
@floreks PTAL

PR #5309, opened 2020-06-25
Update CI

floreks:
Restrict CI for source code to src dir and add license check action for other changes


maciaszczykm:
Should changes in the workflows be tested too? Cancel hold if not.

PR #5311, opened 2020-06-25
Helm Chart CD: create PR instead of trying to push to protected branch

desaintmartin:
This is a workaround from the fact that Github Actions can't push to protected branches for security reasons.

https://github.community/t/allowing-github-actions-bot-to-push-to-protected-branch/16536/2

What it does:
- Run the release script
- Create a gh-pages-pull-request branch (only one can exist at a time)
- Create a PR using this branch (current version of the action will SILENTLY fail (i.e all green) if it can't create the PR, for example if such a PR already exists.

Examples from my fork:
- CD: https://github.com/wiremind/dashboard/actions/runs/147574658
- Creates https://github.com/wiremind/dashboard/pull/6

Note that it is not battle-tested and can be improved/simplified in the future.

Note that on my fork, source branch is automatically deleted, we should see if it is the case here.

### DIFF
--- a/.github/workflows/cd-helm-workflow.yml
+++ b/.github/workflows/cd-helm-workflow.yml
@@ -44,8 +44,23 @@ jobs:
         env:
           TERM: xterm-256color
 
-      - name: Publish
+      - name: Create Pull Request branch
+        run: |
+          git checkout -b gh-pages-pull-request
+
+      # Note that will fail if branch already exists.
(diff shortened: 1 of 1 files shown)


desaintmartin:
cc @floreks

floreks:
Considering the current limitations and security, IMHO this is the best approach.

/lgtm

PR #5353, opened 2020-07-10
Changed the namespace to kube-system

4nx:
Since at least Kubernetes 1.18 bearer token seemed to be stored as secrets in the kube-system namespace not in kubernetes-dashboard. Reference: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#bootstrap-tokens


floreks:
It depends on the cluster provider. Not all of them even have such SA by default. We recommend creating it inside kubernetes dashboard namespace to use it for accessing Dashboard. Namespace in our docs is correct and intentional, and should not be changed.

PR #5383, opened 2020-07-21
Add value for cluster service label

hacktastic:
This is a PR to make the Service object's cluster-service label configurable. This is because having this label hardcoded (as it is today) breaks GKE clusters >=1.15. The GKE addon manager will discover this label and try to delete the Service.

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.3.1
+version: 2.4.0
 appVersion: 2.0.3
 description: General-purpose web UI for Kubernetes clusters
 keywords:
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.3.0
+version: 2.3.1
 appVersion: 2.0.3
 description: General-purpose web UI for Kubernetes clusters
 keywords:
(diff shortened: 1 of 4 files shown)


aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
```diff
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.3.0
+version: 2.3.1
```
desaintmartin, line 17: Could you bump minor version instead of patch? This is a new feature according to semver.

aio/deploy/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -98,6 +98,9 @@ service:
   ## set a specific load balancer Ip if you're using one
   # loadBalancerIP:
 
+  ## enable or disable the kubernetes.io/cluster-service label
+  clusterServiceLabel: "kubernetes.io/cluster-service: true"
```
desaintmartin, line 102: What should be the value for gke? Maybe it is worth mentionning it. I even wonder if this should be a boolean.

aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
hacktastic: Sure thing. Bumped to 2.4.0

aio/deploy/helm-chart/kubernetes-dashboard/values.yaml
hacktastic: I wasn't sure about that. My plan for my own deployments to GKE was to set clusterServiceLabel to an empty string, thereby not adding a label at all. But maybe it should be something like:
`clusterServiceLabel: 
  enabled: true
  key: "kubernetes.io/cluster-service"` 
What do you think of that?

hacktastic: also, the service manifest would then contain this templated label:
`{{ .Values.service.clusterServiceLabel.key | indent 4}}: {{ .Values.service.clusterServiceLabel.enabled | quote }}`

aio/deploy/helm-chart/kubernetes-dashboard/README.md
```diff
@@ -91,6 +91,7 @@ Parameter                                       | Description
 `service.externalPort`                          | Dashboard external port                                                                                                          | `443`
 `service.loadBalancerSourceRanges`              | list of IP CIDRs allowed access to load balancer (if supported)                                                                  | `nil`
 `service.loadBalancerIP`                        | A user-specified IP address for load balancer to use as External IP (if supported)                                               | `nil`
+`service.clusterServiceLabel`                   | The cluster-service label                                                                                                        | `kubernetes.io/cluster-service: true`
```
desaintmartin, line 94: Could you update the README?

hacktastic: Updated! Thanks

desaintmartin:
Thanks!
/lgtm

PR #5390, opened 2020-07-24
Available namespaces

FrenchBen:
Add the ability to specify the namespaces to display in the dashboard. This allows for a clean list of user-centric namespaces, without the clutter of system-like namespaces.

This doesn't prevent you from navigating to the namespace of choice, if you know that it exists. It's a simple UI tweak to show the namespaces that you want your users to easily find.

### DIFF
--- a/aio/deploy/alternative.yaml
+++ b/aio/deploy/alternative.yaml
@@ -183,6 +183,8 @@ spec:
           args:
             - --namespace=kubernetes-dashboard
             - --enable-insecure-login
+          # Uncomment the following line to manually specify the namespaces available (default, kubernetes-dashboard, etc.)
+          # - --available-namespaces=default,kubernetes-dashboard
           # Uncomment the following line to manually specify Kubernetes API server Host
           # If not specified, Dashboard will attempt to auto discover the API server and connect
           # to it. Uncomment only if the default does not work.
--- a/aio/deploy/head.yaml
+++ b/aio/deploy/head.yaml
@@ -182,6 +182,8 @@ spec:
               protocol: TCP
           args:
             - --namespace=kubernetes-dashboard-head
+          # Uncomment the following line to manually specify the namespaces available (default, kubernetes-dashboard, etc.)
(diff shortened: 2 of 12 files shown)


floreks:
This definitely shouldn't be done as a backend flag. We don't want to end up with a bunch of API flags that directly impact data filtering. This should be fully dynamic, i.e. a part of global settings.

FrenchBen:
@floreks the original design was for systems where a user doesn't have the ability to list all namespaces in a cluster (rbac), but was given a limited list of namespace they can use. 
It allowed me to provide them with a Dashboard that doesn't fail the namespace list query, by leveraging the list provided, instead of the API lookup.

Should this still be part of the global settings? Do you have an example of how this can be implemented?

floreks:
Global settings can be modified usually only by admins. They are available on the top of settings view. Settings are saved in a ConfigMap. I think we should have some kind of field (chip list?) that allows us to define multiple values (namespaces) and delete specific entries if needed. By default it would be a list of fallback namespaces that use would see in case namespace list request fails. Then we could add a switch below that "list field" to control if defined list of namespaces should always override a list of namespaces available in namespace selector. This means that even if the user can list all namespaces, only the ones defined on the list would be available, other namespaces would have to be entered manually.

This is definitely a bit bigger feature but could be very useful. Feel free to create an issue for that if you don't want to work on such a big feature.

FrenchBen:
>This is definitely a bit bigger feature but could be very useful. Feel free to create an issue for that if you don't want to work on such a big feature.
Definitely a much bigger feature. With the PR interest in mind, how can I redesign the NS filter (phase 1?), with the above goal in mind ?

floreks:
> Definitely a much bigger feature. With the PR interest in mind, how can I redesign the NS filter (phase 1?), with the above goal in mind ?

This wouldn't be that hard. We'd need to adjust the logic in the namespace selector component to read ns settings from the settings service and use it accordingly. I guess adding new fields to settings view and extending settings structures would take more time than updating namespace selector logic.

PR #5508, opened 2020-09-07
add dashboard settings config in the helm chart

vishaltak:
Fixes https://github.com/kubernetes/dashboard/issues/5507

PR #5524, opened 2020-09-10
Bugfix/dashboard settings config helm chart

vishaltak:
After deploying dashboard through helm chart version 2.5.0, I've noticed that the dashbaord does not pick up settings when it is a multiline string of YAML. It works fine when it is a string of JSON. In essence, helm chart version 2.5.0([link](https://github.com/kubernetes/dashboard/pull/5508)) does not do what it is intended to do and it is my fault as I did not deploy and test it before raising the PR.

To fix this, I've changed the CM of the dashbaord settings to that effect. Tested it out and things are working fine after using JSON format.


desaintmartin:
Thanks for the reactivity, this is indeed how kubernetes-dashboard manages the ConfigMap. You may have to rebase!

One related question that I did not think about when reviewing this feature: if user changes the configuration through UI (and it updates the ConfigMap), then any further release upgrade will reset it. Is that intended?

vishaltak:
Was facing issues with rebase. Tried to delete and psuh it again. PR got closed. So I've created https://github.com/kubernetes/dashboard/pull/5530



> One related question that I did not think about when reviewing this feature: if user changes the configuration through UI (and it updates the ConfigMap), then any further release upgrade will reset it. Is that intended?

Yes, it would reset it. Ideally, I would expect that any change in the dashboard settings should go through a helm upgrade.

A related issue to this is https://github.com/kubernetes/dashboard/issues/5531

PR #5525, opened 2020-09-10
Helm Chart: Add options for extraVolumes and extraVolumeMounts

reubit:
This is to add options for specifying helm values for `extraVolumes` and `extraVolumeMounts`.

I've recently needed to provide a custom `kubeconfig` file to the dashboard, to make it work with `kube-oidc-proxy` (https://github.com/jetstack/kube-oidc-proxy). Without being able to specify `extraVolumes` and `extraVolumeMounts`, you're unable to actually provide a custom `kubeconfig` file using the official helm chart, even though the dashboard binary supports the option.

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.5.1
+version: 2.6.0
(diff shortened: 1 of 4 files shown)


desaintmartin:
Thanks!

Note that I use dex and oauth2-proxy to do the same WITHOUT having to provide anything to kubernetes-dashboard but the bearer token (through ingress configuration). This allows me to avoid having to manage secrets. Downside is that it only works for the current cluster.

/lgtm

PR #5593, opened 2020-10-02
Upgraded to a slightly newer kubernetesui/metrics-scraper

zerkms:
1.0.4 is more than a half year old now. 1.0.5 is marginally better, but is there a reason to not upgrade?

An important note: I'm **NOT** a hacktoberfest hunter, even though this PR is a one character changed only :-D I'm just a person who maintains several k8s clusters and has an alert set if docker containers running are too old.

### DIFF
--- a/aio/deploy/alternative.yaml
+++ b/aio/deploy/alternative.yaml
@@ -253,7 +253,7 @@ spec:
     spec:
       containers:
         - name: dashboard-metrics-scraper
-          image: kubernetesui/metrics-scraper:v1.0.4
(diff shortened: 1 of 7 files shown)


maciaszczykm:
Thanks for the PR, please update it also in the other places then we can merge.

maciaszczykm:
https://github.com/kubernetes/dashboard/search?q=metrics-scraper%3Av1.0.4

zerkms:
Indeed, my apologise, I should have grepped first.

Now I think it's good.

maciaszczykm:
Thanks.

/lgtm
/approve

PR #5613, opened 2020-10-08
Update creating-sample-user.md -- token is encoded with base64

avoidik:
Highlight that returned data is base64 encoded, and it should be decoded to get valid bearer token

### DIFF
--- a/docs/user/access-control/creating-sample-user.md
+++ b/docs/user/access-control/creating-sample-user.md
@@ -47,7 +47,7 @@ EOF
 Now we need to find token we can use to log in. Execute following command:
 
 ```bash
-kubectl -n kubernetes-dashboard get secret $(kubectl -n kubernetes-dashboard get sa/admin-user -o jsonpath="{.secrets[0].name}") -o jsonpath="{.data.token}{'\n'}"
+kubectl -n kubernetes-dashboard get secret $(kubectl -n kubernetes-dashboard get sa/admin-user -o jsonpath="{.secrets[0].name}") -o jsonpath="{.data.token}{'\n'}" | base64 --decode
 ```
 
 It should print something like:


docs/user/access-control/creating-sample-user.md
```diff
@@ -47,7 +47,7 @@ EOF
 Now we need to find token we can use to log in. Execute following command:
 
 ```bash
-kubectl -n kubernetes-dashboard get secret $(kubectl -n kubernetes-dashboard get sa/admin-user -o jsonpath="{.secrets[0].name}") -o jsonpath="{.data.token}{'\n'}"
+kubectl -n kubernetes-dashboard get secret $(kubectl -n kubernetes-dashboard get sa/admin-user -o jsonpath="{.secrets[0].name}") -o jsonpath="{.data.token}{'\n'}" | base64 --decode
```
floreks, line 50: I guess the issue that the previous PR (https://github.com/kubernetes/dashboard/pull/5605) tried to fix with using different commands for Windows/Linux cmd will appear again when using pipe and base64. I am not sure if this is allowed in the PowerShell.

@kohtala unless you have a way to fix this, we'll have to revert to the old version.

kohtala: Ok. Sorry about that. Did not notice the describe had decoded it. The base64 is not there in base Windows 10 powershell. There is a way though, but you'd need to again separate the commands for sh and PowerShell:

PS C:\WINDOWS\system32> [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($(kubectl -n kube-public get secret $(kubectl -n kube-public get sa/default -o jsonpath='{.secrets[0].name}') -o jsonpath="{.data.token}{'\n'}")))

kohtala: Good that it was found so fast. That part was excellent. :-)

kohtala: base64 is part of coreutils, so I think use of it is fine for shell. I checked jsonpath just in case, but no conversion there.

floreks: If we need to split them again I'd just revert to using old commands.

floreks:
I am reverting the previous PR to use the old commands https://github.com/kubernetes/dashboard/pull/5618

/close

kohtala:
@avoidik, could you take a look at the #5621 if that is good?

PR #5660, opened 2020-10-22
Sync pod status logic with kubectl and simplify API

### LINKED ISSUE #5656: A "Terminating" pod is shown incorrectly in the dashboard as "Waiting: ContainerCreating"
##### Environment

```
Installation method: kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/v2.0.4/aio/deploy/recommended.yaml
Kubernetes version: v1.19.0
Dashboard version: 2.0.4
Operating system: CentOS Linux release 7.8.2003 (Core)
Node.js version ('node --version' output): ??
Go version ('go version' output): ??
```

##### Steps to reproduce
I have a deployment that runs 3 pods.

Starting the pod takes a few seconds.

In my test I used this as the basis: https://github.com/nielsbasjes/yauaa/blob/master/webapp/yauaa-k8s-example.yaml

I see the effect when I simply do a restart of the deployment with
`kubectl -n yauaa rollout restart deployment yauaa`

##### Observed result
Looking at the pod status the same moment via command line and in the UI:
`kubectl -n yauaa get pods -owide`
Note: I sorted the results to match the ordering in the screenshot.

```
NAME                     READY   STATUS        RESTARTS   AGE     IP               NODE         NOMINATED NODE   READINESS GATES
yauaa-565956f64f-6dmd9   0/1     Running       0          7s      172.30.100.170   node11.xxx   <none>           <none>
yauaa-565956f64f-ftdxr   1/1     Running       0          26s     172.30.3.67      node10.xxx   <none>           <none>
yauaa-565956f64f-4x6r7   1/1     Running       0          46s     172.30.228.216   node12.xxx   <none>           <none>
yauaa-58fc466c4-vdpmr    1/1     Running       0          2m21s   172.30.100.171   node11.xxx   <none>           <none>
yauaa-58fc466c4-thqcj    0/1     Terminating   0          2m38s   <none>           node10.xxx   <none>           <none>
```

Kubernetes Dashboard 2.0.4
image

Looking at pod yauaa-58fc466c4-thqcj

The commandline shows it as `Terminating` which is correct.
The dashboard shows it as `Waiting: ContainerCreating` which is incorrect.

##### Expected result
The dashboard shows pod yauaa-58fc466c4-thqcj also as `Terminating`.

##### Comments
No comments.

floreks:
- Moved logic that determines the status of a pod to the backend
- Used current kubectl logic to determine the status of the pod
- Simplified status of the pod to a single string instead of using a more complicated structure and adding an additional layer of logic in the frontend
- Updated frontend logic that determines icon display based on pod status

The logic that figures out the icon to use basically checks if the returned status is a known one and only then the specific icon will be used. For unchecked statuses, it will most often be a question mark icon meaning that we simply do not have a specific icon for it or it is not recognized by us.

I have checked the output with `kubectl` and it was the same at all times.

Other resources might still require additional work as they work on much simpler statuses. The pod itself has a wider range of possible statuses.

Zrzut ekranu z 2020-10-22 14-36-29

Zrzut ekranu z 2020-10-22 14-37-49

Fixes #5656


cameronbraid:
I now get green tick icons for pods that aren't ready :(

floreks:
I don't see how it is possible as logic is simple here:
```ts
[Status.Succeeded, Status.Running, Status.Completed].some(s => resource.status === s)
```

If pod status is different than one of those 3, then it can't be green.

cameronbraid:
Here is what I am seeing

Two tabs open
* left showing pod list with a green light for the zeebe-zeebe-gateway
* right showing new pod's details with ready false conditions

image

cameronbraid:
Here is how it looks in dashboard 2.0.2

image

floreks:
I'd say that right now it is consistent. Status is Running therefore green icon is shown. This should be consistent with what kubectl reports.

cameronbraid:
It may well be consistent but the previous user experience was better since, allowing me to see from the list view that the pod isn't ready

floreks:
Our goal is and always have been to be consistent with kubectl output. Previous logic was invalid and is considered a bug by us. If status in Dashboard will be different than the one reporter by kubectl we will fix it.

cameronbraid:
I understand that and dont disagree that it should be consistent with kubectl, however I also think that dashboard can provide a better experience.

kubectl shows this in the default pod list view
```
NAME                                    READY   STATUS    RESTARTS   AGE
zeebe-zeebe-gateway-5d9868cc6c-pmzjq    0/2     Running   0          11h
```

So it is immediately obvious from that list that 0 of 2 pods are ready.  

So to be consistent, dashboard should also show this 'Ready' information. 

Would you consider adding a Ready column to dashboard ? Or better still incorporate this ready state into the icon like like dashboard 2.0.2 does.  You could also add a mouseover/tooltip for the icon which reads "ready: 0/2"

maciaszczykm:
Add an issue for that because otherwise we will lose track of it.

PR #5665, opened 2020-10-28
drop unnecessary capabilities

Constantin07:
Minor security improvement - drop unnecessary Linux capabilities.
PS: no issues detected in UI or container logs.

### DIFF
--- a/aio/deploy/recommended.yaml
+++ b/aio/deploy/recommended.yaml
@@ -217,6 +217,8 @@ spec:
             readOnlyRootFilesystem: true
             runAsUser: 1001
             runAsGroup: 2001
+            capabilities:
+              drop: [ "MKNOD", "DAC_OVERRIDE", "NET_BIND_SERVICE", "NET_RAW", "SYS_CHROOT" ]
       volumes:
         - name: kubernetes-dashboard-certs
           secret:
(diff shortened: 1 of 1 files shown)


Constantin07:
ready for review

Constantin07:
I signed it

floreks:
I am not sure if this actually improves anything. We are using a scratch image as a base. Can you actually provide any valid use case that in our case could benefit from this change? Is there any vulnerability inside our image that can allow attackers to somehow exploit those capabilities?

Constantin07:
I haven't scanned the image for vulnerabilities and unfortunately haven't seen the scan reports. But it doesn't mean they may not happen in feature. There is no 100% guarantee even for stock scratch image.
The point is that if upstream image or any dependencies develop a flaw trying to exploit these capabilities, it will help to prevent taken advantage of them.
It's just following principle of least privilege. Secondly, it free - doesn't cost a penny. I don't see how this can make it worse.

floreks:
> I don't see how this can make it worse.

I haven't suggested anything like that. I am only trying to find out if this actually improves anything. Adding "extra protection" for something that could never be actually exploited in the first place seems like overkill IMO.

`scratch` image is basically empty. You can't do anything with it. Inside the image, there is only a static dashboard binary. Nothing else. Privileges are also stripped to a non-root user. That's why I am asking if you think that this could improve security in any way.

1. What are the potential issues now?
1. How the current configuration could be exploited?
1. What types of attacks/exploits can we avoid by dropping those capabilities that are possible now?

maciaszczykm:
Closing as there was no answer.

Issue #2272, opened 2017-08-16
Websocket auth with multiple replicas


PaulFurtado:
##### Environment

```
Dashboard version: 1.6.3
Kubernetes version: 1.7.3
Operating system: (using official docker image)
Node.js version: (using official docker image)
Go version: (using official docker image)
Docker image:  gcr.io/google_containers/kubernetes-dashboard-amd64:v1.6.3
```

##### Steps to reproduce
Run multiple replicas of the dashboard behind a standard service (not behind apiservers).

##### Observed result
If you attempt to use the shell feature, the websockets may fail to connect. The backend will show errors like:
```
handleTerminalSession: can't find session '58b0f210745ff704af190f0d181a5111'
```
I'm guessing this is because one instance of the backend is not aware of sessions created on another instance of the backend.

##### Expected result
I'd expect to be able to run multiple replicas of the dashboard for redundancy. 

##### Comments
I'm curious if there is any workaround that I'm currently missing. It seems like the dashboard makes a standard XHR request to get a terminal session, and then another to open the websocket. Would it be possible to make it skip the XHR request and do everything over the websocket so only one backend is involved?

rf232:
I guess we should update the dashboard service to include some form of session affinity. IIRC it should be as simple as adding `sessionAffinity: "ClientIP"` to the spec of our service. Could you perhaps confirm that for us @PaulFurtado ?

floreks:
Beside that shell feature is still in early stage. We will be working on improving it.

cheld:
@PaulFurtado I curious about your environment. Do you have so many concurrent users that you need multiple replicas?

floreks:
Closing as we'll track it from #2029.

abursavich:
Can we reopen this issue? #2029 seems pretty dead.

PR #5670, opened 2020-11-04
Fix terminal bug when using multiple replicas in helm

mluds:
Set sessionAffinity to ClientIP on the Helm service when using multiple replicas. This keeps the session intact for clients, preventing the terminal from hanging.

Fixes this: #2272

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubernetes-dashboard
-version: 2.8.2
+version: 2.8.3
 appVersion: 2.0.4
 description: General-purpose web UI for Kubernetes clusters
 keywords:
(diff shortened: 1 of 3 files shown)


mluds:
CLA signed

desaintmartin:
This is interesting. Should we unconditionally set this clientIP? AFAIK It won't harm for 1 replica and will allow for things like HPA.

mluds:
Yeah I didn't consider that, but maybe setting it unconditionally, then adding it the values.yaml with a default of ClientIP would be better.

floreks:
Does it mean that Kubernetes can redirect traffic to a different pod while the SPDY exec stream is open? When does this happen actually? Can you describe when and what exactly happens?

maciaszczykm:
@mluds Can you provide any update?

mluds:
@floreks 
When a request reaches a service it will load-balance between the backend pods using round-robin by default. For the dashboard terminal sessions, it seems like there is some state saved on the pod, which doesn't get shared to the other pods. So if your SPDY requests are sent to two different pods it gives an error. (correct me if I'm wrong on any of this)

This change will make requests coming from the same client IP address go to the same pod (with a configurable "sticky" timeout if needed), avoiding the above situation. This will of course not be ideal if all your clients have the same IP address, but I think there's a cookie-based method in that case.

desaintmartin:
Should we merge this? Indeed, I believe most organization use the dashboard from behind a NAT with very few IPs, making the load-balancing unbalanced.
Proposing the option is the least we should do, but should we set it by default?

jeefy:
I see the use case here and this is just adding a config / sane default.
/approve

desaintmartin:
Great! @mluds, could you rebase, document the value in the README.md and bump a minor version (new feature!) instead of patch?

PR #5672, opened 2020-11-05
Update German translation.

headcr4sh:
Kubernetes-specific technical terms have not been translated.

A minor translation issue that was introduced in PR #5453, which was
spotted by @mkorbi after the PR was already merged has also been
corrected: "Network Policies" should *not* have been translated
as it's a Kubernetes-specific technical term.


floreks:
Thank you for updating the translation.

/lgtm
/approve

floreks (approved):

Issue #5391, opened 2020-07-24
Provide Global admin settings for dashboard config


FrenchBen:
##### Why is this needed

When a cluster is used by users with limited access, it is desirable to have a Dashboard view which limits similar resources for the user. 

User Scenario:
As a user of a cluster, I am given access to 3 namespaces: dev, test, prod. 
When I login to the Kube Dashboard, I should see only that which I can interact with.

In the above scenario, a user logged in with their credentials, and limited access (RBAC), would not be able to list all namespaces in the cluster (list all namespaces would fail). The user should be have matching settings, where an Admin could provide a pre-defined list of namespaces / resources  that a user would have access to. 

##### Comments
Discussion with @floreks in PR #5390 
>Global settings can be modified usually only by admins. They are available on the top of settings view. Settings are saved in a ConfigMap. I think we should have some kind of field (chip list?) that allows us to define multiple values (namespaces) and delete specific entries if needed. By default it would be a list of fallback namespaces that the user would see in case namespace list request fails. Then we could add a switch below that "list field" to control if defined list of namespaces should always override a list of namespaces available in namespace selector. This means that even if the user can list all namespaces, only the ones defined on the list would be available, other namespaces would have to be entered manually.

##### Further Improvements
Although the above would provide some settings that could come handy, there is a greater desire to see this implemented as a core RBAC feature of kubernetes/kubernetes, where a user doing a lookup on all resources would be given resources for which they have access to. 

Using the above user scenario, a user should be able to list namespaces via the cli `kubectl get namespaces`, and for the APi server to return the only namespaces the user has access to: `dev, test, prod`

floreks:
> In the above scenario, a user logged in with their credentials, and limited access (RBAC), would not be able to list all namespaces in the cluster (list all namespaces would fail). The user should be have matching settings, where an Admin could provide a pre-defined list of namespaces / resources that a user would have access to.

Important thing here is that there is no concept of user in Kubernetes API and it will not be possible to define per-user list of namespaces. This list would be global for all users.

FrenchBen:
>Important thing here is that there is no concept of user in Kubernetes API and it will not be possible to define per-user list of namespaces. This list would be global for all users. 
For my use case, it would be ok, since all users would be part of the same team (they share the namespaces & good 1st step). For a proper multi-tenant cluster, it wouldn't.

>It would be best if API could actually do this work for us.
Agreed. Having this concept as a 1st class citizen in API would simplify things in the dashboard.

[...]

Issue #4900, opened 2020-02-12
Set default namespace within request headers


fentas:
##### What would you like to be added

To set the default header the same way as the Authorization Header (via a reverse proxy).

##### Why is this needed
I have multiple users which have no right to list namespaces so _default_ namespace is automatically selected and can't be changed unless they know their default namespace and add a `?namespace=<name>` to the URL.
So it would be really great to select their default namespace via a request header the same way as their auth token.

floreks:
I don't think this is the way it should be done. That's what settings page is for. Global settings should be able to enforce default namespace. It will be much better way to achieve what you described.

fentas:
Does this setting already exists?

Well in my use-case users do not have rights to list namespaces so having the ability to set it for them would be nice.

floreks:
It does not. It is a feature that we have to add.

shu-mutou:
If namespace that users have authority are written in context on kubeconfig like [here](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/) and use it for logging in, we may be able to implement to specify namespace selector with it (also limit namespace selector and stop retrieving namespaces). Does it satisfy your use case? if so, I would consider to implement in #4522.
@floreks HDYT?

floreks:
Reading default context namespace from kubeconfig is ok, but it is a different feature IMHO.

I'd simply add `defaultNamespace` option to the global options. User could then change it to a different namespace and this namespace would be used for all users as default one then.

fentas:
> If namespace that users have authority are written in context on kubeconfig like here and use it for logging in, we may be able to implement to specify namespace selector with it (also limit namespace selector and stop retrieving namespaces). Does it satisfy your use case? if so, I would consider to implement in #4522.

Well I use Authorization header - so no kubeconfig. But this idea goes in the direction of a second header as described above.

> I'd simply add `defaultNamespace` option to the global options. User could then change it to a different namespace and this namespace would be used for all users as default one then.

This is definitely nice to have :+1: 
But not 100% reflects my use-case as the accesstoken is generated for every user the same way as the namespace.
To make it more precise (very specific to my solution right now) I have the following URL `<namespace-name>.dahsboard.example.com`. 
[...]

PR #5674, opened 2020-11-09
Smart default

FrenchBen:
Leverage the target namespace that hosts the dashboard deployment, to determine if the "default" namespace should be used as the first namespace to select.

Use cases:

k8s Dashboard deployed in NS `kube-system` -> default namespace is `default`
k8s Dashboard deployed in NS `kubernetes-dashboard` -> default namespace is `default`
k8s Dashboard deployed in NS `my-namespace` -> default namespace is `my-namespace`

### DIFF
--- a/src/app/backend/handler/confighandler.go
+++ b/src/app/backend/handler/confighandler.go
@@ -67,6 +67,7 @@ func getAppConfigJSON() string {
 	return string(jsonConfig)
 }
 
+// ConfigHandler handles config calls
 func ConfigHandler(w http.ResponseWriter, r *http.Request) (int, error) {
 	configTemplate, err := template.New(ConfigTemplateName).Parse(ConfigTemplate)
 	w.Header().Set("Content-Type", "application/json")
--- a/src/app/backend/handler/confighandler.go
+++ b/src/app/backend/handler/confighandler.go
@@ -20,6 +20,8 @@ import (
 	"net/http"
 	"text/template"
 	"time"
+
+	"github.com/kubernetes/dashboard/src/app/backend/args"
 )
 
 // AppHandler is an application handler.
@@ -28,7 +30,8 @@ type AppHandler func(http.ResponseWriter, *http.Request) (int, error)
 // AppConfig is a global configuration of application.
 type AppConfig struct {
 	// ServerTime is current server time.
-	ServerTime int64 `json:"serverTime"`
+	ServerTime       int64  `json:"serverTime"`
+	DefaultNamespace string `json:"defaultNamespace"`
 }
 
 const (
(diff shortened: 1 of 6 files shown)


floreks:
As it was already discussed in a few issues including https://github.com/kubernetes/dashboard/issues/5391, default namespace selection should be done per-user and saved in a cookie or overridden through the global settings. I don't really want to introduce another backend flag just for this.

FrenchBen:
@floreks I see this as a blocker when the dashboard is deployed to a non-default namespace, the first experience for a user is to find a dashboard that doesn't work, because it's set to "default" namespace. 
Secondly, setting this as a cookie would prevent the setting of being shared across devices, and users, making the experience again less than desirable. 
Global settings is perhaps a way to do this as well, but I don't believe they can be set from the initial deployment. 

The idea of this PR is to have a working dashboard, configured out of the box with a namespace that it can get resources from. This does not echo the feature requests in #4900 or #5391, which are discussing ways to set a "preferred" namespace. These can be added as a preference in the future, as discussed in #5391

floreks:
> @floreks I see this as a blocker when the dashboard is deployed to a non-default namespace, the first experience for a user is to find a dashboard that doesn't work, because it's set to "default" namespace.

The assumption here is invalid as Dashboard itself has almost no privileges. Dashboard acts based on user privileges. You can never guarantee that logged in user will have access to a particular namespace or any namespace actually.

> The idea of this PR is to have a working dashboard, configured out of the box with a namespace that it can get resources from.

For this to be true you'd have to assume that every user has privileges to your preconfigured namespace. There is just no way we can assume this to be always true for everyone. 

In fact, the namespace, where Dashboard is deployed, does not matter at all here. Some users won't even have access to the `kubernetes-dashboard` namespace.

> Secondly, setting this as a cookie would prevent the setting of being shared across devices, and users, making the experience again less than desirable.
Global settings is perhaps a way to do this as well, but I don't believe they can be set from the initial deployment.

It's true that local settings are not shared and in order to share per-user settings across devices, we'd need a way to get the unique user identifier when the user logs in. That's why we only have global settings that can enforce settings for every user that uses Dashboard. They don't have to be set from the initial deployment. They are saved in a config map used by Dashboard and can be dynamically changed while it is running.

The last thing is that we do not want to introduce more flags that have a direct impact on the UI anymore. As we are working on a new API, we want to split it from the UI part and have a standalone API. All the UI configuration stuff will be moved to the ConfigMap, so that we'll be able to dynamically update them without restarting the container.

FrenchBen:
Thanks @floreks for the insight - Closing this in favor of https://github.com/kubernetes/dashboard/pull/5682

PR #5679, opened 2020-11-11
add dir check for default

helight:
add dir existing check for default localdir


floreks:
There is no need for this additional check.

/close

PR #5682, opened 2020-11-12
Add namespace related options to global settings

floreks:
The goal of this PR is to give admin better control over default namespace and namespaces presented to users without proper privileges.

I am currently testing out different ways of implementing this feature. I'd like to offer a way to either validate user input in case of a raw input field or offer a Combobox with a list of available namespaces to select from. I am not sure if we can assume that a user with privileges to update global dashboard settings has privileges to list all namespaces.

### UI
Zrzut ekranu z 2021-02-18 14-19-24

### User without privileges to list namespaces
Zrzut ekranu z 2021-02-18 14-23-17


FrenchBen:
To provide some insights (point of view), I'll provide a few use cases.

## Goal
Upon logging in to the Dashboard, derived from the settings, a User will have access to a pre-define set of NS within the dropdown.

### I am a user with Admin role
As an admin, I am setting up a restricted dashboard for my users.
* I can lookup all Namespaces, thus can validate them, and have a selection (predictive, or dropdown selection)

### I am a user, part of a team, with Limited NS access
As a user, I would like to customize the dashboard to focus on the namespaces for my team (settings shared across all user roles)
* I cannot lookup all Namespaces, but can validate the NS input given (user has access)

### I am a user with Limited NS access
As a user, I would like to customize the dashboard to focus on my namespaces (settings for the user only)
* I cannot lookup all Namespaces, but can validate the NS input given (user has access)

From the above, I can see the following paths:
* Global Settings
    * Set by Admin
    * Set by user (affects all users)
* User Settings
    * Set by user (localStorage setting?)

floreks:
Global settings do not provide a distinction between users. As the name suggests they are global and enforced for everyone. Users with edit privileges are considered an admin here.

We do not have user settings right now and local settings are only stored locally, so there is no way to share them between devices.

The first step is to add global settings options for namespace management. As a next step, we can think about local settings that could override global settings.

omesser:
ping @floreks
Any news / plans to get this in and released ?
A properly namespaced version (which can be limited via chart values/global settings) is something many (well, us at [iguazio](https://www.iguazio.com/) at least) crave for :)

floreks:
Ye, I really wanted to finish it this week but other important issues took over and I didn't have time to finish. I'd like to finish ASAP. Most probably next Thursday I'll have time to push this and then push the release.

omesser:
@floreks no worries of course, there are other things happening at the world atm :] just checking and wanted to mention there are clients for this change
Thanks for the quick reply!

floreks:
Mostly finished now. Namespace selector uses values from settings also. Only some minor styling tweaks left to do and it should be good to go. I will finish this next Thursday and push a new release.

PR #5684, opened 2020-11-13
update zh translation

helight:
1. fix some words
2.translation some word we haven't
3.we did't translate some words because we think we should't do it and those words show keep. but I find that those has already translated in the kubernetes doc,  here is some url for kubernetes zh doc:

* https://kubernetes.io/zh/docs/concepts/storage/storage-capacity/
* https://kubernetes.io/zh/docs/reference/access-authn-authz/rbac/
* https://kubernetes.io/zh/docs/concepts/

You can find more here :  https://kubernetes.io/zh/docs

before this pr, the old UI is like this:
image

now with this pr it is like this:
image

### DIFF
--- a/i18n/zh-Hans/messages.zh-Hans.xlf
+++ b/i18n/zh-Hans/messages.zh-Hans.xlf
@@ -1782,7 +1782,7 @@
       </trans-unit>
       <trans-unit id="c3e2c59f79e1c277706a4106e7ada43e5387a9d0" datatype="html">
         <source>Namespaced</source>
-        <target>Namespaced</target>
+        <target>有命名空间的</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/common/components/resourcelist/crd/template.html</context>
           <context context-type="linenumber">73</context>
@@ -2022,7 +2022,7 @@
       </trans-unit>
       <trans-unit id="6e2329529b1953198c7dfa0edb260554310bc636" datatype="html">
         <source>Nodes</source>
-        <target>Nodes</target>
+        <target>节点</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/common/components/resourcelist/node/template.html</context>
           <context context-type="linenumber">24</context>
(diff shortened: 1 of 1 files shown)


zehuaiWANG:
image

I would rather keep it in English than translate it into Chinese. It's more like our manipulation of kubectl.
What is your opinion? @hwdef

helight:
> image
> 
> I would rather keep it in English than translate it into Chinese. It's more like our manipulation of kubectl.
> What is your opinion? @hwdef

This kind version has some problem： some items are english， some are Chinese. it‘s realy confused

Other reason is that the members in k8smeetup community translated all kubernetes documents to Chinese, and those words are all translated to Chinese. you guys can find here https://kubernetes.io/zh/docs. So I think we should translate it as well

hwdef:
These translations look good, but there is still confusion about things like configmap and secrets. not sure if there is a good solution for this.

@tanjunchen  @tengqm @chenrui333 
Do you have any suggestions?

tengqm:
On website, we don't translate the resource names such as ConfigMap, Secret, Deployment etc. When we are talking about generic terms such as namespace, roles, volumes, we do translate where appropriate.

In this context, I'd vote for not translating the resource names. Strictly speaking, the English source needs some changes as well, because in Kubernetes, nothing is called "Daemon Sets", we only have "DaemonSets". We may want to treat "RoleBindings" as a single term rather than two words "Role Bindings".

Take "Roles" and "ClusterRoles" as examples. If we split the second term into "Cluster Roles", users may see it as a special case of "Roles", so it should be put under the "Roles" group. However, the fact is that "Role" and "ClusterRole" are two separate concepts. If we really want to use "Cluster Roles", we should say "Generic Roles" rather "Roles".

Just my two cents.

helight:
@tanjunchen @tengqm @chenrui333 @zehuaiWANG 
I found that in Japanese, Korean they translated nearly all words, include Deployments, Cron Jobs, Daemon Sets.
In French and German they just translated some of those words。

hwdef:
@helight 
I think  Japanese and Korean cannot be used as a reference. We have different language systems.

For example, `pods` is [ポッド](https://github.com/kubernetes/dashboard/blob/26eaf08939b7229400d8639624e36fafecfbc04e/i18n/ja/messages.ja.xlf#L1475) in Japanese , [파드](https://github.com/kubernetes/dashboard/blob/26eaf08939b7229400d8639624e36fafecfbc04e/i18n/ko/messages.ko.xlf#L399) in Korean. . The reason for this translation is the same pronunciation. 

@tengqm 
Thanks for your suggestion.

helight:
@tanjunchen @tengqm @chenrui333 @zehuaiWANG
hi all guys take a look please, I roll back some translation of resource names.

hwdef:
/lgtm 

Thanks for your contributions!

PR #5693, opened 2020-11-23
Add change language option.

theSinner:
Hi

I added the "choose language" option in the settings.
After choosing a language, the frontend side sets the language code in the cookie and reload the dashboard and the backend side checks the cookie before checking the "Accept-Language" header and load the dashboard based on the selected language.

### DIFF
--- a/i18n/zh-Hans/messages.zh-Hans.xlf
+++ b/i18n/zh-Hans/messages.zh-Hans.xlf
@@ -5517,6 +5517,22 @@
           <context context-type="linenumber">27</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="fe46ccaae902ce974e2441abe752399288298619" datatype="html">
+        <source>Language</source>
+        <target state="new">语言</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
+          <context context-type="linenumber">36</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="170c19b70d40071916e39cad610ac12f269dd473" datatype="html">
+        <source>Change the language of the dashboard</source>
+        <target state="new">更改仪表板的语言</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
+          <context context-type="linenumber">38</context>
+        </context-group>
+      </trans-unit>
       <trans-unit id="dc935c1eb87651baf1b1b0e53119e6dec2db8862" datatype="html">
         <source>
     Shell in
--- a/i18n/zh-Hans/messages.zh-Hans.xlf
+++ b/i18n/zh-Hans/messages.zh-Hans.xlf
@@ -5527,7 +5527,7 @@
       </trans-unit>
       <trans-unit id="170c19b70d40071916e39cad610ac12f269dd473" datatype="html">
         <source>Change the language of the dashboard</source>
-        <target state="new">更改仪表板的语言</target>
+        <target state="new">更改 dashboard 的语言</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
           <context context-type="linenumber">38</context>
--- a/i18n/zh-Hans/messages.zh-Hans.xlf
+++ b/i18n/zh-Hans/messages.zh-Hans.xlf
@@ -5522,7 +5522,7 @@
         <target state="new">语言</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
-          <context context-type="linenumber">36</context>
+          <context context-type="linenumber">37</context>
         </context-group>
       </trans-unit>
       <trans-unit id="170c19b70d40071916e39cad610ac12f269dd473" datatype="html">
@@ -5530,7 +5530,7 @@
         <target state="new">更改 dashboard 的语言</target>
         <context-group purpose="location">
           <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
-          <context context-type="linenumber">38</context>
+          <context context-type="linenumber">39</context>
         </context-group>
       </trans-unit>
       <trans-unit id="dc935c1eb87651baf1b1b0e53119e6dec2db8862" datatype="html">
--- a/i18n/zh-Hant-HK/messages.zh-Hant-HK.xlf
+++ b/i18n/zh-Hant-HK/messages.zh-Hant-HK.xlf
@@ -5517,6 +5517,22 @@
           <context context-type="linenumber">27</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="fe46ccaae902ce974e2441abe752399288298619" datatype="html">
+        <source>Language</source>
+        <target state="new">语言</target>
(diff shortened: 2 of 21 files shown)


floreks (changes requested):
I like the overall idea. I'd propose a few changes:

- Instead of keeping locales in the frontend, I'd move it to the backend and add a new endpoint to list supported locales. It can be a simple structure the same as the one used here.
- I'd remove cookie reading logic in the backend and use angular interceptor to alter the `Accept-Language` request header similar to what we do when [injecting the token](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/services/global/interceptor.ts).

theSinner:
> * I'd remove cookie reading logic in the backend and use angular interceptor to alter the `Accept-Language` request header similar to what we do when [injecting the token](https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/services/global/interceptor.ts).

Because Angular needs the locale list to generate the dashboard for each locale, I think it's better to keep the config together to make future changes easier.

It was my first idea to inject the "Accept-Language" header, and it's okay for the requests that angular send them, but the problem is when we want to load the dashboard itself. Because in this case, the browser sends the request with default parameters, so we can't inject the custom "Accept-Language" header. I had three choices to send the language parameter:

- Query string
- URL param
- Cookie

Using query string and URL params cause more changes and I thought using cookies is a better idea.

src/app/backend/handler/localehandler.go
```diff
@@ -94,7 +94,16 @@ func (handler *LocaleHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		// we want a different index.html (for the right locale) to be served when the page refreshes.
 		w.Header().Add("Cache-Control", "no-store")
 	}
+
```
floreks, line 97: I'd change the code here and make the language cookie the most important, then the env variable and the least important should be the default browser header.

```go
// It does not matter if there was an error. In case the string is empty it will be overriden.
acceptLanguage, _ := r.Cookie("lang")

if len(acceptLanguage) == 0 {
    acceptLanguage = os.Getenv("ACCEPT_LANGUAGE")
}

if len(acceptLanguage) == 0 {
    acceptLanguage = r.Header.Get("Accept-Language")
}
```

floreks:
> Because Angular needs the locale list to generate the dashboard for each locale, I think it's better to keep the config together to make future changes easier.

That's a valid point that I can agree with. I'd then move it to the `index.config.ts` so that we have a single place where this is defined and we can inject it wherever we want.

As for the second point, it's true that the interceptor cannot be called before the app bundle is downloaded from the server. I did not think properly about the data flow here.

Using cookie then seems like the easiest approach as we have to save this information in there anyway.

theSinner:
> > Because Angular needs the locale list to generate the dashboard for each locale, I think it's better to keep the config together to make future changes easier.
> 
> That's a valid point that I can agree with. I'd then move it to the `index.config.ts` so that we have a single place where this is defined and we can inject it wherever we want.
> 
> As for the second point, it's true that the interceptor cannot be called before the app bundle is downloaded from the server. I did not think properly about the data flow here.
> 
> Using cookie then seems like the easiest approach as we have to save this information in there anyway.

Yes, you are right. Using "index.config" is a better idea. I didn't know this file exists.
I moved the languages list there as AVAILABLE_LANGUAGES variable and also make cookie value the most important value for choosing the language in the backend.

i18n/zh-Hans/messages.zh-Hans.xlf
```diff
@@ -5517,6 +5517,22 @@
...
+        <source>Language</source>
+        <target state="new">语言</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
+          <context context-type="linenumber">36</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="170c19b70d40071916e39cad610ac12f269dd473" datatype="html">
+        <source>Change the language of the dashboard</source>
+        <target state="new">更改仪表板的语言</target>
```
helight, line 5530:
```suggestion
        <target state="new">更改 dashbord 的语言</target>
```
we  don't translate the tashboard。

i18n/zh-Hant-HK/messages.zh-Hant-HK.xlf
```diff
@@ -5517,6 +5517,22 @@
...
+        <source>Language</source>
+        <target state="new">语言</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
+          <context context-type="linenumber">36</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="170c19b70d40071916e39cad610ac12f269dd473" datatype="html">
+        <source>Change the language of the dashboard</source>
+        <target state="new">更改仪表板的语言</target>
```
helight, line 5530:
```suggestion
        <target state="new">更改 dashboard 的语言</target>
```

i18n/zh-Hant/messages.zh-Hant.xlf
```diff
@@ -5517,6 +5517,22 @@
...
+        <source>Language</source>
+        <target state="new">語言</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">../src/app/frontend/settings/local/template.html</context>
+          <context context-type="linenumber">36</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="170c19b70d40071916e39cad610ac12f269dd473" datatype="html">
+        <source>Change the language of the dashboard</source>
+        <target state="new">更改儀表板的語言</target>
```
helight, line 5530:
```suggestion
        <target state="new">更改 dashboard 的語言</target>
```

helight:
@theSinner you need to sign the cla(Contributor License Agreement )

📝 Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.

theSinner:
> @theSinner you need to sign the cla(Contributor License Agreement )
> 
> 📝 Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.

Actually, I signed the cla. Maybe I should comment "I signed it" here to recheck.

theSinner:
I signed it

floreks:
@theSinner your commits are missing user/email information. You need to configure git to use proper email when signing the commits.

theSinner:
> @theSinner your commits are missing user/email information. You need to configure git to use proper email when signing the commits.

I fixed the name and email of the last commit, but the cla check still failed. Should I change the previous commits' email/name too?

floreks:
Squash commits into a single commit, rebase and push force.

floreks:
@theSinner I have refactored code a bit and fixed the language caching issue where the browser would not reload language due to naming of output build files. I have also disabled language selector while in dev mode as it would not work anyway.

floreks (approved):

theSinner:
> @theSinner I have refactored code a bit and fixed the language caching issue where the browser would not reload language due to naming of output build files. I have also disabled language selector while in dev mode as it would not work anyway.
> 
> Thank you for your help and for creating the PR. I think it will really help Dashboard users.

Thank you for your time and help to fix the issues. 
I'm glad I was able to contribute to Kubernetes and I will try to contribute more as I can.

PR #5694, opened 2020-11-24
add test intro for i18n 

helight:
When I test for my local language, I found that it didn't work. 
At last I found that we should use this command for my test: `npm start`.

we should use this `npm run start:prod` for my local test for i18n. but this is not mentioned in the document.

So I add some note. Please guy help to check if it is ok.

### DIFF
--- a/docs/developer/getting-started.md
+++ b/docs/developer/getting-started.md
@@ -75,7 +75,7 @@ The Dashboard project can be built for production by using the following task:
 npm run build
 ```
 
-The code is compiled, compressed and debug support removed. The dashboard binary can be found in the `dist` folder.
+The code is compiled, compressed, i18n support is enabled and debug support removed. The dashboard binary can be found in the `dist` folder.
 
 To build and immediately serve Dashboard from the `dist` folder, use the following task:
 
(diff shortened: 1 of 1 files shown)


docs/developer/getting-started.md
```diff
@@ -62,6 +62,8 @@ In the background, `npm start` makes a [concurrently](https://github.com/kimmobr
 
 Once the angular server starts, it takes some time to pre-compile all assets before serving them. By default, the angular development server watches for file changes and will update accordingly.
 
+If you want to test your local language in the web ui by i18n, you should run dashboard with the command: `npm run start:prod`.
```
floreks: There is a section about building prod build a bit lower already. I'd only add a small note there regarding i18n.

Current:

> The code is compiled, compressed and debug support removed. The dashboard binary can be found in the `dist` folder.
To build and immediately serve Dashboard from the `dist` folder, use the following task:

Proposed change
> The code is compiled, compressed, i18n support is enabled and debug features are removed. The dashboard binary can be found in the `dist` folder.
To build and immediately serve Dashboard from the `dist` folder, use the following task:

docs/developer/getting-started.md
```diff
@@ -62,6 +62,8 @@ In the background, `npm start` makes a [concurrently](https://github.com/kimmobr
...
+
 ## Building Dashboard for production
 
 To build dashboard for production, you still need to install `bc`.
```
floreks, line 69: This could also contain an additional small note:

> Please note that i18n support is only available in the production build.

helight: thanks, I put this note into the document.

docs/developer/getting-started.md
helight: ok, I will add this change.

helight: done

PR #5696, opened 2020-11-27
add france for language selector

helight:
const SUPPORTED_LANGUAGES: LanguageConfig[] = [
  {
    label: 'German',
    value: 'de',
  },
  {
    label: 'French',
    value: 'fr',
  },
....

image

### DIFF
--- a/src/app/frontend/index.config.ts
+++ b/src/app/frontend/index.config.ts
@@ -24,6 +24,10 @@ const SUPPORTED_LANGUAGES: LanguageConfig[] = [
     value: 'de',
   },
   {
+    label: 'French',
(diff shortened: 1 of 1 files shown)


src/app/frontend/index.config.ts
```diff
@@ -23,6 +23,10 @@ const SUPPORTED_LANGUAGES: LanguageConfig[] = [
     label: 'German',
     value: 'de',
   },
+  {
+    label: 'France',
```
maciaszczykm, line 27: Change to `French`.

helight: got

PR #5700, opened 2020-12-02
Use perl instead of sed to fix build error in macos.

theSinner:
Hi

In my previous PR for adding change language option, @floreks added some lines to `aio/scripts/build.sh` to prevent caching after changing language using `sed` command which makes build failed in macOS, because in macOS `sed` command is a little bit different from Linux.
I changed `sed` with `perl` which does the same thing and the dashboard builds successfully in macOS and I think it works in Linux, too.

### DIFF
--- a/aio/scripts/build.sh
+++ b/aio/scripts/build.sh
@@ -45,7 +45,7 @@ function build::frontend {
     filename=("$(find "${localeDir}" -name 'main.*.js' -exec basename {} \;)")
 
     mv "${localeDir}/${filename}" "${localeDir}/${language}.${filename}"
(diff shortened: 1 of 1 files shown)


maciaszczykm:
It's true, that macOS has different version of `sed` built-in and it doesn't work with current build script. Normally, I'd recommend to install GNU version of `sed` but if `pearl` will work on Linux too I think we can merge it to make it more convenient for all of us.

@floreks can you verify if it works on Linux? After change there is no more issues on macOS, so I'd be glad to merge it.

floreks:
I will verify and merge if it works.

floreks:
Seems to be working just fine.

/lgtm
/approve

PR #5703, opened 2020-12-05
Fix closing sidemenu issue

theSinner:
Hi

I have an issue with closing the side menu. When I close the side menu, the side menu content went away, but it's space stays.
I think the side menu component code is changed and now the hidden class is assigned to the wrong class.


maciaszczykm:
Good spot. I can confirm that this is a problem on current master. Let me check your solution.

Issue #2464, opened 2017-10-11
Add option/button for OAuth logout


kfox1111:
I have setup oauth2_proxy in front of kube-dashboard and have it working well.

One feature that is suboptomal though, is when the user wants to logout, they must change the url to be /oauth/sign_out manually in the location bar and hit enter.

Could we add an option to the cli to take in a partial url '/oauth/sign_out' to add to an optionally displayed logout link on the UI that would just link to the specified url? Clicking on the link should cause the proxy to do the proper logout.

dbogatov:
I would be interested in this feature too.
For now, I have to manually type the URL.

floreks:
Will be tracked from #2353 
/close

Issue #2353, opened 2017-09-11
Login improvements


floreks:
Tracking issue for further non-critical login improvements.

- [ ] Allow overriding key holder name/namespace with arguments passed to dashboard. This will allow multiple dashboard deployments to use different encryption keys.
- [ ] Token blacklisting feature
   In case user logs out his token should be invalidated. We should blacklist it until it expires.
- [ ] Key rotation feature
- [ ] Consider generating self-signed certificates signed by apiserver CA. This would allow to keep HTTPS enabled by default.
- [ ] Support for external IdP.

cohadar:
It would be nice when I login with kubeconfig file to automatically show me the namespace from kubeconfig file instead of just showing default namespace by default.

pdf:
Is there a ticket on the core side that tracks implementation of the API features/endpoints required for OIDC to work sensibly in dashboard?

floreks:
I haven't been following core roadmap lately, so I don't know. We'll try to revisit it after angular migration.

colemickens:
What's the latest on this? I'm interested in having OIDC auth information published in the `kube-public` namespace for interested clients to discover, for other reasons, in addition to possibly having the dashboard auth scenario be much simpler.

yhaskell:
It would be also awesome if, when you go to some url (for example, `dashboard/#!/log/namespace/pod`), and it required sign, it wouldn't forget your url and send you to the page you wanted, and not to overview of default namespace.

edshelton:
Any progress on OIDC integration w/ the kube-apiserver? 

If I put oauth2_proxy in front of the Kubernetes dashboard, can I pass my auth token to the dashboard somehow so that my users don't have to paste their token all the time?

jurgenweber:
It would be absolutely fantastic if we could get some traction on this.

erandu:
Also interested by the feature that @cohadar suggested :

> It would be nice when I login with kubeconfig file to automatically show me the namespace from kubeconfig file instead of just showing default namespace by default.

Any news/info about this ? 
Thanks

nandonespolo:
Is there any integration on accessing the dashboard and AWS IAM?

Many thanks.

sshishov:
Still waiting for this...

maciaszczykm:
@sshishov You can wait or you can contribute. We are not able to do everything instantly with just a few people.

andloh:
Any news regarding oidc support? :)

[...]

PR #5710, opened 2020-12-09
Add sign out button for Oauth / token

thomastvedt:
-- work in progress --

Problem:
There is no way to sign out from UI when signed in with Authorization header.
In our case we setup oauth proxy under url `/oauth2`, to sign out we need to manually enter url `/oauth2/sign_out`.

Suggested solution:
- Make sign out button visible when signed in with Authorization header.
- Add (optional) parameter `--signout-url` which will be called when signing out

ref issues:
https://github.com/kubernetes/dashboard/issues/2353
https://github.com/kubernetes/dashboard/issues/2464

### DIFF
--- a/src/app/frontend/common/services/global/authentication.ts
+++ b/src/app/frontend/common/services/global/authentication.ts
@@ -109,6 +109,11 @@ export class AuthService {
     this.router_.navigate(['login']);
   }
 
+  logoutWithHeader(): Observable<void> {
+    this.removeAuthCookies();
+    return this.callLogoutUrl();
+  }
+
   /**
    * Sends a token refresh request to the backend. In case user is not logged in
    * with token nothing will happen.
@@ -178,5 +183,13 @@ export class AuthService {
    */
   isLoginEnabled(): boolean {
     return this.isCurrentDomainSecure_() || this.isCurrentProtocolSecure_();
+  }
+
+  private callLogoutUrl(): Observable<void> {
+    const temp = 'oauth2/sign_out';
+
+    // TODO: get sign out url oauth from settings
+
+    return this.http_.get<void>(temp);
   }
 }
--- a/TEMP.md
+++ b/TEMP.md
@@ -1,0 +1,8 @@
+the plan:
+- add sign out button for oauth
+- call sign out url on click (auth service?)
+- configurable sign out url?
+- only call sign out url if config set?
+
+ref: https://github.com/kubernetes/dashboard/issues/2353
+https://github.com/kubernetes/dashboard/issues/2464
--- a/aio/gulp/conf.js
+++ b/aio/gulp/conf.js
@@ -221,6 +221,10 @@ export default {
       argv.enableInsecureLogin :
       false,
     /**
+     * Optional signout url to be called when signing out (if signed in with token)
+     */
+    signoutUrl: argv.signoutUrl,
+    /**
      * Defines token time to live.
      */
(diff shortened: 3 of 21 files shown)


src/app/frontend/common/services/global/authentication.ts
```diff
@@ -179,4 +184,12 @@ export class AuthService {
...
     return this.isCurrentDomainSecure_() || this.isCurrentProtocolSecure_();
   }
+
+  private callLogoutUrl(): Observable<void> {
+    const temp = 'oauth2/sign_out';
+
+    // TODO: get sign out url oauth from settings
+
+    return this.http_.get<void>(temp);
+  }
```
thomastvedt, line 194: Is there a simple way to get cli arguments here? `--signout-url`

floreks:
First we'd need to propose a way it should be done. I think it should be more generic. OAuth is a specific standard that also defines how i.e. logout should be handled.

thomastvedt:
> First we'd need to propose a way it should be done. I think it should be more generic. OAuth is a specific standard that also defines how i.e. logout should be handled.

Great, do you have a suggestion on how it could be implemented?

thomastvedt:
And what about oauth single sign out? dashboard should have a url/hook to support single sign out too?

floreks:
Sorry for the late response. The end of year is always busy.

As Kubernetes Dashboard is something like a proxy for Kubernetes API my initial idea was to offer flags similar to `kube-apiserver`. There are OIDC related flags. In the end, we always have to provide a token acceptable by the apiserver. Based on the OIDC issuer URL we could support login/logout directly from the app without relying on some reverse proxies to handle redirects and inject a token header. I am not sure right now how much we can reuse the k8s API here and what would be the best approach as I haven't investigated this topic in depth.

PS. Having only a flag for `logout` feels very specific and not generic enough.

thomastvedt:
> Sorry for the late response. The end of year is always busy.
> 
> As Kubernetes Dashboard is something like a proxy for Kubernetes API my initial idea was to offer flags similar to `kube-apiserver`. There are OIDC related flags. In the end, we always have to provide a token acceptable by the apiserver. Based on the OIDC issuer URL we could support login/logout directly from the app without relying on some reverse proxies to handle redirects and inject a token header. I am not sure right now how much we can reuse the k8s API here and what would be the best approach as I haven't investigated this topic in depth.
> 

that sounds like a good idea, but also a lot of work?

> PS. Having only a flag for `logout` feels very specific and not generic enough.

I agree, it would be a quick fix but probably not generic enough :) 


I'll close this PR for now, but I'll throw in a last suggestion:
- Add possibility for generic webhooks on certain events, example: `loggedOut`. This could call our oauth2 sign out url, or post a message to slack, etc?
- Always show the sign out button (even if signed in with oauth). 
- Change `isLoggedIn()` in `src/app/frontend/chrome/userpanel/component.ts:45` to return true when signed in with auth header
- in `AuthService.logout()`, call webhook?

dwimsey:
This has been an ongoing issue for several years - while it may be oauth2 specific, that is by far the largest use case for this exact feature so rather than deny ourselves a feature that would be used by many just because it doesn't solve every user problem, why don't we considering using this as a starting point for the more generic version so we get the benefit of it now without waiting for a total solution that is clearly not going to come considering that this https://github.com/kubernetes/dashboard/issues/2464 has been there since 2018.

PR #5713, opened 2020-12-10
change storageClasses to items

helight:
make the storageClasses return with items, some as pv and pvc.

image


maciaszczykm:
Anything new that is added to the API response structures should be used in the UI. Otherwise it is a waste of resources.

PR #5723, opened 2020-12-16
Nodes table now show the number of pods

marcosdiez:
I modified the `nodes` table to also show the amount of pods.

This is quite useful on AWS, where depending on the network setting and machine size on uses, the number of pods is quite limited.

This is how it looks like now:

image


helight:
@marcosdiez  you should also add the langue for i18

marcosdiez:
@helight Done! I kept my changes in different commits to keep the code review easy.

PR #5725, opened 2020-12-16
show pvc storage capacity when it is pending

helight:
show the pvc storage capacity when it is pending, 
I think it is useful to know the pvc storage capacity whether it is pending or not.


floreks:
This changes the underneath logic and displays something else than the actual capacity. If anything, it should be conditionally determined based on status, not changed like that.

/close

helight:
> This changes the underneath logic and displays something else than the actual capacity. If anything, it should be conditionally determined based on status, not changed like that.
> 
> /close

I think the status of pvc and the property of pvc is different,  
status is for running time, 
property is some attribute that the pvc should have wether it is running or not.

helight:
image

helight:
image

shu-mutou:
+1 for @helight
Capacity of PV might be changed by status, i.e. bound PVC, but capacity fo PVC would not be changed, I think.
If this merged, we would not need to memorize PVC capacity anymore.

shu-mutou:
How about simply adding Spec.Resources.Requests. If so, it is easy to notice about difference in PVC request and capacity of bound PV.

PR #5730, opened 2020-12-18
fix display pvc storage when it is appending

helight:
I think the status of pvc and the property of pvc are different,
status is for running time,
property is some attribute that the pvc should have wether it is running or not.

image
image


shu-mutou:
Request and actual capacity are different, so we can not substitute them.
So I said how about simply adding `pvc.Spec.Resources.Requests`.
From point of UX view, requested spec should be shown in detail view at least, because seeing yaml is not comfortable, IMHO.

floreks:
It is correct as it is now.

> // Spec defines the desired characteristics of a volume requested by a pod author.

> // Status represents the current information/status of a persistent volume claim.
> ...
> // Represents the actual resources of the underlying volume.
> Capacity ResourceList

/close

PR #5732, opened 2020-12-18
helm: use helm api v2 version

cpanato:
Helm2 is deprecated and should not be used anymore, we should use helm3 from now on.
this PR update the chart.yaml to use the v2 apiversion

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 3.1.0
+version: 4.0.0
 appVersion: 2.1.0
 description: General-purpose web UI for Kubernetes clusters
 keywords:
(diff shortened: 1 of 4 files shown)


desaintmartin:
Thanks!

Isn't this a breaking change?

Shouldn't the requirements be incorporated into Chart.yaml?

cpanato:
@desaintmartin let me deploy and upgrade to check more carefully, thanks for the reminder, forgot that part

cpanato:
@desaintmartin tested the upgrade and works fine

aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: v1
+apiVersion: v2
 name: kubernetes-dashboard
-version: 3.0.2
+version: 3.1.0
```
desaintmartin, line 17: I feel we should advertise the breaking change by bumping major version, don't you think?

cpanato: sure!

cpanato: done, pta
update the readme as well

desaintmartin:
For reference:

https://helm.sh/blog/helm-v2-deprecation-timeline/

``` The Helm maintainers committed to providing [...] security patches for Helm v2 until November 2020```

Thanks! 

/lgtm

PR #5734, opened 2020-12-20
The Ingress Detail pane now has an extra table which shows all the rules

marcosdiez:
I've always felt the ingress panel has less information than necessary. So I added an extra table for rules.

It can be tested using the following container: `marcosdiez/dashboard:v2.1.0h`

Although the PR is not ready yet to be merged, I do accept comments on my implementation.

Things I am not really sure:

* should the new `template.html` and `components.ts` be located inside ` src/app/frontend/resource/discovery/ingress/rulelist/` or `src/app/frontend/common/components/resourcelist` ? Since rules are not a  resource,, but a part of `ingress`, I am not sure

* what exactly should I do with `ingresses` that don't point to `services` ? For starters, I've never seen it in real life, but the specs do say it's possible.

* please notice that I "flatten" the original structure so that every row has a hostname (hence my `IngressRuleFlat` interface)

This is how it looks like:

image

### DIFF
--- a/src/app/frontend/common/components/ingressrulelist/component.ts
+++ b/src/app/frontend/common/components/ingressrulelist/component.ts
@@ -1,0 +1,80 @@
+// Copyright 2017 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Component, Input} from '@angular/core';
+import {MatTableDataSource} from '@angular/material/table';
+import {IngressSpecRule, IngressSpecRuleHttpPath} from '@api/backendapi';
+import {KdStateService} from '../../services/global/state';
+import {GlobalServicesModule} from '../../services/global/module';
+
+interface IngressRuleFlat {
+  host?: string;
+  path: IngressSpecRuleHttpPath;
+}
+
+@Component({
+  selector: 'kd-ingressruleflat-card-list',
+  templateUrl: './template.html',
+})
+export class IngressRuleFlatListComponent {
+  @Input() initialized: boolean;
+  @Input() ingressSpecRules?: IngressSpecRule[];
+  @Input() namespace: string;
+
+  private readonly kdState_: KdStateService = GlobalServicesModule.injector.get(KdStateService);
+
+  getIngressRulesFlatColumns(): string[] {
+    return ['Host', 'Path', 'Path Type', 'Service Name', 'Service Port'];
+  }
+
+  trackByIngressRuleFlat(index: number, item: any): any {
+    return item.host ? index : index;
+  }
+
+  ingressSpecRuleToIngressRuleFlat(ingressSpecRules: IngressSpecRule[]): IngressRuleFlat[] {
+    const output: IngressRuleFlat[] = [];
+
+    ingressSpecRules.forEach(ingressSpecRule => {
+      let host: string = null;
+      if (ingressSpecRule.host !== undefined) {
+        host = ingressSpecRule.host;
+      }
+
+      ingressSpecRule.http.paths.forEach(path => {
+        const ingressRuleFlat: IngressRuleFlat = {
+          path: path,
+        };
+        if (host !== null) {
+          ingressRuleFlat.host = host;
+        }
+        output.push(ingressRuleFlat);
+      });
+    });
+    return output;
+  }
+
+  getDataSource(): MatTableDataSource<IngressRuleFlat> {
+    const tableData = new MatTableDataSource<IngressRuleFlat>();
+    const the_input =
+      this.ingressSpecRules === undefined || this.ingressSpecRules === null ? [] : this.ingressSpecRules;
+
+    tableData.data = this.ingressSpecRuleToIngressRuleFlat(the_input);
+
+    return tableData;
+  }
+
+  getDetailsHref(name: string, kind: string): string {
+    return this.kdState_.href(kind, name, this.namespace);
+  }
+}
--- a/src/app/frontend/common/components/ingressrulelist/component.ts
+++ b/src/app/frontend/common/components/ingressrulelist/component.ts
@@ -39,25 +39,17 @@ export class IngressRuleFlatListComponent {
   }
 
   ingressSpecRuleToIngressRuleFlat(ingressSpecRules: IngressSpecRule[]): IngressRuleFlat[] {
-    const output: IngressRuleFlat[] = [];
-
-    ingressSpecRules.forEach(ingressSpecRule => {
-      let host: string = null;
-      if (ingressSpecRule.host !== undefined) {
-        host = ingressSpecRule.host;
-      }
-
-      ingressSpecRule.http.paths.forEach(path => {
-        const ingressRuleFlat: IngressRuleFlat = {
-          path: path,
-        };
-        if (host !== null) {
-          ingressRuleFlat.host = host;
-        }
-        output.push(ingressRuleFlat);
-      });
-    });
-    return output;
+    return [].concat(
+      ...ingressSpecRules.map(rule =>
+        rule.http.paths.map(
+          specPath =>
+            ({
+              host: rule.host || '',
+              path: specPath,
+            } as IngressRuleFlat)
+        )
+      )
+    );
   }
 
   getDataSource(): MatTableDataSource<IngressRuleFlat> {
--- a/src/app/frontend/common/components/ingressrulelist/component.ts
+++ b/src/app/frontend/common/components/ingressrulelist/component.ts
@@ -29,17 +29,13 @@ interface IngressRuleFlat {
 })
 export class IngressRuleFlatListComponent {
   @Input() initialized: boolean;
-  @Input() ingressSpecRules?: IngressSpecRule[];
+  @Input() ingressSpecRules: IngressSpecRule[] = [];
   @Input() namespace: string;
 
   private readonly kdState_: KdStateService = GlobalServicesModule.injector.get(KdStateService);
 
   getIngressRulesFlatColumns(): string[] {
     return ['Host', 'Path', 'Path Type', 'Service Name', 'Service Port'];
-  }
-
-  trackByIngressRuleFlat(index: number, item: any): any {
-    return item.host ? index : index;
   }
 
   ingressSpecRuleToIngressRuleFlat(ingressSpecRules: IngressSpecRule[]): IngressRuleFlat[] {
@@ -66,10 +62,7 @@ export class IngressRuleFlatListComponent {
 
   getDataSource(): MatTableDataSource<IngressRuleFlat> {
     const tableData = new MatTableDataSource<IngressRuleFlat>();
-    const the_input =
-      this.ingressSpecRules === undefined || this.ingressSpecRules === null ? [] : this.ingressSpecRules;
-
-    tableData.data = this.ingressSpecRuleToIngressRuleFlat(the_input);
+    tableData.data = this.ingressSpecRuleToIngressRuleFlat(this.ingressSpecRules || []);
 
     return tableData;
   }
--- a/src/app/frontend/common/components/ingressrulelist/template.html
+++ b/src/app/frontend/common/components/ingressrulelist/template.html
@@ -44,14 +44,14 @@ limitations under the License.
         <mat-header-cell *matHeaderCellDef
                          i18n>Path</mat-header-cell>
         <mat-cell *matCellDef="let ingressRuleFlat">
-          <ng-container *ngIf="ingressRuleFlat.path.path">{{ingressRuleFlat.path.path}}</ng-container>
+          {{!!ingressRuleFlat?.path?.path ? ingressRuleFlat?.path?.path : '-'}}
         </mat-cell>
       </ng-container>
       <ng-container [matColumnDef]="getIngressRulesFlatColumns()[2]">
         <mat-header-cell *matHeaderCellDef
                          i18n>Path Type</mat-header-cell>
         <mat-cell *matCellDef="let ingressRuleFlat">
-          <ng-container *ngIf="ingressRuleFlat.path.pathType">{{ingressRuleFlat.path.pathType}}</ng-container>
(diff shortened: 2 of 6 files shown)


maciaszczykm:
Thanks for the contribution. I'll take a look on it later.

floreks:
> Things I am not really sure:
>
> - should the new template.html and components.ts be located inside src/app/frontend/resource/discovery/ingress/rulelist/ or src/app/frontend/common/components/resourcelist ? Since rules are not a resource,, but a part of ingress, I am not sure
>
> - what exactly should I do with ingresses that don't point to services ? For starters, I've never seen it in real life, but the specs do > say it's possible.
>
> - please notice that I "flatten" the original structure so that every row has a hostname (hence my IngressRuleFlat interface)

1. I'd say that this should be a reusable component located in common components. It can be located directly under `common/components`. Similar to `policyrule`.
2.  If there is a rule with empty service I'd simply display `-` instead of `Service Name`

As for the last point. This is a nice work already, but I'd actually extend the API to offer an endpoint that would return ingress rules only and allow reusing our generic list. We do that for most of the resources on the detail views.

We are also using the old `extensions/v1beta1` API and it has to be finally updated to use `networking/v1` as there are differences in the API objects there that will definitely cause some issues between versions.

marcosdiez:
Hey, @floreks , thanks for the comments.

So, about moving this to `common/components`, that's trivial and I can easily do.

According to the specs, rules can't have empty services, they can have a `resource backend` instead, which I believe my code is handling well: https://kubernetes.io/docs/concepts/services-networking/ingress/#resource-backend

About the last point, combined with the fact that you plan to change the API, I am not sure what I should do.
a) nothing
b) implement an endpoint just for the rules and then rewrite whenever we implement the `networking/v1` api ?
c) implement `networking/v1`, then implement the dedicated endpoint.

Since the end result is the same, maybe we want to do a) now (nothing) and c) in a separate PR, for it's a bigger change and under the hood only

what do you think?

floreks:
> About the last point, combined with the fact that you plan to change the API, I am not sure what I should do.
a) nothing
b) implement an endpoint just for the rules and then rewrite whenever we implement the networking/v1 api ?
c) implement networking/v1, then implement the dedicated endpoint.
>
> Since the end result is the same, maybe we want to do a) now (nothing) and c) in a separate PR, for it's a bigger change and under the hood only

Changing  `extensions` to `networking` is rather a small change. Adding a new endpoint is a bit bigger task. If you'd like to do it in a different PR then it is also ok for me. Nevertheless, I'd wait with this PR for the API first and then update this one to do it properly at the first attempt.

The API v2 is a completely different issue not related to the current API. We'll successively migrate functionality "in the background" to match the current one. No worries there.

marcosdiez:
Hi @floreks

As agreed, I

- moved the `rulelist` folder to `src/app/frontend/common/components/ingressrulelist`
- changed the backend to talk to the V1 Networking API ( https://github.com/kubernetes/dashboard/pull/5741 )

It all works well, except that not things only work on k8s >= 1.19

Maybe this change was done too early ?

marcosdiez:
Hi again @floreks !

Problem solved. I created https://github.com/kubernetes/dashboard/pull/5742 , which works on k8s 1.14+

Now everything works perfectly!

Have fun code reviewing it :)

marcosdiez:
Hi again, @floreks ! 
What else is missing for this PR to be merged ?

floreks:
@marcosdiez one CI check fails and you also need to rebase to the latest master as angular version has been bumped and it doesn't compile anymore.

marcosdiez:
Hi @floreks !

I did the merge and the rebase. Somehow, though, `npm run fix i18n` does not modify any files. please take a look of the output:

```
mdiez@batman:~/go/src/github.com/kubernetes/dashboard$ npm run fix:i18n

> kubernetes-dashboard@2.1.0 fix:i18n /home/mdiez/go/src/github.com/kubernetes/dashboard
> ng extract-i18n --outFile ../i18n/messages.xlf && aio/scripts/xliffmerge.sh

✔ Browser application bundle generation complete.

Initial Chunk Files | Names         |      Size
main.js             | main          |  13.70 MB
polyfills-es5.js    | polyfills-es5 | 684.99 kB
polyfills.js        | polyfills     | 266.70 kB

                    | Initial Total |  14.63 MB

Build at: 2021-01-14T12:57:20.182Z - Hash: 1feea0eb816355899af9 - Time: 82606ms

Warning: /home/mdiez/go/src/github.com/kubernetes/dashboard/src/app/frontend/environments/environment.prod.ts is                                                                                                                     part of the TypeScript compilation but it's unused.
Add only entry points to the 'files' or 'include' properties in your tsconfig.


Move translation file messages.de.xlf to be merged by xliffmerge.
Move translation file messages.fr.xlf to be merged by xliffmerge.
Move translation file messages.ja.xlf to be merged by xliffmerge.
Move translation file messages.ko.xlf to be merged by xliffmerge.
Move translation file messages.zh-Hans.xlf to be merged by xliffmerge.
Move translation file messages.zh-Hant.xlf to be merged by xliffmerge.
Move translation file messages.zh-Hant-HK.xlf to be merged by xliffmerge.
Move merged file i18n/messages.de.xlf to i18n/de
Move merged file i18n/messages.fr.xlf to i18n/fr
Move merged file i18n/messages.ja.xlf to i18n/ja
Move merged file i18n/messages.ko.xlf to i18n/ko
Move merged file i18n/messages.zh-Hans.xlf to i18n/zh-Hans
Move merged file i18n/messages.zh-Hant.xlf to i18n/zh-Hant
Move merged file i18n/messages.zh-Hant-HK.xlf to i18n/zh-Hant-HK
mdiez@batman:~/go/src/github.com/kubernetes/dashboard$ git status
On branch ingress_rules
nothing added to commit but untracked files present (use "git add" to track)
mdiez@batman:~/go/src/github.com/kubernetes/dashboard$
```

Is that expected?
Because `npm run docker:build:head` is not happy ?

as a workaround I can merge my code from something older than `origin/master` generate the files from there and "port them back here"

mdiez-modus:
Hi @floreks !

I did the merge and the rebase. Somehow, though, `npm run fix i18n` does not modify any files. please take a look of the output:

```
mdiez@batman:~/go/src/github.com/kubernetes/dashboard$ npm run fix:i18n

> kubernetes-dashboard@2.1.0 fix:i18n /home/mdiez/go/src/github.com/kubernetes/dashboard
> ng extract-i18n --outFile ../i18n/messages.xlf && aio/scripts/xliffmerge.sh

✔ Browser application bundle generation complete.

Initial Chunk Files | Names         |      Size
main.js             | main          |  13.70 MB
polyfills-es5.js    | polyfills-es5 | 684.99 kB
polyfills.js        | polyfills     | 266.70 kB

                    | Initial Total |  14.63 MB

Build at: 2021-01-14T12:57:20.182Z - Hash: 1feea0eb816355899af9 - Time: 82606ms

Warning: /home/mdiez/go/src/github.com/kubernetes/dashboard/src/app/frontend/environments/environment.prod.ts is                                                                                                                     part of the TypeScript compilation but it's unused.
Add only entry points to the 'files' or 'include' properties in your tsconfig.


Move translation file messages.de.xlf to be merged by xliffmerge.
Move translation file messages.fr.xlf to be merged by xliffmerge.
Move translation file messages.ja.xlf to be merged by xliffmerge.
Move translation file messages.ko.xlf to be merged by xliffmerge.
Move translation file messages.zh-Hans.xlf to be merged by xliffmerge.
Move translation file messages.zh-Hant.xlf to be merged by xliffmerge.
Move translation file messages.zh-Hant-HK.xlf to be merged by xliffmerge.
Move merged file i18n/messages.de.xlf to i18n/de
Move merged file i18n/messages.fr.xlf to i18n/fr
Move merged file i18n/messages.ja.xlf to i18n/ja
Move merged file i18n/messages.ko.xlf to i18n/ko
Move merged file i18n/messages.zh-Hans.xlf to i18n/zh-Hans
Move merged file i18n/messages.zh-Hant.xlf to i18n/zh-Hant
Move merged file i18n/messages.zh-Hant-HK.xlf to i18n/zh-Hant-HK
mdiez@batman:~/go/src/github.com/kubernetes/dashboard$ git status
On branch ingress_rules
nothing added to commit but untracked files present (use "git add" to track)
mdiez@batman:~/go/src/github.com/kubernetes/dashboard$
```

Is that expected?

floreks:
I'll try to check later and regenerate them.

src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -0,0 +1,80 @@
...
+  @Input() namespace: string;
+
+  private readonly kdState_: KdStateService = GlobalServicesModule.injector.get(KdStateService);
+
+  getIngressRulesFlatColumns(): string[] {
+    return ['Host', 'Path', 'Path Type', 'Service Name', 'Service Port'];
+  }
+
+  trackByIngressRuleFlat(index: number, item: any): any {
+    return item.host ? index : index;
```
floreks, line 42: This is basically equal to `return index` in the current form.

src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -0,0 +1,80 @@
...
+  @Input() ingressSpecRules?: IngressSpecRule[];
+  @Input() namespace: string;
+
+  private readonly kdState_: KdStateService = GlobalServicesModule.injector.get(KdStateService);
+
+  getIngressRulesFlatColumns(): string[] {
+    return ['Host', 'Path', 'Path Type', 'Service Name', 'Service Port'];
+  }
+
+  trackByIngressRuleFlat(index: number, item: any): any {
```
floreks, line 41: I don't see this function being used in the template. Doesn't seem to be needed.

src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -0,0 +1,80 @@
...
+    return item.host ? index : index;
+  }
+
+  ingressSpecRuleToIngressRuleFlat(ingressSpecRules: IngressSpecRule[]): IngressRuleFlat[] {
```
floreks, line 45: This can be simplified.

```typescript
    return [].concat(
      ...ingressSpecRules.map(rule =>
        rule.http.paths.map(
          specPath =>
            ({
              host: rule.host || '',
              path: specPath,
            } as IngressRuleFlat)
        )
      )
    );
```

src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -0,0 +1,80 @@
...
+  path: IngressSpecRuleHttpPath;
+}
+
+@Component({
+  selector: 'kd-ingressruleflat-card-list',
+  templateUrl: './template.html',
+})
+export class IngressRuleFlatListComponent {
+  @Input() initialized: boolean;
+  @Input() ingressSpecRules?: IngressSpecRule[];
```
floreks, line 32: This shouldn't be optional as it is the most important input for this component.

Should be
```typescript
  @Input() ingressSpecRules: IngressSpecRule[] = [];
```

src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+        output.push(ingressRuleFlat);
+      });
+    });
+    return output;
+  }
+
+  getDataSource(): MatTableDataSource<IngressRuleFlat> {
+    const tableData = new MatTableDataSource<IngressRuleFlat>();
+    const the_input =
```
floreks, line 69: 1. Do not mix camel case with underscore case. We are using only camel case.
2. This is not needed. Can be simplified to just `tableData.data = this.ingressSpecRuleToIngressRuleFlat(this.ingressSpecRules || []);`

src/app/frontend/common/components/ingressrulelist/template.html
```diff
@@ -0,0 +1,92 @@
...
+        <mat-cell *matCellDef="let ingressRuleFlat">
+          <ng-container *ngIf="ingressRuleFlat.host">{{ingressRuleFlat.host}}</ng-container>
+          <ng-container *ngIf="!ingressRuleFlat.host">*</ng-container>
+        </mat-cell>
+      </ng-container>
+      <ng-container [matColumnDef]="getIngressRulesFlatColumns()[1]">
+        <mat-header-cell *matHeaderCellDef
+                         i18n>Path</mat-header-cell>
+        <mat-cell *matCellDef="let ingressRuleFlat">
+          <ng-container *ngIf="ingressRuleFlat.path.path">{{ingressRuleFlat.path.path}}</ng-container>
```
floreks, line 47: When it is uncertain if the property will exist use `ingressRuleFlat?.path?.path` to avoid undefined exceptions.

Second thing is that for properties that are often empty try to provide some default value like
```html
{{!!ingressRuleFlat?.path?.path ? ingressRuleFlat?.path?.path : '-'}}
```

src/app/frontend/common/components/ingressrulelist/component.ts
marcosdiez: That's fancy! When I grow up I want to code typescript like that.

marcosdiez:
@floreks all requested changes were made. Thanks for the code review. I learned a lot with it!

marcosdiez:
Hey, @floreks 

I was thinking over the weekend.

Since in a few of the ingress entries we have both a `hostname` and a `path`, it would be great if I could make them links so the user could click.

My biggest problem, though, is that there is no way I can obtain the protocol (HTTP or HTTPS).

In my setup for example, I have an AWS ALB doing the HTTPS part and forwarding everything to k8s. Hence, with the exception of a special custom flag in my ingress controller saying `trust that all traffic is encrypted somewhere else`, nowhere else my cluster is aware of that.

How to we solve that in the UI level ?

Shall I always put HTTPS in the links in every column of the ingress controller ?
Shall I use the "://" which means the browser will use whatever the k8s dashboard use ?
Shall I add two buttons, one called HTTP and one called HTTPS so the user picks which he wants to click ?
Shall I do nothing ?

Although I don't like neither option, the 3rd seems to be less worse, but your input is appreciated !

One way or another, whatever we decide here is trivial to implement and can easily come in a different PR.

Thanks !
Marcos

floreks:
> My biggest problem, though, is that there is no way I can obtain the protocol (HTTP or HTTPS).
> 
> In my setup for example, I have an AWS ALB doing the HTTPS part and forwarding everything to k8s. Hence, with the exception of a special custom flag in my ingress controller saying trust that all traffic is encrypted somewhere else, nowhere else my cluster is aware of that.
> 
> How to we solve that in the UI level ?
> 
> Shall I always put HTTPS in the links in every column of the ingress controller ?
> Shall I use the "://" which means the browser will use whatever the k8s dashboard use ?
> Shall I add two buttons, one called HTTP and one called HTTPS so the user picks which he wants to click ?
> Shall I do nothing ?

I'd leave it for now and create an issue to improve. We could also improve the list view and try to offer working links instead of IP endpoints. If I'd have to pick, we could probably add `http` by default and let the browser/website redirect user to the `https` endpoint automatically in case `http` is not supported.

PR #5737, opened 2020-12-21
Strip ansi color codes and do not use bypass HTML function

floreks:
- Applied `npm audit fix`
- Slight refactoring of log component
- Removed `ansi-to-html` and added `strip-ansi` instead
- Removed usage of `bypassSecurityTrustHtml`
- Fixed potential issue with not escaping some of the HTML codes properly. Replace was not done globally and in some cases, it was possible to display i.e. a clickable link inside logs. 

Since every log line was first sanitized before applying ANSI to HTML conversion, I do not think there was a potential XSS vulnerability. The converter itself would have to be compromised in order to append insecure code. In order to prevent that logs will be further stripped from ANSI codes.

### Example
Zrzut ekranu z 2020-12-21 13-27-11
Zrzut ekranu z 2020-12-21 13-27-33

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "kubernetes-dashboard",
-  "version": "2.0.4",
+  "version": "2.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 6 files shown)


floreks:
Please also update the notification panel as it also uses inner HTML currently.

Zrzut ekranu z 2020-12-21 15-13-07

MichaelJCole:
Hurray!  Thanks!  Go Team!

MattJeanes:
Does this change mean that colours are no longer supported in the logs view?

floreks:
@MattJeanes correct. They are stripped from the logs to not impact readability.

MattJeanes:
A shame, but understandable given the security concern - thanks for letting me know, not sure if that would be worth a mention in the v2.2.0 release notes?

PR #5738, opened 2020-12-22
Disable codecov check for patch

shu-mutou:
We use codecov for just information as described in `.codecov.yml`.
So it is enough to check only `project` as of now.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,14 +8,15 @@ coverage:
   range: "70...90"
 
   status:
-    project: yes
+    project:
       default:
         target: 75%
         threshold: 1%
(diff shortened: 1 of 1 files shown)


maciaszczykm:
Makes sense for me because it is still flaky. @floreks WDYT?

floreks:
If this will actually work then I'm okay with it.  I have the feeling that parts of this config are not working.

shu-mutou:
@floreks Current yaml is broken. In fact, following configs are ignored:
https://github.com/kubernetes/dashboard/blob/master/.codecov.yml#L12-L18
So I just fixed them 😉 

Because we do check for `project` as `informational`, we don't need check for `patch`. So this config reduces one of CI check results.

PR #5741, opened 2020-12-30
backend: ingress will now use the networking/v1 API instead of extensions/v1beta1 (merge only in 2022?)

marcosdiez:
Hi @floreks 

As we spoke in my other PR ( https://github.com/kubernetes/dashboard/pull/5734 ), here is the code change to use the new v1 API.

You said it was a trivial change and, well, you were right. 
It does work, but only on k8s >= 1.19

Then again, do we want to have a version of the dashboard that does not work with k8s <= 1.18  ?
Can we make it work with both ?

Marcos

### DIFF
--- a/src/app/backend/api/types.go
+++ b/src/app/backend/api/types.go
@@ -204,7 +204,7 @@ var KindToAPIMapping = map[string]APIMapping{
 	ResourceKindDeployment:               {"deployments", ClientTypeAppsClient, true},
 	ResourceKindEvent:                    {"events", ClientTypeDefault, true},
 	ResourceKindHorizontalPodAutoscaler:  {"horizontalpodautoscalers", ClientTypeAutoscalingClient, true},
-	ResourceKindIngress:                  {"ingresses", ClientTypeExtensionClient, true},
+	ResourceKindIngress:                  {"ingresses", ClientTypeNetworkingClient, true},
 	ResourceKindJob:                      {"jobs", ClientTypeBatchClient, true},
 	ResourceKindCronJob:                  {"cronjobs", ClientTypeBetaBatchClient, true},
 	ResourceKindLimitRange:               {"limitrange", ClientTypeDefault, true},
--- a/src/app/backend/resource/ingress/common.go
+++ b/src/app/backend/resource/ingress/common.go
@@ -16,12 +16,12 @@ package ingress
 
 import (
 	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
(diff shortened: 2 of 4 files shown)


floreks:
https://github.com/kubernetes/dashboard/releases

We are mainly focusing on supporting the latest releases and we do not guarantee full backward compatibility. Using `networking/v1` is fine for me if it works for k8s v1.20. It should also work for v1.19.

marcosdiez:
Hi @floreks 

I understand the rationale, but historically, compatibility has been great.

Why not keep it this way ?

If there was any killer feature that the latest API provides, I would agree using the latest API would be great.
But this PR would actually mean using the latest API for latest API's sake. No extra feature for the user.
In other words, for those who can't upgrade k8s for one reason or another (like AWS users who are currently stuck to 1.18), the latest version of the dashboard would actually do less.

How about we merge my other PR and I commit updating my ingress code whenever  k8s 1.22 is released  ?
Let's think of our users. Let's write software that does more, not less.

floreks:
Actually, we do not verify if Dashboard even works with older versions. There might be already some existing issues with older versions. I am wondering about the time schedule for `networking/v1beta1` support. If it won't be deprecated (removed) any time soon (at least half a year) we can use it for now.

Are there any schema changes between `v1beta1` and `v1`?

marcosdiez:
Hi @floreks 

yes, there are some trivial changes (field names) between `networking/v1beta1` and `networking/v1`

That being said, the former will be supported until k8s 1.22.
source: https://v1-19.docs.kubernetes.io/docs/setup/release/notes/

```
Ingress and IngressClass resources have graduated to networking.k8s.io/v1. Ingress and IngressClass types in the extensions/v1beta1 and networking.k8s.io/v1beta1 API versions are deprecated and will no longer be served in 1.22+. Persisted objects can be accessed via the networking.k8s.io/v1 API. Notable changes in v1 Ingress objects (v1beta1 field names are unchanged):
```

According to the release schedule (source: https://en.wikipedia.org/wiki/Kubernetes )


1.16 | 22 October 2019 |  
-- | -- | --
1.17 | 9 December 2019 |  
1.18 | 25 March 2020 |  
1.19 | 26 August 2020[22] | From Kubernetes version 1.19 on, the support window will be extended to one year[21]
1.19.3 | 14 October 2020[23] |  
1.20 | 8 December 2020 | https://kubernetes.io/blog/2020/12/08/kubernetes-1-20-release-announcement/

it should last for more than 6 months. So I would keep it

floreks:
I'd close it for now.

/close

PR #5757, opened 2021-01-12
update comments

jacky68147527:
What type of PR is this?

/kind comment

What this PR does / why we need it:
update comment of MemoryRequests.

### DIFF
--- a/src/app/backend/resource/node/detail.go
+++ b/src/app/backend/resource/node/detail.go
@@ -49,7 +49,7 @@ type NodeAllocatedResources struct {
 	// CPUCapacity is specified node CPU capacity in milicores.
 	CPUCapacity int64 `json:"cpuCapacity"`
 
-	// MemoryRequests is a fraction of memory, that is allocated.
+  // MemoryRequests is number of allocated bytes.
 	MemoryRequests int64 `json:"memoryRequests"`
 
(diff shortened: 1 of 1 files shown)


src/app/backend/resource/node/detail.go
```diff
@@ -49,7 +49,7 @@ type NodeAllocatedResources struct {
 	// CPUCapacity is specified node CPU capacity in milicores.
 	CPUCapacity int64 `json:"cpuCapacity"`
 
-	// MemoryRequests is a fraction of memory, that is allocated.
+  // MemoryRequests is number of allocated bytes.
```
floreks, line 52: Lint check failed.

Also "MemoryRequests is **a** number of allocated bytes.".

floreks: After checking the docs even the new comment isn't correct.

> requests.memory is the maximum combined Memory requests for all the containers in the Namespace. In the above example, you can have 50 containers with 2MiB requests, five containers with 20MiB CPU requests, or even a single container with a 100MiB request. As long as the total requested Memory in the Namespace is less than 100MiB!

floreks: In our case memory requests are also taken from the pods running on the node. It will be a maximum combined memory requests for all the pods on the node.

PR #5760, opened 2021-01-12
var pod collides with the imported package name

twilight327426371:
img


twilight327426371:
I signed it

twilight327426371:
I signed it

floreks:
We are aware of such warnings in many places but it was an intentional decision to ignore them.

/close

twilight327426371:
ok, i get it, thanks

PR #5761, opened 2021-01-13
Reduce the pressure on api server caused by daemonset related pod queries

twilight327426371:
Query pods related to daemonset. If select is not set, all pods in the current namespace of daemonset will be queried. If there are more pods in the current namespace, the pressure on the api server will increase greatly. Set the label when querying to reduce the number of query pods.

### DIFF
--- a/src/app/backend/resource/daemonset/pods.go
+++ b/src/app/backend/resource/daemonset/pods.go
@@ -56,8 +56,14 @@ func getRawDaemonSetPods(client k8sClient.Interface, daemonSetName, namespace st
 		return nil, err
 	}
 
+	selector, err := metaV1.LabelSelectorAsSelector(daemonSet.Spec.Selector)
+	if err != nil {
+		return nil, err
+	}
+	options := metaV1.ListOptions{LabelSelector: selector.String()}
+
 	channels := &common.ResourceChannels{
-		PodList: common.GetPodListChannel(client, common.NewSameNamespaceQuery(namespace), 1),
+		PodList: common.GetPodListChannelWithOptions(client, common.NewSameNamespaceQuery(namespace), options, 1),
 	}
 
 	podList := <-channels.PodList.List


floreks:
IIRC we have intentionally not used this type of filtering as the only certain way to check if pod is controller by deamon set is to do it by the controller ref as it is done at the end of this method `common.FilterPodsByControllerRef(daemonSet, podList.Items)`.

@maciaszczykm do you remember that case?

maciaszczykm:
More or less I think this was the case. Perhaps it has changed, but we would need to have some kind of docs to confirm that.

twilight327426371:
> IIRC we have intentionally not used this type of filtering as the only certain way to check if pod is controlled by deamon set is to do it based on the controller ref as it is done at the end of this method `common.FilterPodsByControllerRef(daemonSet, podList.Items)`.
> 
> @maciaszczykm do you remember that case?


In the actual production process, frequent queries of damonset pod caused the k8s api server pressure to reach 70%，You will encounter performance problems if you don’t find the label when searching. If a nameapce has thousands of pods， This is not to solve accuracy.

maciaszczykm:
But it may break accuracy if the label selector will be missing on the daemon set, isn't it?

floreks:
The first and most important thing that we need to ensure is to always match correct pods with correct daemon sets. The second thing is the performance. If the first point would be somehow impacted by the performance improvement then we cannot allow that.

twilight327426371:
> The first and most important thing that we need to ensure is to always match correct pods with correct daemon sets. The second thing is the performance. If the first point would be somehow impacted by the performance improvement then we cannot allow that.

Label filtering during query to reduce api server pressure，`common.FilterPodsByControllerRef(daemonSet, podList.Items)` 
 alos ensure is to always match correct pods with correct daemon sets.

twilight327426371:
> > The first and most important thing that we need to ensure is to always match correct pods with correct daemon sets. The second thing is the performance. If the first point would be somehow impacted by the performance improvement then we cannot allow that.
> 
> Label filtering during query to reduce api server pressure，`common.FilterPodsByControllerRef(daemonSet, podList.Items)`
> alos ensure is to always match correct pods with correct daemon sets.

企业微信截图_16109345506120

floreks:
Can you guarantee with 100% certainty that filtering by selector will not filter out a pod that does bot match some selector but would be matched by controller ref?

twilight327426371:
> 100% certainty that filtering by selector will not filter out a pod that does not match some selector but would be matched by controller ref


I have created a lot of daemonsets in the system, tested it can guarantee 100% certainty that filtering by selector will not filter out a pod that does not match some selector but would be matched by controller ref.

image

twilight327426371:
> Can you guarantee with 100% certainty that filtering by selector will not filter out a pod that does not match some selector but would be matched by controller ref?

Yes, we use the daemonset selector to match to ensure that the matched pod and controller ref match the same. If there are other pods configured with the same daemonset label, the controller ref shall prevail. I have created many in the system daemonset, can be 100% guaranteed to match the corresponding POD

floreks:
I have done some tests and it seems that a selector is required when creating DaemonSet and it is always propagated to the created Pods. Updating the Pod itself forces it to get "disconnected" from the daemon set and it gets replcaed by a new Pod. It should be safe then to limit the search to matched pods only.

PS. Fix the static check.

floreks:
@twilight327426371 can you fix the CI so we can merge?

twilight327426371:
@floreks  Please help retest

twilight327426371:
@floreks fix static check.

Issue #5751, opened 2021-01-08
Improve the theming options


maciaszczykm:
- [x] change theme switch to button toggle to allow more options in the future
- [x] expand theme interface to add display names and indicator if theme is dark or light
- [x] add configuration file for themes (for example JSON; with list of themes and the default one)
- [ ] add option to easily change the configuration and mount new themes into the container

PR #5763, opened 2021-01-14
Improve theming options

maciaszczykm:
It will allow further improvements as in the future we would like to allow users having more than two themes, where they can add new ones as easy as possible.

Zrzut ekranu 2021-01-14 o 15 42 18

Closes https://github.com/kubernetes/dashboard/issues/5751.


src/app/frontend/common/components/textinput/component.ts
```diff
@@ -68,7 +69,13 @@ export class TextInputComponent implements OnInit, AfterViewInit, OnChanges {
   constructor(private readonly themeService_: ThemeService) {}
 
   ngOnInit(): void {
-    this.theme = this.themeService_.isLightThemeEnabled() ? EditorTheme.light : EditorTheme.dark;
+    switch (this.themeService_.theme) {
```
maciaszczykm, line 72: In the next change this will be switched to just checking if the theme is dark or light. It's planned in https://github.com/kubernetes/dashboard/issues/5751.

src/app/frontend/common/services/global/theme.ts
```diff
@@ -14,19 +14,25 @@
 
 import {EventEmitter, Injectable} from '@angular/core';
 import {ThemeSwitchCallback} from '@api/frontendapi';
+import {Theme} from '@api/backendapi';
 
 @Injectable()
 export class ThemeService {
-  private isLightThemeEnabled_ = true;
-  private readonly onThemeSwitchEvent_ = new EventEmitter<boolean>();
+  private _availableThemes = [Theme.Light, Theme.Dark];
```
maciaszczykm, line 21: In the future this will be loaded from JSON config file. It's planned in https://github.com/kubernetes/dashboard/issues/5751.

src/app/frontend/typings/backendapi.ts
```diff
@@ -714,7 +714,12 @@ export interface CsrfToken {
 }
 
 export interface LocalSettings {
-  isThemeDark: boolean;
+  theme: Theme;
+}
+
+export enum Theme {
+  Light = 'kd-light-theme',
+  Dark = 'kd-dark-theme',
```
maciaszczykm, line 722: We will use display names once the JSON configuration will be done.

FrenchBen:
Semantics, but wouldn't this change be a "Color Scheme" rather than a "Theme"?
We're changing the supported colors of the site, not the actual layout. 
Think of the plethora of Wordpress Themes vs the different colors that can be applied to the same theme.

maciaszczykm:
@FrenchBen, in default "themes" the layout is not changed much or at all, but there is an option to do this. Since we want to enable theming for 3rd parties I'd keep calling it theme. You are basically free to do any CSS changes from you own theme. 

Moreover, keeping switch is not a good option for me as we shouldn't enforce that every 3rd party theme will have dark and white mode. If you want to have dark and white mode then you have to implement 2 custom themes, it will be fairly quick. It gives more freedom to the users IMO.

floreks:
A theme can be a color scheme, but a color scheme is not really a theme. We intentionally call it a theme even though our default themes work only with color changes (that's how angular calls it also by the way). The theming framework that we are trying to introduce here will allow everyone to prepare and dynamically add new themes to the Dashboard by only mounting them into the right directory. We do not restrict it only to color changes. Everything will be possible here. You will be able to prepare a completely custom theme that will change the whole interface. Introduce white-labeling and so on.

src/app/frontend/common/components/textinput/component.ts
maciaszczykm: Done.

src/app/frontend/common/services/global/theme.ts
maciaszczykm: Done.

src/app/frontend/typings/backendapi.ts
maciaszczykm: Done.

maciaszczykm:
I've added some improvements:

- The configuration is right now theme array that can be extended easily. 
- There is a new "System" option that will pick automatically light/dark theme based on system color preference.

The next step in this area would be https://github.com/kubernetes/dashboard/issues/5778.

@floreks PTAL

PR #5766, opened 2021-01-21
Fix i18n extract option

shu-mutou:
`ng extract-i18n` does not work in development container now. This fix enables to do it again.

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -6,78 +6,73 @@
         <source>Edit a resource</source>
         <target>Ressource bearbeiten</target>
         <context-group purpose="location">
-          <context context-type="sourcefile">../src/app/frontend/common/dialogs/editresource/template.html</context>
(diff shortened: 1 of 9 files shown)


src/app/frontend/common/components/resourcelist/secret/component.ts
```diff
@@ -29,7 +29,6 @@ import {ListGroupIdentifier, ListIdentifier} from '../groupids';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class SecretListComponent extends ResourceListBase<SecretList, Secret> {
-  @Input() title = 'Secrets';
```
floreks, line 32: This can't be removed as it is used:
https://github.com/kubernetes/dashboard/blob/bd025c33bd36e214185d91796d12f6030b191c23/src/app/frontend/resource/cluster/serviceaccount/detail/template.html#L22-L23

PR #5771, opened 2021-01-25
Ingress Detail: Now the endpoint(s) is/are shown and the TLS Secret, if any

marcosdiez:
image

Ingress detail pane now has two new items:

the endpoints (which were shown in the list view already)
if there is a TLS section in the ingress, the associated secret is now shown on the "flat rule table"

### DIFF
--- a/src/app/frontend/common/components/endpoint/external/template.html
+++ b/src/app/frontend/common/components/endpoint/external/template.html
@@ -25,14 +25,14 @@ limitations under the License.
       <div *ngFor="let port of endpoint?.ports">
         <div>
           <a href="http://{{endpoint.host}}:{{port.port}}"
-            target="_blank">
+             target="_blank">
             {{endpoint.host}}:{{port.port}}
             <mat-icon class="kd-endpoint-icon">open_in_new</mat-icon>
           </a>
         </div>
         <div *ngIf="!port.port">
           <a href="http://{{endpoint.host}}:{{port.nodePort}}"
-            target="_blank">
+             target="_blank">
             {{endpoint.host}}:{{port.nodePort}}
             <mat-icon class="kd-endpoint-icon">open_in_new</mat-icon>
           </a>
@@ -40,7 +40,7 @@ limitations under the License.
       </div>
       <div *ngIf="!endpoint.ports?.length">
         <a href="http://{{endpoint.host}}"
-          target="_blank">
+           target="_blank">
           {{endpoint.host}}
           <mat-icon class="kd-endpoint-icon">open_in_new</mat-icon>
         </a>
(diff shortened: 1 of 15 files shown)


src/app/frontend/common/components/endpoint/external/template.html
```diff
@@ -20,30 +20,32 @@
...
 <div>
-  <div *ngFor="let endpoint of endpoints">
-    <div *ngFor="let port of endpoint?.ports">
-      <div>
-        <a href="http://{{endpoint.host}}:{{port.port}}"
-           target="_blank">
-          {{endpoint.host}}:{{port.port}}
-          <mat-icon class="kd-endpoint-icon">open_in_new</mat-icon>
-        </a>
+  <ng-container *ngIf="endpoints">
```
marcosdiez, line 23: Just for the record, all I added here was <ng-container *ngIf="endpoints"> and </ng-container>
everything else in this diff is caused by identation.

src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -49,12 +51,24 @@ export class IngressRuleFlatListComponent {
...
         )
       )
-    );
+    ) as IngressRuleFlat[];
+
+    ingressRuleList.forEach((ingressRule, _) => {
+      tlsList.forEach((tls, _) => {
+        tls.hosts.forEach((tlsHost, _) => {
+          if (tlsHost === ingressRule.host) {
+            ingressRule.tlsSecretName = tls.secretName;
```
maciaszczykm, line 60: 3 loops nested in each other don't look good. Can we improve it by using set or a map?

marcosdiez: I removed one loop and now I used `every`, which means I can break out of my loops whenever I get the answer I needed. How about that ?

src/app/frontend/common/components/endpoint/external/template.html
floreks: Why is it needed actually? If array will be empty nothing will be changed.

marcosdiez: The page would not load without that. I assume it's some race condition.

floreks: I'll take a look

floreks:
Slightly updated the UI and optimized code.

Zrzut ekranu z 2021-02-04 15-19-00

src/app/frontend/common/components/endpoint/external/template.html
floreks: Fixed

PR #5794, opened 2021-02-08
Pod security context2

marcosdiez:
It turns out both pods and containers can have their own security context.
To make things worse, although the container one has priority, they only share a few of the properties.

I actually spent some time thinking about how to make the UI less confusing, despite the reality.

This is what I came up with:

image

On the code level, I did something that is either very smart or very dumb depending on the point of view:

The backend just exposes the info. So nothing special here.

On the frontend, we now have 3 similar objects:

- PodSecurityContext  (for pods)
- ContainerSecurityContext (for containers)
- PodContainerMergedSecurityContext (artificial structure which is a merge of the two above, with all fields optional, so we can use the same UI to widget both, since they are very similar)

So although the Pod does expose PodSecurityContext   and the container does expose ContainerSecurityContext, the security context list interprete it as PodContainerMergedSecurityContext and plots its accordinly.

As always, @floreks , please be the judge.

By the way I never explicitly had the chance to thank you for thoughtfully reviewing every single PR of mine and even going to the extra work of rewriting them when needed. Thank you. Really!

Here is a minimalist example of a yaml to test this:

```
kind: Pod
apiVersion: v1
metadata:
  name: test-security-context2
  namespace: default
spec:
  containers:
    - name: test-security-context2
      image: alpine:latest
      command:
        - tail
        - '-f'
        - /dev/null
      securityContext:
        readOnlyRootFilesystem: true
  securityContext:
    runAsGroup: 1000
    runAsUser: 1000

```

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -275,7 +275,7 @@
         <target state="new">SE Linux User </target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/frontend/common/components/securitycontext/component.ts</context>
-          <context context-type="linenumber">26</context>
+          <context context-type="linenumber">32</context>
         </context-group>
       </trans-unit>
       <trans-unit id="1985a4160878273346b2ac67f9f2e5cf69aeb3c9" datatype="html">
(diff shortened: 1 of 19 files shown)


floreks:
Zrzut ekranu z 2021-02-11 12-29-45

Refactored code/ui a bit and moved pod security context to the resource information to be consistent.

PR #5798, opened 2021-02-09
Fix error on the Ingress Detail page

marcosdiez:
Steps to reproduce (works 100% of the time):

a) open any ingress detail pane with a rule
b) click on the rule so it opens the website in a new window
c) click anywhere on the left sidebar ( Services, Deployment, etc)

this is javascript error:

image

### DIFF
--- a/src/app/frontend/common/components/ingressrulelist/component.ts
+++ b/src/app/frontend/common/components/ingressrulelist/component.ts
@@ -49,7 +49,7 @@ export class IngressRuleFlatListComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges(changes: {tlsList: SimpleChange}): void {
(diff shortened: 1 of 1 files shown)


src/app/frontend/common/components/ingressrulelist/component.ts
```diff
@@ -49,7 +49,7 @@ export class IngressRuleFlatListComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges(changes: {tlsList: SimpleChange}): void {
-    if (changes.tlsList.currentValue) {
+    if (changes.tlsList !== undefined && changes.tlsList.currentValue) {
```
floreks, line 52: Go for `changes.tlsList && change.tlsList.currentValue`

PR #5800, opened 2021-02-10
Hosts are now shown in the ingress list

marcosdiez:
image

### DIFF
--- a/src/app/backend/resource/ingress/list.go
+++ b/src/app/backend/resource/ingress/list.go
@@ -32,6 +32,7 @@ type Ingress struct {
 
 	// External endpoints of this ingress.
 	Endpoints []common.Endpoint `json:"endpoints"`
+	Hosts     []string          `json:"hosts"`
 }
 
 // IngressList - response structure for a queried ingress list.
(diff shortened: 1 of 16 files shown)


src/app/backend/resource/ingress/list.go
```diff
@@ -74,11 +75,33 @@ func getEndpoints(ingress *v1beta1.Ingress) []common.Endpoint {
...
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
+
+func getHosts(ingress *v1beta1.Ingress) []string {
+	hosts := make([]string, 0)
+	if len(ingress.Spec.Rules) > 0 {
```
floreks, line 89: There is no need to check for length before loop. `for` does that implicitly by not iterating at all if the input array is empty.

src/app/backend/resource/ingress/list.go
```diff
@@ -74,11 +75,27 @@ func getEndpoints(ingress *v1beta1.Ingress) []common.Endpoint {
 	return endpoints
 }
 
+func getHosts(ingress *v1beta1.Ingress) []string {
```
floreks, line 78: I have simplified this method.

PR #5815, opened 2021-02-22
pods now show their service account

marcosdiez:
image

### DIFF
--- a/src/app/frontend/resource/workloads/pod/detail/template.html
+++ b/src/app/frontend/resource/workloads/pod/detail/template.html
@@ -80,6 +80,16 @@ limitations under the License.
       <div value>{{pod.restartCount}}</div>
     </kd-property>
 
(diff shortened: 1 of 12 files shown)


src/app/frontend/resource/workloads/pod/detail/template.html
```diff
@@ -80,14 +80,24 @@
       <div value>{{pod.restartCount}}</div>
     </kd-property>
 
+    <kd-property *ngIf="pod?.serviceAccountName !== undefined">
```
floreks, line 83: As this is a string then `ngIf="pod?.serviceAccountName` is sufficient here. It will only be true for non-empty strings.

mdiez-modus: Another lesson learned. Thank you.

PR #5912, opened 2021-03-22
Support changing number of columns in container envs block

astorath:
Hey!

Sometimes ENV name is quite long (especially when using spring boot or something). Dashboard UI assumes it could always fit 5 columns if env vars in display, but this is not the case: env names do overlap a lot.

This PR makes it possible to configure number of columns to display.

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -59,7 +59,7 @@
         </context-group>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/frontend/resource/config/secret/detail/edit/component.ts</context>
-          <context context-type="linenumber">82</context>
+          <context context-type="linenumber">81</context>
         </context-group>
       </trans-unit>
       <trans-unit id="8c839e142c68aa402fae1482e3099f883aeae446" datatype="html">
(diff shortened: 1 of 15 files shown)


floreks:
The UI should be improved to be more responsive and not rely on this kind of user setting. I wouldn't want to end up with lots of settings that influence small parts of the UI.

astorath:
@floreks How about making it one-column for the time being? Env variable names may be of arbitrary length. Or actually make some option to change the way variables are displayed on the from itself by some sort of switch?

floreks:
I'd simply use autowrap so that it will not force multiple variables into one line if they do not fit, but rather automatically put them into the next row. Another option is to specify the max-width of the element and use ellipsis with a tooltip.

astorath:
Autowrap makes this a mess (because every line is different). Max-width is also not great:
1. It is impossible to find the env you need with Ctrl+F
2. You don't know in general what part of env is the most valuable, but usually it's the end

floreks:
I do see only 2 options here. Either we forget about aligning each entry in a separate row/column and make them auto-align and auto-wrap or we set max-width + add ellipsis for too long variables. Any kind of user settings that control this behavior is a definite no here.

PR #5914, opened 2021-03-22
feat(helm-chart): adding toggle for read-only role logging

sc250024:
# Description

Fairly simple addition. This adds the ability to enable / disable the read-only role's ability to view `Pod` logs in the dashboard. This came up in my company as we wanted to disable only this part of the dashboard. Figured I'd submit it here for a potential merge.

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 4.0.2
+version: 4.1.0
 appVersion: 2.2.0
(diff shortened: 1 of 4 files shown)


desaintmartin:
Hm, this becomes quite specific and I'm sure there are a lot of other valid use cases for adding a lot of conditions.

What I personally do is disabling ClusterRoles creation and use my own (of course, automatically generated). Would it be acceptable for your use case?

sc250024:
> Hm, this becomes quite specific and I'm sure there are a lot of other valid use cases for adding a lot of conditions.
> 
> What I personally do is disabling ClusterRoles creation and use my own (of course, automatically generated). Would it be acceptable for your use case?

We are also considering that as well, but we thought we'd try this avenue too. The change doesn't break default behavior, but it's useful in case some container logs might have sensitive information in them.

Issue #4254, opened 2019-09-04
UI Enhancement: Add Restart Deployment Feature


ryanelian:
##### What would you like to be added

![image](https://user-images.githubusercontent.com/8892187/64242065-15f06780-cf2f-11e9-9713-89516bfbe287.png)

An additional button labeled Restart should be added to the Deployment item dropdown menu.

##### Why is this needed

Kubernetes does not restart deployments when Secrets / ConfigMap are updated (CMIIW).  This feature is essential for quick restart after updating app secrets via the dashboard.

##### Comments

Related PRs:
https://github.com/kubernetes/kubernetes/pull/76062
https://github.com/kubernetes/kubernetes/issues/13488#issuecomment-481023838

zehuaiWANG:
Do you mean to do a feature just like excuting `kubectl rolling-update deployment xxx`

ryanelian:
Yes but the command is:

```
kubectl rollout restart deployment/$deployment
```

rsjohn:
I have been using this for 3hrs now and the first thing I noticed was the lack of pod restarts.  It is needed

galtonova:
Yes, it would help a lot. Exactly as rsjohn said, first thing when I started using dashboard was the lack of rolling update command for deployments. Especially for my deployments that use :latest tagged images.

floreks:
It is quite an interesting feature. We will have to take a look at kubectl implementation and add that.

PR #4542, opened 2019-11-14
Dashboard rollback

zehuaiWANG:

Issue #2889, opened 2018-03-12
Support rolling back Deployments to previous revisions


timoreimann:
##### Environment

```
Dashboard version: 1.8.0
Kubernetes version: v1.8.7-gke.1
Operating system: 
Node.js version:
Go version:
```

##### Steps to reproduce
Try to roll back a Deployment to a previous revision (i.e., what `kubectl rollout undo` does).

##### Observed result
No such UI interaction is available.

##### Expected result
A UI element is available that allows me to select from a list of recent revisions and roll back to any of them.

##### Motivation
Rolling back to a previous release is a fairly important operation as it allows to remedy critical situations quickly. While the functionality is accessible in `kubectl` as well, I find an integration into the Dashboard very desirable given that it's easier to use, especially for users that are not that familiar with `kubectl`.

##### Comments
I asked for the availability of this feature some time ago on Slack and [got the response](https://kubernetes.slack.com/archives/C09R1JVU6/p1500012587455992) that it's worthwhile creating an issue. So I'm doing this now. :-)

timoreimann:
This issue still seems pretty relevant -- unstaling.

/remove-lifecycle stale

leo1515416043:
This feature is import.

shu-mutou:
Ref: #4542

huweihuang:
Is the rolling back function currently supported on the dashboard page？

shu-mutou:
No, it's not supported yet.

shu-mutou:
This is a function that kubectl implements by adding its own annotations, and it is necessary to maintain data structures that are not directly supported by the K8s API. In Dashboard, we are considering how to handle data that extends the K8s API, and the implementation has stopped.

PR #5917, opened 2021-03-23
Implements rollout APIs for deployment

shu-mutou:
This implements REST APIs for `kubectl rollout deployment` command.
Also, implements frontend for `kubectl rollout restart deployment` command.

Co-authored-by: ZeHuaiWang <874697675@qq.com>

Ref: #2889, #4542
Fixes: #4254

### DIFF
--- a/src/app/backend/handler/apihandler.go
+++ b/src/app/backend/handler/apihandler.go
@@ -272,14 +272,22 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
 			To(apiHandler.handleGetDeploymentNewReplicaSet).
 			Writes(replicaset.ReplicaSet{}))
 	apiV1Ws.Route(
-		apiV1Ws.POST("/{kind}/{namespace}/{deployment}/rollback").
-			To(apiHandler.handleDeployRollback).
+		apiV1Ws.PUT("/{kind}/{namespace}/{deployment}/pause").
+			To(apiHandler.handleDeploymentPause).
+			Writes(deployment.DeploymentDetail{}))
+	apiV1Ws.Route(
+		apiV1Ws.PUT("/{kind}/{namespace}/{deployment}/rollback").
+			To(apiHandler.handleDeploymentRollback).
 			Reads(deployment.RolloutSpec{}).
 			Writes(deployment.RolloutSpec{}))
 	apiV1Ws.Route(
-		apiV1Ws.GET("/{kind}/{namespace}/{deployment}/restart").
-			To(apiHandler.handleDeployRestart).
+		apiV1Ws.PUT("/{kind}/{namespace}/{deployment}/restart").
+			To(apiHandler.handleDeploymentRestart).
 			Writes(deployment.RolloutSpec{}))
+	apiV1Ws.Route(
+		apiV1Ws.PUT("/{kind}/{namespace}/{deployment}/resume").
+			To(apiHandler.handleDeploymentResume).
+			Writes(deployment.DeploymentDetail{}))
 
 	apiV1Ws.Route(
 		apiV1Ws.PUT("/scale/{kind}/{namespace}/{name}/").
@@ -1262,7 +1270,24 @@ func (apiHandler *APIHandler) handleDeployFromFile(request *restful.Request, res
 	})
 }
 
-func (apiHandler *APIHandler) handleDeployRollback(request *restful.Request, response *restful.Response) {
+func (apiHandler *APIHandler) handleDeploymentPause(request *restful.Request, response *restful.Response) {
+	k8sClient, err := apiHandler.cManager.Client(request)
+	if err != nil {
+		errors.HandleInternalError(response, err)
+		return
+	}
+
+	namespace := request.PathParameter("namespace")
+	name := request.PathParameter("deployment")
+	deploymentSpec, err := deployment.PauseDeployment(k8sClient, namespace, name)
+	if err != nil {
+		errors.HandleInternalError(response, err)
+		return
+	}
+	response.WriteHeaderAndEntity(http.StatusOK, deploymentSpec)
+}
+
+func (apiHandler *APIHandler) handleDeploymentRollback(request *restful.Request, response *restful.Response) {
 	k8sClient, err := apiHandler.cManager.Client(request)
 	if err != nil {
 		errors.HandleInternalError(response, err)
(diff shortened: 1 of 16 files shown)


src/app/backend/handler/apihandler.go
```diff
@@ -271,6 +271,15 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
 		apiV1Ws.GET("/deployment/{namespace}/{deployment}/newreplicaset").
 			To(apiHandler.handleGetDeploymentNewReplicaSet).
 			Writes(replicaset.ReplicaSet{}))
+	apiV1Ws.Route(
+		apiV1Ws.POST("/{kind}/{namespace}/{deployment}/rollback").
+			To(apiHandler.handleDeployRollback).
```
floreks, line 276: s/handleDeployRollback/handleDeploymentRollback

src/app/backend/handler/apihandler.go
```diff
@@ -271,6 +271,15 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
...
+			Reads(deployment.RolloutSpec{}).
+			Writes(deployment.RolloutSpec{}))
+	apiV1Ws.Route(
+		apiV1Ws.GET("/{kind}/{namespace}/{deployment}/restart").
```
floreks, line 280: Executing action on a resource should not use GET. It should be POST.

src/app/backend/resource/deployment/rollout.go
```diff
@@ -0,0 +1,137 @@
...
+	Revision string `json:"revision"`
+}
+
+// RollbackDeployment rollback to a specific ReplicaSet revision
+func RollbackDeployment(client client.Interface, rolloutSpec *RolloutSpec, namespace, name string) (*RolloutSpec, error) {
+	deployment, err := client.AppsV1().Deployments(namespace).Get(context.TODO(), name, metaV1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	currRevision := deployment.Annotations["deployment.kubernetes.io/revision"]
```
floreks, line 39: Export annotation string and `1` version string to const values. I'd put them on top.

src/app/backend/resource/deployment/rollout.go
```diff
@@ -0,0 +1,137 @@
...
+			return &RolloutSpec{
+				Revision: res.Annotations["deployment.kubernetes.io/revision"],
+			}, nil
+		}
+	}
+	return nil, errors.New("There is no ReplicaSet that has the requested revision for the Deployment.")
+}
+
+// PauseDeployment is used to pause a deployment
+func PauseDeployment(client client.Interface, namespace, name string) (*v1.Deployment, error) {
```
floreks, line 64: Is it exposed to any endpoint or used anywhere?

src/app/backend/resource/deployment/rollout.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return deployment, nil
+	}
+	return nil, errors.New("The Deployment is already paused.")
+}
+
+// ResumeDeployment is used to resume a deployment
+func ResumeDeployment(client client.Interface, namespace, name string) (*v1.Deployment, error) {
```
floreks, line 81: Same question as to the above.

src/app/backend/resource/deployment/rollout.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &RolloutSpec{
+		Revision: res.Annotations["deployment.kubernetes.io/revision"],
+	}, nil
+}
+
+// GetReplicaSetFromDeployment return all replicaSet which is belong to the deployment
+func GetReplicaSetFromDeployment(client client.Interface, namespace, name string) ([]v1.ReplicaSet, error) {
```
floreks, line 115: Shouldn't this search only for old replica sets?

floreks (changes requested):

shu-mutou:
@floreks Thanks for reviewing! I'll update this PR next week.

src/app/backend/handler/apihandler.go
shu-mutou: done

src/app/backend/handler/apihandler.go
shu-mutou: done. but change to PUT, following other actions.

src/app/backend/resource/deployment/rollout.go
shu-mutou: done

src/app/backend/resource/deployment/rollout.go
shu-mutou: Endpoints for PauseDeployment and ResumeDeployment added.

src/app/backend/resource/deployment/rollout.go
shu-mutou: ditto

src/app/backend/resource/deployment/rollout.go
shu-mutou: I am thinking of creating history, i.e. same as output of `kubectl rollout history deploy <resource name>`,  from all of related ReplicaSets, and displaying the history on the Deployment details view as a list, then putting the Rollback action on a line other than the current revision of history. So I'd like to get all of ReplicaSets here.

src/app/backend/handler/apihandler.go
floreks: PUT makes sense as we will somehow mutate the existing deployment.

src/app/frontend/common/dialogs/restartresource/dialog.ts
```diff
@@ -0,0 +1,32 @@
...
+  selector: 'kd-restart-resource-dialog',
+  templateUrl: 'template.html',
+})
+export class RestartResourceDialog {
+  constructor(
+    public dialogRef: MatDialogRef<RestartResourceDialog>,
+    @Inject(MAT_DIALOG_DATA) public data: ResourceMeta
+  ) {}
+
+  onNoClick(): void {
```
floreks, line 29: This can actually be removed aswell as `dialogRef` injection in the constructor as they are not really used in the template. I am not sure why do we have it in our dialogs when it is not used. Probably missed during initial review and copy-pasted.

floreks:
This is a very useful functionality. Just a last small comment and we can merge.

src/app/frontend/common/dialogs/restartresource/dialog.ts
shu-mutou: done

PR #5967, opened 2021-04-07
Move check and test script to pre-push git hook

damingerdai:
`check` and `test` scripts should be in `pre-push` git hook.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -11,8 +11,10 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
-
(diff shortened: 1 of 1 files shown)


floreks:
Any particular reason tests should be executed before push instead of commit? Seems like a personal preference to me.

damingerdai:
> Any particular reason why tests should be executed before push instead of commit? Seems like a personal preference to me.

I admit that there is a personal preference. but i think pre-push is better than pre-commit. It will takes a lot of time in run `check` and `test` script when we commit code.Usually, there will be multiple commits in a pr. So I suggest that we should use pre-push, this is just to submit the code to execute the check and test commands.

shu-mutou:
/lgtm
Agree. When rebasing own PR to master to resolve conflicts, checks and tests run for each commit and it takes lots of time.

damingerdai:
@shu-mutou thanks for your support

Issue #3941, opened 2019-06-18
Don't include the default in the help text, this is already printed by the help system.


jeefy:
##### What should be cleaned up or changed
Go through all the flags and ensure defaults aren't part of the canned help-text.

##### Provide any links for context
https://github.com/kubernetes/dashboard/pull/3504/files#r277414128

/assign

PR #5993, opened 2021-04-15
Update flag description

maciaszczykm:
Result:

```
Usage of ./dist/amd64/dashboard:
      --alsologtostderr                  log to standard error as well as files
      --api-log-level string             level of API request logging, should be one of 'NONE', 'INFO' or 'DEBUG' (default "INFO")
      --apiserver-host string            address of the Kubernetes API server to connect to in the format of protocol://address:port, leave it empty if the binary runs inside cluster for local discovery attempt
      --authentication-mode strings      enabled authentication options, supports 'token' and 'basic' that should only be used if Kubernetes API server has --authorization-mode=ABAC and --basic-auth-file flags set (default [token])
      --auto-generate-certificates       enables automatic certificates generation used to serve HTTPS
      --bind-address ip                  IP address on which to serve the --port, set to 0.0.0.0 for all interfaces (default 0.0.0.0)
      --default-cert-dir string          directory path containing files from --tls-cert-file and --tls-key-file, used also when auto-generating certificates flag is set (default "/certs")
      --disable-settings-authorizer      disables settings page user authorizer so anyone can access settings page
      --enable-insecure-login            enables login view when the app is not served over HTTPS
      --enable-skip-login                enables skip button on the login page
      --heapster-host string             address of the Heapster API server to connect to in the format of protocol://address:port, leave it empty if the binary runs inside cluster for service proxy usage
      --insecure-bind-address ip         IP address on which to serve the --insecure-port, set to 127.0.0.1 for all interfaces (default 127.0.0.1)
      --insecure-port int                port to listen to for incoming HTTP requests (default 9090)
      --kubeconfig string                path to kubeconfig file with authorization and master location information
      --locale-config string             path to file containing the locale configuration (default "./locale_conf.json")
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files
      --metric-client-check-period int   time interval between separate metric client health checks in seconds (default 30)
      --metrics-provider string          select provider type for metrics, 'none' will not check metrics (default "sidecar")
      --namespace string                 if non-default namespace is used encryption key will be created in the specified namespace (default "kube-system")
      --port int                         secure port to listen to for incoming HTTPS requests (default 8443)
      --sidecar-host string              address of the Sidecar API server to connect to in the format of protocol://address:port, leave it empty if the binary runs inside cluster for service proxy usage
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
      --system-banner string             system banner message displayed in the app if non-empty, it accepts simple HTML
      --system-banner-severity string    severity of system banner, should be one of 'INFO', 'WARNING' or 'ERROR' (default "INFO")
      --tls-cert-file string             file containing the default x509 certificate for HTTPS
      --tls-key-file string              file containing the default x509 private key matching --tls-cert-file
      --token-ttl int                    expiration time in seconds of JWE tokens generated by dashboard, set to 0 to avoid expiration (default 900)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
```

Closes https://github.com/kubernetes/dashboard/issues/3941.


maciaszczykm:
Summary:
- tooltips should start with small case (as the default ones)
- there is not dot at the end
- the defaults are printed by the system, so there is no need for us to write them
- some of the messages were updated

/assign @floreks @shu-mutou 

Please take a look.

PR #6003, opened 2021-04-20
Implements frontend for deployment rollback

shu-mutou:
History list is added into deployment detail view, and `Rollback` action is added in each row except for current revision.

Fixes: #2889

### DIFF
--- a/src/app/backend/resource/deployment/rollout.go
+++ b/src/app/backend/resource/deployment/rollout.go
@@ -20,8 +20,14 @@ import (
 	"time"
 
 	v1 "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	client "k8s.io/client-go/kubernetes"
+
+	"github.com/kubernetes/dashboard/src/app/backend/api"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
+	"github.com/kubernetes/dashboard/src/app/backend/resource/replicaset"
 )
 
 const (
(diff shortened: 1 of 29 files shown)


shu-mutou:
/hold
This would conflict with #6005. Wait for merging it.

src/app/backend/resource/deployment/rollout.go
```diff
@@ -39,17 +45,45 @@ type RolloutSpec struct {
...
+
+	// Unordered list of Replica Sets.
+	HistoryList []History `json:"history"`
+
+	// List of non-critical errors, that occurred during resource retrieval.
+	Errors []error `json:"errors"`
+}
+
+// History is a presentation layer view of Kubernetes Replica Set resource for history.
+type History struct {
```
floreks, line 63: This kind of breaks our pattern of resource being a Kubernetes resource. It is then used in the frontend as a resource also. I'd have to rethink how to incorporate that into our patterns.

shu-mutou: Yes, these are defined in frontend also.
To create dialog contents and submission endpoint URL, the action on history row needed to pass deployment name and namespace to dialog. So I defined new type History that have ReplicaSet and parent resource information. Could you let me know if you have better idea about this?

floreks: I'll think about how we could add this without breaking our patterns. Resource should be mappable to a real Kubernetes resource. History/HistoryList is a very abstract thing.

floreks:
I was on vacation last week. I'll try to take a look at it this week.

cell4rom:
Hello, why You don't using Redux, it's very important in a project like kubernetes UI.

floreks:
It's not important. It's a matter of personal preference. Also, PR is not a good place to discuss such things.

shu-mutou:
@maciaszczykm @floreks PTAL again. Thanks!!

floreks:

/close

PR #6007, opened 2021-04-21
Pods list show images

marcosdiez:
This is 99% copy paste from the Deployment list, which also shows container images

image

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -7,7 +7,7 @@
         <target>Ressource bearbeiten</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/frontend/common/dialogs/editresource/dialog.ts</context>
-          <context context-type="linenumber">56</context>
+          <context context-type="linenumber">58</context>
         </context-group>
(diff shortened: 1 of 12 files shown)


helight (approved):
LGTM

maciaszczykm:
It looks a bit cluttered right now, we cannot show all the information on the list view. Why images are so important to be shown there too?

floreks:
I agree with @maciaszczykm. With `All namespaces` selected it gets even worse as there are already 9 columns without images. I am also not sure if the image is that important. We could maybe replace node with image, but they both seem equally important to me.

mdiez-modus:
Looking at the pod container image is the ultimate way to make sure all my pods have the expected image.
Sometimes deployments are failing and the previous image is still there, sometimes there is a mix of both and so on.

I know I am using more screen space, but in the end of the day it's not that much since it's usually _show al_" that is shown, which uses much less space than the container name:

image

floreks:
I'd close this as I have done some more changes in the linked PR that will save us some space and align all resource lists, not only the Pods.

PR #6038, opened 2021-04-28
Features/oauth2 proxy

pari-p19:
Helm Chart for Oauth2 proxy


desaintmartin:
Thanks for the PR, but I do not believe this is the right place to add oauth2-proxy deployment.

Note that I use it myself and it may still lack documentation. Some common work between https://github.com/oauth2-proxy/manifests and here may be done in order to have a streamlined experience / doc.

/close

PR #6078, opened 2021-05-18
GetCustomResourceDefinitionDetail function can't queries more than 10

twilight327426371:
When the GetCustomResourceDefinitionDetail function queries more than 10, will not be queried.


twilight327426371:
@maciaszczykm

maciaszczykm:
This is used only for loading initial data. There is specific endpoint to get paginated responses that has it implemented. Just check if it works.

PR #6119, opened 2021-05-26
Update node version & fix CI

### LINKED ISSUE #6117: Broken CI and unit tests
Unit tests crash is related to: https://github.com/kubernetes/dashboard/pull/6014

CI does not properly report failures. Even though a bunch of steps returns exit code 1 it still shows up as passed. We have to investigate and fix both CI and the unit test configuration.

floreks:
Fixes #6117


floreks:
There is still an issue with a couple of tests reporting a very strange error that I can't really debug. If anyone wants to take a shot at it then feel free.

```bash
  ● Test suite failed to run

    SyntaxError: The requested module '@api/root.ui' does not provide an export named 'IConfig'

      at Runtime.linkAndEvaluateModule (../../../node_modules/jest-runtime/build/index.js:669:5)

 FAIL  src/app/frontend/resource/discovery/service/detail/component.spec.ts
```

floreks:
Reconfigured Jest to use `commonJS` and `es2015` as build result since it does not fully support es modules just yet.

PR #6143, opened 2021-06-08
Use valid YAML code for snippets

Backhage:
Since it is suggested that the code snippets for ServiceAccount and ClusterRoleBinding are copied to manifest files and applied using kubectl apply -f <filename.yaml> the code should be valid YAML.

### DIFF
--- a/docs/user/access-control/creating-sample-user.md
+++ b/docs/user/access-control/creating-sample-user.md
@@ -10,14 +10,12 @@ For each of the following snippets for `ServiceAccount` and `ClusterRoleBinding`
 
 We are creating Service Account with name `admin-user` in namespace `kubernetes-dashboard` first.
 
-```shell
-cat <<EOF | kubectl apply -f -
+```yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: admin-user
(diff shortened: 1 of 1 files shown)


Backhage:
Individual Contributor License Agreement (CLA) has been signed.

maciaszczykm:
I don't get it. The YAMLs are valid. The current method is quicker because you can copy and paste it to your terminal and then it's done.

Backhage:
> 
> 
> I don't get it. The YAMLs are valid. The current method is quicker because you can copy and paste it to your terminal and then it's done.
Having a leading _cat <<EOF | kubectl apply -f -_ and _EOF_ in the end is not valid YAML. If you copy/paste this into a yaml file and try to execute this with kubectl apply -f <file_name.yaml> you will get a cryptic error message: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context.
The guide clearly states that the code snippet should be copied into a yaml file, so then it is important that the code snippet is valid as is.
Also, for me being on Windows using Powershell, it does not work to copy and paste the current snippets into the terminal.

floreks (approved):

PR #6144, opened 2021-06-08
fix how we display nfs mounts

marcosdiez:
NFS mounts should be displayed as SERVER:PATH

something like `my.nfs.server.com:/share` We are currently displaying it as `my.nfs.server.com//share` which is awful :)

image


floreks (approved):
/lgtm

PR #6145, opened 2021-06-09
Pod HealthCheck

marcosdiez:
Floresk, please feel free to change the layout as you prefer!

Also, I need CSS help removing whitespaces on the `HTTP Healthcheck URI`

ps: an excellent container to test this is kube-system/coredns

image

image

### DIFF
--- a/src/app/frontend/common/components/probe/component.ts
+++ b/src/app/frontend/common/components/probe/component.ts
@@ -1,0 +1,29 @@
+// Copyright 2017 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


floreks:
To be honest I wouldn't want to add so many details to the pod info. There is already a lot of information there and the liveness probe can be easily checked by simply clicking edit and looking at yaml.

marcosdiez:
I understand, but if some users can't read YAML ...
Separating the container from the POD is also not a good idea because they have too much info in common.

What if we get a compromise. Like the healthcheck comes "closed" or "minimized" by default and one needs to click on it to open it ?

Also, to make life worse, sooner or later I might provide you a patch to see the TCP ports of the containers.

My users need that :/

floreks:
We have to rethink how to better manage information display as the pod page tends to get quite long already. We can merge for now but soon we'll have to improve it.

/lgtm
/approve

src/app/frontend/common/components/probe/component.ts
```diff
@@ -22,4 +22,23 @@ import {Probe} from '@api/root.api';
 export class ProbeComponent {
   @Input() initialized: boolean;
   @Input() probe: Probe;
+
```
marcosdiez, line 25: Very smart from your side! And I was fighting for hours with CSS. Thank you once more!

PR #6163, opened 2021-06-11
metricsScraper use logrus instead of klog

pierluigilenoci:
Ref: https://github.com/sirupsen/logrus

https://github.com/kubernetes-sigs/dashboard-metrics-scraper/blob/master/server.go#L9


pierluigilenoci:
@desaintmartin I didn't do a chart bump because I only changed a comment inside a YAML file, there is no actual difference in functionality. It seemed to me that it didn't make sense. But now the test fails. 😢 

How do we solve?

floreks:
From what I see metrics scraper uses glog
https://github.com/kubernetes/dashboard/blob/f06f78d395e0ff331afdda6f895144ae4327133f/go.mod#L8

and glog uses a `--v=N` where N is a number related to the log level.

floreks:
Any change in any chart file requires a version bump of the chart. It is a strict requirement.

pierluigilenoci:
@floreks bump done.

floreks:
Could you please change the last commit message to simple `helm chart version bump`? :slightly_smiling_face:

pierluigilenoci:
@floreks done 😆

pierluigilenoci:
@desaintmartin @maciaszczykm could you please `lgtm` me?

floreks (approved):
/lgtm

PR #6166, opened 2021-06-14
Remove `highlight.js`

shu-mutou:
`highlight.js` is not in use.

Ref: #6136

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -82,7 +82,6 @@
               ]
             },
             "styles": [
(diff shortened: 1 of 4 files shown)


shu-mutou:
/hold
Seems something wrong with installation of `@angular-devkit/build-angular`. Please wait for a while.

floreks:
There is also a related entry in the `angular.json`. It has to be removed also.

shu-mutou:
PTAL again. Thanks!

Issue #6089, opened 2021-05-20
Migrate from @import to @use


floreks:
Related to angular 12 migration and newly added sass features: https://css-tricks.com/introducing-sass-modules/

PR #6181, opened 2021-06-17
Migrate from "import" to "use", avoid division in stylesheets and switch from sass-lint to stylelint

maciaszczykm:
Closes https://github.com/kubernetes/dashboard/issues/6089.
Closes https://github.com/kubernetes/dashboard/issues/6204.


go.mod
```diff
@@ -18,5 +18,6 @@ require (
 	k8s.io/apiextensions-apiserver v0.21.0
 	k8s.io/apimachinery v0.21.0
 	k8s.io/client-go v0.21.0
+	k8s.io/code-generator v0.21.0 // indirect
```
maciaszczykm, line 21: Auto generated by `npm ci`.

package.json
```diff
@@ -189,7 +189,7 @@
   },
   "engines": {
     "node": ">=14.16.1",
-    "npm": ">=6.0.0 <7.0.0"
+    "npm": ">=6.0.0"
```
maciaszczykm, line 192:
```➜ npm --version
7.7.6

 ➜ node --version              
v15.14.0
```

Works for me.

shu-mutou: F.Y.I. Same errors of static check occur in env using `./aio/develop/run-npm-on-container.sh`. In this env, versions are as follows:
node: 14.16.1
npm: 6.14.12

But e2e test and dashboard works properly.

shu-mutou: I could not find the way to fix it today. Have a nice weekend.

maciaszczykm: You too, I'll try to investigate next week too.

maciaszczykm: I've reverted it for now. Let's focus on LTS only.

aio/gulp/backend.js
```diff
@@ -30,8 +30,6 @@ gulp.task('backend', gulp.series((doneFn) => {
   goCommand(
       [
         'build',
-        // Install dependencies to speed up subsequent compilations.
-        '-i',
```
floreks, line 34: Any reason to remove this?

shu-mutou:
LGTM except for floreks questions. thanks!

aio/gulp/backend.js
maciaszczykm: Yes, there is warning in the build logs that this flag is deprecated. From https://golang.org/cmd/go/: "The -i flag installs the packages that are dependencies of the target. The -i flag is deprecated. Compiled packages are cached automatically.".

PR #6220, opened 2021-07-01
Add events list to the global sidebar

### LINKED ISSUE #6186: Implement "kubectl get events" in the side panel
During fires the first thing I need to know is "what's going on".

One way to get that is to type

`kubectl get  events`
or even better
`kubectl get -A  events -w`

The dashboard currently does not have that.

If I implement it, would you guys approve it/accept it ?

It would be a new item in the side panel

marcosdiez:
closes #6186

some design decisions, which we can discuss:

a) events are NOT bound to a namespace, for we already have a UI to see namespace specific events
b) events are NOT shown in neither search nor when one clicks in cluster, because that would pollute more the UI than it would give benefits (since events are not normal/common/tangible k8s objects)
c) we now have a `reason` column
d) I replaced the `SubObject` column with `Object`, it's value is ObjectKind/ObjectName ( example: deployment/kuberntes-dashboard ), we can now click on it!
e) events are now sorted by the `lastSeen` field (newer events show first)

A container with that code is available here: `marcosdiez/dashboard:v2.3.1f`

image

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -858,6 +858,10 @@
           <context context-type="linenumber">70</context>
         </context-group>
         <context-group purpose="location">
+          <context context-type="sourcefile">src/app/frontend/common/components/resourcelist/event/template.html</context>
+          <context context-type="linenumber">50</context>
+        </context-group>
+        <context-group purpose="location">
           <context context-type="sourcefile">src/app/frontend/common/components/resourcelist/horizontalpodautoscaler/template.html</context>
           <context context-type="linenumber">46</context>
         </context-group>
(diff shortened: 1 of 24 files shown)


floreks:
Surely event list component will need some adjustments to work the same as other high-level resource lists. Mostly LGTM, I'll do some minor changes.

floreks:
- Added support for sorting by `reason`, `firstSeen`, `lastSeen` columns to events
- Added dynamic columns `object` and `sub-object`. `object` will be displayed on global event resource list and `sub-object` will be displayed on a resource details views
- Updated `object` link logic to try and not create links for unsupported resources
- Events now respect selected namespace. It works the same as with `kubectl`
- Used resource channels to match our backend logic for list/details resource
- Fixed an issue with `default` namespace being recognized as `All namespaces` in some cases

floreks:
Closes #6186

PR #6236, opened 2021-07-07
Properly decode base64 encoded UTF-8 strings

### LINKED ISSUE #6233: secret decode - Using JavaScript's atob decoder Base64 can't decode UTF-8 strings correctly
```
Dashboard version: v2.3.1+0.g8d9f8e76c
```

##### Steps to reproduce

test-secret.yaml
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test
stringData:
  username: "哈哈"
  password: "123456"
```

##### Observed result

username incorrect

image-20210707101342498

##### Expected result

username: 哈哈

##### Comments

ref: https://github.com/kubernetes/dashboard/blob/a4deef48f50ff9eb975fe574c9f776b22632fbd0/src/app/frontend/resource/config/secret/detail/component.ts#L72

floreks:
Fixes #6233

### DIFF
--- a/src/app/frontend/common/components/container/component.ts
+++ b/src/app/frontend/common/components/container/component.ts
@@ -15,6 +15,7 @@
 import {Component, Input, OnChanges} from '@angular/core';
 import {ConfigMapKeyRef, Container, EnvVar, SecretKeyRef} from '@api/root.api';
 import {Status, StatusClass} from '@common/components/resourcelist/statuses';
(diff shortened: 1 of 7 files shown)


DuC-cnZj:
encode also has the same problem
@floreks 

image


ref: https://github.com/kubernetes/dashboard/blob/b783f69462eadc6aa98c2d16c6d111523111ea6d/src/app/frontend/resource/config/secret/detail/edit/component.ts#L91

PR #6237, opened 2021-07-07
Resource owner now shown in every pane

marcosdiez:
image

I am not sure if this is useful anywhere but in a ReplicaSet. Also, we have the "problem" of having the info shown twice in Pods.

### DIFF
--- a/src/app/frontend/common/components/objectmeta/template.html
+++ b/src/app/frontend/common/components/objectmeta/template.html
@@ -71,6 +71,15 @@ limitations under the License.
              i18n>UID</div>
         <div value>{{objectMeta?.uid}}</div>
       </kd-property>
+      <kd-property *ngIf="objectMeta?.ownerReferences"
+                   [ngClass]="'object-meta-ownerReferences'">
+        <div key
+             i18n>Owner</div>
+        <div value *ngFor="let ownerReference of objectMeta?.ownerReferences">
+          <a [routerLink]="getObjectHref(ownerReference.kind, ownerReference.name)"
+          queryParamsHandling="preserve">{{ownerReference.kind}}/{{ownerReference.name}}</a>
+        </div>
+      </kd-property>
       <kd-property *ngIf="objectMeta?.labels"
                    fxFlex="100"
                    [ngClass]="'object-meta-labels'">
--- a/src/app/frontend/common/components/objectmeta/template.html
+++ b/src/app/frontend/common/components/objectmeta/template.html
@@ -75,9 +75,10 @@ limitations under the License.
                    [ngClass]="'object-meta-ownerReferences'">
         <div key
              i18n>Owner</div>
-        <div value *ngFor="let ownerReference of objectMeta?.ownerReferences">
+        <div value
+             *ngFor="let ownerReference of objectMeta?.ownerReferences">
           <a [routerLink]="getObjectHref(ownerReference.kind, ownerReference.name)"
-          queryParamsHandling="preserve">{{ownerReference.kind}}/{{ownerReference.name}}</a>
+             queryParamsHandling="preserve">{{ownerReference.kind}}/{{ownerReference.name}}</a>
         </div>
       </kd-property>
       <kd-property *ngIf="objectMeta?.labels"
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -1074,6 +1074,14 @@
           <context context-type="linenumber">71</context>
         </context-group>
       </trans-unit>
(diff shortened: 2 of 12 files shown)


src/app/frontend/common/components/objectmeta/template.html
```diff
@@ -71,6 +71,16 @@
              i18n>UID</div>
         <div value>{{objectMeta?.uid}}</div>
       </kd-property>
+      <kd-property *ngIf="objectMeta?.ownerReferences"
+                   [ngClass]="'object-meta-ownerReferences'">
+        <div key
+             i18n>Owner</div>
+        <div value
+             *ngFor="let ownerReference of objectMeta?.ownerReferences">
```
maciaszczykm, line 79: Handling for multiple owner refs is implemented here but I wonder how the view will look like if there will be more than one owner. Can you paste some screenshot?

floreks: I'd maybe rely on the `controlled by` card component that we use i.e. on pod details and try to reuse it for other resources.

marcosdiez: Hi @maciaszczykm,

From my understanding, one can't automatically get two owners, but one can manually set them. This is how you get two of them.

I did it out of curiosity and I got this:
image

But I think @floreks is right. It's better to ditch this and just put the `controlled by` card everywhere.

marcosdiez: I am thinking again. Do we really need the controller card ? Currently it's just used on the pods page, which has already too much content. Maybe removing it would be better. Just thinking outloud.

floreks: I'd still keep it but focus on reorganizing information on the detail pages overall. Use tabs, maybe use some smaller cards that we could put in a single row instead of always using 1 card per row.

marcosdiez: Just for the record, I've been used my forked version of the dashboard that contains this code ( marcosdiez/dashboard:v2.3.1j ) and I personally thing it's quite good as is.

Some objects are created by replica sets, some are created by daemon sets, some by cron jobs, some by stateful sets and this tiny PR covers it all. If I need more info I just go and click (or middle click) on the `Owner` label. 

Maybe we should consider merging it as is.

marcosdiez: @floreks ^^^^

floreks: Having this kind of generic resource owner information and creating internal links based on name/kind is not a good idea since we do not support every potential resource that might be created this way and it will result in dead/not working links inside the Dashboard.

We should support this explicitly on a per resource basis.

marcosdiez: @floreks you are right. I just added the code. I took the liberty of copying what you did the events file.

marcosdiez:
@floreks @maciaszczykm Is there anything I can do for this code to be merged ?

floreks:
My original concerns still exist. I'd prefer to have a separate `controlled by` card that would be added on a per resource basis since there are resources that do not have owner.

Also, we should not display multiple levels of ownership, i.e. pod is controlled by the replica set and replica set is controlled by the deployment. I wouldn't want to see both replica set and deployment (owner of an owner) on the pod details.

We also should not display raw owner information `<kind>/<name>` (raw string) without any additional information if we do not support such resource in the Dashboard.

marcosdiez:
@floreks I am not sure I understand your concerns:

- if a resource does not have an owner, the `Owner` field in `Metadata` is not shown at all.
- we do not display multiple levels of ownership unless the metadata contains that. The only object I've ever seen with two owners is the one I manually modified the YAML to see how the UI would behave so I could answer @maciaszczykm 's question. (which is something legal but usually unnecessary per https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/ . So unless you manually modify your `pod`, you won't see the `Deployment`, only the replicaset.
- I personally prefer showing raw owner information `<kind>/<name>` instead of omitting information. More information is better then omitting information. It's also a nice hint that I should implement another pane :)

Think about it. Even though this is not your dream, the code works, it's helpful, does not consume many resources nor UI space. The kubernetes dashboard is better with it.

marcosdiez:
Hello @floreks !

I am hoping you changed your mind and now thinks this PR is good to be merged. 
An extra card would have to be created per resource and is not as future proof.

Please consider merging this PR as is!

Marcos

floreks:

/close

PR #6278, opened 2021-07-22
Ingress Detail: shows default backend, IngressClass + a small fix

marcosdiez:
In the Ingress detail pane we:
- now show  the IngressClass 
- default backend in bottom the rule list (more intuitive and we were not showing previously because of a bug anyway)
- ingress backend resources are now showing (bugfix)

image

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -2164,7 +2164,7 @@
         </context-group>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/frontend/resource/discovery/ingress/detail/template.html</context>
-          <context context-type="linenumber">51</context>
+          <context context-type="linenumber">36</context>
         </context-group>
       </trans-unit>
       <trans-unit id="7143579180750436311" datatype="html">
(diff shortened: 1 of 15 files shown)


marcosdiez:
easy way to test it:

------------------------
kind: Ingress
metadata:
  name: echoxx
  namespace: echons
  annotations:
    kubernetes.io/ingress.class: kong
spec:
  ingressClassName: someingressclass

  defaultBackend:
    service:
      name: httpbin
      port:
        number: 80
  rules:

  - host: somedomain.com
    http:
      paths:
        - path: /
          pathType: Prefix
          backend:
            service:
              name: someservice
              port:
                number: 90
  - host: otherdomain.com
    http:
      paths:
        - path: /
          pathType: Prefix
          backend:
            service:
              name: otherservice
              port:
                number: 90
  - http:
      paths:
      - backend:
          service:
            name: echo
            port:
              number: 8080
        path: /foo
        pathType: Prefix
  - http:
      paths:
        - path: /icons
          pathType: ImplementationSpecific
          backend:
            resource:
              apiGroup: k8s.example.com
              kind: StorageBucket
              name: icon-assets

maciaszczykm:
Verified.

/lgtm

PR #6284, opened 2021-07-26
[WIP] Dashboard can now list files inside of containers

### LINKED ISSUE #6283: The dashboard should have a way to browse files inside of a pod
Again, I volunteer to develop it, but before I want to discuss acceptance and implementation details.

The idea is that in pod details panel, on the right side of the toolbar, we add a new icon called "file manager".

It should show a "windows explorer" like view of all files within a specific container, allowing one to browse/view/download/edit? files inside of a contaier

Something very similar to https://marketplace.visualstudio.com/items?itemName=sandipchitale.kubernetes-file-system-explorer

Initially, it  could work like the above extension, grabbing the data using the go api with commands equivalentes to 

`kubectl exec -it pod-name -c containername -- ls -f /`

I have ideas to make it work in containers that don't have `ls`, like the kubernetes dashboard, but that's a discussion for a second topic.

marcosdiez:
closes #6283

For now, it only lists paths, and only backend. No file contents yet. Also, very naive method, no checks for errors.
But this is enough to if somebody else wants to do the frontend while I tune the backend. It's only a POC

Example: 
http://192.168.58.176:8080/api/v1/pod/default/ubuntudebug2-6f95594698-bs4vk/ls/ubuntudebug2?path=/

image

### DIFF
--- a/src/app/backend/handler/apihandler.go
+++ b/src/app/backend/handler/apihandler.go
@@ -245,7 +245,7 @@ func CreateHTTPAPIHandler(iManager integration.IntegrationManager, cManager clie
 			To(apiHandler.handleExecShell).
 			Writes(TerminalResponse{}))
 	apiV1Ws.Route(
-		apiV1Ws.GET("/pod/{namespace}/{pod}/ls/{container}/{path}").
+		apiV1Ws.GET("/pod/{namespace}/{pod}/ls/{container}").
 			To(apiHandler.handleExecLs).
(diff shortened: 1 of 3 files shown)


floreks:
This is what exec into container feature is for and more. I don't think we need anything like that.

maciaszczykm:
I would be supporting that idea if the method to get the files wouldn't be a workaround. You basically use exec, do the `ls` and then parse the output. For me it's too hacky to use this.

PR #6342, opened 2021-08-11
Bump k8s modules to 0.22.0 and golang to 1.16.7

shu-mutou:
Also, bump golangci-lint to 1.41.1.


maciaszczykm:
Looks good, I've one remark.

PR #6344, opened 2021-08-11
support pod list filter by hostIp and podIp property

twilight327426371:
Instructions：
filterBy=hostIp,9.xxx.xxx.xxx
filterBy=podIp,9.xxx.xxx.xx

Query a podIp container or host container on hundreds of node daemonsets, the same situation applies to deployment and statefulset


floreks:
Unless you can also prepare the UI part, I don't see how we'd benefit from that.

twilight327426371:
> Unless you can also prepare the UI part, I don't see how we'd benefit from that.

I modified the front-end page and took a screenshot

PR #6367, opened 2021-08-19
Add webpack and bump jest-preset-angular

shu-mutou:
jest-preset-angular removed webpack from its dependency in v9.0.5.
So we need to add webpack into own dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4871,126 +4871,31 @@
       }
     },
     "@jest/environment": {
-      "version": "27.0.3",
-      "resolved": "https://registry.npmjs.org/@jest/environment/-/environment-27.0.3.tgz",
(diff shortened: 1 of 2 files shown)


floreks:
I don't think we need to import webpack directly as angular already does and we do not use it directly anyway.

shu-mutou:
Sure, `@angular-devkit/build-angular` installs `webpack`, but it's underneath `node_modules/@angular-devkit/build-angular/node_modules/` folder. So we need to install `webpack` to keep to use `jest-preset-angular`.

floreks:
Since `jest-preset-angular` dropped webpack as a dependency then it makes no sense that we'd have to install it in order to use it. You should always only install dependencies that are directly used in the project and all peer dependencies should be installed by those direct dependencies silently.

The issue here is that most probably angular versions were bumped in an incorrect way and `package-lock.json` was not properly regenerated. I have bumped only the `jest-preset-angular` and regenerated whole `package-lock.json` and `webpack` is not required. The application builds just fine.

floreks:
Closing in favor of #6375

/close

Issue #5722, opened 2020-12-14
seccomp annotation missing in recommended.yaml manifest


tsde:
##### What should be cleaned up or changed
Hi, it seems that the following annotation 
```
annotations:
   seccomp.security.alpha.kubernetes.io/pod: 'runtime/default'
```
is missing in `recommended.yaml` manifest for the `kubernetes-dashboard` Deployment while it is correctly set in `recommended/06_dashboard-deployment.yaml`. The `dashboard-metrics-scraper` Deployment annotation is correctly set in both `recommended.yaml` and `recommended/08_scraper-deployment.yaml`.

I suppose the annotation should be set in `recommended.yaml` for the `kubernetes-dashboard` Deployment.

##### Provide any links for context
https://github.com/kubernetes/dashboard/blob/v2.1.0/aio/deploy/recommended/06_dashboard-deployment.yaml
https://github.com/kubernetes/dashboard/blob/v2.1.0/aio/deploy/recommended/08_scraper-deployment.yaml
https://github.com/kubernetes/dashboard/blob/v2.1.0/aio/deploy/recommended.yaml

floreks:
Based on docs we can actually move it from annotation to `securityContext`
> As of Kubernetes v1.19, you can use the seccompProfile field in the securityContext of Pods or containers to control use of seccomp profiles. In prior versions, seccomp was controlled by adding annotations to a Pod. The same PodSecurityPolicies can be used with either version to enforce how these fields or annotations are applied.

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccomp

PR #6382, opened 2021-08-23
Swich PodSecurityContext from annotation to securityContext in deploy

shu-mutou:
closes: #5722

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/values.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/values.yaml
@@ -60,9 +60,11 @@ extraVolumeMounts: []
 #   name: dashboard-kubeconfig
 #   readOnly: true
 
-# Annotations to be added to kubernetes dashboard pods
-podAnnotations:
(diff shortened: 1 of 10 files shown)


aio/deploy/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -60,9 +60,11 @@ extraVolumeMounts: []
 #   name: dashboard-kubeconfig
 #   readOnly: true
 
-# Annotations to be added to kubernetes dashboard pods
-podAnnotations:
-  seccomp.security.alpha.kubernetes.io/pod: "runtime/default"
+# SecurityContext to be added to kubernetes dashboard pods
```
floreks, line 63: Helm chart templates have to be also updated. `podAnnotations` can be commented and I'd only remove the single line with seccomp annotation.

shu-mutou: Done. Thanks!
`securityContext` for pod was existed in template already. So fixed param name to it.

PR #6545, opened 2021-10-16
Trim space around the token

### LINKED ISSUE #6543: Token field login space sanitization issue
##### Environment

```
Installation method: Deployed via yaml
Kubernetes version: 1.22.2
Dashboard version (v1.x is not supported. Use v2.0 or later.): Dashboard v2.3.1+0.g8d9f8e76c
Operating system: Ubuntu 20.04.3
Node.js version ('node --version' output):
Go version ('go version' output):
```

##### Steps to reproduce
Proceed to Login page. 
Select authentication via token
Type a space character, then any sort of characters after. (Ex. " asdfa", " apples", " 1232") 
Hit enter to Submit

##### Observed result
User is logged in as system:anonymous user and granted access to the GUI.

##### Expected result
Login should be rejected.

##### Comments
Feel free to reach out, or rename if the issue title is not appropriate. 
Thanks, Sean.

tcnt:
Fixes #6543 
A token is base 64 encoded and should not contains space at start or end.
By trimming spaces the resulting token will be passed to login process
If a token is invalid then a proper message will be issued
  Unauthorized (401): Invalid credentials provided

The current kubernetes login process seems to accept a token starting with
a space and then logs on as an anonymous user.

### DIFF
--- a/src/app/frontend/login/component.ts
+++ b/src/app/frontend/login/component.ts
@@ -133,7 +133,7 @@ export class LoginComponent implements OnInit {
         this.onFileLoad_(event as KdFile);
         break;
       case LoginModes.Token:
-        this.token_ = (event.target as HTMLInputElement).value;
+        this.token_ = (event.target as HTMLInputElement).value.trim();
         break;
       case LoginModes.Basic:
         if ((event.target as HTMLInputElement).id === 'username') {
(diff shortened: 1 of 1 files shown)


tcnt:
At least we should not pass a token containing leading or trailing spaces no matter how the API server handles it.
Since the token is base64 it should not contain spaces.
So this change makes sure a token in base64 format is passed.
If a token is 'empty' then login process will not start (no call to API server and thus block anonymous logon).
Also if we pass the token without leading space, then it will be blocked by the API server if the token is not valid.
So also no anonymous login will take place.

PR #6595, opened 2021-10-29
Update method ToEvent to compatible with new event api

zzzkl:
Fix similar bug like   kubernetes/kubernetes#94019 in 'kubectl printer'.

Because the fields FirstTimestamp&LastTimestamp aren't used in new events api.

Like: 
Fixes for the internal printer are already merged in master:
- lastTimestamp: https://github.com/kubernetes/kubernetes/pull/86557
- firstTimestamp: https://github.com/kubernetes/kubernetes/pull/89999

`firstTimestamp` is present in 1.19 but not to 1.18
`lastTimestamp` is present in both 1.19 and 1.18

`firstTimestamp` 1.18 backport: https://github.com/kubernetes/kubernetes/pull/94252

_Originally posted by @ingvagabund in https://github.com/kubernetes/kubernetes/issues/94019#issuecomment-680838311_
image
image

### DIFF
--- a/src/app/backend/resource/event/common.go
+++ b/src/app/backend/resource/event/common.go
@@ -16,11 +16,8 @@ package event
 
 import (
 	"context"
+	"time"
 
-	"github.com/kubernetes/dashboard/src/app/backend/api"
-	"github.com/kubernetes/dashboard/src/app/backend/errors"
-	"github.com/kubernetes/dashboard/src/app/backend/resource/common"
-	"github.com/kubernetes/dashboard/src/app/backend/resource/dataselect"
 	v1 "k8s.io/api/core/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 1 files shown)


src/app/backend/resource/event/common.go
```diff
@@ -169,6 +171,18 @@ func FillEventsType(events []v1.Event) []v1.Event {
...
+	if !event.EventTime.IsZero() {
+		eventTime := metaV1.NewTime(time.Unix(event.EventTime.Unix(), 0))
+		if event.FirstTimestamp.IsZero() {
+			firstTimestamp = eventTime
+		}
+		if event.LastTimestamp.IsZero() {
+			lastTimestamp = firstTimestamp
+		}
+	}
+
```
floreks, line 185: I'd maybe simplify this code to
```go
	firstTimestamp, lastTimestamp := event.FirstTimestamp, event.LastTimestamp
	eventTime := metaV1.NewTime(event.EventTime.Time)

	if firstTimestamp.IsZero() {
		firstTimestamp = eventTime
	}

	if lastTimestamp.IsZero() {
		lastTimestamp = eventTime
	}
```

If it's already zero then overriding with zero does not matter for us. I know it adds a couple of extra cycles, but it's very minor and the code is simpler.

zzzkl: Yes, this makes the code more readable. I have updated.

floreks:
Thanks for this change! :tada: 

/lgtm

PR #6351, opened 2021-08-12
Migrate Gulp scripts to Makefile, add latest tag for releases and update dependencies

maciaszczykm:
At the beginning I would like to apologize for the size of the it but once I started removing Gulp tasks I wanted to get rid of all of them. After few days of work it's done and most of the logic is moved to Makefile which is a lot shorter and easier to understand than the previous setup. There are still a lot of things that can be improved but at some point it has to be merged and then we can continue.

@floreks @shu-mutou and anyone else who wants to contribute please review and test it so we will avoid any regressions.

Progress:
- [x] add latest tag for releases
- [x] remove all tracks of Gulp and add Makefile
- [x] update documentation
- [x] update dependencies, Angular to v13 and Node to v16

Closes https://github.com/kubernetes/dashboard/issues/5417.

PR #6611, opened 2021-11-05
Bump nodejs to 16

shu-mutou:
NodeJS 16 becomes ACTIVE LTS from 16.13.0. So update to 16.13.0 or later.

https://nodejs.org/en/about/releases/


shu-mutou:
/hold
Checks, tests, build and built binary works well in our development container. But in not CI.

floreks:
E2E do not work for some time now. We have to investigate that. I also see that angular has finally fixed their node/npm engine restrictions and it is now allowed to be used with the new node v16.13 and npm v8. Great :slightly_smiling_face:

maciaszczykm:
I've fixed static check in my PR. Some of dependencies need updates.

shu-mutou:
/close
in favor of #6351 .

PR #6638, opened 2021-11-25
Add _tplvalues.tpl and define commonLabels, podLabels, commonAnnotations and podAnnotations attributes

ecthelion77:
* Add _tplvalues.tpl
* Update values file with default commonLabels, commonAnnotations, podLabels and podAnnotations
* Update all template to render properly labels and annotations

This change allows to use this helm chart with argocd for example and use multistring labels and annotations with the standardized way from bitnami


floreks:
> Add dependency to bitnami common helm chart

Could you provide a little more background behind this?
Why do we need this and why should we rely on this particular external dependency?

ecthelion77:
Today, it's not possible to provide common information to all resources deployed by the chart. It's really usefull to be able to do that for monitoring for example.

While I integrate those new information (commonLabels, podLabels, commonAnnotations and podAnnotations) seems to be really interesting to use the bitnami common chart as subchart to avoid  adding complexity on the concerned chart.

 If we integrate basically those new information on templates with :
```yaml
 {{- toYaml .Values.commonLabels | nindent 4 }}
```
it's always impossible to use a multistring to set information :

```yaml
commonLabels:
  label1: value1
  label2: value2
```
is working, but
```yaml
commonLabels: |
  "
  label1: value1
  label2: value2
  "
```
not working

with the templating from bitnami
```yaml
{{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
```

both possibilites are working.

And the second solution is really usefull to simplify deployment with argocd

floreks:
We are usually quite hesitant to add dependencies of third-party software that is not part of the k8s core (especially when it comes to deployment). We are only providing a helm chart because it is basically the main alternative to deploying apps in K8S. That's why we are not providing i.e. any example reverse-proxy configurations. There are just too many options and we'd rather force users to configure everything on their own without influencing their selection of third-party software.

desaintmartin:
I was hesitating before answering. I personally use `common` but if I'm not wrong, you only use https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_tplvalues.tpl which could be included into this chart instead of adding the whole dependency. What do you think?

(Thanks a lot for the files reformatting)

ecthelion77:
Yes currently I simply use _tplvalues and we can imagine simply add this template to the chart. In my sense, I consider it's interesting to add the dependency to the common to avoid a rewrite of existing template and to let the possibility for future to use others powerfull objects of the dependency.
But if you consider that it's too much and you validate solution based on the integration of the template, I can modify my PR.

desaintmartin:
It would be easy to add this dependency if there are other significant needs later, from now on I think the template integration directly in the chart would be simpler to manage.

ecthelion77:
@desaintmartin change are done, can you review

aio/deploy/helm-chart/kubernetes-dashboard/templates/ingress.yaml
```diff
@@ -79,6 +84,6 @@ spec:
   {{- end -}}
   {{- if .Values.ingress.tls }}
   tls:
-{{ toYaml .Values.ingress.tls | indent 4 }}
+{{ toYaml .Values.ingress.tls | nindent 4 }}
```
desaintmartin, line 87: hehe this could have been indented as well, but honestly thanks for the work so far in streamlining good usage of newlines.

desaintmartin (approved):

desaintmartin:
Could a member trigger the CI, please?

desaintmartin:
Oh, by the way, @ecthelion77, could you update PR name and description to refer to actual content?

ecthelion77:
> Oh, by the way, @ecthelion77, could you update PR name and description to refer to actual content?

done :)

aio/deploy/helm-chart/kubernetes-dashboard/templates/ingress.yaml
ecthelion77: with pleasure :)

ecthelion77:
up

desaintmartin:
@floreks @maciaszczykm Could one of you trigger CI?

ecthelion77:
@desaintmartin can you review again please, the chart version was bumped during waiting for worflow approval so i bumped again to version 5.0.5

ecthelion77:
up

PR #6645, opened 2021-12-01
Use prettier to format HTML files

maciaszczykm:
We indirectly use prettier as a dependency of gts to format TypeScript files. This change makes it our HTML formatter too. It has nice set of features and built-in option to check files, so the temporary file generation and diffs is not needed anymore. The only bigger difference that can be seen is how it handles attributes. It's no longer the case that every element attribute will be in different line unless line width limit is reached. There is an issue and pull request to implement this functionality: https://github.com/prettier/prettier/issues/5501 and https://github.com/prettier/prettier/pull/6644

Edit: The change is going to be done in Prettier v2.6.0 https://prettier.io/docs/en/next/options.html#single-attribute-per-line


jeefy:
:wave: :grin: 
/lgtm

PR #6649, opened 2021-12-03
Fix node event (/node/{name}/event) can not get the whole node event

SihengCui:
Using the kubectl command, I can get the node event like this:
image
However, on the dashboard, I cannot view any node events.
image

The reason of such problem is that, the k8s dashboard transmits `node.UID` to client-go for selection the involvedObject.uid of node event.

Kubectl [kubectl(describe node)](https://github.com/kubernetes/kubernetes/blob/555623c07eabf22864f6147736fa191e020cca25/staging/src/k8s.io/kubectl/pkg/describe/describe.go#L3524), it use `ref.UID=node.name` for searching.

In [kubernetes/kubernetes/issues/100236](https://github.com/kubernetes/kubernetes/issues/100236) , he reported a bug. That, the involvedObject.uid of node event may be node_uid or node_name.

So I set the uid of the search reference to `""`. It will not use involvedObject.uid to searching the node event, so that we can get the whole node event to handel both cases.


SihengCui:
CLA signed

SihengCui:
CLA signed

SihengCui:
CLA signed

SihengCui:
It is a natural way to use `ref.uid=node.uid` to select.
But for some node events, `node.uid` may not be generated. So some components write `node.name` to the `involvedObject.uid` of the node event.
I have now removed the uid filter condition, Is there a better solution?
This is my first pr to cncf community,  it's my pleasure to get your advise. @floreks @shu-mutou

floreks:
I am wondering if keeping an empty UID won't result in getting event information from the unrelated (similar) nodes. It might be better to first try with the UID and if it will return an empty result then retry with UID set to `node.name`.

SihengCui:
> 



> 

In my test environment, there are 9 node events at that time.  8 of them involvedObject.uid (from kubelte) are node_name , and 1 of them is  node_uid.

Is it necessary to search with nodename first(1 result)and then search with uid(8 results)? finally list append the results.

If uid="", just use involvedObject.name=node.name to search. Theoretically, the node name of k8s is uniquely. But I dont know if there are risks elsewhere.

SihengCui:
> I am wondering if keeping an empty UID won't result in getting event information from the unrelated (similar) nodes. It might be better to first try with the UID and if it will return an empty result then retry with UID set to `node.name`.

If so, you have to sort by CreationTimestamp.
If the empty uid can meet the requirements, only use `involvedObject.name=node.name` to find, it might be easier.

Or can we keep the same with `bukectl describe node`,  only set `involvedObject.uid = node.name` ?

SihengCui:
> I am wondering if keeping an empty UID won't result in getting event information from the unrelated (similar) nodes. It might be better to first try with the UID and if it will return an empty result then retry with UID set to `node.name`.

If so, you have to sort by CreationTimestamp.
If the empty uid can meet the requirements, only use `involvedObject.name=node.name` to find, it might be easier.

Or can we keep the same with `bukectl describe node`,  only set `involvedObject.uid = node.name` ?

SihengCui:
> I am wondering if keeping an empty UID won't result in getting event information from the unrelated (similar) nodes. It might be better to first try with the UID and if it will return an empty result then retry with UID set to `node.name`.

@floreks Some new work was submitted. `node,Name` and `node.UID` are both set as `node.UID` for searching. At the end, The results were integrated.

floreks:
Can you elaborate? Will it be fixed on the K8S core side or what?

SihengCui:
> Can you elaborate? Will it be fixed on the K8S core side or what?

@floreks 
Ais in [kubernetes/issues/100236](https://github.com/kubernetes/kubernetes/issues/100236) , node events are two situations. 
I think it is caused by different core components of kubernetes. 
When some node events are generated, the uid of the node is not generated. Therefore the `involvedObject.uid` is set to `node.name`.

The original logic of the k8s dashboard is to query the node info , and directly apply it to the query conditions of the node event. In this way, a part of the node event is lost, which `involvedObject.uid = node.name` . 

In my solution, the two cases were queried separately. Then integrate them.

src/app/backend/resource/event/common.go
```diff
@@ -136,13 +137,20 @@ func GetNodeEvents(client kubernetes.Interface, dsQuery *dataselect.DataSelectQu
...
 
-	eventList = CreateEventList(FillEventsType(events.Items), dsQuery)
+	node.UID = types.UID(node.Name)
+	eventsInvName, err := client.CoreV1().Events(v1.NamespaceAll).Search(scheme, node)
+	_, criticalError = errors.HandleError(err)
+	if criticalError != nil {
+		return &eventList, criticalError
+	}
+
+	eventList = CreateEventList(FillEventsType(append(eventsInvName.Items, eventsInvUID.Items...)), dsQuery)
```
floreks, line 153: Isn't there a chance of getting duplicate entries due to 2 event calls? Shouldn't we check for duplicates and filter them out from the result array?

SihengCui: > Isn't there a chance of getting duplicate entries due to 2 event calls? Shouldn't we check for duplicates and filter them out from the result array?

I don't think it is necessary. One event will not be set to `node.uid` and `node.name` to be saved separately. The `node.name` is the name we entered, and then the uid is randomly generated. Uid consists of 32-bit letters or numbers. It is almost impossible that the two are equal.

floreks: Makes sense

floreks:
Please fix the CI (`npm run fix` should be enough).

SihengCui:
> Please fix the CI (`npm run fix` should be enough).

@floreks Sorry. I clicked the reset review by mistake. Also, please describe Fix CI in detail. This is my first submission and I am not very familiar with the process.

SihengCui:
I have changed the go imports order to fix the CI.

PR #6653, opened 2021-12-07
docs: add pull request template

stu01509:
Add Pull Request Template for the developer to describe their PR changes.


floreks:
This is a 1:1 copy/paste of the PR template from the K8S core. We'd need a custom template that takes into account what we use and do here. Most of it is irrelevant for us. You can try to prepare a custom template just for us and open a new PR if needed.

/close

PR #6656, opened 2021-12-09
Fix the problem that some place in the code can not be translated directly

### LINKED ISSUE #6620: The notification panel's tooltip is difficult to translate to other language.
The notification panel's tooltip is difficult to translate to other language.  The i18n need static text but the text is dynamic.  It is writed by angular code to show or hide one of two sentence. 

I want to improve it. There is a method by using static html element to fix it. :)

image

The code is in https://github.com/kubernetes/dashboard/blob/15300bd43793dd12e8c38feaf4e4bee686d10905/src/app/frontend/chrome/notifications/template.html#L21

qlijin:
Some similar place can not be translated directly, as fellow: 

https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/chrome/notifications/template.html#L21

https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/actionbar/detailactions/pin/template.html#L21

https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/card/template.html#L43

The method is using static html element to fix it.

fix #6620


floreks:
I'd rather move those strings to code and use [$localize](https://angular.io/api/localize) to enable i18n for them.

As mentioned in the issue there are many more places like that in the code that have to be updated to allow i18n. I would prefer to have a single PR that fixes them all rather than many smaller PRs that fix 2-3 places every time.

qlijin:
> I'd rather move those strings to code and use [$localize](https://angular.io/api/localize) to enable i18n for them.

Thank @floreks , I will see the [$localize](https://angular.io/api/localize). 

> As mentioned in the issue there are many more places like that in the code that have to be updated to allow i18n. I would prefer to have a single PR that fixes them all rather than many smaller PRs that fix 2-3 places every time.

And for another suggestion, I may don't have enough time to translate that many places in one PR.  You can specify a single issue to let people in.

shu-mutou:
I found one more untranslated word.
Could you add `i18n-matTooltip` attribute into following too?
https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/list/column/menu/template.html#L65

qlijin:
> I found one more untranslated word. Could you add `i18n-matTooltip` attribute into following too? https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/list/column/menu/template.html#L65



> I found one more untranslated word. Could you add `i18n-matTooltip` attribute into following too? https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/list/column/menu/template.html#L65

Yes, of course. I will fix it

qlijin:
> I found one more untranslated word. Could you add `i18n-matTooltip` attribute into following too? https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/list/column/menu/template.html#L65

Yes, of course. I will fix it.

floreks:
What about using the `$localize` API? This is still not changed.

qlijin:
> What about using the `$localize` API? This is still not changed.
I tried the $localize, but it raise some error, and can not run, for myself it don't work. So I decided still use the template html element.

qlijin:
The way of using the template html element has no error.  Review is required

floreks:
This is not how it should be used. I have tested the `$localize` method and it works as it should without any errors. It should be refactored to a common config or a service where we could build localized messages used in the template in a generic and standardized way.

qlijin:
@floreks hi, I don't know the code you metioned. You can paste the code here or open your PR to fix it. Both is good.

floreks:
I'll reopen a new PR soon.

PR #6660, opened 2021-12-13
Fix container build (make docker-build-head)

marcosdiez:
I always build containers using `make docker-build-head`

when trying to run that (either in k8s or simply with `docker run -ti kubernetesdashboarddev/dashboard-amd64:head`, I was getting the error below:

`standard_init_linux.go:190: exec user process caused "no such file or directory"`

(example: `marcosdiez/dashboard:2.4.0c`)

After doing the changes in the PR from in the `makefile`, which I got from https://stackoverflow.com/a/62123648 , this worked again: example: `marcosdiez/dashboard:2.4.0d`

I use ubuntu 20.04 on a PC

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ clean:
 
 .PHONY: build-backend
 build-backend: ensure-go
-	go build -ldflags "-X $(MAIN_PACKAGE)/client.Version=$(RELEASE_VERSION)" -gcflags="all=-N -l" -o $(SERVE_BINARY) $(MAIN_PACKAGE)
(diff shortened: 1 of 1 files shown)


floreks:
It must have been missed during build pipeline refactoring as it was there before: https://github.com/kubernetes/dashboard/commit/2033f87f7949e9fa0512f3ad0d4c9c844ffb9dfa#diff-0ecc63da20d6cde87280e1e8d1590afb0494852fc6be123895752f34619d8d07L100

Good catch.

/lgtm

floreks (approved):

PR #6658, opened 2021-12-11
Services detail page now shows the ingressess related to it

marcosdiez:
image

requires #6664

PR #6664, opened 2021-12-16
migrate ingress from extensions/v1beta1 to networking/v1

marcosdiez:
Separate PR, as requested by @floreks on #6658


floreks:
I meant that we should do a common migration of all deprecated packages mentioned in the official guide :slightly_smiling_face: 

There is also `apiexteions/v1beta1`, possibly more.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

marcosdiez:
@floreks "apiexteions/v1beta1" is no more

floreks:
We should keep track of some other resources that will be removed a bit later (v1.25).

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125

src/app/backend/resource/common/resourcechannels.go
```diff
@@ -160,7 +160,7 @@ func GetServiceListChannel(client client.Interface, nsQuery *NamespaceQuery,
 
 // IngressListChannel is a list and error channels to Ingresss.
 type IngressListChannel struct {
-	List  chan *extensions.IngressList
+	List  chan *networkingv1.IngressList
```
floreks, line 163: Funny thing is that I think this is actually not used anywhere :joy: `ingress` package directly uses the `client.NetworkingV1` instead of channels. This is probably because it only requires a single call instead of multiple calls.

On the other hand `src/app/backend/client/manager.go` still creates a `k8sClient.ExtensionsV1beta1()` to create the `NewResourceVerber`. I guess this should also be updated.

marcosdiez: Hi @floreks 
According to https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/, `extensions/v1beta1` was only used for:

- `Ingress`, which this PR deals with
- `PodSecurityPolicy`, which is not yet handled by the dashboard
- `DaemonSet`, `Deployment` and `ReplicaSet` which are already handled by `apps/v1`

hence I just removed it's reference. This commit is trivial to commit, but please confirm you agree with the rationale behind it.

About CronJob (or any other thing that will be deprecated in the future, I'd rather wait to do the change. Not because of laziness, but because of backward compatibility. I know we only support officially the latest version, but I see no point in supporting a version that is not yet released ( 1.23 is less than a month old), since not everybody can upgrade k8s easily.

PR #6667, opened 2021-12-19
Fix advanced settings button name on create form

prnam:
This change will fix switch advanced settings button name present on `Create from form`

This what we see currently -
Action to hide is executing perfectly, however button name is always "Show advanced settings".

https://github.com/kubernetes/dashboard/blob/ffd9798fc09449beba0b2831667cbfb04405a525/src/app/frontend/create/from/form/template.html#L371

before-one

before-two

After PR is accepted this how it will look -

https://github.com/kubernetes/dashboard/blob/f8918f3b200b217c0e6239ec8072e4fe055e9677/src/app/frontend/create/from/form/template.html#L371
after-one

after-two

### DIFF
--- a/i18n/de/messages.de.xlf
+++ b/i18n/de/messages.de.xlf
@@ -4891,6 +4891,14 @@
           <context context-type="linenumber">354</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="2986806881378584169" datatype="html">
+        <source>{VAR_SELECT, select, true {Hide advanced options} other {Show advanced options}}</source>
+        <target state="new">{VAR_SELECT, select, true {Erweiterte Optionen ausblenden} other {Erweiterte Optionen anzeigen}}</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">src/app/frontend/create/from/form/template.html</context>
+          <context context-type="linenumber">371</context>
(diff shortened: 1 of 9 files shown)


floreks:
"Hide Advanced Options" is intentional. It indicates what will happen after clicking this button. We do not want to have 2x the same message.

prnam:
> "Hide Advanced Options" is intentional. It indicates what will happen after clicking this button. We do not want to have 2x the same message. It should be dynamic.

I did not understand you, @floreks. Did you mean you support this PR which will change current 2x same message ( i.e Show advanced settings) to getting Hide advanced settings text working on the button? 🤔

prnam:
Hey @chenrui, @helight,
Kindly review this PR and let me know if there is anything you need from my end. 🙂

floreks:
> > "Hide Advanced Options" is intentional. It indicates what will happen after clicking this button. We do not want to have 2x the same message. It should be dynamic.
> 
> I did not understand you, @floreks. Did you mean you support this PR which will change current 2x same message ( i.e Show advanced settings) to getting Hide advanced settings text working on the button? 🤔

Sorry for the late response. Initial issue description was a bit misleading for me also and I thought we already have Show/Hide text.

This change LGTM :)

PR #6672, opened 2021-12-22
Bump k8s modules to 0.23.1

shu-mutou:


maciaszczykm:
/lgtm

Thanks for the constant updates @shu-mutou!

PR #6682, opened 2022-01-07
Fix development container

shu-mutou:
* Use specified docker network, due to latest `kind` uses it.
* Change default running mode to `npm run start:prod`, because almost initial user
  for this development container is assumed to try dashboard or to translate dashboard.


maciaszczykm:
Nice, thanks!

/lgtm

PR #6684, opened 2022-01-12
Update npm packeges in dependencies

shu-mutou:
This PR updates dependencies for production that do not have breakable changes.


shu-mutou:
I would update other dependencies later.

shu-mutou:
Is EasyCLA gone?

Issue #6685, opened 2022-01-12
[BUG] ReplicaSet Error when deploying dashboard from helmchart


janlauber:
##### Environment
Production Kubernetes Cluster staged with RKE2

```
Installation method:
Kubernetes version: 1.22.5
Dashboard version (v1.x is not supported. Use v2.0 or later.): 2.4.0
Operating system: Ubuntu 20.04.3 (each K8S node)
```

##### Steps to reproduce
Just deploy the current helm chart version (5.0.5 on Artifacthub) or the 5.1.0 Version in the source code. The result will be the same. There will be no pods created via the deployment. (vanilla, no values provided)

##### Observed result
Pod cannot be deployed. When describing the replicaset you can see the following error message:
```
message: 'pods "kubernetes-dashboard-local-699c8b95fc-" is forbidden: PodSecurityPolicy:
      unable to admit pod: [pod.metadata.annotations[seccomp.security.alpha.kubernetes.io/pod]:
      Forbidden: seccomp may not be set pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/kubernetes-dashboard]:
      Forbidden: seccomp may not be set]'
```
##### Expected result
Deployment should complete and there should be one running pod.

##### Comments
I tried to fix this locally and it works for me now. I will create and link a PR which you can check out.

floreks:
Not an issue on our side. Explained in https://github.com/kubernetes/dashboard/pull/6686#issuecomment-1011089997

/close

PR #6686, opened 2022-01-12
Fix seccomp error on helmchart deployment

janlauber:
This PR faces the following issue:
- #6685 

It removes the securityContext templating implementation in the `deployment.yaml` at a place where it shouldn't belong.
Please checkout the issue and let me know if this PR works for you. I tested it locally and it works now and the replicaset can be created.

Signed-off-by: Jan Lauber <jan.lauber@protonmail.ch>

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/templates/deployment.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/templates/deployment.yaml
@@ -56,10 +56,6 @@ spec:
         {{- include "common.tplvalues.render" (dict "value" .Values.podLabels "context" $) | nindent 8 }}
         {{- end }}
     spec:
-{{- with .Values.securityContext }}
-      securityContext:
-{{ toYaml . | nindent 8 }}
-{{- end }}
       serviceAccountName: {{ template "kubernetes-dashboard.serviceAccountName" . }}
       containers:
       - name: {{ .Chart.Name }}


floreks:
This is not a correct way to fix this. `securityContext` can be set on the deployment level without any issues. If you do not need the `securityContext` to be set then modify the default values. 

Most probably this is an issue with your Kubernetes cluster. You are using rancher and K8S v1.22 (which is not yet officialy supported by the Dashboard FYI). I think that seccomp is simply enabled inside your cluster and that's why you are unable to deploy the application with custom `securityContext` settings. You should either override it during deployment or update your seccomp configuration.

/close

janlauber:
Hey @floreks 
Thanks for your answer.

So there could be a helm value to disable the securityContexts in the deployment manifest?
Can I try to implement this?

floreks:
Helm already provides a way to provide a custom `values` file or simply override a single value with the `--set` flag, right? Can't you just use that?

janlauber:
@floreks 
I tried to do that, with the following values.yaml configuration but **didn't** work:
```yaml
securityContext: {}
```
or
```yaml
securityContext:
  seccompProfile:
    type: null
```

Now I've implemented a `enabled: true` value in the values.yaml file with an `if` statement of the block I've deleted in this closed PR. This works now for me pretty well and the default of securityContext is set to true, so there should nothing change for every other one using this helm chart. Just people with securityContext disabled in the cluster can now deploy it. 
**Would like to make a new PR to show this to you.** 
If this change is **not welcome**, please provide me an **example** to disable the securityContext with the Helm functionality you mentioned.

thanks and greez

floreks:
Simply, remove the `securityContext` at all from your custom `values.yaml` file and it won't appear in the result deployment file.

janlauber:
@floreks 
So I tried that but this doesn't work either. 
When I get the deployment ressource with `kubectl get deployments.apps kubernetes-dashboard-local -o yaml` this is the output:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
    meta.helm.sh/release-name: kubernetes-dashboard-local
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2022-01-12T16:09:34Z"
  generation: 1
  labels:
    app.kubernetes.io/component: kubernetes-dashboard
    app.kubernetes.io/instance: kubernetes-dashboard-local
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: kubernetes-dashboard
    app.kubernetes.io/version: 2.4.0
    helm.sh/chart: kubernetes-dashboard-5.1.0
  name: kubernetes-dashboard-local
  namespace: default
  resourceVersion: "1332552"
  uid: df7773bc-98a2-49ed-89ce-4dc2de9e4350
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/component: kubernetes-dashboard
      app.kubernetes.io/instance: kubernetes-dashboard-local
      app.kubernetes.io/name: kubernetes-dashboard
  strategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/component: kubernetes-dashboard
        app.kubernetes.io/instance: kubernetes-dashboard-local
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/name: kubernetes-dashboard
        app.kubernetes.io/version: 2.4.0
        helm.sh/chart: kubernetes-dashboard-5.1.0
    spec:
      containers:
      - args:
        - --namespace=default
        - --auto-generate-certificates
        - --metrics-provider=none
        image: kubernetesui/dashboard:v2.4.0
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /
            port: 8443
            scheme: HTTPS
          initialDelaySeconds: 30
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 30
        name: kubernetes-dashboard
        ports:
        - containerPort: 8443
          name: https
          protocol: TCP
        resources:
          limits:
            cpu: "2"
            memory: 200Mi
          requests:
            cpu: 100m
            memory: 200Mi
        securityContext:
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
          runAsGroup: 2001
          runAsUser: 1001
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /certs
          name: kubernetes-dashboard-certs
        - mountPath: /tmp
          name: tmp-volume
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: # here is the line who shouldn't get included, when not adding the securityContext value to the custom values.yaml
        seccompProfile:
          type: RuntimeDefault
      serviceAccount: kubernetes-dashboard-local
      serviceAccountName: kubernetes-dashboard-local
      terminationGracePeriodSeconds: 30
      volumes:
      - name: kubernetes-dashboard-certs
        secret:
          defaultMode: 420
          secretName: kubernetes-dashboard-local-certs
      - emptyDir: {}
        name: tmp-volume
status:
  conditions:
  - lastTransitionTime: "2022-01-12T16:09:34Z"
    lastUpdateTime: "2022-01-12T16:09:34Z"
    message: Created new replica set "kubernetes-dashboard-local-699c8b95fc"
    reason: NewReplicaSetCreated
    status: "True"
    type: Progressing
  - lastTransitionTime: "2022-01-12T16:09:34Z"
    lastUpdateTime: "2022-01-12T16:09:34Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  - lastTransitionTime: "2022-01-12T16:09:34Z"
    lastUpdateTime: "2022-01-12T16:09:34Z"
    message: 'pods "kubernetes-dashboard-local-699c8b95fc-" is forbidden: PodSecurityPolicy:
      unable to admit pod: [pod.metadata.annotations[seccomp.security.alpha.kubernetes.io/pod]:
      Forbidden: seccomp may not be set pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/kubernetes-dashboard]:
      Forbidden: seccomp may not be set]'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
  observedGeneration: 1
  unavailableReplicas: 1
```

greez

floreks:
Looks like it might be automatically set by the seccomp profile enabled inside your cluster. I don't remember all settings we are using but I am 90% sure now that it does not come from our config.

janlauber:
So I deep dive analysed the rendering of the securityContext values and found a way to get it working when setting the following in the custom values.yaml file:
```yaml
securityContext: null
```
So it is indeed your config of rendering this variables. 
It has to be an `if` and not a `with` at the following line: https://github.com/kubernetes/dashboard/blob/1148f7ba9f9eadd719e53fa3bc8bde5b7cfdb395/aio/deploy/helm-chart/kubernetes-dashboard/templates/deployment.yaml#L59

also the following line must contain the whole values path of rendering:
https://github.com/kubernetes/dashboard/blob/1148f7ba9f9eadd719e53fa3bc8bde5b7cfdb395/aio/deploy/helm-chart/kubernetes-dashboard/templates/deployment.yaml#L61

The changes are faced in a new PR.
- #6687

greez

PR #6690, opened 2022-01-14
Update npm packeges in devDependencies

shu-mutou:
This PR updates dependencies for development that do not have breakable changes.


floreks:
We should not constantly be doing this manually. We do have a bot that should keep our dependencies up-to-date. The only time when we should intervene is:
- bulk update of related dependencies (i.e. angular)
- breaking change update where bot created a PR but it fails or requires manual changes in the codebase
- removal of unused dependencies

If the bot is not creating PRs then we should fix the bot configuration.

shu-mutou:
I found that dependabot is stopping to post PR for dashboard, because it does not support nodejs 16 and npm 8. :disappointed: 
So until following PR merged on dependabot, we need to update manually.
https://github.com/dependabot/dependabot-core/pull/4531

floreks:
Are you aware of how exactly is the bot checking which npm version is being used on the repo? We do not have an explicit version in the `package.json` anymore. I see that there is something in the lock file, but potentially we could change it to 7.X and this should allow bot to work maybe?

```json
      "engines": {
        "node": ">=16.13.0 <17",
        "npm": ">=8.1.0 <9"
      }
```

v8 should be mostly backward compatible so maybe it would work.

shu-mutou:
bot doesn't support lockfile v3 actually.

shu-mutou:
we can find errors of dependabot in following page:
https://github.com/kubernetes/dashboard/network/updates/258146930

floreks:
So it indeed take the `engine` from our `package.json`. Maybe we could enable npm v7, for now, to allow the bot to work.

shu-mutou:
I will try to change `engine` in `package.json` in the other PR.
But dependabot have stopped already several weeks, I want to update dashboard's dependency once right now.

PR #6699, opened 2022-01-24
Support npm v7 and nodejs v14 for backward compatibility.

shu-mutou:
This backward compatibility is for dependabot.
Dashboard may not work in nodejs v14.


shu-mutou:
Dependabot revived!! Thanks!!

PR #6734, opened 2022-01-27
Fix logs autorefresh

### LINKED ISSUE #6701: Auto-Refresh (every 5 s.) checkbox gets out of sync with backend state of the checkbox on same tab navigation
##### Environment
```
Installation method: `kubectl apply -f recommended.yaml`, with a locally cached copy of 2.0.5
Kubernetes version: v1.19.13-eks-8df270
Dashboard version (v1.x is not supported. Use v2.0 or later.): Dashboard v2.0.5+0.g8b89a1c7b
Operating system: linux, amd64
Node.js version ('node --version' output): ?? difficult to look up on EKS
Go version ('go version' output): ?? difficult to look up on EKS
```
Browser: Chrome 96.0

##### Steps to reproduce
1. Open a log for a pod, mark it to auto-refresh.
2. Leave the page navigating thru the bread crumbs, and come back to a log view for the pod or a different pod.
3. The checkbox for auto-refresh for the log is checked, but it is not refreshing.
4. Unchecking the checkbox causes it to start auto-refreshing.

Note that pressing refresh on the page clears the checkbox or opening k8s dashboard in a new tab clears the checkbox.

##### Observed result
The checkbox for auto-refresh is sticky for the browser (remains checked or remains unchecked while navigating to and from logs), but the setting doesn't get turned on for the page until it is clicked, regardless of the checkbox state before getting clicked.

##### Expected result
The setting is sticky, and when looking at a new or returning to a previously viewed log, it auto-refreshes the log according to the browser checkbox state for the setting.

##### Comments
My coworker pointed this out to me, and said he has seen this going on for the past year or two with k8s dashboard.
I looked for other issues mentioning this or commits about it, but I didn't see any.

floreks:
Fixes #6701

It does not introduce any kind of keeping logs settings after page refresh. During the view load, the autorefresh option was not checked. Right now it will check if it is set to true and start the refresh interval.

### DIFF
--- a/src/app/frontend/logs/component.ts
+++ b/src/app/frontend/logs/component.ts
@@ -17,13 +17,13 @@ import {Component, ElementRef, OnDestroy, ViewChild} from '@angular/core';
 import {MatDialog} from '@angular/material/dialog';
 import {ActivatedRoute, Router} from '@angular/router';
 import {LogControl, LogDetails, LogLine, LogSelection, LogSources} from '@api/root.api';
+
+import {LogsDownloadDialog} from '@common/dialogs/download/dialog';
 import {GlobalSettingsService} from 'common/services/global/globalsettings';
 import {LogService} from 'common/services/global/logs';
 import {NotificationSeverity, NotificationsService} from 'common/services/global/notifications';
 import {merge, Observable, of, Subject, timer} from 'rxjs';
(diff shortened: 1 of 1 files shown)


peteristhegreat:
Fantastic!  Thanks @floreks for the fix.  I really appreciate it.

PR #6741, opened 2022-01-30
Display more logs in the dashboard UI #6700

### LINKED ISSUE #6700: Display more logs in the dashboard UI
I noticed that only very few logs are displayed when i check the logs of a pod. The amount of logs displayed is very disappointing to me. it is approximately .... 50 lines?
Why not add a possibility to configure the amount of logs that are displayed in the UI? Or the possibility to select a time interval and set some default, like last 2 hours or something?
 
##### Why is this needed
If an exception occurs, it messes it all up and the log view in the dashboard UI is pretty much useless then. To see whats going on i would have to download the log file or use a shell and take a look with `kubectl`. Downloading lots of files is not really a desirable state imo, when you could just display more logs. Why not, i guess?
The same happens when i just want to check what happened, for example, a day ago.

glendsoza:
This PR tries to add a feature to configure the number of lines of logs being displayed in the view 

Fixes  #6700

### DIFF
--- a/src/app/frontend/logs/template.html
+++ b/src/app/frontend/logs/template.html
@@ -153,4 +153,22 @@ limitations under the License.
     <mat-icon>{{logService.getPrevious() ? 'check_box' : 'check_box_outline_blank'}}</mat-icon>
     <span i18n>Show previous logs</span>
   </button>
+  <div mat-menu-item
+       fxLayout="row"
+       fxLayoutAlign="start center"
+       fxLayoutGap="8px"
+       fxFlex="100"
+       (click)="$event.stopPropagation()">
+    <span i18n>Show Last</span>
+    <mat-form-field [style.width.%]="25">
+      <mat-select [(ngModel)]="logsPerView"
+                  (selectionChange)="setLogsPerView($event.value)">
+        <mat-option *ngFor="let logs of getLogsPerViewConfig()"
+                    [value]="logs">
+          {{logs}}
+        </mat-option>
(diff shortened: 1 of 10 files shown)


src/app/frontend/logs/template.html
```diff
@@ -153,4 +153,22 @@
...
+    <mat-form-field [style.width.%]="25">
+      <mat-select [(ngModel)]="logsPerView"
+                  (selectionChange)="setLogsPerView($event.value)">
+        <mat-option *ngFor="let logs of getLogsPerViewConfig()"
+                    [value]="logs">
+          {{logs}}
+        </mat-option>
+      </mat-select>
+    </mat-form-field>
+    <span i18n>Logs</span>
```
shu-mutou, line 172: To translate these words as one sentence, these `i18n` attribute should not be set in each `span` tags.
Could you move these `i18n` attribute to parent tag?

shu-mutou:
Also, please sign CLA. Thanks!

floreks:
I think I'd rather have it configurable through the general settings to allow persisting it across sessions. I will also have some comments regarding the code itself, but right now I don't have time for the proper review. Will do it later this week.

adamency:
Hey @floreks, what about this PR ? Logs are very much a priority for observability of the cluster and kubernetes-dashboard is a central piece of software to a significant proportion of clusters. It would greatly empower the community to be able to actually acess pod logs, instead of simply previewing them. Please consider adding the feature.

PR #6831, opened 2022-02-18
Bump k8s modules to 0.24.4

shu-mutou:
Also, remove `go get vendor` from `postinstall`.
Because the existence of `vendor` directory causes error
on `make fix-codegen` when update codegen version.


maciaszczykm:
/lgtm

Nice!

PR #6840, opened 2022-02-23
Fix bug in multiple removal from crd versions list

shixuyue:
If there are multiple served = false in the crd versions list, the original logic will break since the it will result in out of bond error.

Using the basic for loop to check if the Served field is false, if so, remove it from the version list, and shrink the size of list by 1 to avoid out of bond error.

### DIFF
--- a/src/app/backend/resource/customresourcedefinition/v1/common.go
+++ b/src/app/backend/resource/customresourcedefinition/v1/common.go
@@ -122,18 +122,14 @@ func isServed(crd apiextensions.CustomResourceDefinition) bool {
 }
 
 func removeNonServedVersions(crd apiextensions.CustomResourceDefinition) apiextensions.CustomResourceDefinition {
-	versions := append(crd.Spec.Versions)
-	if len(versions) == 1 && !versions[0].Served {
-		versions = make([]apiextensions.CustomResourceDefinitionVersion, 0)
-	} else {
-		for i := 0; i < len(versions); i++ {
-			if !versions[i].Served {
-				versions = append(versions[:i], versions[i+1:]...)
-				i--
-			}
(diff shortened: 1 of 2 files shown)


src/app/backend/resource/customresourcedefinition/v1/common.go
```diff
@@ -123,17 +123,17 @@ func isServed(crd apiextensions.CustomResourceDefinition) bool {
...
 	versions := append(crd.Spec.Versions)
-	for i, version := range crd.Spec.Versions {
-		if len(versions) == 1 && !version.Served {
-			versions = make([]apiextensions.CustomResourceDefinitionVersion, 0)
-			break
-		}
-
-		if !version.Served {
-			versions = append(versions[:i], versions[i+1:]...)
+	if len(versions) == 1 && !versions[0].Served {
```
floreks, line 126: Honestly, I am not sure why I have complicated it so much. Let's simply reverse the logic here and start with an empty array and add only served versions. Thanks for reporting this!
 
```go
	versions := make([]apiextensions.CustomResourceDefinitionVersion, 0)

	for _, version := range crd.Spec.Versions {
		if version.Served {
			versions = append(versions, version)
		}
	}

	crd.Spec.Versions = versions
	return crd
```

floreks:
Fixes #6632

src/app/backend/resource/customresourcedefinition/v1/common.go
shixuyue: yep, that would work just fine. This is actually my first thought of fix, but I thought you may have some other considerations to use an in-place like removal logic. Then I followed your pattern. 😆

PR #6852, opened 2022-02-24
Add untrim pipe to allow string indent in templates

### LINKED ISSUE #6696: text is missing identation in the log view after upgrade to 2.4.0
##### Environment

```
Installation method: alternative.yaml
Kubernetes version: 1.21.8
Dashboard version (v1.x is not supported. Use v2.0 or later.): 2.4.0
Operating system: Rocky Linux 8.5
```

##### Steps to reproduce

1. Install or upgrade to 2.4.0.
2. Go to any of the applications which log data with spaces before text, like structured JSON array output.

##### Observed result

Log text is completely left aligned and missing indentation.

##### Expected result

The log text should be idented.

##### Comments

I have confirmed that log text spaces actually exist in HTML code, it's probably an issue in CSS which messes alignment.

floreks:
Preview of the logs from the test yaml provided in #6696

Fixes #6696

### DIFF
--- a/src/app/frontend/common/pipes/module.ts
+++ b/src/app/frontend/common/pipes/module.ts
@@ -15,6 +15,7 @@
 import {DecimalPipe} from '@angular/common';
 import {NgModule} from '@angular/core';
 import {FilterByPipe} from '@common/pipes/filterby';
+import {UntrimPipe} from '@common/pipes/untrim';
 
 import {CommaSeparatedPipe} from './commaseparated';
 import {CoreFormatter} from './coreformatter';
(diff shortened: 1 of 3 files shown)


shu-mutou:
/lgtm
It works perfectly!

ViliusS:
I don't have an environment to test this at the moment, but won't this break large line wrapping in the log?

I don't think that it is a good idea to modify the log content in such way. There could be also issues with copying and pasting to various text editors.

floreks:
The character with code `160` is basically a `&nbsp;` character that is normally used by the browser to display leading/trailing spaces. In general, there is no issue with the line break and wrapping of the text.

PR #6868, opened 2022-03-02
Use Monaco Editor 

maciaszczykm:


angular.json
```diff
@@ -94,8 +93,7 @@
             "scripts": [
               "node_modules/sockjs-client/dist/sockjs.min.js",
               "node_modules/systemjs/dist/s.js",
-              "node_modules/systemjs/dist/extras/named-register.js",
-              "node_modules/systemjs/dist/extras/amd.js"
```
maciaszczykm, line 98: It needs to be checked if we can replace/remove it. It is used in the plugin system but it prevents us from using the ngx-monaco-editor module.

PR #6873, opened 2022-03-06
Deploy dashboard using docker out of K8s cluster.

qlijin:
As title, let people deploy dashboard using docker out of K8s cluster. Everyone can execute it very fast in this way even they are not familiar with K8s.


floreks:
To be honest this is not something we want to support nor recommend in any way to users. You are free to do it on your own.

/close

PR #6905, opened 2022-03-17
Fix the problem that some place in the code can not be translated directly

### LINKED ISSUE #6620: The notification panel's tooltip is difficult to translate to other language.

floreks:
A continuation of #6656

> Some similar place can not be translated directly, as fellow: 
> 
> https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/chrome/notifications/template.html#L21
> 
> https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/actionbar/detailactions/pin/template.html#L21
> 
> https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/card/template.html#L43
> 
> The method is using static html element to fix it.

Fixes #6620


maciaszczykm:
LGTM, there are some conflicts.

PR #6917, opened 2022-03-22
Added support for extra K8s manifests in the Helm chart

### LINKED ISSUE #6740: [FR] Allow the Chart to create extra manifest

Sometimes, accessory to the installation there may be manifest to be installed (for example a SecretProviderClass).

It would be nice to be able to do this directly via the Chart using the approach that ArgoCD also uses.

Ref: https://github.com/argoproj/argo-helm/blob/cf57a71a154a1a8a19f5f9d3980d60eb3efd80c8/charts/argo-cd/values.yaml#L62-L87

##### Why is this needed

See above

andreadecorte:
Closes #6740

### DIFF
--- a/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 5.3.1
+version: 5.4.0
 appVersion: 2.5.1
 description: General-purpose web UI for Kubernetes clusters
 keywords:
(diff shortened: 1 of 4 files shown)


aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
```diff
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 5.3.1
+version: 5.3.2
```
desaintmartin, line 17:
```suggestion
version: 5.4.0
```
This is a new feature, according to semver we should bump minor version.

desaintmartin (changes requested):
Thanks for the contribution! Here is a small note.

aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
andreadecorte: makes sense, thanks. Changed

desaintmartin (approved):

desaintmartin:
/lgtm
/approve 
cc @pierluigilenoci

pierluigilenoci:
@desaintmartin super cool! 🥳 
@andreadecorte thank you a lot! 🚀

Issue #5232, opened 2020-06-04
Support all kinds of resources


maciaszczykm:
- [ ] containers
- [ ] cron jobs
- [ ] containers

The list is based on https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/.

zehuaiWANG:
I think service accounts are quite necessary. Because many operations need to bind service accounts to give permissions.

zehuaiWANG:
BTW, I'm willing to contribute any part of the backend code if needed.

maciaszczykm:
I'll take care of roles and role bindings.

maciaszczykm:
@zehuaiWANG You can improve ingresses if you want. We display not all the information that we can.

creatica-soft:
Add the missing spec and status fields for Ingress objects. Currently, they just display the metadata.

marcosdiez:
Hey, @creatica-soft 
Last week v2.2.0 of the dashboard was released. Among other changes, it does have multiple improvements on the Ingress objects. You might want to check it out.

PR #6919, opened 2022-03-23
[#5232] Added Support for PriorityClass

shixuyue:
Added both of backend and frontend support for priority class.

1. I have made a new group called `Metadata` which can be used in the future to show more `metadata` objects. This `Metadata` is not shown and available at the main page. 
2. Added a dedicated card for priority class in the pod view.
3. The name of priority class in the card is clickable and clicking it will redirect to the priorityclass detail view.
4. The breadcumb allows users to view priority class list and all metadata objects. (currently only priorityclass is supported)
5. Added backend support for priorityclass. (brief info, detail info and list)

image
image
image
image


floreks:
Our layout of the left-side menu is based on the official K8S API structure: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/

Since `PriorityClass` is inside a metadata section we would not display it as a list. It should only be visible as part of the object it is applied to.

shixuyue:
> Our layout of the left-side menu is based on the official K8S API structure: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/
> 
> Since `PriorityClass` is inside a metadata section we would not display it as a list. It should only be visible as part of the object it is applied to.

@floreks Do we want to make priorityclass part clickable and then redirect to a detail view page?
Currently I cannot find a proper place for priority class modules in the repo.

maciaszczykm:
> @floreks Do we want to make priorityclass part clickable and then redirect to a detail view?
Currently I cannot find a proper place for priorityclass detail modules in the repo.

The second option could be adding a dedicated card for that. Let me check the docs and the code.

floreks:
I'd use a dedicated card also and embed detailed information directly into it and then we can use it as part of the higher-level resources.

maciaszczykm:
@floreks Agree, but I still think that the list is needed somewhere.

shixuyue:
@floreks  @maciaszczykm 
Thanks for advices.
I added a dedicated card for priorityclass, before I move forward, can you check if this is good? (I am sorry that I suck at frontend lol)
image
Also, I made the name of the priority class clickable.
However, the detail view of priority class needs a proper place. I am thinking of adding a new group called `Metadata` and we can have the list there. And this `Metadata` is not shown at the main page.  How does that sound?
Since there are a few more `Metadata` object are valuable to be viewed  as a list.

package.json
```diff
@@ -122,7 +122,7 @@
     "eslint-plugin-node": "11.1.0",
     "eslint-plugin-prettier": "4.0.0",
     "eslint-plugin-rxjs": "5.0.2",
-    "git-describe": "4.1.0",
+    "git-describe": "^4.1.0",
```
maciaszczykm, line 125: Remove `^`.

src/app/frontend/resource/metadata/routing.ts
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {NgModule} from '@angular/core';
+import {Route, RouterModule} from '@angular/router';
+import {MetadataComponent} from './component';
+
+export const Metadata_ROUTE: Route = {
```
maciaszczykm, line 19: Inconsistent naming. Check how it is done in other places.

maciaszczykm:
I do not have any better idea how to link to the list at the moment so I would say we can proceed and update later if needed. @floreks @shu-mutou do you agree? Any ideas?

shixuyue:
> horizontalpodautoscaler

sorry for the long delay. It was crazy for me for the past a few months. I finally have time to pick this up again.
While I think having `endpoint` is a good idea, but it doesnt apply to `priorityclass` that well. Since like `hpa` we can loop thru and check all hpas in the same `namespace` and then do a comparsion with the `name` and the `kind` of the `ScaleTargetRef`. But `priorityclass` doesnt contain the information of who are using this `priorityclass`. I think using an `endpoint` is kind of a waste to have an extra api call to the backend. We can do it with `PodDetails` which already contains the `priorityclass`, then we can just showing the brief information of the `priorityclass` in a delicated card.

floreks:
We can embed it directly inside the PodDetails. Since this is only a single object it should be fine. The tricky part is that `PriorityClassName` can be a reference to an object, string, or even `0` if there is no default.

floreks:

/close

PR #6920, opened 2022-03-23
Add support for IngressClass [#5232]

andreadecorte:
Adding support for Ingress Classes. I added them under "Cluster" section, let me know if you see them elsewhere instead (under Service?)

image
image

### DIFF
--- a/src/app/frontend/resource/discovery/ingressclass/routing.ts
+++ b/src/app/frontend/resource/discovery/ingressclass/routing.ts
@@ -1,0 +1,47 @@
+// Copyright 2017 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {NgModule} from '@angular/core';
+import {Route, RouterModule} from '@angular/router';
+import {BREADCRUMBS} from '../../../index.messages';
+import {DEFAULT_ACTIONBAR} from '@common/components/actionbars/routing';
+
+import {DISCOVERY_ROUTE} from '../routing';
+
+import {IngressClassDetailComponent} from './detail/component';
+import {IngressClassListComponent} from './list/component';
+
+const INGRESSCLASS_LIST_ROUTE: Route = {
+  path: '',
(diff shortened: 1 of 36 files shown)


floreks:
Our layout is based on the https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#ingressclass-v1-networking-k8s-io

IngressClass is under Service section which means it should be the same in the Dashboard.

andreadecorte:
Thanks @floreks , moved to the Service section, updated screenshot below
Screenshot 2022-03-24 at 11 17 08

src/app/frontend/typings/root.api.ts
```diff
@@ -405,6 +409,11 @@ export interface StorageClass extends Resource {
   parameters: StringMap[];
 }
 
+export interface IngressClass extends Resource {
+  controller: string;
+  parameters: StringMap[];
```
floreks, line 414: I don't see this field on the API side.

andreadecorte: I am flattening here https://github.com/kubernetes/dashboard/pull/6920/files#diff-f2a62a475a29db82a76ba9f4ea8915ed8a175af848d689109ecb25e118e6aa22R46 IngressClassParametersReference to a map of strings as all the contents are strings. Should I rather keep a dedicated object for that?

floreks: Ye, but this is not the `IngressClassDetails` object, it's this one https://github.com/kubernetes/dashboard/pull/6920/files#diff-a6cfbdffa7b2f43b890a274cb3a387fcda23797c8cc90e7bada690411a689d19R38-R42

andreadecorte: and you're of course right, field removed from there

src/app/frontend/resource/discovery/ingressclass/routing.ts
```diff
@@ -0,0 +1,46 @@
...
+import {DISCOVERY_ROUTE} from '../routing';
+
+import {IngressClassDetailComponent} from './detail/component';
+import {IngressClassListComponent} from './list/component';
+
+const INGRESSCLASS_LIST_ROUTE: Route = {
+  path: '',
+  component: IngressClassListComponent,
+  data: {
+    breadcrumb: 'Ingress Classes',
```
floreks, line 28: A new way to localize breadcrumbs should be used here. It was recently merged. There is now `messages.ts` file that localizes such messages.

andreadecorte: thanks, updated and rebased

andreadecorte:
did you have any chance to take a look again at this one @floreks ?

floreks (approved):
/lgtm
/approve

floreks:
@andreadecorte I think we can merge. Thanks! :slightly_smiling_face:

andreadecorte:
> @andreadecorte I think we can merge. Thanks! 🙂

thank you for the merge!

PR #6922, opened 2022-03-25
Add secure domains args

ndunks:
Add argumens on dashboard binary to enable login button on frontend that includ in the domains

default domains is: `localhost` `127.0.0.1`

with `--secure-domains`, developer can customize that list to fit their needed. So frontend will enable login on custom domain like `dashboard.svc.cluster.local`

### DIFF
--- a/src/app/backend/args/builder.go
+++ b/src/app/backend/args/builder.go
@@ -162,6 +162,12 @@ func (self *holderBuilder) SetNamespace(namespace string) *holderBuilder {
 	return self
 }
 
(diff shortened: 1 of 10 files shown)


floreks:
This is not something we want to change. It is done this way for a security reason. Accessing Dashboard over HTTP will be restricted to localhost domain only. We do not want to open up a way to omit this restriction. Any other domain should always be accessed over HTTPS when login functionality is available.

/close

ndunks:
nope, at least it what I need and I add it.

PR #6952, opened 2022-04-02
Normalize graph values after determining the highest suffix in the whole dataset

jamesorlakin:
I noticed the graphs have subtly incorrect normalizing behaviour if latter data points have a higher suffix. For example, crossing the Mi to Gi boundary results in the strange behaviour of the graph being in Gi but some values being unnormalized Mi:
Pre

This patch tweaks normalization to happen on all values after finding the suffix. It also tweaks the 'ballast' added to memory graphs so that the change between Mi and Gi is less jarring:
Post


floreks:
This totally makes more sense than the old version. Nice catch and fix! Thanks 🙂

PR #6964, opened 2022-04-05
ci(helm): remove remote PR branch if already exists.

desaintmartin:
Otherwise, we can't have several PRs opened by CD, or we can't create new PR if old branch has not been deleted.

### DIFF
--- a/.github/workflows/cd-helm-workflow.yml
+++ b/.github/workflows/cd-helm-workflow.yml
@@ -44,18 +44,21 @@ jobs:
         env:
           TERM: xterm-256color
 
+      - name: Remove remote Pull Request branch if it already exists
+        # Note: closes related PR if exists
+        run: |
+          git ls-remote --heads --exit-code origin gh-pages-pull-request && git push origin :gh-pages-pull-request
+
       - name: Create Pull Request branch
         run: |
           git checkout -b gh-pages-pull-request
 
-      # Note that will fail if branch already exists.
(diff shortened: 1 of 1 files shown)


floreks:
To be honest, I am not sure if it will work since branches are protected and only the repo admin can remove them right now.

floreks:
I'll try to do it later today. It would be best if the name of the branch could contain some unique part such as short commit hash or some id so that it would not be blocked by this auto branch protection mechanism. Every time I want to delete the branch I have to first go to repo settings and remove branch protection first.

desaintmartin:
Actually, auto-creating PRs out of order may result in a missing version, and I think using the current way (one single git repository) we can't do much better. Missing a past version is not that much important, but missing a next version is.

Can't you see a "delete branch" from the PR I mentioned?

floreks:
> Can't you see a "delete branch" from the PR I mentioned?

The Kubernetes org has the configuration in place that automatically adds protection to all branches in the repo every couple of minutes. I have to first remove the protection manually before deleting the branch. Unfortunately, it's a bit inconvenient.

desaintmartin:
Okay, I see. Maybe I can actually setup a fork outside of the kubernetes organization so that it really works like a normal PR without any special permission required, but that would mean that CI/CD would push on this one. I suppose this is okay since I would give access from "my" fork to the CI/CD here, but that is a bit... uncommon.

PR #6967, opened 2022-04-05
Update docs for 1.24+

liggitt:
In 1.24+, secret-based tokens are no longer auto-created.

This updates to use the `kubectl create token` command provided instead.


floreks:
Thanks for the update. Let's hold it until the official v1.24 release and once we update our dependencies to support it officially.

/hold

sftim:
What work remains before this is OK to unhold?

liggitt:
@floreks - any update on this? 1.24 has been out for almost a month

sftim (commented):
LGTM

maciaszczykm:
New release should happen today or tomorrow. Let's cancel hold.

/hold cancel

PR #6971, opened 2022-04-06
Remove unused import and useless conditional expression

xiaoyang-sde:
Remove unused import:
- `MatMenu` in `src/app/frontend/chrome/userpanel/component.ts`
- `MatDialogRef` in `src/app/frontend/common/dialogs/config/dialog.ts`
- `SecurityContext` and `DomSanitizer` in `src/app/frontend/common/pipes/stripansi.ts`
- `StringMap` in `src/app/frontend/resource/cluster/persistentvolume/detail/component.ts`

Remove useless conditional expression:
- `route` in `src/app/frontend/common/components/breadcrumbs/component.ts`

### DIFF
--- a/src/app/frontend/common/components/breadcrumbs/component.ts
+++ b/src/app/frontend/common/components/breadcrumbs/component.ts
@@ -101,7 +101,7 @@ export class BreadcrumbsComponent implements OnInit {
 
         // Explore the route tree to the root route (parent references have to be defined by us on
         // each route).
-        if (route && route.data && route.data.parent) {
+        if (route.data && route.data.parent) {
           route = route.data.parent;
         } else {
           break;
--- a/src/app/frontend/common/components/breadcrumbs/component.ts
+++ b/src/app/frontend/common/components/breadcrumbs/component.ts
@@ -101,11 +101,7 @@ export class BreadcrumbsComponent implements OnInit {
 
         // Explore the route tree to the root route (parent references have to be defined by us on
         // each route).
-        if (route.data && route.data.parent) {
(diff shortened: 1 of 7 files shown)


src/app/frontend/common/components/breadcrumbs/component.ts
```diff
@@ -101,7 +101,7 @@ export class BreadcrumbsComponent implements OnInit {
 
         // Explore the route tree to the root route (parent references have to be defined by us on
         // each route).
-        if (route && route.data && route.data.parent) {
+        if (route.data && route.data.parent) {
```
floreks, line 104: I think we could simplify this even further and simply go with `route = route?.data?.parent`. An undefined route will break the loop anyway so `else` is not really needed.

floreks (changes requested):
Overall, I'd prefer to update the `no-unused-vars` linter rule to automatically throw errors when unused imports or variables are found.

xiaoyang-sde:
> Overall, I'd prefer to update the `no-unused-vars` linter rule to automatically throw errors when unused imports or variables are found.

I found that the .eslintrc.yaml file explicitly disabled this rule for some reason. I could try to enable this rule and check if it will introduce unexpected effects.

floreks:
Ye, I am not sure why it is disabled. Below config should be sufficient:
`"@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern":  "^_"}],`

xiaoyang-sde:
> 

Thanks. I enabled it in the latest commit!

PR #6990, opened 2022-04-09
Fix the unexpected confirmation dialog in 'Create from *'

xiaoyang-sde:
The dialog that alerts 'The form has not been submitted yet, do you really want to leave?' (introduced in #6566) will appear even if the user clicked the 'Upload' or 'Deploy' button and the request is successful. This pull request fixes the unexpected confirmation dialog in the 'Create from file', 'Create from form', and 'Create from input' components.

The `this.create_.createContent()` function navigates to the `/overview` page before it returns, but the `this.creating_` variable will be changed to `false` after the function returns. Therefore, the`CanDeactivateGuard` guard will prevent the navigation.

```ts
  create(): void {
    this.creating_ = true;
    this.create_
      .createContent(this.file.content, true, this.file.name)
      .then(() => (this.creating_ = false))
      .finally(() => (this.creating_ = false));
  }
```

The solution is to include a `bypassGuard` state in the navigation of `this.create_.createContent()` function to let it bypass the `CanDeactivate` guard.

### DIFF
--- a/src/app/frontend/create/from/file/component.ts
+++ b/src/app/frontend/create/from/file/component.ts
@@ -43,17 +43,17 @@ export class CreateFromFileComponent extends ICanDeactivate {
     return !this.file || this.file.content.length === 0 || this.create_.isDeployDisabled();
   }
 
-  async create(): Promise<void> {
+  create(): void {
     this.creating_ = true;
-    try {
-      await this.create_.createContent(this.file.content, true, this.file.name);
-      this.file = {
-        name: '',
-        content: '',
-      };
-    } finally {
-      this.creating_ = false;
-    }
+    this.create_
+      .createContent(this.file.content, true, this.file.name)
+      .then(() => {
+        this.file = {
+          name: '',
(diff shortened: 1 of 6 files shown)


src/app/frontend/create/from/file/component.ts
```diff
@@ -43,12 +43,17 @@ export class CreateFromFileComponent extends ICanDeactivate {
     return !this.file || this.file.content.length === 0 || this.create_.isDeployDisabled();
   }
 
-  create(): void {
+  async create(): Promise<void> {
```
floreks, line 46: I am actually not a big fan of using async/await and `Promise<void>` without returning anything from the function. Here it's used more as a workaround or alternative code rather than for any good reason. Using `then`, `catch`, and `finally` is shorter and easier to read for me.

xiaoyang-sde: Yeah this makes sense. Thanks for the suggestion.

PR #7061, opened 2022-05-01
added certificate based authentication using the kubeconfig certs

yuvDev-hub:


floreks:
Certificate-based auth through web browser is extremely insecure and we have not added it on purpose. Private key should never leave your PC.

/close

PR #7047, opened 2022-04-27
Split frontend and backend

maciaszczykm:
Closes https://github.com/kubernetes/dashboard/issues/6799.

PR #7073, opened 2022-05-04
Bump k8s modules to 0.24.0

shu-mutou:


maciaszczykm:
@shu-mutou Can we avoid regular bumps until https://github.com/kubernetes/dashboard/pull/7047 is in progress? It will cause conflicts that we will need to resolve. If you want you can update the dependencies in the branch directly.

shu-mutou:
/hold
https://github.com/kubernetes/dashboard/pull/7047 will be merged

Issue #7080, opened 2022-05-09
Pod information page failed to load when using (generic) ephemeral volume mount


cshubhamrao:
### What happened?

Trying to view a Pod on Dashboard results in an (almost) blank page when the Pod uses a [Generic ephemeral volume](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes)

The observations and JS Console output are same as what was reported in #5831 and https://github.com/kubernetes/dashboard/issues/6381 (same issue but with a `cephFS` volume)
![Screenshot 2022-05-09 at 2 29 19 PM](https://user-images.githubusercontent.com/3601491/167378968-36b3873f-0b51-44d2-9c84-412949979328.png)

![image](https://user-images.githubusercontent.com/3601491/167379418-67bf3de9-d326-49fd-8ed5-9daa25604122.png)

### What did you expect to happen?

To be able to view all the Pod Details (along with PV details). For eg: 
![image](https://user-images.githubusercontent.com/3601491/167380082-aab3e208-c617-452f-8399-9de91cce3db6.png)

### How can we reproduce it (as minimally and precisely as possible)?

Assuming the cluster is on AWS (for the `gp2` storage class) [Not sure if `storageClass` is relevant]
* Deploy this pod (uses a generic ephemeral volume using the default storage class):
```bash
cat << EOF | kubectl apply -f -
kind: Pod
apiVersion: v1
metadata:
  name: test-pod
spec:
  volumes:
    - name: job-folder
      ephemeral:
        volumeClaimTemplate:
          spec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 1Gi
            volumeMode: Filesystem
  containers:
    - name: jobs
      image: ubuntu
      command:
        - sleep
      args:
        - "infinity"
      workingDir: /job
      resources:
        requests:
          cpu: 100m
          memory: 200Mi
      volumeMounts:
        - name: job-folder
          mountPath: /job
      terminationMessagePath: /dev/termination-log
      terminationMessagePolicy: File
      imagePullPolicy: IfNotPresent

EOF
```
* Navigate to Kubernetes Dashboard to view the pod details

### What browsers are you seeing the problem on?

Chrome, Safari

### Kubernetes Dashboard version

v2.5.1+0.g520013042

### Kubernetes version

v1.22.6-eks-7d68063

cshubhamrao:
Logs from `kubernetes-dashboard` pod:
[...]

PR #7093, opened 2022-05-10
Add barebones support for generic ephemeral volume

cshubhamrao:
Adds an "Ephemeral" volume source with blank ("-") values so that the page loads atleast

image

See #7080


cshubhamrao:
@maciaszczykm PTAL

floreks (approved):
/lgtm

PR #7110, opened 2022-05-17
Show Cluster status with cumulative node resources

### LINKED ISSUE #7109: Show cluster summary of available node resources

I would like the node resources to be aggregated in the "Cluster" view, the same way that pods are in the "Workload" view.

This would show a summary of _all_ the nodes, with their status and their cpu/memory/pods allocation.

Here is a screenshot from my implementation, using pie charts (different colors for requests and for limits):

kubernetes-cluster-status

### Why is this needed?

Currently there is a list of nodes, and a detail page of each. But there is no cluster summary.

The other pages are also missing the capacity information, but that will be a separate issue.

node-capacity

afbjorklund:
Add a summary of nodes and resources at the top of Cluster,
similar to the existing summary of pods at the top of Workload.

Refactor the calculation of fractions, to reduce duplication
and to make it possible to apply after adding up all the nodes.

Closes #7109

### DIFF
--- a/src/app/backend/resource/node/detail.go
+++ b/src/app/backend/resource/node/detail.go
@@ -197,39 +197,58 @@ func getNodeAllocatedResources(node v1.Node, podList *v1.PodList) (NodeAllocated
 	cpuRequests, cpuLimits, memoryRequests, memoryLimits := reqs[v1.ResourceCPU],
 		limits[v1.ResourceCPU], reqs[v1.ResourceMemory], limits[v1.ResourceMemory]
 
-	var cpuRequestsFraction, cpuLimitsFraction float64 = 0, 0
-	if capacity := float64(node.Status.Allocatable.Cpu().MilliValue()); capacity > 0 {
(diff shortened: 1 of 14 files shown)


maciaszczykm:
@afbjorklund Thanks for contributions! We'll release new version today, so you will be able to use it soon.

maciaszczykm (commented):
After discussion with @floreks we agreed that this needs to be reverted until it will be refactored. We'd like to avoid sending this data twice for node list as each node already has all the required data. You can create cumulative statuses in the frontend code. I hope that you will be interested in finishing it, sorry for confusion!

afbjorklund:
Will give it a try, and make a new PR.

maciaszczykm:
@afbjorklund Great. 

> Hopefully we can still add Ready somewhere.

https://github.com/kubernetes/dashboard/blob/master/src/app/frontend/common/components/resourcelist/node/template.html#L80

It's in the list already.

afbjorklund:
Ok, it seems to be _only_ attribute. Need to find the Resource.

PR #7152, opened 2022-06-03
Support custom rules for read-only mode

korjek:
When dashboard is started with read-only role, CRDs are not shown in UI due to permissions issue.
This PR adds ability to provide custom rules for read-only role


korjek:
@desaintmartin sorry for the late reply, it fell out of my radar.

>Wouldn't it be better to add the ability in this case to add in the clusterrole itself the missing permission?
Not sure that I got the idea. Could you please explain a bit more?

>in the case where someone wants to have its own customized rbacs, it's probably better to do it outside of the chart
If someone uses helm to deploy everything, that would mean (if I got your idea correct) creating a custom in-house chart and using it to create role with custom permissions for k8s dashboard. I don't think this is a good way to go.

korjek:
> Instead of setting a new option and telling users to add CRD permissions in extra RBAC to have a working dashboard, shouldn't it be done by default?

I don't think it's possible, as every CRD has it's own apiGroup.
So we with the PR I propose we have the next lines in values.yaml to be able to see `sparkapplications` in k8s dashboard:
```
    clusterReadOnlyRoleAdditionalRules:
      - apiGroups:
        - apiextensions.k8s.io
        resources:
        - customresourcedefinitions
        verbs:
        - get
        - list
        - watch
      - apiGroups:
        - sparkoperator.k8s.io
        resources:
        - sparkapplications
        - sparkapplications/status
        - scheduledsparkapplication
        - scheduledsparkapplication/status
        verbs:
        - get
        - list
        - watch
```

korjek:
@desaintmartin thoughts?

korjek:
@desaintmartin could we proceed this way? or what other options do we have instead of forking?

korjek:
@shu-mutou maybe you can help here?

desaintmartin:
Sorry for the huge delay. Indeed this makes sense for CRDs.

aio/deploy/helm-chart/kubernetes-dashboard/Chart.yaml
```diff
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 5.5.2
+version: 5.5.3
```
desaintmartin, line 17:
```suggestion
version: 5.6.0
```
(after a rebase)

korjek:
@desaintmartin thank you!
Resolved conflicts

PR #7189, opened 2022-06-16
fix: fix panic while RsaKeyHolder secret is nil

ismdeep:


floreks:
That's a good catch. In theory, a secret could be inaccessible and in sporadic cases, it could be nil.

/lgtm
/approve

PR #7244, opened 2022-07-09
fix: close sizeChan to avoid memory leak.

lixd:


floreks (approved):

floreks:
Thank you for the fix and sorry for the late response. Finally found some time :slightly_smiling_face: 

/lgtm
/approve

PR #7335, opened 2022-08-07
Alternative create token command for Kubernetes prior to v1.24

flavienbwk:
The currently suggested command [doesn't work](https://stackoverflow.com/questions/72481655/creating-a-kubernetes-dashboard-token) for Kubernetes < v1.24.

Signed-off-by: Flavien Berwick <flavien@berwick.fr>


floreks:
Unfortunately, since we do not officially support K8S versions older than v1.24 (as of Dashboard v2.6.0) I don't think we should put such information in our documentation. See our [compatibility matrix](https://github.com/kubernetes/dashboard/releases/tag/v2.6.0).

/close

PR #7372, opened 2022-08-18
Add support for ephemeral-storage display

### LINKED ISSUE #7371: Add support for ephemeral storage display

Beyond CPU and Memory, the dashboard should also show the Ephemeral storage (requests/limits/capacity-allowed)

It is available as a resource, both for nodes and for pods. It is measured in bytes, and displayed similar to memory.

### Why is this needed?

In version 1.10, Kubernetes also learned how handle local ephemeral storage:

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

afbjorklund:
Closes #7371

### DIFF
--- a/src/app/backend/resource/node/detail.go
+++ b/src/app/backend/resource/node/detail.go
@@ -51,6 +51,22 @@ type NodeAllocatedResources struct {
 	// CPUCapacity is specified node CPU capacity in milicores.
 	CPUCapacity int64 `json:"cpuCapacity"`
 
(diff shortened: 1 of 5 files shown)


romdhanisam (approved):

floreks:
https://github.com/kubernetes/dashboard#-important-

afbjorklund:
Consider it a feature request for the next generation, then. 🙂

Kubernetes itself is in a similar freeze, so the metric won't be fixed either...

PR #7414, opened 2022-08-31
Fix tags in German translation

shu-mutou:


shu-mutou:
/cc @maciaszczykm 
Build error in master branch will fixed by this PR.

PR #7430, opened 2022-09-08
ci: add minimum GitHub token permissions for workflows

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using https://github.com/step-security/secure-workflows.

GitHub Actions workflows have a GITHUB_TOKEN with write access to multiple scopes.
Here is an example of the permissions in one of the workflows:
https://github.com/kubernetes/dashboard/runs/7993046968?check_suite_focus=true#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflows:

- cd-helm-workflow.yml

The following workflow files already have the least privileged token permission set:

- cd-workflow.yml
- ci-helm-workflow.yml
- ci-workflow.yml
- codeql-analysis.yml
- depsreview.yaml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- GitHub recommends defining minimum GITHUB_TOKEN permissions.
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>


.github/workflows/cd-helm-workflow.yml
```diff
@@ -22,8 +22,14 @@ on:
     - 'aio/scripts/helm-release-chart.sh'
     - '.github/workflows/cd-helm-workflow.yml'
 
+permissions:
+  contents: read
```
maciaszczykm, line 26: Is it necessary since we have only one job and you change the permissions below?

boahc077: Yes, the contents: read is not strictly needed here. Setting that at the workflow level helps if a new job is added to the workflow. In that case, the new job will have contents: read, which is a secure-by-default setting.

Issue #7480, opened 2022-09-23
Get http panic runtime error: invalid memory address or nil pointer dereference when viewing a service


shankerwangmiao:
### What happened?

In a namespace, when clicking on any of the services, the browser displays a 502 error, and the following error is given in the  output log of the dashboard process:

```
2022/09/23 19:02:42 http: panic serving 10.36.11.72:43794: runtime error: invalid memory address or nil pointer dereference
goroutine 4534 [running]:
net/http.(*conn).serve.func1()
 /opt/hostedtoolcache/go/1.19.0/x64/src/net/http/server.go:1850 +0xbf
panic({0x177be80, 0x2890960})
 /opt/hostedtoolcache/go/1.19.0/x64/src/runtime/panic.go:890 +0x262
github.com/kubernetes/dashboard/src/app/backend/resource/ingress.ingressMatchesServiceName({{{0x0, 0x0}, {0x0, 0x0}}, {{0xc0010227e0, 0x17}, {0x0, 0x0}, {0xc0008d9580, 0x4}, ...}, ...}, ...)
 /home/runner/work/dashboard/dashboard/src/app/backend/resource/ingress/filter.go:39 +0x1b4
github.com/kubernetes/dashboard/src/app/backend/resource/ingress.FilterIngressByService(...)
 /home/runner/work/dashboard/dashboard/src/app/backend/resource/ingress/filter.go:25
github.com/kubernetes/dashboard/src/app/backend/resource/service.GetServiceIngressList({0x1c45500, 0xc000180a80}, 0x199a011?, {0xc0004de014, 0x4}, {0xc0004de019, 0x19})
 /home/runner/work/dashboard/dashboard/src/app/backend/resource/service/ingress.go:39 +0x348
github.com/kubernetes/dashboard/src/app/backend/handler.(*APIHandler).handleGetServiceIngressList(0x68?, 0xc00057c120, 0x1a?)
 /home/runner/work/dashboard/dashboard/src/app/backend/handler/apihandler.go:1136 +0x10c
github.com/emicklei/go-restful/v3.(*FilterChain).ProcessFilter(0xc0004de100?, 0x68?, 0xc000358150?)
 /home/runner/go/bin/pkg/mod/github.com/emicklei/go-restful/v3@v3.7.4/filter.go:23 +0x58
github.com/kubernetes/dashboard/src/app/backend/handler.restrictedResourcesFilter(0xc00057c120, 0x0?, 0x2?)
[...]

PR #7481, opened 2022-09-23
check the existence of the http member of a ingress rule before iterating through it

shankerwangmiao:
Fixes: #7480


floreks (approved):

floreks:
Thanks for the fix
/lgtm

PR #7489, opened 2022-09-29
Resource owner now shown in every pane

marcosdiez:
This is a duplicate of #6237

image

I am not sure if this is useful anywhere but in a `ReplicaSet` and in a `Job` started by a `CronJob` Also, we have the "problem" of having the info shown twice in Pods.

@floreks , my team and I use this every day. It's quite useful. Consider merging, please!

### DIFF
--- a/modules/web/i18n/fr/messages.fr.xlf
+++ b/modules/web/i18n/fr/messages.fr.xlf
@@ -930,6 +930,10 @@
         <source>unset</source>
         <target>non défini</target>
       </trans-unit>
+      <trans-unit id="3715596725146409911" datatype="html">
+        <source>Owner</source>
+        <target state="new">Owner</target>
+      </trans-unit>
       <trans-unit id="374277779600579508" datatype="html">
         <source>Resource Limits</source>
         <target>Limites de ressources</target>
--- a/modules/web/src/common/components/objectmeta/component.ts
+++ b/modules/web/src/common/components/objectmeta/component.ts
@@ -14,6 +14,9 @@
 
 import {Component, Input} from '@angular/core';
 import {ObjectMeta} from '@api/root.api';
+import {GlobalServicesModule} from '../../services/global/module';
+import {KdStateService} from '../../services/global/state';
+import {Resource} from '@common/services/resource/endpoint';
 
 @Component({
   selector: 'kd-object-meta',
@@ -23,6 +26,8 @@ export class ObjectMetaComponent {
   @Input() initialized = false;
 
   private objectMeta_: ObjectMeta;
+  private readonly kdState_: KdStateService = GlobalServicesModule.injector.get(KdStateService);
+
   get objectMeta(): ObjectMeta {
     return this.objectMeta_;
   }
@@ -34,5 +39,13 @@ export class ObjectMetaComponent {
     } else {
       this.objectMeta_ = val;
     }
+  }
+
+  getObjectHref(kind: string, name: string): string {
+    if (!Object.values(Resource).includes(kind.toLowerCase() as Resource)) {
+      return '';
+    }
+
+    return this.kdState_.href(kind.toLowerCase(), name, this.objectMeta_.namespace);
   }
 }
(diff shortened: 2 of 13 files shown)


maciaszczykm:
@marcosdiez CI doesn't pass.

What do you mean by duplicate?

marcosdiez:
@floreks @maciaszczykm do you guys know why the tests are failing ?
the quick way to reproduce it is to type `yarn test src/resource/discovery/ingress/detail/component.spec.ts`

The issue is caused by this line:
https://github.com/kubernetes/dashboard/blob/95a06ec8ad16f6a04e2062d25512aa1ccf2e337d/modules/web/src/resource/discovery/ingress/detail/component.spec.ts#L167

but I am having trouble understanding why. Suggestions are welcome. Thank you!

modules/web/src/common/components/objectmeta/component.ts
```diff
@@ -23,6 +26,8 @@ export class ObjectMetaComponent {
   @Input() initialized = false;
 
   private objectMeta_: ObjectMeta;
+  private readonly kdState_: KdStateService = GlobalServicesModule.injector.get(KdStateService);
```
maciaszczykm, line 29: Why not using constructor for that?

maciaszczykm: ^ @marcosdiez

marcosdiez: @maciaszczykm I tried
```
  private readonly kdState_: KdStateService;

  constructor() {
    this.kdState_ = GlobalServicesModule.injector.get(KdStateService);
  }
```

and it fails just like without the constructor.

```
mdiez@batman:~/code/dashboard/modules/web$ yarn test src/resource/discovery/ingress/detail/component.spec.ts
 FAIL  src/resource/discovery/ingress/detail/component.spec.ts
  ● Test suite failed to run

    src/common/services/pluginloader/clientloader.service.ts:21:25 - error TS2339: Property 'System' does not exist on type 'Window & typeof globalThis'.

    21 const systemJS = window.System;
                               ~~~~~~
    src/common/services/pluginloader/clientloader.service.ts:31:14 - error TS2339: Property 'define' does not exist on type 'Window & typeof globalThis'.

    31       window.define(externalKey, [], () => {
                    ~~~~~~
    src/common/services/pluginloader/clientloader.service.ts:51:52 - error TS7006: Parameter 'm' implicitly has an 'any' type.

    51       return systemJS.import(dependency.path).then(m => {
                                                          ~
    src/common/services/pluginloader/clientloader.service.ts:52:16 - error TS7015: Element implicitly has an 'any' type because index expression is not of type 'number'.

    52         window['define'](dep, [], () => m.default);
                      ~~~~~~~~
    src/common/services/pluginloader/clientloader.service.ts:57:48 - error TS7006: Parameter 'module' implicitly has an 'any' type.

    57       return systemJS.import(plugin.path).then(module => module.default.default);
                                                      ~~~~~~

Test Suites: 1 failed, 1 total
```

I am out of ideas. I'll try again next week.

maciaszczykm: @marcosdiez I mean like here: 

https://github.com/kubernetes/dashboard/blob/master/modules/web/src/common/components/actionbar/detailactions/exec/component.ts#L26

marcosdiez: @maciaszczykm it did not work either :(

maciaszczykm: @marcosdiez Sorry, I do not have time this and perhaps next month to work on the project. Once I will find some I will take a look on it.

modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -930,6 +930,10 @@
         <source>unset</source>
         <target>non défini</target>
       </trans-unit>
+      <trans-unit id="3715596725146409911" datatype="html">
+        <source>Owner</source>
+        <target state="new">Owner</target>
```
azro352, line 935:
```suggestion
        <target state="new">Propriétaire</target>
```

marcosdiez:

floreks:
It has been added in #9483

/close

Issue #6597, opened 2021-10-29
Kubernetes Dashboard and Metrics Scrapper on Apple Silicon (Arm64v8) starting AMD64 images


goafabric:
##### Environment
Kubernetes Dashboard 2.4.0
Apple Silicon M1
Docker Desktop

##### Steps to reproduce
Basically just
docker run kubernetesui/dashboard:v2.4.0

##### Observed result
It will allways pull and start an amd64 image.

##### Expected result
It should run an arm64(v8) image

##### Comments
The Problem can be observed directly in docker, as well inside of kubernetes in docker.
Instead of using an arm64 image, docker will always retrieve the wrong AMD64 image and running in emulation.
Even adding the correct DIGEST/SHA256 does not seem to help.

What I also observed is that the architecture type is "linux/arm64", in constrast for example to postgres
where the type is "linux/arm64/v8" and i think Docker expects the additional v8 .. not an expert here though

gmuth:
check supported architectures:
 
```
% docker manifest inspect kubernetesui/dashboard:v2.4.0|grep arch
            "architecture": "amd64",
            "architecture": "arm",
            "architecture": "arm64",
            "architecture": "ppc64le",
            "architecture": "s390x",
```

pull the image for arm64: 

```
% docker pull --platform inux/arm64 kubernetesui/dashboard:v2.4.0
v2.4.0: Pulling from kubernetesui/dashboard
2f554c6f6f2c: Pull complete 
476e0d029a85: Pull complete 
Digest: sha256:526850ae4ea9aba360e72b6df69fd3126b129d446efe83ac5250282b85f95b7f
Status: Downloaded newer image for kubernetesui/dashboard:v2.4.0
docker.io/kubernetesui/dashboard:v2.4.0
```

verify the platform: 

```
% docker inspect kubernetesui/dashboard:v2.4.0|grep Arch   
        "Architecture": "amd64",
```

I pulled arm64 - so why is it now amd64?
I think the images ARE amd64 images!
It looks like the multiplatform image building process has issues.
For some reason the digests don't correlate as expected. 
Maybe there is something wrong with the manifest.

floreks:
The real question here is if it runs correctly? I have recently tested it on the centos with `aarch` CPU and even though it pulled `amd64` arch image it started without any issues. I wouldn't spend too much time on this investigating if there is no real issue in running the application.

maciaszczykm:
[...]

PR #7491, opened 2022-09-30
Add support for multiarch docker build and deploy

floreks:
- Some cleanup work for the makefiles. Still, there are some leftovers that should be migrated/checked if they are needed.
- Added 2 new global make targets `make deploy` and `make deploy-dev` that will build binaries for all supported architectures and then build and push containers to configured container registries. We should now be able to easily deploy to multiple registries if needed.
- Updated `cross-build` to actually produce `os/arch` directories in the dist directory.
- Fixed an issue with the angular build since the `outputPath` argument was replaced by the `output-path` and the build was failing.

Tested all important targets: `make build`, `make build-cross`, `make run`, `make serve`, `make deploy-dev`.

## Supported architectures
We are cross-building go binaries and using `docker buildx` to build and push the images for the listed architectures.

Here are our test dev images: [dashboard-api](https://hub.docker.com/repository/registry-1.docker.io/kubernetesdashboarddev/dashboard-api/tags?page=1&ordering=last_updated), [dashboard-web](https://hub.docker.com/repository/registry-1.docker.io/kubernetesdashboarddev/dashboard-web/tags?page=1&ordering=last_updated)

**Note:** Do not try to run them manually as they need some kind of gateway that will "connect" them properly. For the local prod run we are using `nginx`. Our deployment files will be updated soon to support new architecture.
| OS     | ARCH    |
|--------|---------|
| darwin | amd64   |
| darwin | arm64   |
| linux  | amd64   |
| linux  | arm64   |
| linux  | arm     |
| linux  | ppc64le |
| linux  | s390x   |


goafabric:
@floreks 
after once in while i created this issue: https://github.com/kubernetes/dashboard/issues/6597
which fixed the dashboard .. metrics scrapper and server are still amd64 on apple silicon until dashboard 2.7.0 and its adherent helm charts

does this mean that this merge request will resolve this ?
and when and in which version will it be available ?

thank you

floreks:
@goafabric the metrics scraper has its' own repository but we can also manage it so I will definitely try to update the build system there to provide native images. Unfortunately, when it comes to the `metrics-server` it is out of our control. It is handled by other kubernetes sig and you would have to ask them to provide support. We are only using it as a general metrics provider.

goafabric:
@floreks ok thank you

PR #7513, opened 2022-10-31
Replace deprecated k8s registry references.

jmhbnz:
Problem:
Previously all of Kubernetes' image hosting has been out of gcr.io. There were significant egress costs associated with this when images were pulled from entities outside gcp.  Refer to https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io)

Solution:
As highlighted at KubeCon NA 2022 k8s infra SIG update, the replacement for k8s.gcr.io which is registry.k8s.io is now ready for mainstream use and the old k8s.gcr.io has been formally deprecated and projects are requested to migrate off it. This commit migrates all remaining references for kubernetes/dashboard to registry.k8s.io.

### DIFF
--- a/modules/api/pkg/resource/pod/detail_test.go
+++ b/modules/api/pkg/resource/pod/detail_test.go
@@ -161,7 +161,7 @@ func TestEvalEnvFrom(t *testing.T) {
 		{
 			container: v1.Container{
 				Name:  "echoserver",
-				Image: "k8s.gcr.io/echoserver",
+				Image: "registry.k8s.io/echoserver",
 				EnvFrom: []v1.EnvFromSource{
 					{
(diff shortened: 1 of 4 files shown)


modules/api/pkg/resource/pod/detail_test.go
```diff
@@ -161,7 +161,7 @@ func TestEvalEnvFrom(t *testing.T) {
 		{
 			container: v1.Container{
 				Name:  "echoserver",
-				Image: "k8s.gcr.io/echoserver",
+				Image: "registry.k8s.io/echoserver",
```
maciaszczykm, line 164: It doesn't have latest version available, but it's only for testing so we can keep it like it is.

PR #7514, opened 2022-11-01
feat: update Korean localization

jc01rho:
Update of the some missing translated texts.

### DIFF
--- a/modules/web/i18n/ko/messages.ko.xlf
+++ b/modules/web/i18n/ko/messages.ko.xlf
@@ -1554,7 +1554,7 @@
       </trans-unit>
       <trans-unit id="6119876613180878188" datatype="html">
         <source>Dropped Capabilities</source>
-        <target state="new">Dropped Capabilities</target>
+        <target state="new">제거된 기능(Dropped Capabilities)</target>
       </trans-unit>
       <trans-unit id="6140249581799344433" datatype="html">
         <source>Subresources</source>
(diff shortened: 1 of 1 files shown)


seokho-son:
Hi @shu-mutou 
Could you enable workflows for this PR from @jc01rho who is a first-time contributor ?

modules/web/i18n/ko/messages.ko.xlf
```diff
@@ -1554,7 +1554,7 @@
       </trans-unit>
       <trans-unit id="6119876613180878188" datatype="html">
         <source>Dropped Capabilities</source>
-        <target state="new">Dropped Capabilities</target>
+        <target state="new">제거된 기능(Dropped Capabilities)</target>
```
maciaszczykm, line 1557: Why did you keep english version?

jc01rho: I thought that there is no exact specific word for "Capabilities" ,  it looks better that for save it for prevent confuses.

seokho-son: Hi @jc01rho 
Thank you for the contribution ! :)

There are few Korean terms that provide original English term in k8s/dashboard in order to deliver exact meaning of the term. 

In case of "Capabilities", I know "Capabilities" is not exactly matched with "기능", but it is acceptable translation. Not to distort existing UI, I think we need to minimize long texts.

If my opinion is acceptable to you, please remove the English section. :)

jc01rho: sure. I'd agree.

seokho-son:
Thanks @jc01rho 

/lgtm

(a maintainer needs to approve running workflows)

PR #7521, opened 2022-11-08
Add support for NPM dep, and compatibility for "rm"

FrenchBen:
Current code base didn't build on cli bin that didn't support `rm` flags after the path.
* replace perl with grep for `make help`
* add npm dependency check (similar to golang version check)
*


floreks:
Nice! Thanks for this. There is still some work that needs to be done around tooling and dev setup but it's getting better.

License check failure is not related. 

/lgtm
/approve

floreks (approved):

maciaszczykm:
We are using 'yarn' instead od 'npm' now.

floreks:
> We are using 'yarn' instead od 'npm' now.

totally forgot 😆

FrenchBen:
Moved the check over to validate node version, ensured yarn was installed, and updated the docs to reflect the existence of the install tools target, along with the yarn setup.

```
$ make --directory modules/common/tools install
[tools] golang version requirements met
[tools] node version older than required v16.14.0+
make: *** [ensure-node] Error 1
$ nvm use lts/gallium
Now using node v16.15.0 (npm v8.5.5)
$ make --directory modules/common/tools install
[tools] golang version requirements met
[tools] node version requirements met
[tools] license-eye already exists
[tools] air already exists
[tools] golangci-lint already exists
[tools] client-gen already exists
[tools] yarn already exists
```

docs/developer/getting-started.md
```diff
@@ -6,16 +6,17 @@ This document describes how to setup your development environment.
...
-* Docker 1.13.1+ ([installation manual](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/))
-* Golang 1.19+ ([installation manual](https://golang.org/dl/))
-    * Dashboard uses `go mod` for go dependency management.
-* Node.js 16.14.2+ and yarn ([installation with nvm](https://github.com/creationix/nvm#usage))
+- Curl 7+ ([installation manual](https://curl.se/docs/install.html))
+- Git 2.13.2+ ([installation manual](https://git-scm.com/downloads))
+- Docker 1.13.1+ ([installation manual](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/))
+- Golang 1.19+ ([installation manual](https://golang.org/dl/))
+  - Dashboard uses `go mod` for go dependency management.
+- Node.js 16.14.2+ and yarn ([installation with nvm](https://github.com/creationix/nvm#usage))
```
maciaszczykm, line 14: We should remove part with `nvm` installation. @FrenchBen can you do it?

modules/common/tools/Makefile
```diff
@@ -62,3 +89,12 @@ ifndef CLIENT_GEN_BINARY
 else
 	@echo "[tools] client-gen already exists"
 endif
+
+.PHONY: install-yarn
+install-yarn:
+ifndef YARN_BINARY
+	@echo "[tools] downloading yarn..."
+	@npm install yarn
```
maciaszczykm, line 97: I'm not sure if we should force install like that. I'm using brew to install most things, for sure others have their preference too. Yarn is well-known tool, I'd just check if it's installed and if not print some feedback message.

floreks: Agree. We shouldn't install this for the user. We also do not install Go for them. Tools installation is only for Go dependencies and we make sure that Go is installed first.

floreks:
In general, I don't think we should install `yarn` automatically. It should be similar to Go/Node. The user has to do it. We can maybe point in the error to the link with instructions. That could be helpful and speed up the initial setup.

package.json
```diff
@@ -0,0 +1,6 @@
+{
```
floreks, line 1: There should be no `package.json` in the root.

modules/go.work.sum
```diff
@@ -0,0 +1,499 @@
+bitbucket.org/creachadair/shell v0.0.6 h1:reJflDbKqnlnqb4Oo2pQ1/BqmY/eCWcNGHrIUO8qIzc=
```
floreks, line 1: You might want to run `git rm --cached ...` on this file since it was added to the `.gitignore`.

FrenchBen: Cleaned up the files left behind

FrenchBen:
Updated the logic to show yarn as missing and needing to be install via:
```
corepack enable
corepack prepare yarn --activate
```
Npm will install the older version of yarn (yarnV1) by default, which isn't desirable.

floreks:
> Updated the logic to show yarn as missing and needing to be install via:
> 
> ```
> corepack enable
> corepack prepare yarn --activate
> ```
> 
> Npm will install the older version of yarn (yarnV1) by default, which isn't desirable.

In general, this does not matter to us since the "correct" version is uploaded to the [repo](https://github.com/kubernetes/dashboard/tree/master/modules/web/.yarn/releases). The user only needs to have yarn installed and it will use our local version anyway. It doesn't matter if it is v1 or v3.

docs/developer/getting-started.md
romdhanisam, line 14:
```suggestion
- Node.js 16.15.0+ and yarn ([installation with nvm](https://github.com/creationix/nvm#usage))
```
In line with `package.json` specifications

```json
 "engines": {
    "node": ">=16.15.0"
  }
```

maciaszczykm:
I've updated it to reflect current state.

/lgtm

Issue #7522, opened 2022-11-08
Makefile pattern breaks on OS other than Linux


FrenchBen:
### What happened?

When trying to target `run` on OS other than linux, the strategy isn't correct.
Using `make run` will target this target:

```
--compose: --ensure-certificates build
	@KUBECONFIG=$(KUBECONFIG) \
	SYSTEM_BANNER=$(SYSTEM_BANNER) \
	SYSTEM_BANNER_SEVERITY=$(SYSTEM_BANNER_SEVERITY) \
	ENABLE_SKIP_LOGIN=$(ENABLE_SKIP_LOGIN) \
	SIDECAR_HOST=$(SIDECAR_HOST) \
	TOKEN_TTL=$(TOKEN_TTL) \
	ARCH=$(ARCH) \
	OS=$(OS) \
	docker compose -f $(DOCKER_COMPOSE_PATH) --project-name=$(PROJECT_NAME) build
```
On a macOS, this will target a darwin build (no cross-build), which it will try to then bundle into the docker image. 
Additionally most `Dockerfile` are targeting a matching platform `FROM --platform=$BUILDPLATFORM`, when it should target a known platform (linux/amd64).

### What did you expect to happen?

I expect `make run` to work as it should, and launch the expected stack

### How can we reproduce it (as minimally and precisely as possible)?

Get macOS (m1 preferred to add more arch diff), try to execute `make run`

### Kubernetes Dashboard version

master

### Kubernetes version

all

### Dev environment

required node 16 & golang 1.19

floreks:
@maciaszczykm can you reproduce this? I think you said it was working for you.

maciaszczykm:
Here is partial output of `make run` on my Mac:
```
Building dashboard-web for darwin/arm64
Building dashboard-api for darwin/arm64
```
This part seems to work fine. However, it fails further due to lack of arm64 version of Ubuntu image.

Also, linking https://github.com/kubernetes/dashboard/pull/7537 as it has some needed improvements.

PR #7537, opened 2022-11-18
Cleanup & build system updates

floreks:
Some cleanup stuff and minor fixes for the build system.

After further investigation with @maciaszczykm we found out that there is no native support for the `darwin/arm64` arch in docker on macOS, therefore we've decided to drop the `darwin/arm64` and `darwin/amd64` from officially supported architectures. New docker image releases will be built for `linux/amd64 linux/arm64 linux/arm linux/ppc64le linux/s390x`.


maciaszczykm (approved):
Verified, now it works fine on M1 Mac.

floreks:
Fixes #7522

FrenchBen:
Do we still want https://github.com/kubernetes/dashboard/pull/7521 since this made similar changes? @floreks

floreks:
I think it is still valuable to have it merged.

PR #7544, opened 2022-11-25
Update deployment files to support new architecture

floreks:
```
TBD
```


README.md
```diff
@@ -31,14 +31,22 @@ Alternatively, you can install Dashboard using Helm as described at [`https://ar
 
 To access Dashboard from your local workstation you must create a secure channel to your Kubernetes cluster. Run the following command:
 
+[//]: # (To access Dashboard from your local workstation you forward port to your Kubernetes cluster. Run the following command:)
```
floreks, line 34: Can be uncommented once everything is finalized.

charts/recommended.yaml
```diff
@@ -167,63 +166,149 @@ subjects:
...
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: dashboard-web
+                port:
+                  name: http
+          - path: /api/v1
```
marcosdiez, line 254: Hi @floreks 

Here, path should point to `/api` else `/api/sockjs` for _exec into pod_  does not work.

marcosdiez:
From what I understood, 
the main dashboard container will be split into two:

one for the backend and one for the frontend.

the backend will receive /api/* ( or /api/v1/*)
the rest should go to the frontent

that means a router will be needed in front of it, so the setup will be more complex.

floreks:
@romdhanisam
> You can explain more about new architecture!

@marcosdiez is correct.
> From what I understood, the main dashboard container will be split into two:
> 
> one for the backend and one for the frontend.
> 
> the backend will receive /api/* ( or /api/v1/*) the rest should go to the frontent
> 
> that means a router will be needed in front of it, so the setup will be more complex.

We will utilize `cert-manager` and `ingress-nginx` to expose dashboard by default on a localhost domain. Since API and Web UI are in separate containers now we are using Ingress as a gateway that connects them together. In the future, we will probably use a dedicated gateway to leverage API and merge multiple API services into a single API. `cert-manager` is used to allow HTTPS access as it can generate self-signed cert used by ingress controller to expose the application.

```yaml
kind: Ingress
metadata:
  labels:
    k8s-app: kubernetes-dashboard
  annotations:
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    cert-manager.io/issuer: selfsigned
  name: kubernetes-dashboard
spec:
  ingressClassName: nginx
  tls:
  - hosts:
    - localhost
    secretName: kubernetes-dashboard-certs
  rules:
  - host: localhost
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: dashboard-web
            port:
              name: http
      - path: /api/v1
        pathType: Prefix
        backend:
          service:
            name: dashboard-api
            port:
              name: API
```

maciaszczykm:
@desaintmartin Hey, I was wondering if you would be willing to help us with getting this done as you have most experience with maintaining Dashboard chart. The base is already done but we are missing a lot of possible customizations. If you would like ot help then you can create a PR against this one (`update/deployment` branch on @floreks fork).

desaintmartin:
Sure ! I'm quite busy those days but I'll be happy to help, let me try to have a look in the next days!

maciaszczykm:
@desaintmartin Great! Thanks!

bhutkovskyysos:
any updates on this?

gergokee:
> any updates on this?

It's WIP, you can see the commits above your comment...

charts/kuberentes-dashboard.yaml
```diff
@@ -0,0 +1,440 @@
+# Copyright 2017 The Kubernetes Authors.
```
desaintmartin, line 1: Note: name of the file contains a typo

Idea: generate it regularly from the Helm Chart with default values.

charts/helm-chart/kubernetes-dashboard/ci/extra-manifests-values.yaml
```diff
@@ -1,21 +0,0 @@
-# Copyright 2017 The Kubernetes Authors.
```
desaintmartin, line 1: All the `ci/.*-values.yaml` files are tested by chart-testing, AND can act as valid use cases: I think we should keep/adapt them.

charts/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -12,288 +12,286 @@
...
+## Optional Cert Manager sub-chart configuration
+## Enable this if you don't already have cert-manager enabled on your cluster.
+cert-manager:
+  enabled: true
+  installCRDs: true
+
+## Optional Nginx Ingress sub-chart configuration
+## Enable this if you don't already have nginx-ingress enabled on your cluster.
+nginx:
+  enabled: true
```
desaintmartin, line 278: I wonder if nginx & cert-manager should be enabled by default: those are great demo values, but in production almost everybody will manage those two.

Would having sane but safe defaults for production in values.yaml, and having another "demo" values file on top of that, acceptable?

charts/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -12,288 +12,286 @@
...
-  # loadBalancerSourceRanges: []
-
-  # clusterIP: ""
-
-  ## A user-specified IP address for load balancer to use as External IP (if supported)
-  # loadBalancerIP:
-
-  ## Additional Kubernetes Dashboard Service annotations
+# General configuration shared across resources
+app:
```
desaintmartin, line 16: Aren't there some use cases where you want to differentiate some values between the api and the components? i.e should we avoid too much intelligence in here and just have two blocks?

charts/helm-chart/kubernetes-dashboard/templates/_helpers.tpl
```diff
@@ -49,30 +49,16 @@ Create chart name and version as used by the chart label.
 Common labels
 */}}
 {{- define "kubernetes-dashboard.labels" -}}
-app.kubernetes.io/name: {{ include "kubernetes-dashboard.name" . }}
```
desaintmartin, line 52: We could adapt those templates to be closer to what Helm recommends in the newest versions:

```
{{/*
Common labels
*/}}
{{- define "monchart.labels" -}}
helm.sh/chart: {{ include "monchart.chart" . }}
{{ include "monchart.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{/*
Selector labels
*/}}
{{- define "mychart.selectorLabels" -}}
app.kubernetes.io/name: {{ include "mychart.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}
```

It was not decided yet when the "old" template was written.

Would you like to keep your naming or should we use "full-upstream" mode given in my example and add `app.kubernetes.io/component` label to individual components?

charts/helm-chart/kubernetes-dashboard/templates/networking/post-install-ingress.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if eq .Values.app.ingress.issuer "selfsigned" -}}
```
desaintmartin, line 1: file naming proposition: `post-install-ingress-issuer`

desaintmartin (commented):
Here are some first thoughts. I can create a PR on top of this one to include some of my ideas if you want!

charts/kuberentes-dashboard.yaml
floreks: I was thinking about that but I am not sure if we can generate it reliably without custom post-processing.

charts/helm-chart/kubernetes-dashboard/ci/extra-manifests-values.yaml
floreks: We definitely can. We just didn't have time to do this. We were mostly focused on making it work.

charts/helm-chart/kubernetes-dashboard/values.yaml
floreks: We always wanted to provide a fully working installation by default. Technically even if the user already has both installed it would still work for both groups of users. On the other hand, disabling them by default won't work for users without those dependencies installed.

If there is a simple way to provide some default templates on install that could be selected and used by the users, it would be best. I am not that familiar with helm though.

charts/helm-chart/kubernetes-dashboard/values.yaml
floreks: I have tried to extract only the parts that should be the same across all components. There still might be some room for improvement, but we will indeed have more and more microservices in the future. A generic configuration for security and image handling might be a good idea.

charts/helm-chart/kubernetes-dashboard/templates/_helpers.tpl
floreks: Yes, but the component label might differ from what I understand. It could be `web`, `database`, `api`, etc. I have tried to move it directly to resources and use proper values there. It might still need to be improved. I tried to follow this: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

PR #7575, opened 2023-01-05
Add go.mod update check for tools and web

shu-mutou:
- modules/common/certificates/go.mod
- modules/common/tools/go.mod
- modules/web/go.mod

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,11 @@ updates:
       interval: "daily"
     open-pull-requests-limit: 5
   - package-ecosystem: "gomod"
+    directory: "modules/common/certificates/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 5
+  - package-ecosystem: "gomod"
     directory: "modules/common/tools/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -10,6 +10,16 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 5
+  - package-ecosystem: "gomod"
+    directory: "modules/common/tools/"
```
maciaszczykm, line 14: I'd change it to whole `modules/common` dir. We include K8s dependencies in `modules/api` anyway. In my opinion it makes sense as we will be notified about any releases.

shu-mutou: Unfortunately dependabot does not check recursively. So if we want to check whole `module/common` directory, we need to add one more directive for `module/common/certificates`.

maciaszczykm: Let's add it then.

shu-mutou: Got it. Done.

Issue #7577, opened 2023-01-05
The Delete a Resource UI should have an opt-in "delete now" checkbox


marcosdiez:
### What would you like to be added?

On the delete a resource dialog, there should be an opt-in checkbox called "delete now".

In case it's checked, the command should be equivalent to

kubectl delete --now -n kubernetes-dashboard pod kubernetes-dashboard-6df5cdb65-ch4m4

![image](https://user-images.githubusercontent.com/297498/211221232-e948712b-903d-43da-be32-37f196919344.png)

### Why is this needed?

Sometimes you need to make a resource disappear ASAP from the cluster so it can be respawned.

### Notes

Initially, I wanted the `--force` parameter, but after reading `kubectl`'s source code, I realized it's artificial and by itself does not do anything. Kubernetes itself does not have such concept. It turns out `--now` is the real deal.

Source:
* https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/delete/delete.go#L179
* https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions

Issue #7503, opened 2022-10-14
New option to force kill a pod


marcosdiez:
### What would you like to be added?

On the POD delete confirmation window, there should be a checkbox called "force"

If enabled, the command should change to

kubectl delete pods/THE_POD_NAME     --wait=false --timeout=0 --force=true

I can implement if you guys agree it's a good idea. Next month only, though.

### Why is this needed?

So we can easily delete crashed pods!

marcosdiez:
/remove-lifecycle rotten

Il 11 Febbraio 2023 17:18:51 GMT-03:00, Kubernetes Triage Robot ***@***.***> ha scritto:
>The Kubernetes project currently lacks enough active contributors to adequately respond to all issues.
>
>This bot triages un-triaged issues according to the following rules:
>- After 90d of inactivity, `lifecycle/stale` is applied
>- After 30d of inactivity since `lifecycle/stale` was applied, `lifecycle/rotten` is applied
>- After 30d of inactivity since `lifecycle/rotten` was applied, the issue is closed
>
>You can:
>- Mark this issue as fresh with `/remove-lifecycle rotten`
>- Close this issue with `/close`
>- Offer to help out with [Issue Triage][1]
>
>Please send feedback to sig-contributor-experience at [kubernetes/community](https://github.com/kubernetes/community).
>
>/lifecycle rotten
>
>[1]: https://www.kubernetes.dev/docs/guide/issue-triage/
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/kubernetes/dashboard/issues/7503#issuecomment-1426868476
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #7583, opened 2023-01-09
The Delete a Resource UI should have an opt-in "delete now" checkbox

marcosdiez:
Closes https://github.com/kubernetes/dashboard/issues/7577
Closes https://github.com/kubernetes/dashboard/issues/7503


floreks:
I'd use some kind of `opts ...Option` param here to be future compatible and support both `deleteNow` and `force` options similar to kubectl. I think the only difference is that `force` sets the grace period to 0 instead of 1.

modules/web/src/common/dialogs/deleteresource/template.html
```diff
@@ -22,14 +22,24 @@
     <span *ngIf="data.objectMeta.namespace"> &nbsp;in namespace <i>{{ data.objectMeta.namespace }}</i> </span>
     <span>?</span>
   </ng-container>
+  <ng-container>
+    <div class="kd-equivalent-block">
+      <mat-checkbox color="primary"
+                    [(ngModel)]="isDeleteNowChecked"
+                    name="isDeleteNowChecked"
+                    matInput
+                    i18n>Delete now (sets delete grace period to 1 second instead of the default, which is usually 30 seconds)</mat-checkbox>
```
floreks, line 31: I'd add the description to the info icon with tooltip.

marcosdiez: Happy new year @floreks !

I am not sure it looks better:

image


image

Also, as I mentioned, I did not put both `force` and `now` because they are basically the same, except that `force` is more a `kubectl` thing that a `kubernetes` thing.

marcosdiez: I think this is fine:

image

marcosdiez: Or how about this ?
image

we could even add the `force`, although I don't personally see any advantage.

marcosdiez: or

image

modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -542,6 +542,10 @@
         <source>Add Namespace</source>
         <target state="new">Add Namespace</target>
       </trans-unit>
+      <trans-unit id="2640307640152196302" datatype="html">
+        <source>Delete now (sets delete grace period to 1 second instead of the default, which is usually 30 seconds)</source>
+        <target state="new">Delete now (sets delete grace period to 1 second instead of the default, which is usually 30 seconds)</target>
```
azro352, line 547:
```suggestion
        <target state="new">Supprimer immédiatement (Défini la période de grâce de suppression à 1 seconde au lieu de la valeur par défaut, habituellement de 30 secondes)</target>
```

marcosdiez:

maciaszczykm:
/lgtm

Let's add force later.

modules/web/src/common/dialogs/deleteresource/template.html
aaemula: @marcosdiez if I did not clicks neither Delete nor Cancel but escape the dialog box, the delete request still send to the backend.

PR #7584, opened 2023-01-11
Fix Scale Resource UI

marcosdiez:
The callback UI for the `Scale a Resource` panel was wrong, therefore it was not showing the correct number of `actual replicas`.

This PR fixed that. And since I am only deleting code, I am technically increasing our test coverage :)

image


floreks (approved):
/lgtm

PR #7585, opened 2023-01-11
Fix for situation when ingress has no http rules

duckhawk:
If there is ingress without http rules exists in namespace:

```
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/server-snippet: |
      return 301 https://my.another.pretty.domain$request_uri;
  name: my-pretty-ingress
  namespace: my-pretty-namespace
spec:
  ingressClassName: nginx
  rules:
  - host: my.pretty.domain
  tls:
  - hosts:
    - my.pretty.domain
```

There is panic in https://github.com/kubernetes/dashboard/blob/master/modules/api/pkg/resource/ingress/filter.go#L39, because rule.IngressRuleValue is nil.

This PR fix it.

### DIFF
--- a/modules/api/pkg/resource/ingress/filter.go
+++ b/modules/api/pkg/resource/ingress/filter.go
@@ -36,6 +36,9 @@ func ingressMatchesServiceName(ingress networkingv1.Ingress, serviceName string)
 	}
 
 	for _, rule := range spec.Rules {
+		if rule.IngressRuleValue == nil {
+			continue
+		}
 		if rule.IngressRuleValue.HTTP == nil {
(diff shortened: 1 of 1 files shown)


maciaszczykm:
@duckhawk Thanks for the fix. Can you mark PR as ready for review and sign the CLA?

duckhawk:
@jeefy @shu-mutou @floreks can somebody approve this PR please?

maciaszczykm:
@duckhawk It's already approved. Checks are not passing and it is blocking merge. You need to fix them.

PR #7587, opened 2023-01-12
Set new container image in deployments (former Deployment UI)

### LINKED ISSUE #7576: UI for kubectl set image DEPLOYMENT_NAME CONTAINER_NAME=IMAGE_URL

On the deployment pane, we should have a new button next to scale resource, restart, edit resource and delete resource called "deploy new image" or "change container image" or something like that.

It should open a dialog box similar to the "scale a resource" button, except that it should ask for a new container image.

To make it even better, it should, by checking the existing replica set, show the previous versions in a table.

something like this:

image

The buttons should be links. When one clicks, it it populates the container image to the text box.

### Why is this needed?

Sometimes one needs to deploy/revert something FAST.

That's a 3 click revert. Hard to beat :)

ps: I am implementing this and should finish soon.

marcosdiez:
closes #7576 

image
---
image
---
image
---

For this to work, I had to an unexpected API verb:

`apiV1Ws.GET("/deployment/{namespace}/{deployment}/allreplicasets")`, which lists all replica sets from a deployment so we can show the previous deployment images

Also, on the replicasets, I had to add a two new maps called:
* ContainerImagesMap
* InitContainerImagesMap

their key is the container name, their value is the container image, again, so I can list them.

That does add redundancy, for currently all objects have two lists: ContainerNames and InitContainerNames that don't have the container names.
Should I open a separate PR to refactor so that we just use ContainerNames  and InitContainerNames but as maps instead of lists ?

On the new Set Image dialog, maybe we should have link with the image image instead of buttons ?

I explicitly ignrored init container images, because since they can have the same name as container images, I though that would clutter the UI.

Also, I could have made a table instead of changing one image at a time, but I can't see any real use case for that.

Suggestions are welcome


modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -1402,6 +1414,10 @@
         <source>Egress Rules</source>
         <target state="new">Egress Rules</target>
       </trans-unit>
+      <trans-unit id="5643209019947694032" datatype="html">
+        <source>container image</source>
+        <target state="new">container image</target>
```
azro352, line 1419:
```suggestion
        <target state="new">image du conteneur</target>
```

modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -1504,6 +1520,10 @@
         <source>Exit Code</source>
         <target state="new">Exit Code</target>
       </trans-unit>
+      <trans-unit id="5966254721262198145" datatype="html">
+        <source>Previous Image Name (click to use)</source>
+        <target state="new">Previous Image Name (click to use)</target>
```
azro352, line 1525:
```suggestion
        <target state="new">Nom de l'image précédente (cliquer pour utiliser)</target>
```

modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -1964,6 +1992,10 @@
         <source>NAMESPACES</source>
         <target>ESPACES DE NOMS</target>
       </trans-unit>
+      <trans-unit id="7574968363077451152" datatype="html">
+        <source>Previous Deployment Date</source>
+        <target state="new">Previous Deployment Date</target>
```
azro352, line 1997:
```suggestion
        <target state="new">Date du déploiement précédent</target>
```

modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -1884,6 +1908,10 @@
         <source>Init Containers</source>
         <target>Conteneurs d&apos;initilisation</target>
       </trans-unit>
+      <trans-unit id="7286769344436324609" datatype="html">
+        <source>Set Container Image</source>
+        <target state="new">Set Container Image</target>
```
azro352, line 1913:
```suggestion
        <target state="new">Définir l'image d'un conteneur</target>
```

modules/web/i18n/fr/messages.fr.xlf
```diff
@@ -1600,6 +1620,10 @@
         <source>TCP Socket</source>
         <target state="new">TCP Socket</target>
       </trans-unit>
+      <trans-unit id="6406804063343897494" datatype="html">
+        <source>A container image will be replaced on <x id="INTERPOLATION" equiv-text="&quot;\uFFFD0\uFFFD&quot;"/> <x id="INTERPOLATION_1" equiv-text="&quot;\uFFFD1\uFFFD&quot;"/>.</source>
+        <target state="new">A container image will be replaced on <x id="INTERPOLATION" equiv-text="&quot;\uFFFD0\uFFFD&quot;"/> <x id="INTERPOLATION_1" equiv-text="&quot;\uFFFD1\uFFFD&quot;"/>.</target>
```
azro352, line 1625:
```suggestion
        <target state="new">L'image d'un conteneur sera remplacée sur le <x id="INTERPOLATION" equiv-text="&quot;\uFFFD0\uFFFD&quot;"/> <x id="INTERPOLATION_1" equiv-text="&quot;\uFFFD1\uFFFD&quot;"/>.</target>
```

azro352 (changes requested):
Add french translations

marcosdiez:

marcosdiez:
test

PR #7621, opened 2023-02-15
Fix build in development container

shu-mutou:
Remove `modules/web/node_modules` directory before `yarn`, to do clean install as a default behavior.


maciaszczykm (approved):
/lgtm

PR #7658, opened 2023-03-15
Add flag to prevent ingress

jcpunk:
To help ensure someone doesn't expose the dashboard, I'd like to set a network policy to block ingress.  I figured while I was at it, I'd make it more generic so this could be used more broadly.

Folks who want to be sure no one comes along and adds an ingress to the dashboard now have an easy toggle to lock it down.


maciaszczykm (approved):
/lgtm

PR #7688, opened 2023-03-30
[helm chart] make sure the ServiceMonitor API is available before installing one

asherf:
I think this is a good practice to have since it avoid errors when the same chart with the same values file is installed into multiple clusters, some of them might not have the relevant CRDs.
I have seen other chart use this pattern: https://github.com/fluent/helm-charts/blob/d808425a73580891a77e115561fe3ea2d2e24347/charts/fluent-bit/templates/servicemonitor.yaml#LL1-L1C106
and I think it is a good one to follow.

### DIFF
--- a/charts/helm-chart/kubernetes-dashboard/Chart.yaml
+++ b/charts/helm-chart/kubernetes-dashboard/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 6.0.6
+version: 6.1.0
 appVersion: "v2.7.0"
 description: General-purpose web UI for Kubernetes clusters
 keywords:
--- a/charts/helm-chart/kubernetes-dashboard/templates/servicemonitor.yaml
+++ b/charts/helm-chart/kubernetes-dashboard/templates/servicemonitor.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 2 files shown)


desaintmartin:
Thanks!

Since this is a disabled-by-default feature, that would actually silently breaking the user expectations and not install a ServiceMonitor despite the user explicitly asking to enable it.

As a heavy user of dozens of Helm charts, I prefer when something explicitly breaks.

I would prefer to have something that check and raise than this.

What do you think?

asherf:
> Thanks!
> 
> Since this is a disabled-by-default feature, that would actually silently breaking the user expectations and not install a ServiceMonitor despite the user explicitly asking to enable it.
> 
> As a heavy user of dozens of Helm charts, I prefer when something explicitly breaks.
> 
> I would prefer to have something that check and raise than this.
> 
> What do you think?


I have seen this pattern in many other charts of 3rd party SW, but not all of them, some charts only check the values file and don't check if the API is available.

Our use case, which I am not sure that is super common, is that we want to enable the service monitor in our values file. (we use the same values file for multiple clusters we install the dashboard on). so we want to enable the service monitor in the values file and not have it installed on clusters that don't have the prometheus operator. 

I do agree that using a fail fast method instead of a silent failure which this change will cause has its advantages.
Personally, I am a big fan of the fail-fast approach.
I would say, that when if a user tries to enable the service monitor on a cluster that doesn't have prometheus operator running on it, is not a very common case, considering how basic/common or part of the lower level infra is prometheus operator (for organizations that use it)....

So I am conflicted and can see both sides of this argument.
And I am perfectly fine if maintainers of this chart reject this change.

PR #7729, opened 2023-05-11
fix: use dl.k8s.io, not kubernetes-release bucket

rjsadow:
#### What type of PR is this?
/kind cleanup
/kind deprecation

#### What this PR does / why we need it:
There are still references to https://storage.googleapis.com/kubernetes-release instead of https://dl.k8s.io/

dl.k8s.io is the correct advertised download host and will eventually move to be fastly shielding a fully community-owned bucket

ref: https://github.com/kubernetes/k8s.io/issues/2396


shu-mutou:
/lgtm
/approve
Nice catch! Thanks!!

PR #7756, opened 2023-06-02
Specify bit size to int32

shu-mutou:
Default bit size for `atoi` is int64, but using `int32(c)` for the result can cause a violation.
To parse string to int32, switch from `atoi` to `ParseInt` with specifying bit size.

Fixing security alert from CodeQL.


shu-mutou:
This will fix the security alert.
See https://github.com/kubernetes/dashboard/security/code-scanning/196

maciaszczykm (approved):
/lgtm

PR #7760, opened 2023-06-04
create access token by secret which bound the service account

huweihuang:
The token creation described in the current document is temporary and cannot be queried. Therefore, adding the secret method to create tokens that can be queried.

### DIFF
--- a/docs/user/access-control/creating-sample-user.md
+++ b/docs/user/access-control/creating-sample-user.md
@@ -38,7 +38,7 @@ subjects:
   namespace: kubernetes-dashboard
 ```
 
-## Getting a Bearer Token
+## Getting a Bearer Token for ServiceAccount
 
 Now we need to find the token we can use to log in. Execute the following command:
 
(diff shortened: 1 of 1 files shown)


docs/user/access-control/creating-sample-user.md
```diff
@@ -40,10 +40,10 @@ subjects:
 
 ## Getting a Bearer Token
 
-Now we need to find the token we can use to log in. Execute the following command:
+Now we need to find the token we can use to log in. Execute the following command to create a token with a custom expiration:
 
 ```shell
-kubectl -n kubernetes-dashboard create token admin-user
+kubectl -n kubernetes-dashboard create token admin-user --duration 8760h
```
maciaszczykm, line 46: Setting it to one year is not good security practice.

huweihuang: Do I need to keep this parameter `--duration` so that the people know that an expiration time can be specified, and then set the time parameter to 24 hours?

maciaszczykm: I don't think we have to mention that. Everyone can look for options using `kubectl create token --help`.

huweihuang: ok, i remove it

huweihuang: i have already remove `--duration`, and the img `/images/overview.png` is too old, is there needs to use dashboard-ui.png insteads of it?

maciaszczykm:
@huweihuang I've added some updates.

/lgtm

Issue #7817, opened 2023-06-06
Switch `DEPNEDABOT_HOME` to `DEPENDABOT` for dependency check


shu-mutou:
### What should be cleaned up or changed?

As a flag for judging whether Dependabot is running, instead of the existing DEPENDABOT_HOME, a new environment variable called DEPENDABOT will be introduced, so we will support it when it is released.

https://github.com/dependabot/dependabot-core/pull/7407

### Why is this needed?

Since the current coping method is a kind of hack, we should move to an officially recognized method.

shu-mutou:
https://github.com/dependabot/dependabot-core/pull/7407 was merged yesterday.
I'll check releases in following, If it is released I'll switch to use env `DEPENDABOT`.
https://github.com/dependabot/dependabot-core/releases

PR #7930, opened 2023-06-23
Use `DEPENDABOT` instead of `DEPENDABOT_HOME`

shu-mutou:
`DEPENDABOT` env was introduced for presenting that dependabot is running. So switch to this from `DEPENDABOT_HOME` that we use now.

Fixes: #7817

### DIFF
--- a/modules/web/hack/postinstall.sh
+++ b/modules/web/hack/postinstall.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if [ -z "${DEPENDABOT_HOME}" ]; then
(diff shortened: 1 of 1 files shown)


shu-mutou:
/hold
until new version of dependabot released

shu-mutou:
I'll check following (v0.220.0 is latest now) and will cancel `/hold` if new dependabot released.
https://github.com/dependabot/dependabot-core/releases

shu-mutou:
/close
The cause is unknown, but `cncf-cla: no` was added and I can't remove it, so I will close it once.
`cncf-cla: yes` was given in another newly created PR, so I'll try to recreate this as well.

PR #7984, opened 2023-07-05
Use `DEPENDABOT` instead of `DEPENDABOT_HOME`

shu-mutou:
`DEPENDABOT` env was introduced for presenting that dependabot is running. So switch to this from `DEPENDABOT_HOME` that we use now.

Fixes: https://github.com/kubernetes/dashboard/issues/7817

### DIFF
--- a/modules/web/hack/postinstall.sh
+++ b/modules/web/hack/postinstall.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


shu-mutou:
/hold
until new version (>0.220.0) of dependabot released in following:
https://github.com/dependabot/dependabot-core/releases

shu-mutou:
I confirmed this release tag has my commits.

shu-mutou:
/hold cancel
New version was released :tada:

PR #8018, opened 2023-07-08
Fix ingress nginx dependency condition

chamburr:
This fixes ingress-nginx's dependency condition such that when `nginx.enabled` is set to `false` in values, the subchart is actually disabled.

On a separate note, I installed the new version of kubernetes-dashboard without ingress-nginx at all. I'm using traefik for my ingress and the dashboard is working normally. Is the dependency therefore really mandatory, and perhaps a note could be added if not?

### DIFF
--- a/charts/helm-chart/.helm-chart-testing.yaml
+++ b/charts/helm-chart/.helm-chart-testing.yaml
@@ -17,7 +17,7 @@ chart-dirs:
 chart-repos:
   - metrics-server=https://kubernetes-sigs.github.io/metrics-server/
   - cert-manager=https://charts.jetstack.io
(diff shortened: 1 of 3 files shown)


floreks:
Thanks for testing and fixing that. Please bump chart minor version otherwise checks will not pass.

floreks (approved):
/lgtm

floreks:
@chamburr that's a good point about traefik. In general any ingress controller or API gateway should work here. I have used nginx as it was the most popular in my opinion and kubernetes org officially provides a chart for it. In the future we might switch to something else if we migrate to GraphQL API.

PR #8031, opened 2023-07-10
fix(dashboard): adjust api web order in ingress to prevent 404

leyinetwork:
This commit adjusts the order of api web in ingress configuration to avoid path matching overlapping which causes 404 errors.


maciaszczykm:
@leyinetwork You need to sign the CLA to continue.

PR #8046, opened 2023-07-12
add ability to disable ingress creation

enneitex:
This PR adds the ability to disable ingress creation when deploying the helm chart. We are using ingressRoute CR provided by traefik so the ingress is not needed, neither the issuer.


maciaszczykm:
/lgtm

Leaving @floreks for final approval.

floreks (approved):
/lgtm

maciaszczykm:
Thanks @enneitex for contribution!

PR #8103, opened 2023-07-21
fix: multiple helm chart improvements

### LINKED ISSUE #8048: [Helm] Chart updates
We will use this issue to track all approved changes to the helm chart. It will be easier to find and see what needs to be changed and what was already changed. Feel free to discuss further changes here and we can update it along the way. 

- [x] Allow `cert-manager` and `nginx` to be optional
  - [x] Allow disabling `cert-manager` annotations and providing secret with custom certificate
- [x] Allow disabling Ingress resource completely to allow usage of custom proxies (#8046)
- [x] Update default configuration and use internal nginx service by default
- [x] Fix NetworkPolicy configuration (#8045)
- [ ] Update documentation to inform that all chart dependencies (`cert-manager`, `nginx`, `metrics-server`) can be replaced with other alternatives and are only required by the default deployment.

Ref: #8038, #8043, #8044

floreks:
Fixes #8048

- `cert-manager` can be fully skipped now. Set `cert-manager.enabled=false` and `app.ingress.issuer.scope=disabled` to fully disable it. Note that `secretName` with the certificate must still be configured to point to the correct secret with a valid certificate.
- `nginx` can be fully disabled. Set `nginx.enabled=false` and `app.ingress.enabled=false` to fully disable it. Note that you have to provide a custom configuration that matches what we do with ingress otherwise, Dashboard will not work.
- updated default `nginx` configuration to use `ClusterIP` nginx service and custom `internal-nginx` ingress class.
- updated `NetworkPolicy` config to match correct ingress port names
- updated helm chart versioning to start with alpha. When releasing the helm chart we will remove previous "official" releases without alpha suffixes.

TODO:
- [x] Update documentation to cover new changes and inform users about our default configuration and that our dependencies can be replaced.


floreks:
@mlbiam does this cover everything mentioned in #8038 or did I miss anything?

mlbiam:
@floreks i think so, i'll get it running my lab

charts/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -85,14 +85,15 @@ app:
     # https://localhost:8443
     - localhost
     # - kubernetes.dashboard.domain.com
-    ingressClassName: nginx
+    ingressClassName: internal-nginx
```
mukulgit123, line 88: @floreks - Private ingress class name can be anything for the users. e.g. "nginx-private" or "private-nginx" or "nginx-pvt" etc. How would the helm chart behave in such scenarios?

floreks: This would mean that you already have nginx installed and you should not use our defaults, but rather override this with your custom value.

PR #8106, opened 2023-07-21
Update Angular to v16

### LINKED ISSUE #8101: Bump Angular and switch to Vite builds
### What should be cleaned up or changed?

See https://blog.angular.io/angular-v16-is-here-4d7a28ec680d.

### Why is this needed?

To keep dependencies up to date and speed up builds.

maciaszczykm:
Closes #8101.


floreks (approved):
/lgtm

shu-mutou:
I also checked this in my development environment and this works fine.
But I also faced error as same as CI. So I tried to increase timeout value for cypress, it was not effective. I'm keeping on investigating this error now.

shu-mutou:
I found the Air starts to rebuild while cypress running. We may need to configure `.air.toml`, I guess.

shu-mutou:
I confirmed that the number of the Air rebuilds is same as the number of screenshots (includes its folder).

shu-mutou:
Also, I confirmed `touch modules/web/cypress/screenshots/test` triggered rebuild backend server that was started by `make serve`.

shu-mutou:
I tried some changes in  https://github.com/kubernetes/dashboard/blob/master/modules/api/.air.toml#L21 , but the `.air.toml` file didn't seem to work as documented, not also as we expected. :disappointed: 
It's time to finish work. See you!

maciaszczykm:
Thanks @shu-mutou for taking care of e2e.

Issue #7470, opened 2022-09-23
Use renovate instead of dependabot


maciaszczykm:
### What would you like to be added?

Switch from dependabot to renovate (see: https://docs.renovatebot.com/getting-started/installing-onboarding/).

### Why is this needed?

Dependabot doesn't support Yarn v2 which we are using (see: https://github.com/dependabot/dependabot-core/issues/1297).

shu-mutou:
Now dependabot seems to support yarn v2.
https://github.com/dependabot/dependabot-core/issues/1297#issuecomment-1285998196

Would you like to migrate to renovate?

maciaszczykm:
@shu-mutou We can keep using dependabot as request for renovate is still not approved at org level.

shu-mutou:
@maciaszczykm Got it. I would like to investigate restoring npm dependency checking and yarn support.

shu-mutou:
Re-enabled npm update check with #7574

maciaszczykm:
Let's close it for now. If we will get approval for renovate we can reconsider. Most important thing is fixed already.

PR #8134, opened 2023-07-28
Ignore updates for angular staffs

shu-mutou:
Angular staffs should be manually updated at once.
So, add a setting to ignore updating angular staffs by dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,4 +29,7 @@ updates:
     directory: "modules/web/"
     schedule:
       interval: "daily"
-    open-pull-requests-limit: 30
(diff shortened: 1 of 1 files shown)


floreks:
We should switch to renovate since it can bump multiple packages at once and is more flexible. Right now we are getting way too many dependency PRs

shu-mutou:
@floreks 
For now, I added a setting to ignore Angular which is too much and needs to be updated manually.

maciaszczykm:
/lgtm

It's good for now. We can reopen https://github.com/kubernetes/dashboard/issues/7470 but we will need to get approval for Renovate bot as we cannot do it on our own.

PR #8149, opened 2023-08-03
Allow disabling ssl redirect on Helm chart (nginx) Ingress

### LINKED ISSUE #8148: Allow disabling ssl redirect on Helm chart (nginx) Ingress

Currently the annotation `nginx.ingress.kubernetes.io/ssl-redirect: true` is added by default to the `Ingress` resource created by the Helm chart. It would be nice to allow overriding this to allow access via HTTP instead of HTTPS.

### Why is this needed?

Easier access for e.g. local development environments, where HTTPS is not always needed.

jnoordsij:
Fixes #8148


floreks:
Will be covered by #8735 since a list of default annotations will get bigger. It will be possible to simply disable all of them and provide custom ones with `app.ingress.annotations`.

/close

PR #8173, opened 2023-08-14
fix: Correct path in serviceMonitor

jlec:
According to

https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.Endpoint

the path is part of the endpoint list


maciaszczykm (approved):
/lgtm

PR #8222, opened 2023-08-25
[helm] Allow to configure additional RBAC cluster role-bindings

jkroepke:
For example, give reader permissions to custom CRDs


jkroepke:
Okay

jkroepke:
still relevant

PR #8223, opened 2023-08-25
[helm] Enable issuer, if scope is set to default

jkroepke:
if `.app.ingress.issuer.scope` is set to disabled, it make no sense to create the issuer


jkroepke:
Merged from main (to fix CI issues)

PR #8225, opened 2023-08-25
[helm] Add tpl function to config and ingress for supporting umbrella charts

jkroepke:
Having the tpl function in place here is a real benefit, since a central helm value can be used for ingress hostname across all subscharts.

Various other helm charts like kube-prometheus-stack, grafana, blackbox_exporter, grafana loki are supporting this style, too.


desaintmartin:
Thanks! Could you bump chart version? From experience, this is indeed beneficial, but can be hard to properly setup.

jkroepke:
Rebased and bumped

desaintmartin:
@floreks
- what is the policy now? Is it okay to merge / bump?
- Could you trigger tests?

jkroepke:
How we can proceed here?

jkroepke:
Merged from main (to fix CI issues)

PR #8227, opened 2023-08-26
Bump k8s modules to 0.28.1

shu-mutou:
Also, exclude dependency check for k8s modules, since k8s modules should be at once.
New version of kindest/node is not released, so that will be updated later.


maciaszczykm (approved):
/lgtm

Issue #8211, opened 2023-08-24
Provide context to `GetPluginSource`


artemmarkaryan:
### What should be cleaned up or changed?

Now `GetPluginSource` uses `context.TODO()` stub to call both dashboard and k8s api client. 

```go
// GetPluginSource has the logic to get the actual plugin source code from information in Plugin.Spec
func GetPluginSource(client pluginclientset.Interface, k8sClient kubernetes.Interface, ns string, name string) ([]byte, error) {
	plugin, err := client.DashboardV1alpha1().Plugins(ns).Get(context.TODO(), name, v1.GetOptions{})
	...
	cfgMap, err := k8sClient.CoreV1().ConfigMaps(ns).Get(context.TODO(), plugin.Spec.Source.ConfigMapRef.Name, v1.GetOptions{})
	...
}
```

`GetPluginSource` is only called in `servePluginSource` handler, which has a request context. Context of the request, as far as I can judge, could be passed to `GetPluginSource` as argument.

### Why is this needed?

It is useful to control application's state through context at least not to waste resources on canceled requests.

Sajiyah-Salat:
Hello @artemmarkaryan I am new to this repo it would be good if you can provide me context like where change is required. and how or any details about this issue. It would be highly helpful for me to work on this. Thank you!

artemmarkaryan:
Hi @Sajiyah-Salat! 

Thanks for your willingness to help but I actually was going to fix this myself. 

Actually i came across several place in dashboard codebase where context.TODO() is used, although it is quite possible to provide real context from request. If you'd like, you may try to find such places yourself. I'm not sure that fixes in other places should be related to this issue. It seems more reasonable to create a separate issue for each.

artemmarkaryan:
@maciaszczykm, hi! There are several other places, where `context.TODO()` is used instead of request's context. Is it okay to fix them as a part of this issue or I should create an issue for each? 

Also, by any chance, could these `context.TODO()` have been left intentionally? Maybe they should not be replaced? I couldn't find a reason for that, but maybe there's some point i'm missing.

maciaszczykm:
>  Is it okay to fix them as a part of this issue or I should create an issue for each?
@artemmarkaryan Sure.

> Also, by any chance, could these context.TODO() have been left intentionally? Maybe they should not be replaced? I couldn't find a reason for that, but maybe there's some point i'm missing.
I don't think there is specific reason. Let's do PR and review then.

[...]

PR #8319, opened 2023-09-25
Replace todo ctx with request ctx in GetPluginSource

vlasebian:
**What type of PR is this?**

/kind cleanup

This PR replaces the TODO context with the context of the request in GetPluginSource. If the context is canceled, processing the request stops, thus improving resource utilisation.

Part of #8211.

**Special notes for your reviewer**:

I'm not sure if someone is currently working on this at the moment. I will make a draft PR and continue if it's still needed.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

### DIFF
--- a/modules/api/pkg/plugin/detail_test.go
+++ b/modules/api/pkg/plugin/detail_test.go
@@ -41,12 +41,12 @@ func TestGetPluginSource(t *testing.T) {
 	pcs := fakePluginClientset.NewSimpleClientset()
 	cs := fakeK8sClient.NewSimpleClientset()
 
-	_, err := GetPluginSource(pcs, cs, ns, pluginName)
+	_, err := GetPluginSource(context.Background(), pcs, cs, ns, pluginName)
 	if err == nil {
 		t.Errorf("error 'plugins.dashboard.k8s.io \"%s\" not found' did not occur", pluginName)
 	}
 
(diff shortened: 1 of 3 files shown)


modules/api/pkg/plugin/detail_test.go
```diff
@@ -41,7 +41,7 @@ func TestGetPluginSource(t *testing.T) {
 	pcs := fakePluginClientset.NewSimpleClientset()
 	cs := fakeK8sClient.NewSimpleClientset()
 
-	_, err := GetPluginSource(pcs, cs, ns, pluginName)
+	_, err := GetPluginSource(context.TODO(), pcs, cs, ns, pluginName)
```
maciaszczykm, line 44: Use `context.Background()` instead in this file.

vlasebian: @maciaszczykm I replaced `context.TODO()` with `context.Background()` in the test.

In the issue it was also mentioned that there are more instances of `context.TODO()`. I searched the files and in the plugin package there is also `GetPluginList` that uses `context.TODO()`. Should I replace that one too?

vlasebian:
Hey, @maciaszczykm! When you have a moment, could you please review this PR? Thank you very much :D

maciaszczykm (approved):
/lgtm

PR #8396, opened 2023-10-19
Update nodejs install method

shu-mutou:
GPG key is needed for installation now.
`npm` does not seems to be installed with `nodejs` packages, so it is specified.


shu-mutou:
/close
EasyCLA seems to have problem... 
I'd recreate PR by former experience...

PR #8372, opened 2023-10-15
using takeUntilDestroyed after upgrading to Angular 16

romdhanisam:

Issue #5623, opened 2020-10-12
Add lint rule to better manage subscription cancellation


floreks:
In example namespace list request still happens after the user logs out due to wrongly used `takeUntil` pipe. We should add lint rules to check such things and also check if subscriptions are properly destroyed overall on changing views.

`src/app/frontend/common/components/namespace/component.ts`
```typescript
this.namespaceUpdate_
      .pipe(takeUntil(this.unsubscribe_))
      .pipe(startWith({}))
      .pipe(switchMap(() => this.namespace_.get(this.endpoint_.list())))
      .subscribe(...);
```

romdhanisam:
Related subject, [Avoid takeUntil subscription leaks (no-unsafe-takeuntil)](https://github.com/cartant/eslint-plugin-rxjs/blob/main/docs/rules/no-unsafe-takeuntil.md#avoid-takeuntil-subscription-leaks-no-unsafe-takeuntil)

PR #8410, opened 2023-10-21
feat(Test) Fixed namespace-settings-global setting - added tests

romdhanisam:
Navigating out of setting causes an error.

Screenshot 2023-10-21 at 17 40 54

Version **v3.0.0-alpha0-233-g7a059ac99**


romdhanisam:
#5623

romdhanisam:
The use of takeUntilDestroyed after upgrading to Angular 16 https://github.com/kubernetes/dashboard/pull/8372 shows us that some component are not properly destroyed overall on changing views.

Unit tests can help demonstrate that.

Issue #8496, opened 2023-11-20
[Multi-Tenant Auth] Ability to customize the namespaces dropdown behavior/contents


mecampbellsoup:
### What would you like to be added?

We run a multi-tenant k8s cluster and use namespaces to segregate customers from one another.

I've setup the k8s dashboard UI to proxy auth to the logged-in user and use its credentials. This part is working as expected.

The remaining glaring issue with rolling this out to customers is that I don't have the ability to configure or customize the behavior of the namespaces dropdown list:

<img width="1420" alt="image" src="https://github.com/kubernetes/dashboard/assets/2043821/6484f8e3-07b1-42c2-93af-43a76ec7b3bf">

`default` is confusing here and that namespace is most definitely not something our customers will be able to access.

We worked with kubeapps team and [implemented a fallback mechanism](https://github.com/vmware-tanzu/kubeapps/blob/main/chart/kubeapps/values.yaml#L1614-L1626) there for something similar - would something like this be possible in k8s dashboard UI? Or is there some existing mechanism by which we can configure or even _disable_ the namespaces dropdown?

### Why is this needed?

For better multi-tenant cluster support, we need the ability to configure some sort of fallback so that our customers can rely on the UI's namespaces dropdown to contain the list of namespaces they are able to access.

This feature is necessary for multi-tenant clusters in which isolation is done via namespaces (i.e. customer A has namespace `tenant-a`, customer B has namespace `tenant-b`, etc., and neither has much cluster-scoped RBAC as it is a shared cluster).

Here is how the api namespace endpoint currently responds in e.g. a shared cluster where users' namespaces are segregated (and those users do not have list namespaces RBAC in the cluster):

```json
{
 "listMeta": {
  "totalItems": 0
 },
 "namespaces": [],
 "errors": [
  {
   "ErrStatus": {
    "metadata": {},
    "status": "Failure",
    "message": "namespaces is forbidden: User \"mcampbell+dev+11-16-2023@coreweave.com\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope",
    "reason": "Forbidden",
    "details": {
     "kind": "namespaces"
    },
    "code": 403
   }
  }
 ]
}
```
With this response, the user sees the following namespace dropdown:
![image](https://github.com/kubernetes/dashboard/assets/2043821/1a7bebd2-a855-48c9-9fca-86ef9c795d58)

[...]

PR #8526, opened 2023-12-04
Use proxied trusted request header to filter namespaces list in multi-tenant clusters

mecampbellsoup:
Resolves https://github.com/kubernetes/dashboard/issues/8496.

Also modifies the behavior of Impersonation such that a request header `Authorization: Bearer foo` token is not required. We need this change because we want to combine in-cluster configuration with trusted `Impersonate*` headers set by our authenticating reverse proxy which sits in front of dashboard api, and it seemed silly to have to hardcode inside my ingress the value of the dashboard api service account's token.


modules/api/pkg/client/manager.go
```diff
@@ -358,25 +363,38 @@ func (self *clientManager) buildCmdConfig(authInfo *api.AuthInfo, cfg *rest.Conf
...
 	// Authorization header will be more important than our token
 	token := self.extractTokenFromHeader(authHeader)
-	if len(token) > 0 {
-
+	if len(token) > 0 || len(impersonationHeader) > 0 {
+		if token == "" {
+			// If Impersonate* header(s) provided, but Authorization token is missing,
+			// fallback to in-cluster config by reading the serviceaccount token from the filesystem
+			var dat, _ = os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+			token = string(dat)
```
mecampbellsoup, line 375: @floreks is there a way to activate Impersonation when no explicit `Authorization: Bearer` token has been included in the request, i.e. "fallback" to using in-cluster config if the reverse proxy has added Impersonation headers only?

mecampbellsoup: Note: would also like to cache the result of `os.ReadFile` so as to not read from the filesystem on every request.

modules/api/pkg/client/manager.go
```diff
@@ -358,25 +363,38 @@ func (self *clientManager) buildCmdConfig(authInfo *api.AuthInfo, cfg *rest.Conf
...
+		}
 		authInfo := &api.AuthInfo{Token: token}
 
 		if len(impersonationHeader) > 0 {
-			//there's an impersonation header, lets make sure to add it
+			// Request includes an impersonation header, let's make sure to add it
 			authInfo.Impersonate = impersonationHeader
 
-			//Check for impersonated groups
-			if groupsImpersonationHeader := req.Request.Header["Impersonate-Group"]; len(groupsImpersonationHeader) > 0 {
```
mecampbellsoup, line 375: I changed this to support both of the following formats:

```
Impersonate-Group: one,two,three

-- as well as...

Impersonate-Group:one
Impersonate-Group:two
Impersonate-Group:three
```

mecampbellsoup:
@desaintmartin @maciaszczykm do you guys expect to have a chance to review this in the near future? Thanks 😄

mecampbellsoup:
> @desaintmartin @maciaszczykm do you guys expect to have a chance to review this in the near future? Thanks 😄

Hey guys! Just following up - what can I do to get some 👀 and feedback on this?! Thank you 🙏

PR #8570, opened 2023-12-24
Update config.mk with supported architectures

vivekkoya:


floreks:
Any news that docker images are being run natively now on MacOS instead of being virtualized as a linux VM?

I don't think this has changed yet.

maciaszczykm:
We tried to set it up recently, but there is limited number of images including Go and having support for all archs. If you have any good idea how to set it up with current pipeline please open new pull request. It would be welcomed.

PR #8582, opened 2023-12-30
Misguide the timezone for logs

Milanc35:
Currently, pod start time and end time are showing in local (browser/system) timezone,
This word misguide the time in UTC. So, I propose to remove UTC word. 
Please verify and merge pull


Milanc35:
lgtm

Issue #7633, opened 2023-02-26
CRD does not show the output like kubectl


anannaya:
### What happened?

We use flinkApplication operator , which shows the output status using kubectl like this 
`
k get flinkApplication -A
NAMESPACE        NAME                                           PHASE           CLUSTER HEALTH   JOB HEALTH   JOB RESTARTS   AGE
agentproc        agent-snapshot-metrics                         Running         Green            Green        2              4d10h
agentproc        agent-statistics-metric-processor              Running         Green            Green        1              11d
agentproc        agent-summary-processor                        Running         Green            Green        1              4d10h
agentproc        site-snapshot-metrics                          Running         Green            Green        2              4d10h
agentproc        team-snapshot-metrics                          Running         Yellow           Yellow       4              430d
agentproc        usage-metric-processor                         Running         Green            Green        2              73d
agentstorage     raw-data-agent-storage-agent-activity          Running         Green            Green        3              4d10h
agentstorage     raw-data-agent-storage-agent-error             Running         Yellow           Yellow       11             533d
agentstorage     raw-data-agent-storage-agent-event             Running         Yellow           Yellow       11             533d
agentstorage     raw-data-agent-storage-agent-statistics        Running         Yellow           Yellow       11             533d
agentstorage     refined-data-agent-storage-agent-state         Running         Green            Green        1              21d
billing          agent-pstn-snapshot-metrics                                                                                 2d12h
`

But in a kubernetes dashboad does not shows these information , Is there any way to customize in UI?
<img width="1696" alt="image" src="https://user-images.githubusercontent.com/19487932/221398115-e449f601-d25f-46d7-80c9-91571aa6beaa.png">

### What did you expect to happen?

Does not show the proper output in UI,

### How can we reproduce it (as minimally and precisely as possible)?

Any CRD can be tested .

### Kubernetes Dashboard version

v2.7.0

### Kubernetes version

1.23

floreks:
This is unfortunately not supported right now. We do not have any logic to read custom print columns as described in the k8s docs. It would be an enhancement.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns

anannaya:
@floreks  Thanks for responding .

ashutosh887:
@floreks I would like to work on this

anannaya:
@ashutosh887 i can test and confirm.once the changes are ready

RafikFarhad:
[...]

PR #8583, opened 2023-12-31
Additional printer columns for crd

RafikFarhad:
Closes: #7633

Hello,
This is my first PR. Help me find where I need to write tests for this addition.

------
The CRD objects are shown with `additionalPrinterColumns`.

image

### DIFF
--- a/modules/api/pkg/resource/customresourcedefinition/types/types.go
+++ b/modules/api/pkg/resource/customresourcedefinition/types/types.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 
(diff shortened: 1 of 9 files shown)


maciaszczykm (commented):
Good idea. Looks good to me. We would need to test it.

@floreks Any thoughts?

RafikFarhad:
@maciaszczykm I have attempted to write a test, but I am blocked. Can you please suggest some guidelines here?

modules/api/pkg/resource/customresourcedefinition/v1/detail_test.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2017 The Kubernetes Authors.
```
Jont828, line 1:
```suggestion
// Copyright 2023 The Kubernetes Authors.
```

modules/api/pkg/resource/customresourcedefinition/types/types.go
```diff
@@ -59,31 +59,48 @@ type CustomResourceDefinitionDetail struct {
 	Errors []error `json:"errors"`
 }
 
+type AdditionalPrinterColumn struct {
+	Name     string `json:"name"`
+	Type     string `json:"type,omitempty"`
+	Priority int32  `json:"priority,omitempty"`
+	JSONPath string `json:"jsonPath"`
```
Jont828, line 66: Maybe we can add a description field as shown [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns) that can pop up when you hover over the column name?

PR #8613, opened 2024-01-11
[helm] add if for ingressClassName

### LINKED ISSUE #8593: [helm] cannot use default ingressClass

using default ingressClass by default

### Why is this needed?

`internal-nginx` is not always the default `IngressClassName`.
```
ingressClassName: internal-nginx
```

similar issue: https://github.com/prometheus-community/helm-charts/issues/3823

jmnote:
- fixes #8593

using default ingressClass by default.

AS-IS
```
$ helm template . --set app.ingress.ingressClassName=hello | grep ingressClassName:
  ingressClassName: hello
$ helm template . --set app.ingress.ingressClassName="" | grep ingressClassName:
  ingressClassName: 
$ helm template . | grep ingressClassName:
  ingressClassName: internal-nginx
$
```

TO-BE
```
$ helm template . --set app.ingress.ingressClassName=hello | grep ingressClassName:
  ingressClassName: hello
$ helm template . --set app.ingress.ingressClassName="" | grep ingressClassName:
$ helm template . | grep ingressClassName:
$
```

If ingressClassName is not specified, the ingressClassName block is omitted. This allows us to use the cluster's default IngressClass.

similar case(helm chart):
https://github.com/prometheus-community/helm-charts/blob/alertmanager-1.7.0/charts/alertmanager/templates/ingress.yaml#L16-L18

k8s docs:
https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class

### DIFF
--- a/charts/helm-chart/kubernetes-dashboard/values.yaml
+++ b/charts/helm-chart/kubernetes-dashboard/values.yaml
@@ -85,7 +85,7 @@ app:
     # https://localhost:8443
     - localhost
     # - kubernetes.dashboard.domain.com
-    ingressClassName: internal-nginx
+    ingressClassName: ""
     pathType: ImplementationSpecific
     secretName: kubernetes-dashboard-certs
     issuer:
--- a/charts/helm-chart/kubernetes-dashboard/templates/networking/ingress.yaml
+++ b/charts/helm-chart/kubernetes-dashboard/templates/networking/ingress.yaml
@@ -40,7 +40,9 @@ metadata:
     {{- end }}
   name: {{ template "kubernetes-dashboard.name" . }}
 spec:
+  {{- if .Values.app.ingress.ingressClassName }}
   ingressClassName: {{ .Values.app.ingress.ingressClassName }}
+  {{- end }}
   {{- if .Values.app.ingress.hosts }}
   tls:
     - hosts:


charts/helm-chart/kubernetes-dashboard/values.yaml
```diff
@@ -85,7 +85,7 @@ app:
     # https://localhost:8443
     - localhost
     # - kubernetes.dashboard.domain.com
-    ingressClassName: internal-nginx
+    ingressClassName: ""
```
maciaszczykm, line 88: If `ingressClassName` is ommited, , a [default Ingress class](https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class) should be defined.

jmnote: I confirmed that it can be applied normally even if the value of `ingressClassName:` is omitted in the YAML file as follows.

```
kind: Ingress
metadata:
  name: example
spec:
  ingressClassName:
  rules:
    - host: example.com
      http:
        paths:
          - pathType: ImplementationSpecific
            backend:
              service:
                name: release-name-kubernetes-dashboard-web
                port:
                  name: web
```

But I've rarely seen values left blank in Kubernetes manifest or Helm Chart values yamls.

floreks:
It will be covered by #8735 in [here](https://github.com/kubernetes/dashboard/pull/8735/files#diff-1b4f557f3d3e21667c87094d3b32902a21950212cee66094ecd940ff4061a20bR92).

/close

PR #8766, opened 2024-03-08
feat(api): add full proxy mode to api module

floreks:
Ensure that there are no in-cluster client calls when running in proxy mode.


maciaszczykm (approved):
/lgtm

Issue #8767, opened 2024-03-09
unable to login with "kubectl proxy" way


gergokee:
### What happened?

im trying to use the kubectl proxy method of logging in through the url below, but after login i receive **Invalid credentials provided**
http://localhost:8443/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard-kong-proxy:443/proxy/

### What did you expect to happen?

I expected to be able to login.

### How can we reproduce it (as minimally and precisely as possible)?

`kubectl proxy -p=8443`
and then try:
`http://localhost:8443/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard-kong-proxy:443/proxy/`

### Anything else we need to know?
The "port-forward" way is working fine, but if this is not an "out-of-the box" supported way of login anymore, is there a way to still enable it ?

### What browsers are you seeing the problem on?

Chrome

### Kubernetes Dashboard version

7.1.1

### Kubernetes version

1.28

floreks:
That is most likely an issue with how kubectl proxy handles headers passthrough. 

I will check to confirm, but I remember this being an issue in the past.

floreks:
Ye, that seems to be exactly the case: https://github.com/kubernetes/kubernetes/issues/38775

kubectl proxy strips authorization header from the request and we are no longer using custom JWE token for auth. We are now using standard JWT token and standard authorization bearer token header.

There is nothing we can do about that. It is just how kubectl proxy works.

/close

gergokee:
@floreks how about:
https://github.com/kubernetes/dashboard/pull/8768

i've seen this solution in for example https://github.com/kubesphere/kubesphere

PR #8768, opened 2024-03-10
Fix kubectl stripping auth header

gergokee:
Fixes https://github.com/kubernetes/dashboard/issues/8767


floreks:
I wouldn't want to use workarounds only to support kubectl proxy. This is intended behavior as described here https://github.com/kubernetes/kubernetes/issues/38775#issuecomment-277915961. There are many different ways Dashboard can be exposed and accessed.

Using port-forward, service nodeport/loadbalancer, ingress will work. I don't see any benefits here and potentially it could be a security vulnerability as described in the linked issue.

PR #8770, opened 2024-03-11
fix(ingress): use web as backend when kong is not enabled

sjoukedv:
It's possible to disable the kong proxy, but the ingress backend is hardcoded to the kong proxy


sjoukedv:
@desaintmartin do you have time for a quick review?

shu-mutou:
@sjoukedv we fixed ci, so could you rebase to current master branch?

sjoukedv:
> @sjoukedv we fixed ci, so could you rebase to current master branch?

@shu-mutou done, don't know where the invalid commit messages are coming from. I can create an entire new branch if it doesn't work with this one.

floreks:
I think it would be best if you apply your changes to the latest master and create a new branch.

sjoukedv:
Closed in favor of https://github.com/kubernetes/dashboard/pull/8853

PR #8827, opened 2024-03-23
ci: Fix CodeQL check and ignore codecov token error

shu-mutou:
As of now, we need to manually setup golang for 1.22.
https://github.com/github/codeql/issues/15647#issuecomment-2003768106

Also, ignore Codecov token not found error.
PR from dependabot causes an error, so comment this out until the problem is resolved.
See https://github.com/codecov/codecov-action/issues/1274

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,9 @@ jobs:
       - uses: codecov/codecov-action@v4.1.0
         with:
           directory: ./.tmp
-          fail_ci_if_error: true
+          #fail_ci_if_error: true
+          # PR from dependabot causes an error, so comment this out until the problem is resolved.
+          # See https://github.com/codecov/codecov-action/issues/1274
           token: ${{ secrets.CODECOV_TOKEN }}
 
 # TODO: needs to be fixed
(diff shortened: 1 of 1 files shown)


shu-mutou:
/retitle ci: Fix CodeQL check and ignore codecov token error

.github/workflows/ci.yml
```diff
@@ -132,7 +136,9 @@ jobs:
       - uses: codecov/codecov-action@v4.1.0
         with:
           directory: ./.tmp
-          fail_ci_if_error: true
+          #fail_ci_if_error: true
+          # PR from dependabot causes an error, so comment this out until the problem is resolved.
+          # See https://github.com/codecov/codecov-action/issues/1274
```
maciaszczykm, line 141: Let's remove it completely. `fail_ci_if_error` defaults to false and we should keep it that way. Having reports uploaded each time is not that important.

maciaszczykm:
@shu-mutou Let's change this one little thing and it's good to merge.

shu-mutou:
@maciaszczykm PTAL again, thanks!

maciaszczykm (approved):
/lgtm

Issue #8820, opened 2024-03-20
Unable to login after update to chart 7.1.2


vittoriocanilli:
### What happened?

On my AKS cluster the authentication for the dashboard is managed with OAuth2 proxy and an Azure app. It worked correct till the version `7.0.0-alpha1` of the dashboard's Helm chart.

Unfortunately after the update to version `7.1.2` I land on this screen after the login:
![Screenshot 2024-03-20 at 15 57 40](https://github.com/kubernetes/dashboard/assets/6400531/4be97332-c989-4e3e-aeb3-5330d66406c4)
I guess that I would get the same error with any version greater than `7.0.0-alpha1`, i.e. after the **auth** and **kong** modules were added

### What did you expect to happen?

I would expect to be logged successfully on the dashboard. I have checked similar issues here and I have opened the `/api/v1/me` endpoint of my dashboard, which simply returns this:
`{"authenticated":true}`
If I try to open the dashboard again after that I have checked the `/api/v1/me` endpoint on the browser, then I am logged in the dashboard. So I guess that there is some step missing.

### How can we reproduce it (as minimally and precisely as possible)?

Here are the `app` values which could are relevant for the authentication:
```yaml
app:
  ingress:
    enabled: true
    ingressClassName: nginx
    issuer:
      name: letsencrypt-production
      scope: cluster
    tls:
      secretName: le-kubernetes-dashboard-tls
    annotations:
      nginx.ingress.kubernetes.io/auth-signin: "https://login.mycluster.azure.mycompany.com/oauth2/start?rd=$scheme://$host$request_uri"
      nginx.ingress.kubernetes.io/auth-url: "https://login.mycluster.azure.mycompany.com/oauth2/auth"
      nginx.ingress.kubernetes.io/proxy-buffer-size: "64k"
      nginx.ingress.kubernetes.io/auth-response-headers: "X-Auth-Request-Access-Token"
      nginx.ingress.kubernetes.io/configuration-snippet: |
        auth_request_set $token $upstream_http_x_auth_request_access_token;
        proxy_set_header Authorization "Bearer $token";
    hosts:
      - dashboard.mycluster.azure.mycompany.com
  settings:
    global:
      clusterName: "mycluster.azure.mycompany.com"
```

### Anything else we need to know?

For the `auth` module I have simply set the node selector, while all other values are set to the default:
```yaml
auth:
  nodeSelector:
    node-selector-key: system
```
For the `kong` module I did not overwrite any value.

[...]

Issue #8813, opened 2024-03-18
Login page not consistently skipped with bearer token


mrlhansen:
### What happened?

I have a reverse proxy (Nginx) in front of the dashboard where I pass the Bearer token for authentication. The first time I load the dashboard the login page is shown, but after refreshing the page a couple of times, the login page is skipped and the dashboard loads correctly.

I can consistently reproduce this with different browsers, but it's more pronounced in Firefox. Sometimes the login page is skipped completely, but most of the time the login page will be shown and I have to refresh the page a couple of times before the dashboard loads. Once it loads, there are no problems.

I can even see that `/api/v1/me` is called when the login page is shown and the endpoint returns `authenticated: true`.

This was never a problem with the older versions of the Dashboard, but it is present with the new 7.x release.

### What did you expect to happen?

I expect the login page to never be shown.

### How can we reproduce it (as minimally and precisely as possible)?

I use [nginxinc](https://docs.nginx.com/nginx-ingress-controller/) ingress for Kubernetes, with the following annotation to pass the Bearer token.
```
    nginx.org/location-snippets: |
      proxy_set_header Authorization "Bearer <token>";
      proxy_pass_header Authorization;
```

### What browsers are you seeing the problem on?

Chrome, Safari, Firefox

### Kubernetes Dashboard version

7.x

### Kubernetes version

1.27.9

jaynis:
I am experiencing exactly the same issue. In my case I use a OIDC proxy in front of the dashboard and it almost always needs an additional page refresh after the OIDC redirect in order to skip the login. The `/api/v1/me` endpoint in the auth service indicates no problems which means my token seems to be valid all the time.

PR #8840, opened 2024-03-28
fix: login guard and me service initialization

floreks:
### Auth
- Improved `me` endpoint response to use the correct response code for unauthorized requests.

### Web
- Improved application initialization to ensure that user service is initialized before the application. Login/Auth guards should now work correctly.

Fixes https://github.com/kubernetes/dashboard/issues/8820
Fixes https://github.com/kubernetes/dashboard/issues/8813


maciaszczykm (approved):
/lgtm

PR #8853, opened 2024-04-02
fix(ingress): use web as backend when kong is disabled

sjoukedv:
Replaces https://github.com/kubernetes/dashboard/pull/8770

### DIFF
--- a/charts/kubernetes-dashboard/templates/networking/ingress.yaml
+++ b/charts/kubernetes-dashboard/templates/networking/ingress.yaml
@@ -69,10 +69,19 @@ spec:
         {{- end }}
             pathType: {{ $.Values.app.ingress.pathType }}
             backend:
+              {{- if $.Values.kong.enabled }}
               service:
                 name: {{ template "kong.fullname" (index $.Subcharts "kong") }}-proxy
                 port:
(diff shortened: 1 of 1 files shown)


floreks:
Have you tested it? I am not sure if it would actually work.

floreks:
Kong is used as an API gateway. Simply forwarding all the traffic to the web container won't work since it won't have access to the API and Auth containers. Kong is there for a reason. It ensures that the web and API are all hosted under a single path and when the web tries to access `api/v1` endpoint relative to its location, it will be able to.

Kong can be disabled only if you know what you are doing and can replace our API gateway with something else. Ingress should be used in front of kong, not instead of it.

sjoukedv:
> Kong is used as an API gateway. Simply forwarding all the traffic to the web container won't work since it won't have access to the API and Auth containers. Kong is there for a reason. It ensures that the web and API are all hosted under a single path and when the web tries to access `api/v1` endpoint relative to its location, it will be able to.
> 
> Kong can be disabled only if you know what you are doing and can replace our API gateway with something else. Ingress should be used in front of kong, not instead of it.

Okay, thank you for reviewing. The dashboard stopped working for us behind Treafik, that's why I wanted to get rid of kong. I'll close this MR.

PR #8869, opened 2024-04-03
feat(chart): add automountServiceAccountToken to deployment specs in Helm chart

jnoordsij:
This MR adds the option to set the value of `automountServiceAccountToken` for deployments.

Moreover, it uses `false` as default value for the `auth` deployment, given that it has no service account and not automounting the token of the `default` account is considered best practice by some people, including me.


floreks (approved):

floreks:
/override "CI / Helm / Lint and test"

PR #8870, opened 2024-04-03
feat(chart): allow setting priorityClassName on deployments

jnoordsij:
This MR adds the option to set the value of `priorityClassName` for deployments.


floreks (approved):
/lgtm

PR #8971, opened 2024-04-26
feat(chart): Disable Kong Manager by default

jnoordsij:
With the current defaults, a type `NodePort` service with name `kubernetes-dashboard-kong-manager` is created. Given that Kong Manager is a GUI for managing Kong resources that is as far as I'm aware not in any way required for functionality of this chart, I think disabling it is a more reasonable default.


floreks (approved):
/lgtm

PR #8973, opened 2024-04-28
fix: escape the dialog should not trigger delete resource

jenting:
Escape the delete resource dialog should not trigger delete resource.

### DIFF
--- a/modules/web/src/common/dialogs/deleteresource/dialog.ts
+++ b/modules/web/src/common/dialogs/deleteresource/dialog.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


floreks (changes requested):
You are basically allowing undefined to be recognized as cancel. I think this should be fixed directly in the dialog to make it return doCancel as it should in the first place.

jenting:
> You are basically allowing undefined to be recognized as cancel. I think this should be fixed directly in the dialog to make it return cancelDelete as it should in the first place.

@floreks 
Thanks for your feedback. I updated the code to handle escape event and backdrop clicks in the dialog to make it return cancelDelete. Let me know if it's better than before.

jenting:
@floreks any chance to review this? Thanks.

PR #9138, opened 2024-06-12
docs: warn users about https requirement

darmbrust:


floreks (approved):
Thanks! That's helpful. We might want to try to detect it and display a more suitable error message.

floreks:
@darmbrust could you sing the CLA?

PR #9199, opened 2024-06-27
docs: Added `kubectl proxy` access method

gerardnico:
I searched a little bit for it. So I make the proxy services URL available for anybody to see.


floreks (approved):
/lgtm

PR #9212, opened 2024-07-02
fix: add missing port name for metrics-scraper service

elohmeier:


floreks:
I'd have to recheck but IIRC this was intentional. I think there was an issue accessing metrics scraper service in code when it was using a named port.

https://github.com/kubernetes/dashboard/blob/9a476333d4f392a5b85d80a51c8149b4c2ec011d/modules/api/pkg/integration/metric/sidecar/restclient.go#L48-L55

floreks:
Yep, we can't change that as connecting to the service via API service proxy does not work with named ports.

/close

PR #9226, opened 2024-07-10
fix: add divisor to the deployment manifets

cHiv0rz:
# Context

Although you can apply the manifest without the divisor, it will set a `divisor: 0` in the cluster. If you uses tools like `ArgoCD`, it could cause reconciliation problems with it.

# Solution proposed

Just added a divisor with the most common way to put the resources, `1m` for the CPU and `1Mi` for the memory


floreks (approved):
lgtm

blackliner:
@cHiv0rz I guess you need to bump to latest master for the failing linter? Looking forward to your PR getting merged!
image

floreks:
No need to bump version. We'll do it when we will release a new helm chart version. We should update helm lint CI check.

PR #9317, opened 2024-07-31
Allow helm users to disable kong and use the ingress

mkhpalm:
This change allows helm users to disable kong and enable the app.ingress without getting an error. 

Fixes: https://github.com/kubernetes/dashboard/issues/8972

https://kubernetes.io/docs/concepts/services-networking/ingress/#examples

Example values:
```yaml
app:
  ingress:
    enabled: true
    hosts:
    - example.tld
kong:
  enabled: false
```

### DIFF
--- a/charts/kubernetes-dashboard/templates/_helpers.tpl
+++ b/charts/kubernetes-dashboard/templates/_helpers.tpl
@@ -106,3 +106,63 @@ private.key: {{ randBytes 256 | b64enc | quote }}
 {{- fail "value of .Values.app.ingress.issuer.scope must be one of [default, cluster, disabled]"}}
 {{- end -}}
 {{- end -}}
+
+{{- define "kubernetes-dashboard.ingress.paths" -}}
+paths:
+{{- if eq .Values.app.mode "dashboard" }}
+  - pathType: Exact
+    path: {{ .Values.app.ingress.path }}api/v1/login
+    backend:
+      service:
+        name: {{ template "kubernetes-dashboard.name" . }}-{{ .Values.auth.role }}
+        port:
+        {{- with (index .Values.auth.containers.ports 0) }}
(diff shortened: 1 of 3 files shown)


floreks:
That's not something we'd want to support directly. The reason is that ingress has its limitations and can't support multiple backend protocols in a single ingress instance, i.e. gRPC and HTTP(s) (rest or gql). We know for sure that we will be migrating our backend architecture further and this configuration will eventually stop working. It would require multiple ingresses and that's not something we want to promote. That is why we want to use API Gateway which will handle everything and expose a single public-facing API.

floreks:
Gateway API is not the same as API Gateway. We need API Gateway to stitch together our services and expose them as a single API.

dmaphy:
> Gateway API is not the same as API Gateway. We need API Gateway to stitch together our services and expose them as a single API.

And for this, you could define a Gateway and let the user decide for the tool to use for the Gateway (Traefik, Istio, $younameit) which possibly already exists on a users cluster instead of enforcing the user to pull Kong in.

floreks:
> > Gateway API is not the same as API Gateway. We need API Gateway to stitch together our services and expose them as a single API.
> 
> And for this, you could define a Gateway and let the user decide for the tool to use for the Gateway (Traefik, Istio, $younameit) which possibly already exists on a users cluster instead of enforcing the user to pull Kong in.

That's why it is already possible to disable kong via helm values and replace our gateway configuration for services with configuration matching your gateway of choice.

Unfortunately in that case we cannot guarantee that the gateway configuration will not change. Not every gateway i.e. can support multiple protocols and different types of apis being exposed under the same subpath.

When selecting our gateway of choice we have had a list of requirements and kong checked all the boxes for us. Most of the other gateways we have considered did not.

PR #9320, opened 2024-07-31
feat(chart): remove kong configuration if kong disabled in values

bear-san:
This change aims to remove unnecessary configuration if Kong subchart disabled by Helm Chart values.
I have inserted condition to ConfigMap, that requires `.Values.kong.enabled` set to `true`.


mkhpalm:
FWIW: The same change is added to this PR which allows kong to be disabled and still get an ingress. https://github.com/kubernetes/dashboard/pull/9317

floreks (approved):
Good catch. It does not make sense to create this CM when kong is disabled. Thanks!

PR #9380, opened 2024-08-16
feat(chart): add supported settings to values.yaml based on settings.go

krichter722:
Based on https://github.com/kubernetes/dashboard/blob/master/modules/web/pkg/settings/settings.go. The values all seem to work, they're just missing in the "docs" (`values.yaml` is kind of docs for a chart).


floreks (approved):
/lgtm

PR #9436, opened 2024-09-05
Update values.yaml

siva9941166:
I am facing the issue with kube ui in my cluster. so i have removed the values AAAA.


floreks:
We want to keep it enabled by default. You can override this value during chart installation as required by your environment.

/close

PR #9486, opened 2024-09-26
Fix incorrect resource divisors in Helm templates

### LINKED ISSUE #9485: Incorrect `GOMAXPROCS` and `GOMEMLIMIT` being set on latest Helm chart
### What happened?

After upgrading from 7.5.0 to the latest Helm chart version (7.6.1), the `GOMAXPROCS` and `GOMEMLIMIT` values set in my pod are no longer correct.

### What did you expect to happen?

The values to remain as before.

### How can we reproduce it (as minimally and precisely as possible)?

Set CPU limits and/or memory limits using the latest Helm chart version (7.6.1).

E.g.: setting CPU limit of `250m` and a memory limit of `200Mi` on the `api` containers with values
```
  api:
    containers:
      resources:
        limits:
          cpu: 250m
          memory: 200Mi
```
will result in a pod with the incorrect values `GOMAXPROCS=250` (should be 1) and `GOMEMLIMIT=200` (should be 209715200).

### Anything else we need to know?

This regression was caused by #9226, which introduced a `divisor` value to the chart of `1m` for CPU and `1Mi`, which results in strictly different behavior than before. The Kubernetes default value is (plain) `1` for both; see also [the docs](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/):

> The `divisor` field is optional and has the default value of `1`. A divisor of 1 means cores for `cpu` resources, or bytes for `memory` resources.

Side note: I'm unsure why this change was introduced, as in my eyes it looks like this is a bug with `ArgoCD` defaulting to an invalid value for an unset field, rather than an issue with the chart.

### What browsers are you seeing the problem on?

_No response_

### Kubernetes Dashboard version

N/A

### Kubernetes version

N/A

### Dev environment

_No response_

jnoordsij:
Fixes #9485

### DIFF
--- a/charts/kubernetes-dashboard/templates/deployments/api.yaml
+++ b/charts/kubernetes-dashboard/templates/deployments/api.yaml
@@ -74,14 +74,14 @@ spec:
               valueFrom:
                 resourceFieldRef:
                   resource: limits.cpu
-                  divisor: 1m
+                  divisor: 1
             {{- end }}
             {{- if .Values.api.containers.resources.limits.memory }}
             - name: GOMEMLIMIT
               valueFrom:
                 resourceFieldRef:
                   resource: limits.memory
-                  divisor: 1Mi
+                  divisor: 1
             {{- end }}
 
           {{- with .Values.api.containers.env }}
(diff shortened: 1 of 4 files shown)


floreks (approved):
/lgtm

floreks:
Thanks for the fix. I will do a release tomorrow.

Issue #8894, opened 2024-04-09
Allow hiding "All namespaces" in namespaces dropdown list


chri4774:
### What would you like to be added?

We would like to be able to globally hide the "All namespaces" list item in the namespace drop-down list. Either through global settings or some other method such as command line arguments.

### Why is this needed?

On larger clusters with thousands of pods, selecting "All namespaces" results in huge memory consumption and possible OOMs of the Dashboard API pod. As I've read in other bug reports here, there is no real solution to this problem in the foreseeable future. 
For normal users, we've already removed the ability to list resources cluster-wide in k8s so that clicking "All namespaces" doesn't list anything, but this is not a viable option for people who need cluster admin privileges.
Instead of trying to fiddle with dashboard refresh times and pod resource limits, we would like to hide the "All Namespaces" list item altogether. This would eliminate the possibility of someone accidentally clicking on it and triggering an OOM in the dashboard.

sephialaureenciaa:
Hi, can I contribute to this?

PR #9547, opened 2024-10-17
Allow hiding "All namespaces" in namespaces dropdown list

avanish23:
This PR aims to fix issue #8894 
With this PR a new argument is introduced `**--hide-all-namespaces**` which when passed will hide the all-namespaces option from thenamespaces dropdown list in UI. 
**Briefly explain why this is required**
Hides all namespace fields in the namespace selection dropbox to avoid accidental selection in large clusters and prevent OOM errors.


floreks:
This feels like a hacky workaround. We are working on introducing a dynamic cache that should greatly improve responsiveness in large clusters. However, storing cached data will probably require greater memory limits.

avanish23:
> This feels like a hacky workaround. We are working on introducing a dynamic cache that should greatly improve responsiveness in large clusters. However, storing cached data will probably require greater memory limits.

Yes, I agree. This could be treated as a temporary workaround until we have the fix you are talking about. 
Do you have any potential issues/concerns this could bring up?

floreks:
Ye, I don't like introducing temporary flags that will be deprecated and removed shortly after being introduced. It can be confusing. Also, we are trying to avoid container flags that directly manipulate the UI. It should probably be done in a different way. Possibly via settings.

avanish23:
Ok, this makes sense to be using setting instead. I think it should be added as part of the global setting. 
Let me do it that way.

charts/kubernetes-dashboard/Chart.yaml
```diff
@@ -14,7 +14,7 @@
 
 apiVersion: v2
 name: kubernetes-dashboard
-version: 7.8.0
```
floreks, line 17: You can revert this change. We do not want to trigger the helm release. Do not worry about failing CI check, it has been fixed in a different PR.

charts/kubernetes-dashboard/Chart.yaml
```diff
@@ -14,7 +14,7 @@
...
+version: 7.8.1
```
floreks, line 17:
```suggestion
version: 7.8.0
```

floreks:
For some reason namespace switch on settings view is always set to false even after switching to true and saving. Probably the value read from settings config map is not properly propagated to it.

avanish23:
Oh, I see.
Thanks for the update @floreks, let me take a look at this.

floreks (approved):
/lgtm

PR #9548, opened 2024-10-18
fix: the divisor field in resourceFieldRefs should be a string

elohmeier:
This fixes the type in the yaml, introduced in #9486.

See also: https://github.com/kubernetes/dashboard/blob/1b58a29c161150bf0c151c69cdff9f6e48344ffe/modules/web/schema/schema.graphql#L3843-L3852


floreks (approved):
/lgtm

Issue #7547, opened 2022-11-29
Add ability to filter by pod status and by labels


yasinzaehringer-paradime:
### What would you like to be added?

Add more functionality for filtering (e.g.) pods:
- by status
- by labels
- etc

Also add functionality not only to select matching rows, but also remove matching rows from result.

The selector should also be part of the URL - so it can be shared.

### Why is this needed?

To understand bigger clusters it's important to be able quickly filter. The most common use case I have is this one:
- we have lot's of cronjobs (and we retain the 3 latest runs) -> lot's of `Completed` pods
- we have a few pods -> quite a few `Running` pods

Atm, our dev cluster reports 724 pods and I am interested in let's say failed pods, how do I get there? Sorting by Status and click through a few pages until I find `Status=Error`? It's better to be able to write more complex filters.

avanish23:
These are too many asks for one issue. I have opened a PR that provides backend support to start filtering the pods with various other parameters. 
I think it will be better to track the remaining asks in separate issues; I will probably create them once this PR is merged (might take time). 
Tagging @shu-mutou @floreks for suggestions.

PR #9572, opened 2024-10-23
feat: Add ability to filter by resource labels and annotations

avanish23:
Closes: #7547
Currently, the filtering is allowed by name only, for any resource.
With this PR intends to introduce support for filtering via various parameters 
- name
- status
- label
- annotations
and more. 

How can this be done?
This will be done by comma ',' separated values. If the input does not have a comma-separated value, it will be filtered by name by default. 
If a comma-separated string is provided then the convention followed should be `property,value` for example `status,Running` `label,app=dashboard` `annotations,sidecar.istio.io/inject=true`
For example, if the user input is:
 - `dashboard` the pods filtered will be those that will contain "dashboard" in their name.
 - `name,dashboard,label,app=dashboard,status,Running` the pods filtered will be have "dashboard" in their name, a label with key value pair "{app:dashboard}" and status is "Running"


floreks:
The initial implementation was only meant to work with names. We wanted to take another pass at it. The main flaws of the current approach are:
- Filtering different fields based on string parsing does not make a good API. We should probably use an encoded array of objects that can define fields like: operation, field/property, value, etc. This way we could create more complex filtering queries by multiple fields with i.e. AND/OR operations.
- It does not use server-side filtering. The K8S API supports field-level filtering (name, status, labels). Annotations would probably be handled on our side.

I would prefer not to rely on current logic to not encourage anyone to extend it as it should be rewritten and optimized.

PR #9606, opened 2024-10-28
feat(config map): add binary data to config map detail page

### LINKED ISSUE #9321: Add support of BinaryData fields to configmap display

Currently only data fields is supported in configmap display. binaryData (even if it exist) is not shown.
```kubectl -n percona-xtradb describe configmaps test1```
```
Name:         test1
Namespace:    percona-xtradb
Labels:       <none>
Annotations:  <none>

Data
====

BinaryData
====
readiness-check.sh: 17 bytes

Events:  <none>
```

### Why is this needed?

If would be nice if all standard types of standard kubernetes objects is shown in UI.

### PR DESCRIPTION
**Pull request by lsq645599166**
close #9321 

Add binary data to config map detail page

## Screenshot
image

### DIFF
--- a/modules/web/src/resource/config/configmap/detail/component.ts
+++ b/modules/web/src/resource/config/configmap/detail/component.ts
@@ -66,4 +66,20 @@ export class ConfigMapDetailComponent implements OnInit, OnDestroy {
 
     return JSON.stringify(cm.data);
   }
+
+  getConfigMapBinaryData(cm: ConfigMapDetail): {
+    name: string;
+    size: number;
+  }[] {
+    if (!cm) {
+      return [];
+    }
+
+    return Object.entries(cm.binaryData).map(([name, value]) => ({
+      name,
+      size: atob(value).length,
+    }));
+  }
+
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: modules/web/src/resource/config/configmap/detail/component.ts [hunk 1]
```diff
@@ -66,4 +66,20 @@ export class ConfigMapDetailComponent implements OnInit, OnDestroy {
...
+    name: string;
+    size: number;
+  }[] {
+    if (!cm) {
+      return [];
+    }
+
+    return Object.entries(cm.binaryData).map(([name, value]) => ({
+      name,
+      size: atob(value).length,
```
maciaszczykm, line 80: If we display only the length of it then I think it should be calculated on the server side.

lsq645599166: Thanks. Fixed in the latest commit

lsq645599166:
image
image
@maciaszczykm The api has changed as you last comment, can you take a review again

PR #9610, opened 2024-10-29
fix(signin button): signin button is clickable when token is autofilled by browser

### LINKED ISSUE #8755: Reactivity bug in web UI login form
### What happened?

When logging in (latest version of web UI) and autofilling the token field with my password manager plugin, the value is filled but the Sign in button remains disabled. Only after manually focussing on the field and pressing enter the value is detected and the Sign in button becomes enabled.

### What did you expect to happen?

The Sign in button to become enabled as soon as a value is present in the token input field.

### How can we reproduce it (as minimally and precisely as possible)?

Programmatically set the token value in any manner (e.g. through a browser plugin).

### Anything else we need to know?

I'm pretty sure this worked previously; I have not yet attempted to look for the differences in the UI compared to previous versions myself, but that might be a good place to start.

### What browsers are you seeing the problem on?

_No response_

### Kubernetes Dashboard version

7.0.0

### Kubernetes version

1.29

### Dev environment

_No response_

lsq645599166:
close #8755

Keep sign in button clickable when token is autofilled by browser

### Firefox

https://github.com/user-attachments/assets/ed2a21df-ec4e-4fb5-be1b-370d8a68e2f2

### Chrome

When Chrome, must have an interaction with the browser.

https://github.com/user-attachments/assets/776a86c0-20d5-44c9-a2fa-773c4ae3140a

### DIFF
--- a/modules/web/src/login/component.spec.ts
+++ b/modules/web/src/login/component.spec.ts
@@ -153,7 +153,7 @@ describe('LoginComponent', () => {
       // set inputs and fire change events to trigger onChange()
       const token = fixture.debugElement.query(By.css(queries.token)).nativeElement;
       token.value = loginToken;
-      token.dispatchEvent(new Event('keyup'));
+      token.dispatchEvent(new Event('input'));
 
       submit();
 
--- a/modules/web/src/login/template.html
+++ b/modules/web/src/login/template.html
@@ -47,7 +47,7 @@ limitations under the License.
             placeholder="Bearer token"
             type="password"
             required
-            (keyup)="onChange($event)"
+            (change)="onChange($event)"
           />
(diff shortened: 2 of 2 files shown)


lsq645599166:
> @maciaszczykm wasn't there some kind of issue when using `input` event type too? What about copy-paste action?

It all works ok with `input` event

## Paste with `CMD + v`

https://github.com/user-attachments/assets/56c8c6d7-da91-4efb-aa27-49c046f045b5

## Paste with mouse


https://github.com/user-attachments/assets/eacb48a6-1391-411c-a9e7-6fcc824f7d38

floreks (approved):
Awesome. Thanks for the fix.

/lgtm

PR #9741, opened 2024-12-06
fix: typo error in args for api module

warjiang:
for all eth interface, the `insecure-bind-address` should be set as `0.0.0.0`


warjiang:
another problem i want to metioned is that, why the default value for `argInsecureBindAddress` is `127.0.0.1`, but for `argBindAddress` is `0.0.0.0`, should these two bindaddress be set as `0.0.0.0` as default value.

warjiang:
@floreks @maciaszczykm can you help me to review this one~

floreks (approved):
/lgtm

PR #9913, opened 2025-02-03
Publish Helm chart to GitHub Container Registry

wkbrd:
I have tested this without the chart-releaser step in my GitHub fork and it seems to be working properly.


floreks:
We cannot push our images to this organization's GHCR registry. Due to internal policies, we would not be able to make them public.

/close

PR #10012, opened 2025-03-03
feat(auth api): Allow to pass a path with the Root/Intermediate CA used for the apiserver

### LINKED ISSUE #9715: Allow for custom CA to be installed in api/auth pods

Hi,

I would like to be able to customize (for example, via mounting a configmap) the certificates that the API and Auth pods are using to connect to the Kubernetes API. In my current use case, we use [kube-oidc-proxy](https://github.com/TremoloSecurity/kube-oidc-proxy/blob/master/README.md) as a proxy between all our clusters and Keycloak, which has the SSO information needed for the Kubernetes Dashboard users to properly authenticate and get their authorization details to interact with the underlying Kubernetes API.

However, kube-oidc-proxy is secured by a TLS certificate internally (i.e. in the Service) as well as externally (through the Ingress), both of which are signed by an untrusted authority (the internal cert is auto generated, and the external one with a private CA that would need to be installed in the containers).

As far as I can tell, the only certificate configurations are for the API and the Web containers to expose their HTTP layers with a specific certificate, but there is no way to allow installation of custom certs for Kubernetes API access. The workaround is fully disabling TLS verification on these pods by using `--apiserver-skip-tls-verify`, but that could expose us to MITM attacks.

### Why is this needed?

Would allow usage of kube-oidc-proxy or other scenarios where the Kubernetes API is accessed with a custom certificate chain.

josemaia:
Fixes #9715 

Tested in a real cluster with kube-oidc-proxy with a custom-signed certificate.


josemaia:
@shu-mutou @maciaszczykm @floreks 

would it be possible to get some eyes on this PR? please let me know if I missed any point of the contributing guidelines

PR #9863, opened 2025-01-16
fix: use correct Kong service port for ingress based on TLS settings

hamadodene:
fix #9861

PR #10086, opened 2025-03-27
Update ingress.yaml to accomodate kong service ports

mrjoshuap:
This template built an ingress that incorrectly set `$servicePort` based on `app.ingress.tls.enabled`.

This fixes that by using the `kong.proxy.tls.enabled` for the check instead, thereby selecting the correct service port.

It also adjusts the default annotations to check for `app.ingress.tls.enabled` so unnecessary annotations are avoided.

Finally, it properly sets the backend protocol based on `kong.proxy.tls.enabled`.

I see a few different scenarios:

* SSL Passthrough
  * `app.ingress.tls.enabled` == `false`
  * `kong.proxy.tls.enabled` == `true`

* Ingress terminated TLS
  * `app.ingress.tls.enabled` == `true`
  * `kong.proxy.tls.enabled` == `false`

* Ingress TLS to Kong TLS, though not sure this works because the default ingress annotation `nginx.ingress.kubernetes.io/ssl-redirect: "true"`
  * `app.ingress.tls.enabled` == `true`
  * `kong.proxy.tls.enabled` == `true`

I don't think HTTP to HTTPS proxying is a good idea, so I won't talk more about it.

To fix TLS to TLS, there might be a need to add an additional parameter enabling `ssl-passthrough` behavior.  In this implementation, if `kong.proxy.tls.enabled=true` then it assumes you want to perform SSL-passthrough.  I haven't verified as I'm using Ingress terminated TLS and can't test the needful.

I suspect this has been the culprit of many threads regarding `Error 400 HTTP request to HTTPS`, though I think that might also have something to do with http2 requirements on the proxy -- at least on my deployment.

### DIFF
--- a/charts/kubernetes-dashboard/templates/networking/ingress.yaml
+++ b/charts/kubernetes-dashboard/templates/networking/ingress.yaml
@@ -17,9 +17,14 @@
 
 
 # Determine the service port to use for the ingress configuration
-# If TLS is enabled in the ingress configuration, use the TLS service port.
+# If TLS is enabled in the kong proxy configuration, use the TLS service port.
 # Otherwise, fall back to the HTTP service port.
-{{- $servicePort := (ternary $.Values.kong.proxy.tls.servicePort $.Values.kong.proxy.http.servicePort $.Values.app.ingress.tls.enabled) }}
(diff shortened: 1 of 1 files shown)


mrjoshuap:
This sort of extends the fix applied in #9863

PR #10171, opened 2025-05-12
chore(deps): Bump k8s modules to 0.33.0

shu-mutou:
Also, bump golang to 1.24 and kind to 0.27.0


shu-mutou:
* Angular 19.2 is needed for typescript 5.8.3 that we use now.
  Angular 20.0.0 will be released 20th may 2025.
* k8s.io/api/core/v1 Endpoints was deprecated
  k8s.io/api/discovery/v1 EndpointSlice should be used instead.

PR #10278, opened 2025-08-02
Decide which kong proxy port to use in ingress based on kong setting

tomty89:
Coupling the ingress setting with which kong proxy port to use is not exactly logical in at least some scenarios. One example would be that TLS passthrough is not supported / enabled / desired (but TLS termination is).

Since the kong proxy http port needs to be enabled for it to be usable anyway, this change should not cause current setup that have TLS ingress disabled to break, while it should at the same time offer some extra flexibility (and compatibility, for e.g. "nginx/kubernetes-ingress", which apparently does not support the nginx.ingress.kubernetes.io/ssl-passthrough annotation).

It does break TLS passthrough setup that for some reason have http kong proxy enabled though.

/cc @hamadodene @floreks @maciaszczykm


mrjoshuap:
Yes indeed, @tomty89.  I put in #10086 to accomodate my installation and have been maintaining my own fork... not ideal, but it works.

PR #10368, opened 2025-09-24
fix(api): add GPU resource allocation support and fix pod request/limit calculations

floreks:
- Fixes an issue where pod allocated resource requests/limits were calculated incorrectly and double
- Adds initial support for GPU requests/limits to pod allocated resources


maciaszczykm (approved):
/lgtm

